PR #2, opened 2015-12-14
replace opensoc-streaming version 0.4BETA with 0.6BETA 8e7a6b4ad9febbc…

charlesporter:
…4ea47ba7810c42cc94d4dee37


charlesporter:
Can someone have a look this? I would like it if this could get merged very soon. 

It simply contains the tip of opensoc-streaming from the cisco opensource

This is the point from which the hortonworks forks started.

Owen had indicated that we cannot easily start from the hortonworks repo . 

He had created metron from a very old versio because that is what was hard-linked as sub-repos in the top-level opensoc repo.

We need to get this Pull Request  ok's and merged before we can start on even the most basic improvements, like making it work with recent builds of the hadoop infrastructure .

james-sirota:
I actually don't have permissions to approve this pull request. let me see if i can figure out how to add the PMC members to a group that has these privs

omalley:
Since you guys picked RTC, you should get a review of it to ensure that it matches the code from https://github.com/OpenSOC/opensoc-streaming at 8e7a6b4ad9febbc4ea47ba7810c42cc94d4dee37

ptgoetz:
@james-sirota The github mirrors of apache repos are read-only. You can comment, etc. but you can't use the "merge" button. To merge you have to pull from the repo of the person that opened the request, then push the result to the official apache repo.

omalley:
Yeah, you need a commit that includes "closes apache/#2" in the commit message.

opensoc-streaming/OpenSOC-Alerts/src/main/java/com/opensoc/alerts/TelemetryAlertsBolt.java
```diff
@@ -185,10 +184,10 @@ public void execute(Tuple tuple) {
 			JSONArray uuid_list = new JSONArray();
 
 			if (alerts_list == null || alerts_list.isEmpty()) {
-				LOG.trace("[OpenSOC] No alerts detected in: "
+				System.out.println("[OpenSOC] No alerts detected in: "
```
sirsean: Why would we replace `LOG.trace` with `System.out.println`? Isn't that exactly the opposite of what we want?

james-sirota: I think this is a piece of debugging that someone forgot to comment out.  We can ignore this

dev-warlord: James, do you mean that this particular line will be reverted back to LOG.trace?  We as a group should not have code in the master branch with System.out.*

opensoc-streaming/OpenSOC-Alerts/src/main/java/com/opensoc/alerts/TelemetryAlertsBolt.java
```diff
@@ -196,11 +195,11 @@ public void execute(Tuple tuple) {
 					LOG.trace("[OpenSOC] Checking alerts cache: " + alert);
 
 					if (cache.getIfPresent(alert) == null) {
-						LOG.trace("[OpenSOC]: Alert not found in cache: " + alert);
+						System.out.println("[OpenSOC]: Alert not found in cache: " + alert);
```
dev-warlord: see above

james-sirota: yeah it just needs to be removed as it serves no purpose.  the log is already logging the same line. there is no reason to print it out

sirsean: Well the log isn't logging that same line any more -- the diff shows that the call to the logger was replaced by System.out, which we probably want to switch back.

opensoc-streaming/OpenSOC-Alerts/src/main/java/com/opensoc/alerts/adapters/AllAlertAdapter.java
```diff
@@ -23,64 +21,96 @@
...
+			int _MAX_TIME_RETAIN_MINUTES = Integer.parseInt(config
+					.get("_MAX_TIME_RETAIN_MINUTES"));
 
+			cache = CacheBuilder.newBuilder().maximumSize(_MAX_CACHE_SIZE_OBJECTS_NUM)
+					.expireAfterWrite(_MAX_TIME_RETAIN_MINUTES, TimeUnit.MINUTES)
+					.build();
+		} catch (Exception e) {
+			System.out.println("Could not initialize Alerts Adapter");
+			e.printStackTrace();
+			System.exit(0);
```
sirsean: We should really avoid ever, ever calling `System.exit`

james-sirota: The reason this was done is because this bolt has dependencies on Hbase and Zookeeper.  If something went wrong in the initialization the bolt didn't explicitly fail, but timed out instead.  So the topology would half start and then started blasting packets into a bolt that wasn't fully initialized or had an error.  Having it explicitly exit like this failed the whole topology right away so you immediately knew you had a problem and didn't have to go looking for it in the logs when the topology blew up 5 minutes later

opensoc-streaming/OpenSOC-Alerts/src/main/java/com/opensoc/alerts/TelemetryAlertsBolt.java
james-sirota: I think it's ok to have system.outs in the runners that launch the topology, but it shouldn't exist in the bolts themselves. you are correct. we should change it back to log.trace

omalley:
Yeah, System.exit is exactly right when a thread has failed and can't recover. You want to bring down the whole process and not leave the process partially running.

opensoc-streaming/OpenSOC-Alerts/src/main/java/com/opensoc/alerts/adapters/AllAlertAdapter.java
ptgoetz: An alternative approach that the idiomatic storm way of doing things, is to throw a RuntimeException, which will cause (currently) the worker JVM to exit.

This could be important in the future as storm will likely switch to a model where uncaught  the RuntimeExceptions only cause the specific task to be restarted, not the entire worker JVM.

james-sirota:
hi can anyone walk me through how to commit this the first time i do it?  then i'll publish a wiki so everyone can do it.  i want to make sure i don't screw it up the first time

ptgoetz:
@james-sirota 

Here's the process:

1. Add the official Apache repo as a remote in your local clone (one time step):
`git remote add apache https://git-wip-us.apache.org/repos/asf/incubator-metron.git`
2. checkout whatever branch you want to merge it into:
`git checkout master`
3. Merge the pull request:
`git pull git@github.com:charlesporter/incubator-metron.git master`
4. Push the results to the Apache repo:
`git push apache master`

A git hook will automatically close the github pull request when the repos get synced.

If there is a pull request that you want to close without merging, you can simply ask the person who opened it to close it. If that doesn't work, you can use a commit message to close it.

To do that, make a trivial change, like adding a space to README.md, and commit and push the change. In the commit message add the text "this closes #N" where N is the number of the pull request. When the repos sync, the pull request will get closed.

PR #6, opened 2016-01-06
[METRON-9] Adding a gitignore file

sirsean:
The MVN build process creates a bunch of files (ie, the target directories) that we don't need filling up our source control.


ehashman:
Someone seems to have added a `.gitignore` file since you submitted this pull request. I just did a clean build and the only line I needed to add was

```
opensoc-streaming/OpenSOC-DataLoads/dependency-reduced-pom.xml
```

So perhaps you can reopen this PR by adding that to the new `.gitignore`?

sirsean:
Good idea.

sirsean:
So now there is only one file that's created by the build process and wasn't in the .gitignore file. So that's added here.

reaperhulk:
:+1:

ehashman:
LGTM!

PR #10, opened 2016-01-15
Storm Flux

sirsean:
This switches from the old `TopologyRunner` model written in Java code, to use Storm Flux to configure the topologies entirely in YAML.

This makes the topologies more flexible and the config files are more apparent. It also reduces the number of config files we need in order to run a topology.

I've verified that all these topologies run in local mode.

The command to run a topology has changed:

```
storm jar metron-streaming/Metron-Topologies/target/Metron-Topologies-0.6BETA.jar org.apache.storm.flux.Flux --local metron-streaming/Metron-Topologies/src/main/resources/Metron_Configs/topologies/bro/local.yaml --filter metron-streaming/Metron-Topologies/src/main/resources/Metron_Configs/etc/env/config.properties
```

Note that the "main class" is _always_ the same now, and you specify which topology to run entirely through the YAML.


ptgoetz:
+1 
Nice work migrating everything to flux. That will hopefully make topology configuration more manageable.

sirsean:
I expect this will have a substantial effect on the Ansible scripts, which now have to construct entirely different configuration files.

james-sirota:
thanks for checking in. we are in the process of converting our scripts to encompass your changes. will be checking in shortly

PR #11, opened 2016-01-15
Bro Parser Improvements

sirsean:
The old Bro parser required that the data came wrapped in a key that
specified the protocol. But not every Bro produces data in that format;
now the parser supports that format AND a format that is flatter and the
protocol is just another attribute in the message.

This also contains some minor code cleanup in the Bro parser.

### DIFF
--- a/metron-streaming/Metron-MessageParsers/src/main/java/org/apache/metron/parsing/parsers/BasicBroParser.java
+++ b/metron-streaming/Metron-MessageParsers/src/main/java/org/apache/metron/parsing/parsers/BasicBroParser.java
@@ -22,130 +22,122 @@ import org.json.simple.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.apache.metron.tldextractor.BasicTldExtractor;
-
 @SuppressWarnings("serial")
 public class BasicBroParser extends AbstractParser {
 
-	protected static final Logger _LOG = LoggerFactory
-			.getLogger(BasicBroParser.class);
-	private JSONCleaner cleaner = new JSONCleaner();
-	private BasicTldExtractor tldex = new BasicTldExtractor();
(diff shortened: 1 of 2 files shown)


metron-streaming/Metron-MessageParsers/src/test/java/org/apache/metron/parsing/test/BasicBroParserTest.java
```diff
@@ -29,6 +29,27 @@ public BasicBroParserTest() throws Exception {
 		jsonParser = new JSONParser();		
 	}
 
+    public void testUnwrappedBroMessage() throws Exception {
```
ehashman: Should this be `ParseException`, like the rest of the tests here?

metron-streaming/Metron-MessageParsers/src/main/java/org/apache/metron/parsing/parsers/BasicBroParser.java
```diff
@@ -22,130 +22,122 @@
...
+            payload.put("original_string", originalString);
+
+            replaceKey(payload, "timestamp", new String[]{ "ts" });
+
+            if (payload.containsKey("timestamp")) {
+                try {
+                    long timestamp = Long.parseLong(payload.get("timestamp").toString());
+                    payload.put("timestamp", timestamp);
+                } catch (NumberFormatException nfe) {
+                    _LOG.error(String.format("[Metron] tiemstamp is invalid: %s", payload.get("timestamp")));
```
ehashman: Typo: `tiemstamp -> timestamp`

ehashman:
Other than comments above, LGTM! +1

ehashman:
Note for other reviewers: the way GitHub flowed this is not ideal. Try a `git diff -w` for easier reading.

sirsean:
@ehashman Thanks for the review, I fixed those things.

PR #13, opened 2016-01-22
Falconhose Topology

sirsean:
This adds a FalconHose topology, including a parser that understands the FalconHose API data format, and a new alerts adapter that understands that everything from FalconHose should be considered an alert.


nickwallen:
@sirsean Hi Sean - Even though Github thinks this merges, this no longer merges due to reorg of the code base.  Was this still something that you wanted to address?  If not, mind closing the PR?  Many thanks for your contribution.

PR #14, opened 2016-01-26
Deployment

merrimanr:
This pull request contains the Ansible deployment scripts.  Currently the scripts support 2 deployment types:  deploying Metron to a single node VM (full install including Ambari and cluster) for development and testing purposes and deploying Metron to a cluster that already exists.  The Pcap source works end to end and other sources will be added in future iterations.


merrimanr:
https://issues.apache.org/jira/browse/METRON-21

dbhirko:
A significant part of this code is from another pending pull request from a different committer.  If you have additional code to use, pull the prior pull request into a branch thereby preserving the original author's commits, and you can commit on top of that.

merrimanr:
Sorry was so focused on getting it to work that I wasn't even thinking about the commit history.  I will get this corrected asap.

merrimanr:
This is no longer valid so creating a new pull request now that we have METRON-2 in the commit history.

PR #18, opened 2016-02-09
METRON-28: Metron should compile with JUnit 4

dossett:
This PR will allow metron-streaming to compile with JUnit 4 and puts unit tests in `src/test` instead of `src/main`, but not all of the unit tests will execute, because they are not annotated with `@Test`, do not have the correct constructor signatures, etc.  I would propose additional subtasks under METRON-27 to remediate unit tests in each sub-module.  This change makes it possible to do the work one module at a time.


cestella:
This is great, definitely +1.  Thanks so much!

PR #20, opened 2016-02-11
METRON-29: Remediate Metron-Alerts unit tests for JUnit 4

dossett:
The main unit test here had to be marked `@Ignore` because it is really an integration test, which assumes a zookeeper cluster is available.  However, the main test is class is remediated and ready for JUnit 4 style unit tests to be added.


nickwallen:
@dossett Hi Aaron - This no longer merges due to reorganization of the code base.  Does this still need address?  Mind closing the PR if it has been addressed?  Thanks for your contribution.

dossett:
No problem closing @nickwallen, probably doesn't make sense anymore.

PR #22, opened 2016-02-13
METRON-35 Implement threat intelligence message enrichment

cestella:
Create the infrastructure to
* Bulk ingest threat intelligence feeds from CSV and Stix data sources into HBase
* Enrich messages who have fields which match the threat intelligence data in HBase
* Create the infrastructure to remove unused threat intelligence data
* Augment the Packet capture topology to incorporate a malicious IP threat intel tagger

The tagging infrastructure much meet the following criteria:
* They are downstream of the enrichments
* The threat intelligence bolts execute in parallel with a similar architecture as the enrichments (i.e. split and join).


cestella:
I want to point out a couple of other things this PR provides that aren't strictly associated with the feature above, but are general cleanup tasks:

* Removed lingering hbase-site.xml which have a bad habit of finding their way onto the classpath and confusing HBase in integration tests
* The split of integration tests (defined as a test that ends with "IntegrationTest") into the integration-test maven lifecycle phase
* Using the shade maven plugin to relocate our guava dependency so that we can use a more recent version of Guava than 12 (which is the most recent that HBase will allow due to google's habit of aggressive removal of deprecated code).  This comes up when running HBase in minicluster mode as well as in situations when running bolts which have to package the hbase-client.
* General cleanup of the build to use the version properties instead of hard coding different versions of common components (e.g. hbase-client, storm-core, etc.)

ddutta:
Is there a design doc for this large checkin? Then it will be easier to review. Or maybe if we split this into more manageable chunks.... since this is a new feature beyond infra/devops.

cestella:
Of course, I have attached a design doc to the [JIRA](https://issues.apache.org/jira/browse/METRON-35) .

This is really a single feature as leaving out any part will leave the whole feature nonfunctional.  It can seem a bit complex, but it fits within the overall architecture built for the enrichments.  I detailed this and how it fits within the overall architecture in the design doc.

merrimanr:
Looks good to me.  +1

james-sirota:
+1 from me as well.  Great job

PR #26, opened 2016-02-19
METRON-42 Replace GPL license language with ASF for custom ansible module

mbittmann:
This code is not part of ansible and therefore not subject to GPL. As
the original author, I am including the Apache license as part of the
Metron project.


cestella:
Thanks Mark.  I'm going to merge this one right now

PR #28, opened 2016-02-23
METRON-43: Refactor Deployment Scripts for Correctness and Ease of Use

dlyle65535:
Runs with single node Vagrant, multi-node Vagrant and multi-node cloud instance of 12 nodes.

Includes:
    METRON-44: passwd_less_ssh.yml overwrites authorized_keys and leaves artifacts on local host
    METRON-45: Remove host references from global variables
    METRON-43: Deployment scripts should fail fast
    METRON-47: Vagrant multi-node deployment script only deploys to one node
    METRON-32: Added Ansible role to install and load data into MySQL


dlyle65535:
Sorry- forgot to run RAT last night. Fixed it this morning. There was one file, geoip_ddl.sql, that did not pass. I corrected that by adding a license to it.

I also excluded .iml files as they're .gitiignored.

nickwallen:
Why pull all variables, even those unlikely to change, up to the inventory level?  For example, why define `metron_jar_name` in `deployment/inventory/singlenode-vagrant/group_vars/all`? 

Previously these sorts of vars were pushed down into the role itself.  The logic being that these will rarely change and that the user need not worry about them.  For example, `metron_jar_name` is defined in `incubator-metron/deployment/roles/metron_streaming/vars/main.yml`.

nickwallen:
Why install OpenJDK as part of `deployment/roles/yaf/tasks/main.yml`?  Previously the scripts were standardized on Oracle Java 8, which I believe is preferred for HDP, no?

dlyle65535:
That makes sense. Unless there's something badly broken here, can we push it to a separate issue? It was a major effort getting a passing run on all three integration tests and now that I'm there, I'm reluctant to  re-run them- it's a 2 hour end-to-end and I'm a little concerned about getting back into merge-land while that's happening.

dlyle65535:
That wasn't my change. Oracle is used when it's provided, open-jdk is
selected otherwise. In cases where there was no jdk without a side effect,
I continued that pattern.

nickwallen:
Also, seeing that promisc enablement was turned off as part of this change in `deployment/vagrant/multinode-vagrant/Vagrantfile`.  This needs to be on to sniff network traffic.

dlyle65535:
I don't see that change- which line?

nickwallen:
```   
# enable the hostmanager plugin
    config.hostmanager.enabled = true
 @@ -44,11 +44,6 @@ Vagrant.configure(2) do |config|
        node.vm.provider "virtualbox" do |vb|
          vb.memory = host[:memory]
          vb.cpus = host[:cpus]
 -
 -        # enable promisc mode on the network interface
 -        if host.has_key?(:promisc)
 -          vb.customize ["modifyvm", :id, "--nicpromisc#{host[:promisc]}", "allow-all"]
 -        end
        end
```

dlyle65535:
Got it, thanks- I'll get that back in, move the variables and re-test.

dlyle65535:
Testing passed with latest commit. Thanks for looking it over!

PR #29, opened 2016-02-24
METRON-50 Ingest threat intel data from Taxii feeds

cestella:
Ingest Taxii threat intel feeds via a polling service that will write into HBase as an companion to the BulkLoadThreatIntel job.


merrimanr:
Looks good.  +1

PR #31, opened 2016-02-25
METRON-52 Bro Plugin Header has comma instead of colon

dlyle65535:
Verifed correct in integration testing and JSONLint


nickwallen:
+1 Change looks good.  Thanks for the fix.

PR #35, opened 2016-03-02
Added Checkstyle xml and classes to each submodule

joshuad2:


nickwallen:
@joshuad2 I believe we have this functionality covered now.  Is there anything we're missing from your contribution here?  If not, would you mind closing this PR?  Thanks for your contribution and looking forward to many more.

PR #38, opened 2016-03-08
METRON-61: Allow selection of topologies and sensors in deployment.

dlyle65535:
Allows selection/deselection of topologies and sensor deployment from group vars.

Spun up local vm.
Rat Check passes.
Integration Tests pass.


cestella:
+1

PR #46, opened 2016-03-15
METRON-69 EC2 Volume Enhancements

nickwallen:
### Changes
- The EC2 volume sizes are now configurable
- The root volume is only expanded if needed.

### Validation
- [x] Positive and negative tests on live EC2 cluster.
- [x] License Check
- [x] Integration Tests

### DIFF
--- a/deployment/amazon-ec2/conf/defaults.yml
+++ b/deployment/amazon-ec2/conf/defaults.yml
@@ -16,12 +16,15 @@
 #
 ---
 # ec2
+env: metron-test
 region: us-west-2
 instance_type: m4.xlarge
 image: ami-05cf2265
 volume_type: standard
 key_name: metron-key
-env: metron-test
+xvda_vol_size: 50
+xvdb_vol_size: 100
+xvdc_vol_size: 100
 
 # ambari
 ambari_host: "{{ groups.ambari_master[0] }}"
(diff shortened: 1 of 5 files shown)


nickwallen:
@dlyle65535 - Please take a gander

dlyle65535:
Overall looks good. I'll run it up. Originally, I only intended this to run once at first provisioning. You may want to consider enforcing that. I wouldn't move a partition end on a system after initial buildout.

nickwallen:
Yes, this is a gap, but one that did already exist.  At the very least this change is not making things worse.

Personally I would not be too concerned about this. if conditions ever exist for the 'expand-volume' logic to fire when it shouldn't then there are likely larger problems in the environment.  The kind of 'larger' problems that might cause someone to just shoot the node in the head and start over anyways.  I could be wrong though.

dlyle65535:
I think the gap was that since we still have some work to do on idempoentence in the scripts, I mistakenly believed that no one would run these twice and expect to be happy with the result. It that holds (and potentially is documented somewhere) I'm good.

nickwallen:
Are you saying that we should not merge this PR until everything is completely idempotent?  Totally open to a -1 if you think there is a better way that we should do this.

dlyle65535:
No, not at all. Just don't want to introduce more places where it breaks.

dlyle65535:
+1 as is. Thanks!

PR #48, opened 2016-03-17
METRON-74: Remove Mysql RPM from Project.

dlyle65535:
Remove RPM pull from dev.mysql.com.

Also cleaned up a warning and a textural error in ambari_common that I noticed while testing.


nickwallen:
+1

cestella:
+1 this looks good, merging now

PR #50, opened 2016-03-18
Metron-72 Update UI to support pcap download

merrimanr:
The UI is now built from source and installed instead of being installed from a remote npm repository.  Tests completed:
- single node vagrant script
- integration tests


deployment/roles/metron_ui/tasks/main.yml
```diff
@@ -23,46 +23,33 @@
...
-  file: path=/tmp/nvm-install.sh mode=0755
-
-- name: Install nvm
-  shell: /tmp/nvm-install.sh creates=~/.nvm/nvm.sh
-
-- name: Install Metron UI
-  shell: source ~/.bashrc && nvm install 0.10.35 && nvm use 0.10.35 && nvm alias default 0.10.35 && npm install -g opensoc-ui creates=/root/.nvm/v0.10.35/lib/node_modules/opensoc-ui/package.json
-
-- name: Add Metron UI config
+- name: Copy Metron UI source
```
nickwallen: I am deploying Metron with your PR out to AWS/EC2.  One thing I have noticed so far is that this recursive copy takes FOREVER.  Sorry, I don't have a specific duration to quote, but it seems excessive.

One thing I remember reading is that the 'copy' module in Ansible does not scale well to lots of files.  We may need to tarball up the code and copy it over that way.  Or perhaps you can think of a better alternative.

http://docs.ansible.com/ansible/copy_module.html

> Note
> The “copy” module recursively copy facility does not scale to lots (>hundreds) of files. For alternative, see synchronize module, which is a wrapper around rsync.

merrimanr: Completely agree.  I will add that in.

metron-streaming/Metron-Pcap_Service/pom.xml
```diff
@@ -114,15 +114,24 @@
...
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+			<version>${global_hbase_guava_version}</version>
+		</dependency>
+		<dependency>
 			<groupId>org.apache.hbase</groupId>
 			<artifactId>hbase-client</artifactId>
 			<version>${global_hbase_version}</version>
-			<scope>provided</scope>
+			<!--scope>provided</scope-->
```
dlyle65535: Rather than comment this out, can you remove it? Unless you think it should stay.

metron-streaming/Metron-Topologies/src/main/resources/Metron_Configs/etc/env/config.properties
```diff
@@ -17,8 +17,8 @@
 
 ##### Kafka #####
 
-kafka.zk=zkpr1:2181,zkpr2:2181,zkpr3:2181
-kafka.broker=kfka1:6667
+kafka.zk=node1:2181
+kafka.broker=node1:6667
```
dlyle65535: I think these are going to break the common deploy. Shouldn't they remain zkpr[1-3] and kfka1?

deployment/roles/metron_ui/tasks/main.yml
dlyle65535: Didn't @cestella have a patch out there with a tar.gz'd deploy?

PR #53, opened 2016-03-21
METRON-77: Refactor Deployment Scripts to Accept Changes from METRON-72

dlyle65535:
PCAP topology filename is now remote.yaml.
YAF indexing can sometimes auto-code risn as a number.
Metron-UI copy needs to be an archive rather than recurse directories.
The Kafka client is installed as a side-effect of other tasks. Make it a dependency of each task that requires it.
The single-node deployment can only run 4 topologies. I recommend running Bro, Snort, Yaf and Enrichment by default. It can be overridden should one desire.

This has been run on both single-node Vagrant and EC2.


cestella:
This looks good.  I especially like the single-node deployment only running bro, snort, yaf and enrichment.  I think we could even trim it down to just bro and yaf to give us some headroom.

merrimanr:
Dave commented in Metron-72 that we should revert /incubator-metron/metron-streaming/Metron-Topologies/src/main/resources/Metron_Configs/etc/env/config.properties back to use zkpr[1-3] and kfka1.  I agree with him and think we should but don't see it in this PR.  Can we include that change as well?

dlyle65535:
No thanks @merrimanr. I had that wrong. The regexes match on the lhs and replace the rhs. It works like a champ.

cestella:
+1 from me now.  I believe the travis error is due to travis not being set up yet, which will happen external to this PR.

PR #54, opened 2016-03-21
METRON-79 PcapService is hanging when called from the UI

merrimanr:
This pull request includes changes to the pcapservice Ansible script so that the hbase-site.xml file is put in the correct location inside the pcap jar.  Also includes a UI change to only display the Search button when all required fields are populated.


dlyle65535:
+1. Ran this on EC2 and verified pcap was searchable from the UI.

cestella:
+1 from me

PR #55, opened 2016-03-21
METRON-81: Metron-UI Deployment should exclude node_modules

dlyle65535:
Excluded node_modules from the metron-ui tar call.


cestella:
+1.  This should avoid pulling over local copies node modules, which we're seeing periodically.

PR #58, opened 2016-03-22
METRON-83 Create 'sensor test mode'

nickwallen:
The 'sensor test mode' is an Ansible role that configures each of the sensors to produce the maximum amount of telemetry data.  This role is useful only for testing.  It can be useful to support functional, performance, and load testing of Apache Metron.

The role does the following to maximize the amount of telemetry data produced by each Metron sensor.

- Plays a packet capture file through a network interface to simulate live network traffic.
- Configures [YAF](https://tools.netsa.cert.org/yaf/yaf.html) to produce a flow record for every network packet received.
- Configures [Snort](https://www.snort.org/) to produce an alert for every network packet received.

To enable 'sensor-test-mode' apply the role to the `sensors` host group in your Ansible playbook.

```
- hosts: sensors
  roles:
    - role: sensor-test-mode
```

The role has also been added to the default `metron_install.yml` playbook so that it can be turned on/off with a property in both the local Virtualbox and the remote EC2 deployments.

```
sensor_test_mode: True
```


nickwallen:
@cestella Can we use this mechanism to load the threat intel or other forms of data that are currently being loaded elsewhere?  

This mechanism will really help in creating complementary test data sets. For example where the same IPs need to be used in pcap, threat intel, asset enrichment, etc data sets to enable correlation within Metron.

nickwallen:
@dlyle65535 @cestella @merrimanr I'd love to see this go in.  It is a great way to extract test setup from the cluster deployment process.  I'd like to use this mechanism to setup some additional tests when I return.  Please review and let me know.

deployment/playbooks/metron_install.yml
```diff
@@ -75,7 +75,8 @@
     - { role: flume,  when: install_snort | default(True) == True }
     - { role: snort , when: install_snort | default(True) == True }
     - { role: yaf, when: install_yaf | default(True) == True }
-    - { role: pcap_replay , when: pcap_replay | default(False)  == True }
+    - { role: pcap_replay , when: (pcap_replay | default(False)) or (sensor_test_mode | default(False)) == True }
+    - { role: sensor-test-mode, when: sensor_test_mode | default(False) == True }
```
nickwallen: @dlyle65535 I changed this so `pcap_replay` gets installed if `sensor_test_mode` is on.  Is that the concern you had?  Any other concerns?

nickwallen:
A nice benefit with this change is that the default deployment produces Snort alerts in the dashboard.  I think I've been asked how to "turn on snort alerts" quite a few times. ;)

deployment/playbooks/metron_install.yml
nickwallen: Although sensor-test-mode has pcap_replay as a dependency in `deployments/roles/sensor-test-mode/meta/main.yml`, so maybe I do not need need the extra logic in `deployment/playbooks/metron_install.yml`?

dlyle65535:
+1 looks good!

deployment/playbooks/metron_install.yml
dlyle65535: Yeah, I see what you mean. If it's functionally the same and you'd prefer to omit the or clause, I'm all for it.

james-sirota:
The docs on how to turn on Snort alerts are documented in the wiki here: https://cwiki.apache.org/confluence/display/METRON/Adding+Dummy+Snort+Data+for+Load+Testing

PR #62, opened 2016-03-28
METRON-89: Rerunning playbook halts due to hdfs put.

yibter:
The playbook for AWS fails to rerun due to a hdfs put throwing “File exists” errors. Not seeing a hdfs put with override capabilities, so using the “ignore_errors” flag pattern seen elsewhere just so the playbook will complete on subsequent runs...

ignore_error

(Other than that, due to having to restart the playbook a couple of times, I was able to get an AWS environment up and running with one command -- which is awesome!)

### DIFF
--- a/deployment/roles/metron_streaming/tasks/grok_upload.yml
+++ b/deployment/roles/metron_streaming/tasks/grok_upload.yml
@@ -31,8 +31,7 @@
   become_user: hdfs
 
 - name: Upload Grok Patterns to hdfs://{{ metron_hdfs_output_dir }}
-  command: hdfs dfs -put {{ metron_directory }}/config/patterns  {{ metron_hdfs_output_dir }}
+  command: hdfs dfs -put -f {{ metron_directory }}/config/patterns  {{ metron_hdfs_output_dir }}
   become: yes
   become_user: hdfs
-  ignore_errors: yes
 
--- a/deployment/roles/metron_streaming/tasks/grok_upload.yml
+++ b/deployment/roles/metron_streaming/tasks/grok_upload.yml
@@ -34,4 +34,5 @@
   command: hdfs dfs -put {{ metron_directory }}/config/patterns  {{ metron_hdfs_output_dir }}
   become: yes
   become_user: hdfs
+  ignore_errors: yes
 


james-sirota:
Thanks for the contribution.  Will verify shortly

dlyle65535:
Good catch! Thanks for the PR. 

Lately, I'm a little reluctant to add ignore_errors without a fail block immediately following. In this case, you could add a register: result to the Upload grok patterns task followed by a fail task that fails if "File exists" isn't in stdout. That would make sure it didn't ignore different types of errors.

Since we're interacting with hdfs immediately prior to the file copy, I think ignore_errors would be okay here, but if you had the time to add the fail block, it would be preferred. It demonstrates the preferred pattern and gives others a better example to follow.

What do you think?

yibter:
Sounds like a plan @dlyle65535! So, based on what you were saying about the register and the fail block, I found this guy...

http://docs.ansible.com/ansible/playbooks_error_handling.html#controlling-what-defines-failure

Looks like the "ignore_errors + register + the separate fail block w/ the explicit message" pattern can be collapsed into a "register + failed_when" pattern, so something like this maybe...

```yaml
register: put_result
failed_when: "put_result | failed and 'File exists' not in put_result.stderr"
```

And then in a completely different direction, looks like the "[copyFromLocal](https://hadoop.apache.org/docs/r2.7.2/hadoop-project-dist/hadoop-common/FileSystemShell.html#copyFromLocal
)" (instead of the put) command supports the force (-f) flag... which would just overwrite the destination without throwing an error.

So thoughts on which you'd prefer...
1. Separate Fail Block w/ Messaging.
2. Failed_when that will do a normal fail if it's not a fail exists issue
3. Use a copyFromLocal w/ the override flag (instead of the put)

dlyle65535:
From my perspective, #3 would be a good approach. If you have a different opinion, I'm happy to support it.

Thanks for digging into it!

nickwallen:
In some of my own branches I had been using `hdfs dfs -put -f ...` successfully.  I'd recommend that.  I don't really understand the difference between that and `copyFromLocal`, but either should probably work.

https://github.com/nickwallen/incubator-metron/blob/METRON-80/deployment/roles/metron_streaming/tasks/grok_upload.yml

yibter:
Nice @nickwallen! Just verified the -f on the -put totally works. And for fun, I looked at the hadoop code and you can see the [-copyFromLocal](https://github.com/apache/hadoop/blob/948b75807068c304ffe789e32f2b850c0d653e0a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/shell/CopyCommands.java#L290) is just an alias for put now...

```java
  public static class CopyFromLocal extends Put {
    public static final String NAME = "copyFromLocal";
    public static final String USAGE = Put.USAGE;
    public static final String DESCRIPTION = "Identical to the -put command.";
  }
```

And the [-put ](https://github.com/apache/hadoop/blob/948b75807068c304ffe789e32f2b850c0d653e0a/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/shell/CopyCommands.java#L231)supports all kinds of nice undocumented flags...

```java
  public static class Put extends CommandWithDestination {
    public static final String NAME = "put";
    public static final String USAGE =
        "[-f] [-p] [-l] [-d] <localsrc> ... <dst>";
    public static final String DESCRIPTION =
      "Copy files from the local file system " +
      "into fs. Copying fails if the file already " +
      "exists, unless the -f flag is given.\n" +
      "Flags:\n" +
      "  -p : Preserves access and modification times, ownership and the mode.\n" +
      "  -f : Overwrites the destination if it already exists.\n" +
      "  -l : Allow DataNode to lazily persist the file to disk. Forces\n" +
      "       replication factor of 1. This flag will result in reduced\n" +
      "       durability. Use with care.\n" +
        "  -d : Skip creation of temporary file(<dst>._COPYING_).\n";
```

I've updated this PR to use the -f instead of the ignore_errors, but it looks like Nick already has the fix in his other branch. Up to you guys on if you want to close this and/or wait for the other [commit](https://github.com/nickwallen/incubator-metron/commit/24e3688326d4764e7ef69cec9a7d9e1a40bafe46) to make its way over.

nickwallen:
No, please go ahead with your fix.  Thanks for all the research.

dlyle65535:
+1 Thank you for your contribution!

I've opened METRON-89: Rerunning playbook halts due to hdfs.

PR #63, opened 2016-03-28
METRON-88: Expose rotation policy in Flux configuration of HDFSWriter

dlyle65535:
Defaults to a single file per day.


cestella:
+1 for sure.

merrimanr:
+1.  Eventually we want to move this setting to zookeeper.

PR #66, opened 2016-04-01
METRON-93: Generalize the HBase threat intel infrastructure to support enrichments

cestella:
As it stands, the threat intel infrastructure is awkward. Namely, different threat intelligence sources must be pushed into separate hbase tables (malicious_ips separate form malicious_hosts, for instance). We'd rather have one table where the type is brought into the rowkey. Since this infrastructure is generalized, also add a simple hbase enrichment adapter.
Furthermore, the configuration for a new enrichment should be added to zookeeper as part of the data load.


metron-streaming/Metron-DataLoads/src/main/java/org/apache/metron/dataloads/hbase/mr/PrunerMapper.java
```diff
@@ -48,7 +48,7 @@ public void setup(Context context) throws IOException
         }
         try {
             tracker = AccessTrackerUtil.INSTANCE.loadAll(AccessTrackerUtil.INSTANCE.loadAll(table, atCF, atName, timestamp));
-        } catch (Exception e) {
+        } catch (Throwable e) {
```
dlyle65535: Was this changed to deliberately catch subclasses of Error?

cestella: This is just a style thing.  It's going to throw a runtime exception
regardless and kill the task, but I wanted it to be clear in the logs that
it was due to not bing able to load the access tracker stuff (as opposed to
an OOM exception).

metron-streaming/Metron-Common/src/main/java/org/apache/metron/enrichment/EnrichmentConfig.java
```diff
@@ -0,0 +1,203 @@
...
+          fieldToTypeMap = new HashMap<>();
+          config.setFieldToEnrichmentTypeMap(fieldToTypeMap);
+        }
+      }
+      if(fieldToTypeMap == null  || fieldMap == null) {
+        _LOG.debug("fieldToTypeMap is null or fieldMap is null, so skipping");
+        continue;
+      }
+      //Add the additional fields to the field list associated with the hbase adapter
+      {
```
merrimanr: Are these inline braces necessary?

cestella: Just as logical separation between handling parts of this method handling the additional enrichment types and the part above handling the additional fields that might be added.

metron-streaming/Metron-Common/src/test/java/org/apache/metron/enrichment/EnrichmentConfigTest.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class EnrichmentConfigTest {
+  @Test
+  public void testThreatIntel() throws Exception {
+    /*
+    {
+      "index": "bro",
```
merrimanr: We should probably remove the commented blocks.

cestella: Normally I'd agree, but those comments are there because multiline strings in java are lacking, so if changes to the JSON structure happen, the changes generally happen to the comment and then the string is replaced, letting the IDE handle escaping quotes.  One thing that we could consider is incorporating something like https://github.com/benelog/multiline, which i've seen in other projects for these kinds of situations.

dlyle65535: I'd get rid of them too. I suspect they'll quickly get out of sync and become misleading.  Maybe go ahead and add in multiline or just use a resource file?

cestella: Ok, fair enough, I'm adding the multiline string utility for the unit tests and moving the config JSON to there, so we don't have to maintain them in 2 different places.  This is a similar pattern as they use in DataFu for their unit tests.

metron-streaming/Metron-Common/pom.xml
```diff
@@ -34,8 +34,11 @@
             <name>Metron Kraken Repository</name>
             <url>https://raw.github.com/opensoc/kraken/mvn-repo</url>
         </repository>
+
     </repositories>
     <dependencies>
+
+
```
dlyle65535: Did you mean to add these?

metron-streaming/Metron-Common/src/main/java/org/apache/metron/bolt/BulkMessageWriterBolt.java
```diff
@@ -81,8 +81,7 @@ public void execute(Tuple tuple) {
       sensorMessageMap.put(sensorType, messageList);
     } else {
       try {
-        String esType = sensorType + "_doc";
-        bulkMessageWriter.write(esType, configurations, tupleList, messageList);
+        bulkMessageWriter.write(sensorType, configurations, tupleList, messageList);
```
dlyle65535: This is going to break the ui query.

metron-streaming/Metron-Common/pom.xml
cestella: the newlines?  Nope, I'll remove them.

metron-streaming/Metron-Common/src/main/java/org/apache/metron/bolt/BulkMessageWriterBolt.java
cestella: Yeah, I moved the _doc modification down into the ES writer rather than having it at the bolt level.  Should still work with the UI.

dlyle65535: Disregard. I see it below. :)

dlyle65535:
+1

Ran this up last night on ec2 and let it soak. Looks good!

deployment/roles/metron_streaming/tasks/main.yml
```diff
@@ -109,8 +111,10 @@
...
     - { regexp: "threat.intel.tracker.cf=", line: "threat.intel.tracker.cf=t" }
-    - { regexp: "threat.intel.ip.table=", line: "threat.intel.ip.table={{ threatintel_ip_hbase_table }}" }
-    - { regexp: "threat.intel.ip.cf=", line: "threat.intel.ip.cf=t" }
+    - { regexp: "threat.intel.simple.hbase.table=", line: "threat.intel.simple.hbase.table={{ threatintel_hbase_table }}" }
+    - { regexp: "threat.intel.simple.hbase.cf=", line: "threat.intel.simple.hbase.cf=t" }
+    - { regexp: "enrichment.simple.hbase.table=", line: "enrichment.simple.hbase.table={{ enrichment_hbase_table }}" }
+    - { regexp: "enrichment.simple.hbase.cf=", line: "enrichment.simple.hbase.cf=t" }
     - { regexp: "mysql.ip=", line: "mysql.ip={{ groups.mysql[0] }}" }
     - { regexp: "mysql.password=", line: "mysql.password={{ mysql_root_password }}" }
     - { regexp: "index.hdfs.output=", line: "index.hdfs.output={{ metron_hdfs_output_dir }}/enrichment/indexed" }
```
dlyle65535: Sorry for the late comment- I just noticed. Outside of being duplicate files outside of the name, I think there needs to be a spout.kafka.topic.yaf and spout.kafka.topic.snort in the above. Right now it defaults okay, but it cannot be overridden.

cestella: I think the problem is deeper than this.  We need separate properties files for the sensor topologies than we have for the enrichment topologies.  As it stands, we're using the same config for everything and that's just confusing.  I'm creating https://issues.apache.org/jira/browse/METRON-99 to capture it.

PR #67, opened 2016-04-01
Metron 86 - SolrWriter

merrimanr:
This pull request adds support for Solr.  Adding Solr required a change to how code is built and deployed because a) Solr and Elasticsearch both depend on Apache Lucene and b) the supported versions of Solr and Elasticsearch and very far apart in terms on when they were released.  Now Solr and Elasticsearch are contained in separate projects and deployed as separate artifacts.


metron-streaming/Metron-Common/src/test/java/org/apache/metron/pcap/PcapUtilsTest.java
```diff
@@ -6,17 +6,26 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
  * Unless required by applicable law or agreed to in writing, software
```
dlyle65535: I'm a little surprised that checkstyle didn't object to this. I think this was added by a format code call in your dev IDE.

metron-streaming/Metron-Solr/src/main/java/org/apache/metron/writer/solr/MetronSolrClient.java
```diff
@@ -0,0 +1,83 @@
...
+        request(getCreasteCollectionsRequest(name, numShards, replicationFactor));
+      } catch (SolrServerException | IOException e) {
+        LOG.error(e, e);
+      }
+    }
+  }
+
+  public QueryRequest getCreasteCollectionsRequest(String name, int numShards, int replicationFactor) {
+    ModifiableSolrParams params = new ModifiableSolrParams();
+    params.set("action", CollectionParams.CollectionAction.CREATE.name());
```
cestella: Maybe make these keys a constant?

metron-streaming/Metron-Solr/src/main/java/org/apache/metron/writer/solr/MetronSolrClient.java
```diff
@@ -0,0 +1,83 @@
...
+  public MetronSolrClient(String zkHost) {
+    super(zkHost);
+  }
+
+  public void createCollection(String name, int numShards, int replicationFactor) {
+    if(!listCollections().contains(name)) {
+      try {
+        request(getCreasteCollectionsRequest(name, numShards, replicationFactor));
+      } catch (SolrServerException | IOException e) {
+        LOG.error(e, e);
```
cestella: Should we be swallowing this exception?  Seems like this should bubble up.

metron-streaming/Metron-Solr/src/main/java/org/apache/metron/writer/solr/MetronSolrClient.java
```diff
@@ -0,0 +1,83 @@
...
+    return request;
+  }
+
+  public List<String> listCollections() {
+    List<String> collections = new ArrayList<>();
+    try {
+      NamedList<Object> response = request(getListCollectionsRequest(), null);
+      collections = (List<String>) response.get("collections");
+    } catch (SolrServerException | IOException e) {
+      LOG.error(e, e);
```
cestella: Can we bubble this exception up?

deployment/inventory/singlenode-vagrant/group_vars/all
```diff
@@ -62,7 +62,7 @@ threat_intel_bulk_load: False
 #Sensors
 install_pycapa: False
 install_bro: True
-install_snort: True
+install_snort: False
```
cestella: Why are we not installing snort?  This should be set to True.

deployment/inventory/multinode-vagrant/group_vars/all
```diff
@@ -49,17 +49,23 @@ snort_version: "2.9.8.0-1"
...
-journalnode_edits_dir: "/newdir/hadoop/hdfs/journalnode"
-nodemanager_local_dirs: "/newdir/hadoop/yarn/local"
-timeline_ldb_store_path: "/newdir/hadoop/yarn/timeline"
-timeline_ldb_state_path: "/newdir/hadoop/yarn/timeline"
-nodemanager_log_dirs: "/newdir/hadoop/yarn/log"
-jhs_recovery_store_ldb_path: "/newdir/hadoop/mapreduce/jhs"
-storm_local_dir: "/newdir/hadoop/storm"
-kafka_log_dirs: "/newdir/kafka-log"
-elasticsearch_data_dir: "/newdir1/elasticsearch"
+#zookeeper_data_dir: "/newdir/hadoop/zookeeper"
```
cestella: These should be removed, not commented out.

metron-streaming/Metron-Common/src/main/java/org/apache/metron/bolt/ConfiguredBolt.java
```diff
@@ -77,15 +82,37 @@ public void childEvent(CuratorFramework client, PathChildrenCacheEvent event) th
     } catch (Exception e) {
       throw new RuntimeException(e);
     }
+
+    long timeElapsed = 0;
```
cestella: We should load the configs directly from zookeeper synchronously rather than do a while, sleep until they get loaded asynchronously.

metron-streaming/Metron-Elasticsearch/pom.xml
```diff
@@ -0,0 +1,202 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.metron</groupId>
+            <artifactId>Metron-Topologies</artifactId>
+            <version>${project.parent.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <version>1.9.5</version>
```
cestella: We should make this version a global variable

metron-streaming/Metron-Solr/src/main/java/org/apache/metron/writer/solr/MetronSolrClient.java
```diff
@@ -0,0 +1,83 @@
...
+    if(!listCollections().contains(name)) {
+      try {
+        request(getCreasteCollectionsRequest(name, numShards, replicationFactor));
+      } catch (SolrServerException | IOException e) {
+        LOG.error(e, e);
+      }
+    }
+  }
+
+  public QueryRequest getCreasteCollectionsRequest(String name, int numShards, int replicationFactor) {
```
cestella: Misspelled create

metron-streaming/Metron-Solr/src/main/java/org/apache/metron/writer/solr/SolrWriter.java
```diff
@@ -0,0 +1,115 @@
...
+      SolrInputDocument document = new SolrInputDocument();
+      document.addField("id", getIdValue(message));
+      document.addField("sourceType", sourceType);
+      for(Object key: message.keySet()) {
+        Object value = message.get(key);
+        document.addField(getFieldName(key, value), value);
+      }
+      UpdateResponse response = solr.add(document);
+    }
+    solr.commit(collection);
```
bbende: I don't have a great of understanding of how Metron uses Solr & ES, so my comment may not be relevant... but would you want to consider allowing a commit to be optional based on a flag provided through something like withShouldCommit(boolean shouldCommit)?  

The reason being that a hard commit is generally expensive, and many times people rely on autoCommit/autoSoftCommit in solrconfig.xml to let Solr perform commits on a time interval, rather than every add. Looks like the current solrconfig.xml has soft commits turned off and hard commits every 15 seconds.

merrimanr:
Casey and David, just pushed a commit to address your comments (integration tests ran).  

Bryan, thanks for commenting, that's a great point.  It's a realtime streaming application so handling a high volume of writes and seeing updates in near real time are important.  What strategy would you suggest?

bbende:
@merrimanr I think using the auto commits would be the way to go and gives you the flexibility to tune the values to your needs. You could still leave the optional commit in the writer in case it is needed for testing or anything else. 

In terms of the appropriate commit intervals, a lot of it is trial and error depending on how much latency the application can tolerate to see updates, vs. speed & amount of incoming data. Soft commits are lightweight operations (some overhead) that will make updates visible, but the updates aren't flushed to disk until a hard commit. 

This post has a good description of the tread-offs:
https://lucidworks.com/blog/2013/08/23/understanding-transaction-logs-softcommit-and-commit-in-sorlcloud/

Metron might be the "Index-Heavy, Query-light" scenario? but not sure.

merrimanr:
Thanks for you input @bbende.  I've committed the change.

metron-streaming/Metron-Common/src/main/java/org/apache/metron/bolt/BulkMessageWriterBolt.java
```diff
@@ -66,23 +67,21 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
-      sourceTupleMap.put(sourceType, tupleList);
-      sourceMessageMap.put(sourceType, messageList);
+      sensorTupleMap.put(sensorType, tupleList);
+      sensorMessageMap.put(sensorType, messageList);
     } else {
       try {
-
-        String esType = sourceType + "_doc";
-        bulkMessageWriter.write(esType, configuration, tupleList, messageList);
+        bulkMessageWriter.write(sensorType, configurations, tupleList, messageList);
```
dlyle65535: I think this will break the UI. The query it launches looks is ```"query": "_type:sensorType_doc"``` rather than sensorType. That esType was the fix I put in back when I first got here to make the UI work. I had added it to the writer to keep the ES/UI implementation details out of the rest of the topology.

Alternatively, we could change the UI query. But right now, I don't see a way we will see documents in the UI with this as is.

metron-streaming/Metron-Solr/src/main/java/org/apache/metron/writer/solr/SolrWriter.java
```diff
@@ -65,7 +71,9 @@ public void write(String sourceType, Configurations configurations, List<Tuple>
       }
       UpdateResponse response = solr.add(document);
     }
-    solr.commit(getCollection(configurations));
+    if (shouldCommit) {
+      solr.commit(getCollection(configurations));
+    }
   }
```
dlyle65535: There may be a little more work to do on this. I ran it up last night and after 1/2 an hour had to issue a manual commit to see any documents in Solr. IIRC, openSearcher is false by default which would explain that.

bbende: I think in solrconfig.xml the autoSoftCommit just needs to be set to something:

https://github.com/apache/incubator-metron/pull/67/files#diff-1a81ac5587a67fc61fda38750c61ffa3R195

If you can tolerate some amount of delay from when new documents are added to when they are visible in a search, then maybe 60 seconds would be a good starting point?

PR #68, opened 2016-04-04
METRON-95: Create data purging script for HDFS

dlyle65535:
Add HDFS purge class with example 30 day age off example cronjob in deployment.

Tested in singlenode and ec2.

Travis checks passed in my repo.


deployment/roles/metron_streaming/defaults/main.yml
```diff
@@ -34,3 +39,6 @@ storm_topologies:
     - "{{ metron_directory }}/config/topologies/yaf/remote.yaml"
     - "{{ metron_directory }}/config/topologies/pcap/remote.yaml"
     - "{{ metron_directory }}/config/topologies/enrichment/remote.yaml"
+hdfs_retention_days: 30
+hdfs_bro_purge_cronjob: "yarn jar {{ metron_directory }}/lib/{{ metron_dataloads_jar }} org.apache.metron.dataloads.bulk.HDFSDataPruner -f {{ hdfs_url }} -g '/apps/metron/enrichment/indexed/bro_doc/*enrichment-*' -s $(date -d '{{ hdfs_retention_days }} days ago' +%m/%d/%Y) -n 1 >> /var/log/bro-purge/cron-bro-purge.log 2>&1"
```
cestella: Please make this an actual shell script and put it in an appropriate src/main/bash, not the direct call to yarn jar.

metron-streaming/Metron-DataLoads/pom.xml
```diff
@@ -149,7 +149,12 @@
             <version>${project.parent.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>1.10.16</version>
```
cestella: Ryan's PR from last week includes mockito as well and I believe there's a global mockito variable.  When you get around to merging master (after ryan's PR is merged), please make sure that this is referring to the global mockito version.

dlyle65535: Rgr, will do.

deployment/roles/metron_streaming/defaults/main.yml
dlyle65535: Okay, why is that preferred?

cestella: Because `/usr/metron/0.1BETA/bin/purge_hdfs.sh -f ... -g ... -s ...` is a lot easier to run outside of automated scripts than remembering `yarn jar /usr/metron/0.1BETA/lib/Metron-DataLoads-0.1BETA.jar org.apache.metron.dataloads.bulk.HDFSDataPruner -f ... -g ... -s ...`

dlyle65535: I'd prefer not to shadow the command line options that the main class exposes. That seems overly complicated to me. If you want me to wrap the yarn shell script in shell script that is shorter, I can see the value in that, but running yarn jar foo.jar main.class args is kind of a regular thing, no?

cestella: Not quite sure what shadow the command line options means.  Yeah, some people do explicitly call `yarn jar foo.jar main.class args`, but I suspect they'd rather be calling `my_script_name.sh args` every single time.  Before we had this for the zookeeper config utility, I'd grep the ansible scripts every time I needed to update the sourceconfigs in zookeeper.  When I needed to submit flux topologies (similar pattern) I'd have to grep ansible to remember the Flux command (similarly obtuse as the yarn jar command).  I don't find that to be very good user experience.  Another reason is that I can put $METRON_HOME/bin on the path and have the script autocomplete, whereas I'm never getting that main.class autocompleted. :)

dlyle65535: By shadowing the command line arguments I mean having to maintain the command line arguments in both a bash script and the java class. One of those two places to change one thing deals. 

How would you feel about a runnable jar?

cestella: No no, I don't mean maintaining the command line args in two places. heh, no, I just mean a shell script with yarn jar /usr/metron/0.1BETA/lib/Metron-DataLoads-0.1BETA.jar org.apache.metron.dataloads.bulk.HDFSDataPruner "$@" in it.

This fits the pattern in https://github.com/apache/incubator-metron/tree/master/metron-streaming/Metron-DataLoads/src/main/bash

cestella: also, can't do a runnable jar because the yarn jar command assembles the classpath for hadoop, which we don't bundle with the jar.

dlyle65535: That's okay, didn't really want to do a runnable jar anyway. :)

I can live with "$@", I'll get on it tomorrow.

dlyle65535:
I don't like how the master commits are listed in this PR. I'm going to close and resubmit once I figure this out.

PR #69, opened 2016-04-06
Metron 98 - Travis does not fail a build if the integration tests fail

merrimanr:
Travis will now fail if integration tests do not pass.  Also added logging config files to make the log data generated by the integration tests much less verbose.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,4 @@ target
 *.classpath
 *.settings
 *hbase-site.xml
-log4j.properties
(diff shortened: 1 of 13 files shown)


nickwallen:
Looks good.  Mind squashing your commits?

cestella:
+1 and I'm itching to get travis back and in working order.  Just for posterity, the integration tests were failing and not reporting the failure back up to travis.

PR #72, opened 2016-04-11
METRON-107: Altered travis config to include both maven targets in th…

merrimanr:
Commands need to be run so that a failure in either mvn apache-rat:check or mvn integration-test package returns the right error code


nickwallen:
+1 Change successfully detected missing license files

PR #76, opened 2016-04-14
METRON-110 Create lightweight packet capture script

nickwallen:
The fast packet capture process is difficult to manage and deploy in test and development environments.  There are difficulties in integrating it with the existing deployment environments which makes it difficult to test and demo functionality that is dependent on packet capture data.  Need to create a lightweight packet capture process that replicates what the fast packet capture process does.  This would effectively be a replacement for Pycapa; one that is compatible with the Apache license.  

Pycapa has two primary runtime modes.

### Producer Mode

Pycapa can be configured to capture packets from a network interface and then forward those packets to a Kafka topic.  The following example will capture packets from the `eth0` network interface and forward those to a Kafka topic called `pcap` running on `localhost`.

```
pycapa --producer --kafka localhost:9092 --topic pcap -i eth0
```

To output debug messages every 100 captured packets, run the following.

```
pycapa --producer --kafka localhost:9092 --topic pcap -i eth0 --debug 100
```

### Consumer Mode

Pycapa can be configured to consume packets from a Kafka topic and then write those packets to a [libpcap-compliant file](https://wiki.wireshark.org/Development/LibpcapFileFormat).  To read 100 packets from a kafka topic and then write those to a [libpcap-compliant file](https://wiki.wireshark.org/Development/LibpcapFileFormat), run the following command.  The file `out.pcap` can then be opened with a tool such as Wireshark for further validation.

```
pycapa --consumer --kafka localhost:9092 --topic pcap --n 100 > out.pcap
```

To consume packets from Kafka continuously and print debug messages every 10 packets, run the following command.  

```
pycapa --consumer --kafka localhost:9092 --topic pcap --debug 10
```


nickwallen:
@cestella I was able to use this tool to build valid libpcap-compliant files from packet data in Kafka.  This was successful when using either pycapa or the 'fast packet capture' code as the source of packet data.  This kind of test has not been automated, but at least now I have the means to automate it.

Also, this should help show exactly what kind of headers we need added by your 'pcap extraction topology' effort.

cestella:
I don't think the license for argparse is kosher. Please investigate and
remove if inappropriate or adjust the license file if that's required.

nickwallen:
Argparse is part of the standard library and does not need to be separately installed.  Removed the need to install it separately via pip.  Python standard lib is license compatible with Apache.

cestella:
Ahh I see I thought there was some weirdness with GPL, but as I look it's
Python licensed now. It only became part of the standard package as of 2.7
and I think ambari requires 2.6...is there any issue installing this on the
vagrant vm?

Sorry I'm on my phone since wifi on my computer is slightly busted and it's
a bit hard to investigate fully or type easily ;)

nickwallen:
Interesting.  I manually deployed on CentOS 6 with the latest change and it worked.  But it appears `argparse` was already installed by something else.  This node went through a partial Metron install, but no pycapa.

```
[vagrant@node1 pycapa]$ ls -l /usr/lib/python2.6/site-packages/argparse*
-rw-r--r--. 1 root root 87791 Mar 31  2011 /usr/lib/python2.6/site-packages/argparse.py
-rw-r--r--. 1 root root 66661 Jul 23  2015 /usr/lib/python2.6/site-packages/argparse.pyc
-rw-r--r--. 1 root root 66491 Jul 23  2015 /usr/lib/python2.6/site-packages/argparse.pyo

/usr/lib/python2.6/site-packages/argparse-1.2.1-py2.6.egg-info:
total 16
-rw-r--r--. 1 root root    1 Jul 23  2015 dependency_links.txt
-rw-r--r--. 1 root root 2927 Jul 23  2015 PKG-INFO
-rw-r--r--. 1 root root  250 Jul 23  2015 SOURCES.txt
-rw-r--r--. 1 root root    9 Jul 23  2015 top_level.txt
```

Whether it is standard lib or installed via pip, the license issue is moot now right?  So I could just revert the last commit since I know the original code worked on 2.6 and 2.7?

cestella:
Yeah the license is fine. We are going to have it installed in the vagrant
box as part of ansible, right?

nickwallen:
Right.  Next step is to update `deployment/roles/pycapa` to use this instead of OpenSOC's pycapa.

cestella:
Sounds good!

nickwallen:
Ok good - Leaving argparse in the requirements.txt.

Does "Sound good" == +1?

cestella:
Yes

PR #79, opened 2016-04-15
METRON-96: Create data purging script for ES Index

dlyle65535:
Here's the initial version- I have a few longer running tests in progress, but I wanted to give everyone a chance to look it over while they are running.

##  Left to do: 
- [X] Singlenode Vagrant - Passed, new jobs run, data reaches the UI.
- [X] Multi Node Soak Test - Passed, deployment creates correct crontab entries and logfiles.
- [X] Code Reviews from interested parties


merrimanr:
I noticed there are a coupe sample config json files (bro.json and global.json) being added to Metron-Common.  There are already sample configs in Metron-Testing.  Are these needed?

dlyle65535:
Yes, the global.json is different, targeted at the exact test I am running. I don't need the bro.json, but unfortunately the config loader dies without a sensors directory and git wants it to be populated. 
I'm going to be putting up a Jira on that behavior shortly, once it's completed, that file can (and the enclosing directory) can go away.

merrimanr:
+1

metron-streaming/Metron-DataLoads/pom.xml
```diff
@@ -167,56 +225,14 @@
...
                 <configuration>
-                    <!-- Skip the default running of this plug-in (or everything is run twice...see below) -->
-                    <argLine>-Xmx2048m -XX:MaxPermSize=256m -Djava.util.logging.config.file=logging.properties</argLine>
+                    <argLine>-Xmx2048m -XX:MaxPermSize=256m -XX:-UseSplitVerifier</argLine>
                     <skip>true</skip>
-                    <!-- Show 100% of the lines from the stack trace (doesn't work) -->
                     <trimStackTrace>false</trimStackTrace>
-
                 </configuration>
-                <executions>
```
cestella: So, I don't mind this, but this change will make the "integration tests" run as part of the unit test cycle and people have gotten touchy in the past due to the length that it adds to mvn package.  I'm generally for it as it ensures we get the full suite, but I wanted to make sure it was intended and we're ok with it in general.

cestella:
+1, looks great.

metron-streaming/Metron-DataLoads/pom.xml
dlyle65535: I don't mean it to. The effective pom gets the executions info from the top level pom.

dlyle65535: :)

PR #86, opened 2016-04-22
METRON-110 Add Metron Pycapa to Standard Deployment 

nickwallen:
Metron deployment now deploys Metron's Pycapa instead of the legacy OpenSOC Pycapa.

### DIFF
--- a/metron-deployment/roles/pycapa/defaults/main.yml
+++ b/metron-deployment/roles/pycapa/defaults/main.yml
@@ -15,9 +15,10 @@
 #  limitations under the License.
 #
 ---
-pycapa_repo: https://github.com/OpenSOC/pycapa.git
 pycapa_home: /usr/local/pycapa
-pycapa_main: "{{ pycapa_home }}/pycapa/pycapa_cli.py"
+pycapa_venv: pycapa-venv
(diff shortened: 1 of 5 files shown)


deployment/roles/pycapa/tasks/dependencies.yml
```diff
@@ -15,20 +15,22 @@
...
-  yum:
-    name: "{{item}}"
+- name: Install dependencies
+  yum: name={{ item }}
   with_items:
-    - python
-    - tcpdump
-    - git
+    - "@Development tools"
+    - python27
```
cestella: Is that coinciding ok with ambari?  I know it requires 2.6.  It must be since you verified against the single node vagrant.

cestella: Oh, this puts python 2.7 in /opt/rh/python27, doesn't it?  That should get out of the way of the default 2.6 installed.

nickwallen: Yes, this installs python 2.7 at a different location and does not change the default 2.6 install.

cestella:
+1

merrimanr:
+1

PR #89, opened 2016-04-26
METRON-119 Move PCAP infrastructure from HBase

cestella:
As it stands, the existing approach to handling PCAP data has some issues handling high volume packet capture data.  With the advent of a DPDK plugin for capturing packet data, we are going to hit some limitations on the  throughput of consumption if we continue to try to push packet data into HBase at line-speed.

Furthermore, storing PCAP data into HBase limits the range of filter queries that we can perform (i.e. only those expressible within the key).  As of now, we require all fields to be present (source IP/port, destination IP/port and protocol), rather than allowing any wildcards.

To address these issues, we should create a higher performance topology which attaches the appropriate header to the raw packet and timestamp read from Kafka (as placed onto kafka by the packet capture sensor) and appends this packet to a sequence file in HDFS.  The sequence file will be rolled based on number of packets or time (e.g. 1 hrs worth of packets in a given sequence file).

On the query side, we should adjust the middle tier service layer to start a MR job on the appropriate set of sequence files to filter out the appropriate packets.  NOTE: the UI modifications to make this reasonable for the end-user will need to be done in a follow-on JIRA.

In order to test this PR, I would suggest doing the following as the "happy path":

1. Install the pycapa library & utility via instructions [here](https://github.com/apache/incubator-metron/tree/master/metron-sensors/pycapa)
2. (if using singlenode vagrant) Kill the enrichment and sensor topologies via `for i in bro enrichment yaf snort;do storm kill $i;done`
3. Start the pcap topology via `/usr/metron/0.1BETA/bin/start_pcap_topology.sh`
4. Start the pycapa packet capture producer on eth1 via `/usr/bin/pycapa --producer --topic pcap -i eth1 -k node1:6667`
5. Watch the topology in the [Storm UI](http://node1:8744/index.html) and kill the packet capture utility from before when the number of packets ingested is over 1k.
6. Ensure that at at least 2 files exist on HDFS by running `hadoop fs -ls /apps/metron/pcap`
7. Choose a file (denoted by $FILE) and dump a few of the contents using the `pcap_inspector` utility via `/usr/metron/0.1BETA/bin/pcap_inspector.sh -i $FILE -n 5`
8. Choose one of the lines and note the source ip/port and dest ip/port
9. Go to the kibana app at [http://node1:5000](http://node1:5000) on the singlenode vagrant (ymmv on ec2) and input that query in the kibana PCAP panel.
10. Wait patiently while the MR job completes and the results are sent back in the form of a valid PCAP payload suitable for opening in wireshark
11. Open in wireshark to ensure the payload is valid.

If the payload is not valid PCAP, then please look at the [job history](http://node1:19888/jobhistory) and note the reason for job failure if any.


cestella:
Oh, please see the new updated docs for
* [Metron PCAP Service](https://github.com/cestella/incubator-metron/tree/pcap_extraction_topology/metron-streaming/Metron-Pcap_Service)
* [Metron PCAP Backend](https://github.com/cestella/incubator-metron/tree/pcap_extraction_topology/metron-streaming/Metron-Pcap_Backend)

The backend docs should answer questions as to just how the topology is architected (it's spout-only) and how it's configured.

kidmose:
The two links in the latest comment produces a 404. 
Something I miss or did you maybe forget to add the docs, @cestella ?

dlyle65535:
Can we rebase this post METRON-113 being pushed to master? Having both changesets makes it difficult to isolate your changes.

cestella:
I'm not done merging 113 into this PR branch yet. Took me all day yesterday
and I'm still testing. I'll let you know when I'm done. That should make
things easier.

cestella:
In retrospect, this PR is a hot mess.  Let me close and reopen once merging and validation has been completed with a cleaner commit log.

PR #94, opened 2016-04-28
METRON-123 Pycapa requires 'kafka_broker_url'

nickwallen:
Pycapa needs to know the URL of the Kafka Broker to send packet data to. During a standard deployment, such as single node vagrant or amazon-ec2, this URL needs to be retrieved from Ambari.  When using Pycapa to test other functionality, like fastcapa, Ambari is not used.  In these cases the URL need to be specified by the user instead of being retrieved from Ambari.

The Pycapa role at `metron-deployment/roles/pycapa` previously had a dependency on `ambari_gather_facts` through which it retrieved the Kafka Broker URL. This dependency was removed in METRON-110 which introduced this bug.

### Changes
- Changed the `ambari_gather_facts` Ansible role so that it only retrieves information from Ambari if the information is not already known.
- Updated Pycapa so that the service start script can be installed or not depending on a variable.  As a sensor, we need the start script installed.  As a testing tool, we do not need the start script installed.
- Update Fastcapa test environment so extra dummy variables are provided instead of having the process reach out to a nonexistent Ambari server.

### Validation
Running Pycapa in Single Node Vagant provided the "positive" test, using Ambari to retrieve value, and using the Fastcapa Test Environment provided the "negative" test, using a user-provided value.
- Single Node Vagrant Environment `cd metron-deploymet/vagrant/singlenode-vagrant; vagrant up`
- Fastcapa Test Environment `cd metron-deployment/vagrant/fastcapa-vagrant; vagrant up`

### DIFF
--- a/metron-deployment/roles/ambari_gather_facts/tasks/main.yml
+++ b/metron-deployment/roles/ambari_gather_facts/tasks/main.yml
@@ -15,7 +15,10 @@
 #  limitations under the License.
 #
 ---
-- name: Ambari rest get cluster name
+#
+# cluster_name
+#
+- name: "Ask Ambari: cluster_name"
   uri:
     url: "http://{{ groups.ambari_master[0] }}:{{ ambari_port }}/api/v1/clusters"
     user: "{{ ambari_user }}"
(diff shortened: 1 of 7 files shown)


merrimanr:
Looks good to me +1

cestella:
+1

PR #96, opened 2016-04-29
Metron 122

merrimanr:
During the creation of the blogs detailing how to add parsers, several issues surfaced.  This pull request fixes those issues.  Includes:
- a couple new parsers (bluecoat and squid) along with unit and integration tests
- change to the grokparsertest to make it reusable for other grok based parsers
- changes to the metron-parser jar to include site files and hadoop dependency
- additional ansible scripts to rerun the metron-streaming role for the purpose of pushing out parser changes
- fix to the enrichment topology to pass along data to elasticsearch when no enrichments have been configured


dlyle65535:
tl;dr I'm -1 on a small piece of this changeset, see below for why.

Looking over this PR, it seems to be an aggregation of 4 JIRAs, METRON-122, METRON-91 (expose tags) and 2 new JIRAs, one for Bluecoat and 1 to fix the enrichment topology.

It's a bummer because I'm +1 on everything but how exposing tags/skip_tags was handled. So although everything else looks great, I have to be -1 on the whole changeset.

I'm -1 on that because it operates outside of the Vagrant paradigm and requires the user to enter the Vagrant password. Here's a more vagrantish solution:

Add:
```
require 'getoptlong'

opts = GetoptLong.new(
  [ '--ansible-tags', GetoptLong::OPTIONAL_ARGUMENT ],
  [ '--ansible-skip-tags', GetoptLong::OPTIONAL_ARGUMENT ]
)

ansibleTags=''
ansibleSkipTags=''

opts.each do |opt, arg|
  case opt
    when '--ansible-tags'
      ansibleTags=arg
    when '--ansible-skip-tags'
      ansibleSkipTags=arg
  end
end

puts " Running with ansible-tags: " + ansibleTags.split(",").to_s if ansibleTags != ''
puts " Running with ansible-skip-tags: " + ansibleSkipTags.split(",").to_s if ansibleSkipTags != ''
```
To the top of the Vagrantfile (under the license).

Then make the config loop look like this:

```
  # provisioning
  config.vm.provision :ansible do |ansible|
    ansible.playbook = "../../playbooks/metron_full_install.yml"
    ansible.sudo = true
    ansible.tags = ansibleTags.split(",") if ansibleTags != ''
    ansible.skip_tags = ansibleSkipTags.split(",") if ansibleSkipTags != ''
    ansible.inventory_path = "../../inventory/singlenode-vagrant"
  end
```

That should (we'd have to test it) allow us to remove the new ansible.cfg unless you wanted to keep the custom control_path. You could also get rid of run_ansible_role.sh and run_enrichment_role.sh (it would be ``` vagrant --ansible-tags=enrichment provsion ``` )

merrimanr:
@dlyle65535 thanks for the suggestions, they make sense and I will add them in.  

However, I think we should keep the scripts and change them to call the vagrant command.  The idea is for someone to be able to rerun a role without having to remember or look up the vagrant syntax.  From my perspective running "run_enrichment_role.sh" is much more intuitive than running "vagrant --ansible-tags=enrichment provision".  The whole purpose of adding these scripts is to make it easy for a developer to test and deploy changes without needing detailed knowledge of how metron-deployment works.

merrimanr:
@dlyle65535 I added the proposed changes and it almost works.  Running an individual role worked as expected but when I run "vagrant destroy -f" I get:  

Message: GetoptLong::InvalidOption: invalid option -- f.  

We probably want to still allow other vagrant options to be passed in.  Any ideas?

merrimanr:
I just thought of another potential issue with this approach.  If we decide to publish a pre-built image of the single node vagrant machine, running this script won't work if the private key file generated by vagrant up is missing.

cestella:
@merrimanr Hmm, I liked that feature (being able to run individual roles against either a vagrant image or a prebuilt image).  That will be useful, I think for letting people get metron and try stuff out on it.

dlyle65535:
No specific ideas on the -f issue. I can work on it a bit Sunday or Monday. Of course, everything should work when it's over including a pre-built snapshot. I'd just prefer we use the existing Vagrant capability for VM Provider -> Provisioner handoff. 

@cestella - I think we're good on a generic solution, just use the regular ansible-playbook command with --tags and/or --skip-tags. I don't really see any reason to abstract that. Not abstracting it has a lot of advantages: it's well documented by Ansible, anyone with a background in Ansible already knows about it, we don't have to maintain it- stuff like that. In my opinion, our Vagrant provisioning is the oddball here as we currently have to modify the Vagrantfile to leverage tags. That's why I'm recommending that we do something specific in the Vagrantfile to expose those arguments.

All that said, I'm happy to include the script as well. Though I don't agree that run_enrichment_role.sh is more intuitive than doing it using Vagrant and/or ansible-playbook. Mostly for the reasons stated above.  Though, I probably should have suggested the arguments for Vagrant be --tags and --skip-tags to make it consistent.

cestella:
I'm not sure that `ansible-playbook -i metron-deployment/inventory/singlenode-vagrant/hosts metron-deployment/playbooks/metron_install.yml --tags enrichment -u vagrant -k` is more intuitive than `run_enrichment.sh` for the relatively common task of redeploying.

There is enough differing opinion here, though, that I think perhaps the prudent thing to do is punt any vagrant changes to another PR.  Can we just keep the `run_enrichment.sh

dlyle65535:
Heh, yeah, we're in full agreement there. That command looks horrible. Glad I didn't suggest it.

I did suggest ```vagrant --tags=enrichment provision``` which, I think, is at least as intuitive as run_enrichment.sh and perhaps more so if one has a background in Vagrant and/or Ansible.

I also think that ```ansible-playbook -i ec2.py playbook.yml --tags=enrichment``` is at least as intuitive and perhaps more so if one has a background in Ansible. It's an up arrow-ctrl-e-few keystrokes away. :)

Both of those have the added advantages of supporting other tags as well and in the case of ansible-playbook are fully documented already.

That said- I'm _still_ not against the script, I just don't see it as a viable replacement for exposing tags and skip-tags.

dlyle65535:
Upon further review- I think I have something working that doesn't need the extra '--'.

What would you guys think of the following?

- Remove the singlenode-vagrant stuff from this PR
- I'll push what I've been able to work up for that under METRON-91 (unless @merrimanr has sorted it, then he can)
- For @merrimanr 's concern about pre-built dev images, let's see what it would take to make something like that vagrant upable (I think it's just adding a push provider to the Vagrantfile and setting up an Atlas account) so whatever we do will work with all singlenode dev-type images

merrimanr:
@dlyle65535, I think that's a good plan but I agree with @cestella that we should punt it to a separate Jira.  Sounds like there are some unknowns and things we need to work out to arrive at a solution everyone is comfortable with.  The script included in this pull request works fine and is valuable for testing new parsers in the single node VM.  Any harm in leaving it in until we come up with something better?

dlyle65535:
With @merrimanr 's latest commit, we can consider the METRON-91 stuff separately, so I'm removing my -1 vote. Thanks!

cestella:
+1

PR #104, opened 2016-05-04
METRON-133: ParserBolt is not failing tuples when parsing fails

merrimanr:
Added tuple fail to ParserBolt on a failed parse.  Also added verbose logging in the GrokParser.


dlyle65535:
+1 - ran it up in Vagrant on top of METRON-130. Ran without issues and logging worked well. Looks great!

PR #106, opened 2016-05-05
METRON-135: Ambari Start needs to set state to 'restarted'

dlyle65535:
Set state to restarted and tested on single node vagrant.


nickwallen:
+1

cestella:
Big +1, applied this change last night and it worked like a charm on SNV

PR #108, opened 2016-05-06
METRON-141: The ability to do threat triage

cestella:
We have the ability to mark messages as part of the enrichment topology as threat alerts, but we have no ability to prioritize those alerts.

We should allow for the prioritization of messages that have some threat intelligence alert via a scoring mechanism. 

The general idea is for each message from each sensor:
1. Map threat conditions as defined by a lightweight query language to threat levels (a number).
2. More than one condition may match a given message, so allow for the choice of aggregation functions.
3. If the aggregated score is positive, then add a field (`threat.triage.level`) with the level to the JSON message.

The above configuration is done at the per-sensor level.

As part of this PR, the following were added out of necessity:
* A query language DSL which is used in defining the conditions which map to threat levels.  Docs at [here](https://github.com/cestella/incubator-metron/tree/METRON-141/metron-platform/metron-common#query-language)
* The addition of a `triageConfig` section to the sensor enrichment config.  Description at [here](https://github.com/cestella/incubator-metron/tree/METRON-141/metron-platform/metron-common#the-threatintel-configuration)
* Actual documentation of our enrichment configuration.  Documentation at [here](https://github.com/cestella/incubator-metron/tree/METRON-141/metron-platform/metron-common#enrichment-configuration)
* A rewrite of the `zk_load_configs.sh` utility to support both pushing and pulling configs (the threat triage rules are part of the configs and until we have a UI, we need to add them).  Docs at [here](https://github.com/cestella/incubator-metron/tree/METRON-141/metron-platform/metron-common#management-utility)


metron-platform/metron-common/src/test/java/org/apache/metron/common/cli/ConfigurationsUtilsTest.java
```diff
@@ -74,7 +74,7 @@ public void test() throws Exception {
 
   }
 
-  @Test
+  /*@Test
```
merrimanr: Should we remove this instead of commenting out?

metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/EnrichmentJoinBolt.java
```diff
@@ -84,7 +88,7 @@ public JSONObject joinMessages(Map<String, JSONObject> streamMessageMap) {
     if(sourceType != null) {
       SensorEnrichmentConfig config = configurations.getSensorEnrichmentConfig(sourceType);
       if (config != null) {
-        return config.getEnrichmentFieldMap();
+        return config.getEnrichment().getFieldMap();
```
merrimanr: Does this introduce the possibility of a null pointer exception?

merrimanr:
+1

metron-platform/metron-common/src/test/java/org/apache/metron/common/cli/ConfigurationsUtilsTest.java
cestella: Whoops, absolutely right!  Will do.

metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/EnrichmentJoinBolt.java
cestella: I think that the enrichment variable is initialized, but a null check is never a bad idea.

dlyle65535:
+1 this is really great!

PR #109, opened 2016-05-07
METRON-140 handle empty line in CSVExtractor

mmiklavc:
https://issues.apache.org/jira/browse/METRON-140

### DIFF
--- a/metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/extractor/csv/CSVExtractor.java
+++ b/metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/extractor/csv/CSVExtractor.java
@@ -85,7 +85,7 @@ public class CSVExtractor implements Extractor {
 
   private boolean ignore(String line) {
     if(null == line) {
-      return false;
+      return true;
     }
     String trimmedLine = line.trim();
     return trimmedLine.startsWith("#") || isEmpty(trimmedLine);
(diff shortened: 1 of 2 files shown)


metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/extractor/csv/CSVExtractor.java
```diff
@@ -82,6 +83,14 @@ public LookupConverter getConverter() {
     return Arrays.asList(new LookupKV(key, converter.toValue(values)));
   }
 
+  private boolean ignore(String line) {
+    if(null == line) {
+      return false;
```
cestella: I think you mean to return "true" here

metron-platform/metron-data-management/src/test/java/org/apache/metron/dataloads/extractor/csv/CSVExtractorTest.java
```diff
@@ -91,5 +85,9 @@ public void validate(ExtractorHandler handler) throws IOException {
       Iterable<LookupKV> results = handler.getExtractor().extract("#google.com,1.0,foo");
       Assert.assertEquals(0, Iterables.size(results));
     }
+    {
+      Iterable<LookupKV> results = handler.getExtractor().extract("");
+      Assert.assertEquals(0, Iterables.size(results));
+    }
```
cestella: Cool!  Can you add a case for null and for a "  "?

cestella:
This looks great outside of the couple of things I mentioned.  Definitely like the null check.  +1 imminent if you get those changes.

mmiklavc:
I also tested this on single-node Vagrant

cestella:
+1

dlyle65535:
+1

PR #111, opened 2016-05-07
METRON-140 CSVExtractor gives index out of bounds error and added test

jjonez:


cestella:
Looks like we got 2 for the same JIRA, this PR and PR 109.  Cool!  It's great when people contribute. :)

PR #113, opened 2016-05-09
Metron 136 logrotate pcapservice

jjonez:
I tested this by starting the full-dev-platform vagrant image. The logrotate.d config file was created. I ran logrotate with the -f option (force) to test the configuration and it worked properly.

The configuration is consistent with the other logrotate configs in metron, but I have a couple of questions, which perhaps should be posted elsewhere. These questions apply to most of the default Metron logrotate configurations:

#1 - Should the pcap logs  be in their own directory like most other metron logs instead of /var/log/metron_pcapservice.log?

#2 - Should we change to compress (gzip)?

#3 - Regarding all logrotate configs, should we change them to use dateext? Currently it changes files like this:
/var/log/metron_pcapservice.log.1
/var/log/metron_pcapservice.log.2

Where, using dateext makes them:
/var/log/metron_pcapservice.log.20160507
/var/log/metron_pcapservice.log.20160508

The disadvantage to this is that even if you try to force the rotate  (-f), I don't think it will rotate because the date it would rotate to already exists. With the number, if just adds a new number.

### DIFF
--- a/metron-deployment/roles/metron_pcapservice/defaults/main.yml
+++ b/metron-deployment/roles/metron_pcapservice/defaults/main.yml
@@ -24,3 +24,5 @@ pcapservice_port: 8081
 hbase_config_path: "/etc/hbase/conf"
 query_hdfs_path: "/tmp"
 pcap_hdfs_path: "/apps/metron/pcap"
+metron_pcapservice_logrotate_frequency: daily
+metron_pcapservice_logrotate_retention: 30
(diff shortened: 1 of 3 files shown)


cestella:
This looks good to me, but I'd very much like @dlyle65535  to chime in.

dlyle65535:
+1. Thanks for the contribution and the suggestions. Good stuff.

nickwallen:
@jjonez We need your email address to merge your PR. You do not have one associated with your Github account, that I can see at least.

jjonez:
@nickwallen, Thanks. I just updated my profile and made my email public. And, I think I fixed the author for future patches as well.

PR #114, opened 2016-05-10
Updating document to have the correct script name, instructions on in…

apsaltis:
Updated the README to correct the following:

1. The script name in the directory is launch_image.sh not launch_dev_image.sh
2. The ports for the Ambari and Metron UI were flipped

Also added instructions for installing vagrant hostmanager, if the user does a fresh vagrant install this is not installed by default and they will get errors when they run the launch_image script.


dlyle65535:
+1 looks good, thanks!

nickwallen:
+1

PR #115, opened 2016-05-10
[METRON-150] Adding WebSphere parser, unit tests, and integration tests

DomenicPuzio:
This PR is for [METRON-150](https://issues.apache.org/jira/browse/METRON-150). It satisfies the specified requirements of the WebSphere parser, passes unit tests, and passes integration tests. The parser uses Grok to pull out the fields present in all message types and uses Java to handle message-specific parsing.


merrimanr:
@DomenicPuzio, thanks this looks pretty good.  Are you able to leverage the Grok parser that already exists (org.apache.metron.parsers.GrokParser) or is there some custom behavior needed?  I see a lot of overlap between the custom Java Grok parser you wrote and what's currently there.  If the current Grok parser is not flexible enough I think we should either update it to make it more flexible or extend it to provide custom behavior (rather than writing a new one).  Thoughts?

metron-platform/metron-integration-test/src/main/resources/sample/config/sensors/websphere.json
```diff
@@ -0,0 +1,22 @@
+{
+  "index": "websphere",
+  "batchSize": 1,
+  "enrichmentFieldMap":
```
cestella: So, this structure got refactored a bit as of the last PR.  Check out the example [here](https://github.com/apache/incubator-metron/tree/master/metron-platform/metron-common#example)  The upshot is that instead of `enrichmentFieldMap` and `threatIntelFieldMap` there are `enrichment` and `threatIntel` sections and `fieldMap` is in both.

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/websphere/GrokWebSphereParserTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.json.simple.JSONObject;
+import org.junit.Test;
+
+public class GrokWebSphereParserTest {
+	
+	@Test
+	public void testParseLoginLine() throws Exception {
+		
+		//Set up parser, parse message
+		GrokWebSphereParser parser = new GrokWebSphereParser();
```
cestella: We probably want a negative case for each of these scenarios (login, logout, rbm and other).  What happens when you try to parse malformed lines or an empty line..don't want it to crash the topology.

cestella:
Thanks so much for the contribution.  This is going to be great when it gets in!

DomenicPuzio:
@cestella, thank you so much for the feedback! I will get to work on fixing the 'websphere.json' file that was causing some failed tests. I agree that a negative case in the test class is a really good call.

@merrimanr, I agree that there is a good amount of overlap between this parser and the GrokParser class. There was some custom behavior needed to handle different log types, so I used GrokAsaParser (which also extends BasicParser) as a model. Off the top of my head, there are several sources that have different log types within a single source - Infoblox DNS, Checkpoint Firewall, Big IP VPN - so we should determine a strategy to address those. I like your idea of making the GrokParser more flexible so that we can easily extend it and add custom behavior for sources like WebSphere. Is this something that I should work to address in this PR or in a future refactoring? Either way, I am more than happy to work on that!

merrimanr:
@DomenicPuzio, I think we should address it now since there are likely more Grok parsers being contributed soon.  I submitted a pull request against your branch with an example.

DomenicPuzio:
@merrimanr, thank you so much for the feedback and assistance. I have incorporated your changes! I have also added negative test cases (trying to parse empty or malformed lines) to help ensure that the topology will not crash when attempting to parse incorrectly formatted messages.

merrimanr:
Looks good to me.  Nice job @DomenicPuzio.  +1

cestella:
Great Job!  Have you tested this on the full-dev-vagrant at all?

cestella:
Ok, tested this.  Sorry it took so long (vagrant drama).

Ok, so this is what I did to test this.
Spin up the full-dev-vagrant:
1. Do a build by running `mvn clean integration-test` from the
  `metron-platform` directory
2. From the `metron-deployment/vagrant/full-dev-platform` directory run `vagrant up`
   * Note: If you already have a vagrant machine running and you just want to
     redeploy code, you should be able to do `vagrant provision --tags enrichment`

Log into the full-dev-vagrant and do the following:

* Optionally, stop some things to give myself enough memory headroom:
   * Kill every running parser topology via the [storm UI](http://node1:8744/index.html) to proceed (that VM is pretty packed atm).  
   * Stop pcap replay via `/etc/init.d/pcap-replay stop`
   * Kill bro via `/usr/local/bin/broctl` and type `stop` and then `exit` at the broctl prompt
   * Marvel at the lack of fans running on your laptop
* Create the kafka topic for the sensor: `/usr/hdp/current/kafka-broker/bin//kafka-topics.sh --zookeeper localhost:2181 --create --topic websphere --partitions 1 --replication-factor 1`
* Create text file with sample data in it called `WebsphereOutput.txt`.  This is essentially just [WebsphereOutput.txt](https://raw.githubusercontent.com/DomenicPuzio/incubator-metron/af50623dcb764fda0281b3657c0f40c993c958f3/metron-platform/metron-integration-test/src/main/resources/sample/data/SampleInput/WebsphereOutput.txt)
* Start the parser topology using `/usr/metron/0.1BETA/bin/start_parser_topology.sh`, in this case `/usr/metron/0.1BETA/bin/start_parser_topology.sh websphere`
* Add your data to the kafka queue you created earlier via `cat WebsphereOutput.txt | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic websphere`
* Wait a minute and then check the elasticsearch head plugin (if you haven't installed it, you can via `/usr/share/elasticsearch/bin/plugin -install mobz/elasticsearch-head`).  
   * Browse to the `Browser` tab and click on the index named `${sensor_name}_index_${DATE}` (in my case: `websphere_index_2016.05.13.13`
   * You should see some messages.  Click on each one and inspect the JSON to ensure it looks right.

For future parsers, this is a good template for acceptance testing.  You
shouldn't have to go through a full maven provisioning each time if you
keep the VM around and just run `vagrant provision -tags enrichment`.

cestella:
+1

DomenicPuzio:
@merrimanr and @cestella, thank you SO MUCH for the help! I really appreciate it! I will certainly follow the steps to test on vagrant for future parsers.

I'm very excited to be a contributor, and I'm already working on my next parser!

PR #116, opened 2016-05-11
Metron 146 topology workers

jjonez:
This allows you to configure the number of workers in the enrichment topology. The variable name is pretty long, for the sake of being verbose (storm.topology.enrichment.workers, or in Ansible it is storm_topology_enrichment_workers). 

I tested this using the vagrant/quick-dev-platform image. After starting, I ran the ./run_enrichment_role.sh script to start the enrichment topology. I verified that the file has the configured value and the topology started correctly. You can see in the ansible output that it tried to replaced the variable here but the default value is the same since as the original value. I can only run this on my laptop:
'''
TASK [metron_streaming : Configure Metron Solr topology] ***********************

ok: [node1] => (item={u'regexp': u'storm.topology.enrichment.workers=', u'line': u'storm.topology.enrichment.workers=1'})
'''

If someone could test changing the value on a real cluster that would be great. I don't have the hardware for it.


cestella:
Ah, ok, so this is failing the integration tests because you're not setting the topology.workers property.  You can either set it
* In [this](https://github.com/apache/incubator-metron/blob/master/metron-platform/metron-integration-test/src/main/java/org/apache/metron/integration/EnrichmentIntegrationTest.java#L151) block of code
* In [this](https://github.com/apache/incubator-metron/blob/master/metron-platform/metron-elasticsearch/src/main/config/elasticsearch.properties) and [this](https://github.com/apache/incubator-metron/blob/master/metron-platform/metron-solr/src/main/config/solr.properties) file

The first will just fix it for the integration test.  The second will provide a default for the deployed properties.  That being said, it should be being set via ansible, so providing a default isn't strictly required.

I don't have a strong opinion either way.

jjonez:
I had set the value in those files. The problem was that I set it as String in the Integration test rather than as an integer. 

I also had merged with master, but the build failed on something that I *think* is totally unrelated to my changes, so I reverted the merge. [Here](https://travis-ci.org/apache/incubator-metron/builds/130322731) is the failed build.

metron-platform/metron-integration-test/src/main/java/org/apache/metron/integration/EnrichmentIntegrationTest.java
```diff
@@ -164,6 +164,7 @@ public void test() throws Exception {
       setProperty("es.ip", "localhost");
       setProperty("index.date.format", dateFormat);
       setProperty("index.hdfs.output", hdfsDir);
+      setProperty("storm.topology.enrichment.workers", 1);
```
dlyle65535: I think this is what's failing the test. IIRC, that's setProperty(string, string).

metron-deployment/roles/metron_streaming/tasks/main.yml
```diff
@@ -130,6 +130,8 @@
     - { regexp: "bolt.hdfs.rotation.policy=", line: "bolt.hdfs.rotation.policy={{ metron_hdfs_rotation_policy }}" }
     - { regexp: "bolt.hdfs.rotation.policy.count=", line: "bolt.hdfs.rotation.policy.count={{ metron_hdfs_rotation_policy_count}}" }
     - { regexp: "bolt.hdfs.rotation.policy.units=", line: "bolt.hdfs.rotation.policy.units={{ metron_hdfs_rotation_policy_units }}" }
+    - { regexp: "storm.topology.enrichment.workers=", 
+          line: "storm.topology.enrichment.workers={{ storm_topology_enrichment_workers if storm_topology_enrichment_workers is defined else 1 }}" }
```
dlyle65535: Rather that use ```is defined``` here, you can set the default in metron_streaming/defaults/main.yml by adding ```storm.topology.enrichment.workers: 1``` somewhere in the file.

nickwallen:
@jjonez Are you good with closing this PR?  It has been a good amount of time since its been touched.  Thanks

jjonez:
Yes. Please close it. Thanks

nickwallen:
I am unable to.  You should be able to close it though.  Thanks.

jjonez:
Sorry. I'll take care of it.

PR #119, opened 2016-05-12
METRON-155 Added query filtering capability for PCAP via Metron REST API

mmiklavc:
https://issues.apache.org/jira/browse/METRON-155

Have not run the Vagrant deploy yet, but this is big PR so i wanted to get this reviewed asap.
- Added PCAP filter by query.
- Use Java's Stream, Function, and Predicate impls.
- Added QueryPcapFilterTest
- Added tests in REST api for new query option
- Added integration tests to PCAP topology
- Fixed bug in QueryParser - handle empty string as pass-through impl like WHERE TRUE


metron-platform/metron-api/src/main/java/org/apache/metron/pcapservice/PcapReceiverImplRestEasy.java
```diff
@@ -97,6 +97,66 @@ private static boolean isValidPort(String port) {
     }
     return false;
   }
+	  /*
+	   * (non-Javadoc)
+	   *
+	   * @see
+	   * com.cisco.opensoc.hbase.client.IPcapReceiver#getPcapsByIdentifiers(java.lang
```
cestella: Could you change the javadocs here to reflect the method?

mmiklavc: Looks like older leftover auto-generated javadoc - I'll revise the packaging and make a proper javadoc for the methods in that class.

metron-platform/metron-common/src/main/java/org/apache/metron/common/query/PredicateProcessor.java
```diff
@@ -18,12 +18,20 @@
...
+import org.antlr.v4.runtime.TokenStream;
+import org.apache.metron.common.query.generated.PredicateLexer;
+import org.apache.metron.common.query.generated.PredicateParser;
+
+import static org.apache.commons.lang3.StringUtils.isEmpty;
 
 
 public class PredicateProcessor {
   public boolean parse(String rule, VariableResolver resolver) {
+    if (isEmpty(rule)) {
```
cestella: what about rule != null && isEmpty(rule.trim())?

metron-platform/metron-common/src/test/java/org/apache/metron/common/query/QueryParserTest.java
```diff
@@ -67,6 +67,7 @@ public void testSimpleOps() throws Exception {
     Assert.assertTrue(run("foo== foo", v -> variableMap.get(v)));
     Assert.assertTrue(run("empty== ''", v -> variableMap.get(v)));
     Assert.assertTrue(run("spaced == 'metron is great'", v -> variableMap.get(v)));
+    Assert.assertTrue(run("", v -> variableMap.get(v)));
```
cestella: How about a test case for null too?

metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/filter/PcapFilters.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.function.Function;
+
+public enum PcapFilters {
+  FIXED(x -> new FixedPcapFilter()),
+  QUERY(x -> new QueryPcapFilter());
+
+  Function<Void, PcapFilter> filter;
+
+  PcapFilters(Function<Void, PcapFilter> filter) {
```
cestella: It'd be cool if this was an actual interface like, Creator<T> { T create(); } and it lived in metron-common, as opposed to Function<Void, T>.  I bet we have use for this pattern elsewhere.

metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/filter/query/QueryPcapFilter.java
```diff
@@ -0,0 +1,74 @@
...
+  public static class Configurator implements PcapFilterConfigurator<String> {
+    @Override
+    public void addToConfig(String query, Configuration conf) {
+      conf.set(QUERY_STR_CONFIG, query);
+      conf.set(PCAP_FILTER_NAME_CONF, PcapFilters.QUERY.name());
+    }
+
+    @Override
+    public String queryToString(String fields) {
+      return fields.replaceAll("\\w", "_")
```
cestella: I think you mean "\\s" there, not "\\w".  \\w matches word, so the output of this will be "___ ___" for input "foo bar".  Might be worth while to add a unit test around that too.

metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/mr/PcapJob.java
```diff
@@ -33,36 +33,39 @@
...
+      filter.configure(context.getConfiguration());
       start = Long.parseUnsignedLong(context.getConfiguration().get(START_TS_CONF));
       end = Long.parseUnsignedLong(context.getConfiguration().get(END_TS_CONF));
     }
 
     @Override
     protected void map(LongWritable key, BytesWritable value, Context context) throws IOException, InterruptedException {
       if(Long.compareUnsigned(key.get() ,start) >= 0 && Long.compareUnsigned(key.get(), end) <= 0) {
-        boolean send = Iterables.size(Iterables.filter(PcapHelper.toPacketInfo(value.copyBytes()), filter)) > 0;
+        boolean send = PcapHelper.toPacketInfo(value.copyBytes()).stream().filter(filter).collect(Collectors.toList()).size() > 0;
```
cestella: That `collect(Collectors.toList())` is going to cause the whole stream to be processed (gotta make that list).  I think maybe what we want here is `!Iterables.isEmpty(Iterables.filter(...))` as that will shortcut after the first Iterator.hasNext() returns true.

cestella: It's also clearer of intent (not empty as opposed to size > 0)

cestella:
On the whole, this is great.  Definitely a great feature and an impressive 2nd PR.  Thanks for the contribution!

Please make sure you didn't inadvertently regress the existing pcap functionality on the `full-dev-vagrant` image by following the testing plan in the original PR to move PCAP to hbase [here](https://github.com/apache/incubator-metron/pull/93)

metron-platform/metron-common/src/main/java/org/apache/metron/common/query/PredicateProcessor.java
mmiklavc: It appears that Hadoop Configuration resolves empty values to null - this handled both circumstances. I can add this for the PredicateProcessor and then add an additional translation in the client where the val is pulled from the Hadoop conf.

cestella: Right, I wasn't worried about empty values (your case covered that); I was worried about whitespace-only queries.  In that case `PredicateProcessor.parse("  ") != PredicateParser.parse("")`

mmiklavc: On second thought, what about rule == null || isEmpty(rule.trim())? I can't think of a reason to handle null different from "" or " ". The trim would only be evaluated if rule != null, and now we're also covering the null condition. Thoughts?

cestella: That works. || or && are equivalent in this situation

mmiklavc: What about NPE on rule.trim()? re: &&

cestella: wouldn't be an NPE on `rule != null && isEmpty(rule.trim())`  Java would shortcircuit before the `rule.trim()` call.

metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/filter/query/QueryPcapFilter.java
mmiklavc: Will definitely want feedback on the formatting on this one since we're using the string rep for filenames.

metron-platform/metron-common/src/main/java/org/apache/metron/common/query/PredicateProcessor.java
mmiklavc: To clarify, I've added the following test cases:
```
Assert.assertTrue(run(null, v -> variableMap.get(v)));
Assert.assertTrue(run("", v -> variableMap.get(v)));
Assert.assertTrue(run(" ", v -> variableMap.get(v)));
```
The null results in an NPE from ANTLR. 
Either we:
1. Add an else to return true 
2. Do this(my vote): if (rule == null || isEmpty(rule.trim())) { return true; }
3. Or throw a new exception for the parser

cestella: I vote for 2

metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/filter/query/QueryPcapFilter.java
cestella: Actually the safest thing here is to return "". There is already a UUID in
the file name to ensure uniqueness.

mmiklavc:
Ok, made a number of changes for review. Thanks for the feedback! Should have the remaining fixes/improvements avail soon and will also test per comments on full-dev-vagrant.

nickwallen:
For some reason, I cannot comment on the original JIRA, so I will comment here.

I think using a custom created language for this purpose is going to be confusing for our users.  Most network tools support what is called Berkely Packet Filter (BPF) syntax.  The user community is going to be fairly familiar with this syntax.  I would suggest that we use BPF syntax for this purpose.

cestella:
@nickwallen Agreed we should support BPF.  This PR makes the filter pluggable and we already have the query language.  We can have a follow-on PR for BPF support IMO.

nickwallen:
I think that makes sense, especially since this creates the 'hook' for later contributions of BPF.

The only downside is that we don't want to confuse users by having 8 ways to do things.  But I don't think that will be a problem as long as we eventually get the BPF functionality.

mmiklavc:
fyi, I was able to confirm no regressions in the existing PcapJob query via the instructions located here - https://github.com/apache/incubator-metron/pull/93

Note: I had to up map/reduce heap to 1GB and container mem to 1.2GB to avoid OOM errors.

cestella:
+1, got my vote.

dlyle65535:
+1 worked in EC2.

PR #122, opened 2016-05-17
Specify shorter control_path to avoid 'unix domain socket too long' problem

tom-meyer:
Paired with Ryan Merriman on this fix.  This issue is reported in the following Jira:  https://issues.apache.org/jira/browse/METRON-168

### DIFF
--- a/metron-deployment/amazon-ec2/ansible.cfg
+++ b/metron-deployment/amazon-ec2/ansible.cfg
@@ -25,4 +25,4 @@ log_path = ./ansible.log
 
 # fix for "ssh throws 'unix domain socket too long' " problem
 [ssh_connection]
-control_path = %(directory)s/%%h-%%p-%%r
+control_path = /tmp/%%h-%%p-%%r
--- a/metron-deployment/amazon-ec2/ansible.cfg
+++ b/metron-deployment/amazon-ec2/ansible.cfg
@@ -25,4 +25,4 @@ log_path = ./ansible.log
 
 # fix for "ssh throws 'unix domain socket too long' " problem
 [ssh_connection]
-control_path = /tmp/%%h-%%p-%%r
+control_path = ~/.ssh/ansible-ssh-%%C


metron-deployment/amazon-ec2/ansible.cfg
```diff
@@ -25,4 +25,4 @@ log_path = ./ansible.log
 
 # fix for "ssh throws 'unix domain socket too long' " problem
 [ssh_connection]
-control_path = %(directory)s/%%h-%%p-%%r
+control_path = /tmp/%%h-%%p-%%r
```
dlyle65535: I wouldn't use temp. DoD stigs require noexec on tmp. Alternatively, see if the socket will work when /tmp is mounted noexec.

nickwallen:
+1 Tested successfully with EC2 deployment.  

This same issue has bit us twice now.  It was difficult to diagnose this issue because it only impacted a few users.  It seems that OpenSSH imposes a 104 character limit on the length of the path to the domain socket.  

The previous configuration embedded things like the username and working directory in the path of the domain socket.  It would fail for people who had very specific platform differences, like a long username or highly nested path to Metron, which would then exceed the 104 character limit.  I myself was never able to replicate the problem until now.

This is our second attempt at a fix and I think this is the right approach.  Using `%%C` results in a fixed-length hash. Using `~/.ssh/` also ensures that the path is also of fixed-length.  This removes any variability that might cause it to exceed 104 characters.  [Seems like others agree too.](https://github.com/ansible/ansible/issues/11536)

cestella:
+1 this looks good

nickwallen:
This fix was not working for @dlyle65535 today.  It would error out because it did not understand `%C`.  

I dug a little and found that this option was added in [OpenSSH 6.7](http://www.openssh.com/txt/release-6.7).  I wanted to provide a warning for other users who have older versions of OpenSSH.

PR #125, opened 2016-05-19
Metron-126 added dfs.replication=1 for single node vm

ajayydv:
Added dfs.replication=1 for single node vm in single_node_vm.yml
Tested it by running full-dev-platform.


cestella:
+1 looks good

nickwallen:
+1

PR #129, opened 2016-05-20
METRON-178 Expose the filter capability in the Parser topologies.

cestella:
Allow users to specify the MessageFilter to use in the parser topologies. We already have this capability, we need to expose it via the zookeeper configs.


metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -59,10 +61,22 @@ public ParserBolt withMessageFilter(MessageFilter<JSONObject> filter) {
...
+    else if(filter == null) {
+      filter = Filters.get(getSensorParserConfig().getFilterClassName()
+              , getSensorParserConfig().getParserConfig()
+      );
+    }
     parser.init();
     writer.init();
   }
 
+  protected SensorParserConfig getSensorParserConfig() {
```
merrimanr: Would it make more sense to move this to the ConfiguredParserBolt class?  Don't think it matters much now because ParserBolt is the only class that extends ConfiguredParserBolt but other bolts may be created in the future and it would be nice to have this method available.

merrimanr:
As this PR stands, instantiation and configuration of MessageFilter and MessageParser classes are done differently in different places.  Would it make sense to create and configure them both the same way?  I imagine this will apply to MessageWriter too as it becomes more configurable.  Functionally it doesn't matter much so this is more of a style issue.  Not critical and either approach is fine but a common pattern would make the code easier to understand I think.

Overall looks good to me.  +1

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
cestella: yeah, that's a good catch

cestella:
Yeah, upon further consideration, I think that probably an init method is probably the best thing for MessageFilter.  I'll refactor.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
cestella: When I went to change it, I realized why I put it in the ParserBolt rather than the ConfiguredParserBolt.  The sensorType is part of the ParserBolt rather than the ConfiguredParserBolt.  I am going to make them both part of ConfiguredParserBolt.

james-sirota:
Works.  Great job! +1

PR #132, opened 2016-05-23
METRON-184 Fixed 'creates' path to avoid re-downloading GeoIP data

nickwallen:
Prevents the GeoIP data from being re-downloaded when it already exists.  The 'creates' path was specifying the wrong file.


cestella:
+1

nickwallen:
@dlyle65535 Can you take a look?  Any concerns?

dlyle65535:
+1- good catch!

PR #133, opened 2016-05-24
METRON-180 Granular Control of Component Installation with Tags

nickwallen:
To allow users to make the most use of the deployment scripts, they should be able to choose whether each component is installed or not.  This is important very highly customized environments where Metron will be installed.  This is currently possible in most cases, but there are a few scenarios that are exceptions.

For example, the PCAP service will only be installed if `install_elasticsearch` is set to `true`. This was a good coupling back when PCAP searches were hosted in the index. Now that PCAP searches come from HDFS, we need to be able to install PCAP service separately.

### DIFF
--- a/metron-deployment/playbooks/metron_install.yml
+++ b/metron-deployment/playbooks/metron_install.yml
@@ -116,66 +116,17 @@
 - hosts: sensors
   become: true
   roles:
-    - role: ambari_gather_facts
-    - role: tap_interface
+    - { role: ambari_gather_facts, tags: [ 'always'] }
+    - { role: tap_interface, tags: ['tap'] }
+    - { role: pycapa, tags: ['pycapa'] }
+    - { role: bro, tags: ['bro'] }
+    - { role: flume,  tags: ['snort','flume'] }
+    - { role: snort, tags: ['snort'] }
+    - { role: yaf, tags: ['yaf'] }
+    - { role: pcap_replay, tags: ['pcap-replay'] }
+    - { role: sensor-test-mode, tags: ['sensor-test-mode'] }
   tags:
-    - tap
-
-- hosts: sensors
-  become: true
-  roles:
-    - role: ambari_gather_facts
-    - role: pycapa
-  tags:
-    - pycapa
-
-- hosts: sensors
-  become: true
-  roles:
-    - role: ambari_gather_facts
-    - role: bro
-  tags:
-    - bro
-
-- hosts: sensors
-  become: true
-  roles:
-    - role: ambari_gather_facts
-    - role: flume
-  tags:
-    - flume
-
-- hosts: sensors
-  become: true
-  roles:
-    - role: ambari_gather_facts
-    - role: snort
-  tags:
-    - snort
-
-- hosts: sensors
-  become: true
-  roles:
-    - role: ambari_gather_facts
-    - role: yaf
-  tags:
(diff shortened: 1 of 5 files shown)


cestella:
Yeah, this makes sense.  Do we need to adjust the `run_ansible_role.sh` to pass in the `--skip-tags` arg as well?

cestella:
Also, is there an ec2.py somewhere that didn't get checked in?

mmiklavc:
Looks to me like ec2.py never made its way into source control. These queries return no results:
`git log --all --diff-filter=D --summary | grep delete|grep ec2`
`git log --all -- "**ec2.py"`

nickwallen:
https://community.hortonworks.com/comments/30932/view.html

nickwallen:
I couldn't resist. :)  

TLDR; The `ec2.py` that you are seeing refers to a file distributed with Ansible.  It will not exist in our repo.

nickwallen:
I do not think we need any changes to `run_ansible_role.sh` because we are explicitly defining which tags to run in that script.

nickwallen:
@cestella @mmiklavc Did I answer your question?  Any other ways I can make this better?

cestella:
+1

james-sirota:
@nickwallen what set of commands should i run to validate this?

nickwallen:
You can test this by, for example, choosing to install only one of the 'sensor' components and not the rest.  Before you had to install all 'sensor' components.

Should you want to explore packet capture in your test environment this command will only install pcap-replay and pycapa, but not the rest of the sensor components.

```
cd metron-deployment/vagrant/quick-dev-platform
vagrant --ansible-tags="hdp-deploy,metron" --ansible-skip-tags="solr,bro,flume,snort,yaf,sensor-test-mode"  up
```

metron-deployment/playbooks/metron_install.yml
```diff
@@ -74,47 +83,115 @@
...
+    - pcap-replay
+
+- hosts: sensors
+  become: true
+  roles:
+    - role: ambari_gather_facts
+    - role: sensor-test-mode
+  tags:
+    - sensor-test-mode
+
```
dlyle65535: I think I remembered my comment from last week- you may be able to combine these into a single hosts/roles group by using the following construct:
```yaml
hosts: sensors
become: true
tags: sensors
roles:
     -{ role: ambari_gather_facts, tags: [ 'always'] }
    - { role: tap_interface, tags: ['tap'] }
    - { role: pycapa, tags: ['pycapa'] }
    - { role: bro, tags: ['bro'] }
    - { role: flume,  tags: ['snort','flume'] }
    - { role: snort , tags: ['snort'] }
    - { role: yaf, tags: ['yaf'] }
    - { role: pcap_replay, tags: ['pcap-replay'] }
    - { role: sensor-test-mode, tags: ['sensor-test-mode'] }
```
This is per the docs, so ymmv. But if it works as advertised, I think it's a little bit more readable. Otherwise do what you got to do.

nickwallen: Yes, that is nice.  I like it.

nickwallen: Those changes tested successfully.  Added those in.  Good suggestion.  Much better looking.

dlyle65535:
👍 looks good, ship it! :)

PR #134, opened 2016-05-25
METRON-100 GeoIP errors out silently in vagrant

cestella:
When we transitioned from passing the value to the adapters to passing a CacheKey object, we never adjusted the SQL statements.  Also, this error may cause downstream issues whereby messages error out because enrichments are short circuited if one fails and the source.type never gets properly set.

The JIRA is wrong, this bug is actually present in every environment and prevents GeoIP enrichment from working.

### DIFF
--- a/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/jdbc/JdbcAdapter.java
+++ b/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/jdbc/JdbcAdapter.java
@@ -53,8 +53,7 @@ public abstract class JdbcAdapter implements EnrichmentAdapter<CacheKey>,
   }
 
   protected boolean resetConnectionIfNecessary() {
-    if(isConnectionClosed())
-    {
+    if(isConnectionClosed()) {
       this.cleanup();
       return this.initializeAdapter();
     }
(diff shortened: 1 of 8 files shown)


dlyle65535:
image
+1!

Thanks

metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/jdbc/JdbcAdapter.java
```diff
@@ -38,6 +39,27 @@
...
+        _LOG.error("Unable to maintain open JDBC connection: " + e.getMessage(), e);
+        isClosed = true;
+      }
+    }
+    return isClosed;
+  }
+
+  protected boolean resetConnectionIfNecessary() {
+    if(isConnectionClosed())
+    {
```
nickwallen: Sorry, nit-picky, but we have a new line between the if and the '{' which you don't do elsewhere in this code.  Would be nice to stick to one convention.

We need to enhance our code style checks.  Hard to catch and enforce manually.

cestella: So, I'll definitely make the class consistently K&R (I must've fat-fingered the newline).  We don't have a coherent style at the moment in this respect.  I prefer K&R bracing, so you'll see that in my code, but yeah, I agree, we need to correct style as we see it.

Also, don't worry about nitpicking..we all have to live in this code. :)

nickwallen:
Good catch.  Looks great. +1

AromalAS:
Tested with the latest code changes, ..Good Fix

PR #136, opened 2016-05-25
METRON-186: Create a fieldMapping functionality which allows for parsed fields to be transformed

cestella:
Currently the parsers take care of transforming raw data to the parsed JSON representation. Allow for a layer to be placed at a sensor level to transform input fields from the parsed messages to create new fields. For instance, mapping IANA protocol numbers to a standardized textual representation (i.e. 6 maps to TCP).

This should be generalized, so that we can provide common functions and users can define the input fields and the output field to use.  This configuration should live with the sensor configuration in zookeeper.


metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -61,18 +62,32 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
   public void execute(Tuple tuple) {
     byte[] originalMessage = tuple.getBinary(0);
+    SensorParserConfig sensorParserConfig = configurations.getSensorParserConfig(sensorType);
     try {
       List<JSONObject> messages = parser.parse(originalMessage);
       for(JSONObject message: messages) {
         if (parser.validate(message)) {
           if (filter != null && filter.emitTuple(message)) {
             message.put(Constants.SENSOR_TYPE, sensorType);
+            List<MappingHandler> fieldMappings = sensorParserConfig == null?null:sensorParserConfig.getFieldMappings();
```
mmiklavc: If I'm interpreting correctly, org.apache.metron.common.configuration.SensorParserConfig won't vary with this implementation (i.e. not a user-provided impl). That said, I believe `sensorParserConfig.getFieldMappings()` should always return a non-null `List<MappingHandler>` (per the default init in that class -> `fieldMappings = new ArrayList<>()`)
Maybe simplify line 82-84 to this?
```
if(sensorParserConfig != null) {
  for(MappingHandler handler: sensorParserConfig.getFieldMappings()) {
...
  }
}
```
Also, might be worth a small test to see if line 85 handler null check can also go away.

nickwallen:
Can you provide some documentation, README or otherwise, on how this will work for a user.  Hard to visualize with what you've submitted.  What steps do I need to take to use this functionality?  What configs do I need to deploy?

nickwallen:
I am wondering if `mapping` is the most descriptive word we can use here.  We are also using map in a lot of different contexts in the configuration that makes it a bit confusing; `fieldToTypeMap`, `fieldMap`.

I am sure there are better alternatives, but `transform` is the word that comes first to my mind.  I think if we write the docs, the most descriptive word will more easily fall out.  That's what tends to happen for me at least.

cestella:
Yeah, I was thinking about that too.  I think transformation is a more descriptive term.  You're the 2nd person (in addition to myself) who has made that comment.  I think that means I should call it transformation.

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/RemoveTransformationTest.java
```diff
@@ -0,0 +1,100 @@
...
+      handler.transformAndUpdate(input, new HashMap<>());
+      Assert.assertTrue(input.containsKey("field1"));
+    }
+    {
+      JSONObject input = new JSONObject(new HashMap<String, Object>() {{
+        put("field1", "foo");
+        put("field2", "bar");
+      }});
+      handler.transformAndUpdate(input, new HashMap<>());
+      Assert.assertTrue(input.containsKey("field1"));
```
nickwallen: Don't we also need to check that 'field2' still exists too?

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/RemoveTransformationTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+    {
+      JSONObject input = new JSONObject(new HashMap<String, Object>() {{
+        put("field1", "bar");
+        put("field2", "foo");
+      }});
+      handler.transformAndUpdate(input, new HashMap<>());
+      Assert.assertEquals("bar", input.get("field1"));
```
nickwallen: Does this test ensure that 'field2' has been removed?

nickwallen: Actually, I think you're intending that 'field1' should be removed?  I'm not sure.  Comments would be awesome.  I'm not that bright.

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/RemoveTransformationTest.java
nickwallen: I misunderstood the test.  Since field2 exists and its value is 'foo', then the code should not remove 'field1'.  And this is what you've asserted.  So looks good, but quick comments for those of us who are less bright would be awesome.

nickwallen:
👍

merrimanr:
Nice work.  +1

PR #138, opened 2016-05-27
METRON-189: Add the ability to do global validations on messages passing through the parser.

cestella:
Allow the user to specify field level or message level validations to ensure messages coming from the parser are valid. For instance, allow the ability ensure that a field is an IPv4 address.
If a field is invalid, send to a separate stream from the parser bolt. Follow-on work should be done to send this stream to the index for after-the-fact inspection.

I added the following validation functions:
* `MQL` : Execute a Query Language statement.  Expects the query string in the `condition` field of the config.
* `IP` : Validates that the input fields are an IP addres.  By default, if no configuration is set, it assumes `IPV4`, but you can specify        the type by passing in the config by passing in `type` with either `IPV6` or `IPV4`.
* `DOMAIN` : Validates that the fields are all domains.
* `EMAIL` : Validates that the fields are all email addresses
* `URL` : Validates that the fields are all URLs
* `DATE` : Validates that the fields are a date.  Expects `format` in the config.
* `INTEGER` : Validates that the fields are an integer.  String representation of an integer is allowed.
* `REGEX_MATCH` : Validates that the fields match a regex.  Expects `pattern` in the config.
* `NOT_EMPTY` : Validates that the fields exist and are not empty (after trimming.)

Because of the nice overlap, I also added these functions to the query language, so query language rules can take advantage of `IS_IP(field1)` for instance.


james-sirota:
I think the documentation is misplaced.  https://github.com/cestella/incubator-metron/tree/validation/metron-platform/metron-common is really good, but probably does not belong in common.  I would put this with the parsers.  What was the reasoning for putting it here?

james-sirota:
I spot checked the validators and the validators work.  I checked IP, domain, url and date using YAF and Bro.  I did not test the negative case since the dead letter q does not yet exist.  One thing of note I see that you are using commons validator and in my previous experience this library has pretty significant performance issues validating IPs.  I think checking it with regex was 10 times faster or something around there.  Once we get to performance tuning this may be something to keep in mind.  Otherwise great job!  +1

cestella:
Yeah, the documentation situation is a bit incorrect.  Documentation for configuration started to be placed where the configuration objects existed, rather than where they were used.  Since we put the configuration in commons, then the documentation went in commons.  I suggest strongly that we do a follow-on after we flush the PR queue to move docs around.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -140,18 +142,25 @@ public void execute(Tuple tuple) {
...
-              message.put(Constants.SENSOR_TYPE, getSensorType());
-              for (FieldTransformer handler : sensorParserConfig.getFieldTransformations()) {
-                if (handler != null) {
-                  handler.transformAndUpdate(message, sensorParserConfig.getParserConfig());
+            if(!isGloballyValid(message, fieldValidations)) {
+              message.put(Constants.SENSOR_TYPE, getSensorType()+ ".invalid");
+              collector.emit(Constants.INVALID_STREAM, new Values(message));
+            }
+            else if (filter != null && filter.emitTuple(message)) {
+              if (filter != null && filter.emitTuple(message)) {
```
merrimanr: Isn't this a duplicate of the previous line?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -165,8 +174,17 @@ public void execute(Tuple tuple) {
     }
   }
 
+  private boolean isGloballyValid(JSONObject input, List<FieldValidator> validators) {
+    boolean ret = true;
+    for(FieldValidator validator : validators) {
+      ret &= validator.isValid(input, getConfigurations().getGlobalConfig());
```
merrimanr: Would it be more efficient to break on a failed validation?

cestella: Yep, agreed.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
cestella: Yep, bad merge.  Fixed.

merrimanr:
+1

PR #139, opened 2016-05-27
METRON-190: Make start_parser_topology.sh more adaptable regarding storm topology configuration

cestella:
Add the ability to specify the message timeout, the max task parallelism, the number of ackers, the number of workers. Also allow the user to pass in JSON file containing a map of extra storm configs.

I added the following extra options to `start_parser_topology.sh`:
```
 -e,--extra_options <JSON_FILE>            Extra options in the form of a
                                           JSON file with a map for
                                           content.
 -mt,--message_timeout <TIMEOUT_IN_SECS>   Message Timeout in Seconds
 -mtp,--max_task_parallelism <MAX_TASK>    Max task parallelism
 -na,--num_ackers <NUM_ACKERS>             Number of Ackers
 -nw,--num_workers <NUM_WORKERS>           Number of Workers
```

For instance, if you wanted to set some storm property on
the config called `topology.ticks.tuple.freq.secs` to 1000 and `storm.local.dir` to `/opt/my/path`
you could create a file called `custom_config.json` containing
```
{
  "topology.ticks.tuple.freq.secs" : 1000,
  "storm.local.dir" : "/opt/my/path"
}
```
and pass `--extra_options custom_config.json` to `start_parser_topology.sh`.

### DIFF
--- a/metron-platform/metron-parsers/README.md
+++ b/metron-platform/metron-parsers/README.md
@@ -201,21 +201,23 @@ The usage for `start_parser_topology.sh` is as follows:
 
 ```
 usage: start_parser_topology.sh
- -e,--extra_options <JSON_FILE>            Extra options in the form of a
-                                           JSON file with a map for
-                                           content.
- -h,--help                                 This screen
- -k,--kafka <BROKER_URL>                   Kafka Broker URL
- -mt,--message_timeout <TIMEOUT_IN_SECS>   Message Timeout in Seconds
- -mtp,--max_task_parallelism <MAX_TASK>    Max task parallelism
- -na,--num_ackers <NUM_ACKERS>             Number of Ackers
- -nw,--num_workers <NUM_WORKERS>           Number of Workers
- -pp,--parser_p <PARSER_PARALLELISM>       Parser Parallelism
- -s,--sensor <SENSOR_TYPE>                 Sensor Type
- -sp,--spout_p <SPOUT_PARALLELISM>         Spout Parallelism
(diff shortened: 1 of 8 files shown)


nickwallen:
Just watch out for monit's use of this script.  Let me know if you need specific pointers.

cestella:
Ok, I verified this on full dev vagrant.

james-sirota:
The settings work, but I think you forgot to expose the parallelism hint.  See this blog entry: https://storm.apache.org/releases/1.0.0/Understanding-the-parallelism-of-a-Storm-topology.html

Notice how they are setting 

topologyBuilder.setBolt("green-bolt", new GreenBolt(), 2)
               .setNumTasks(4)
               .shuffleGrouping("blue-spout");

 This will run two executors and four associated tasks (two tasks per executor)

cestella:
@james-sirota Right you are!  Just updated exposing the numTasks for the spout and the parser bolt.

```
-pnt,--parser_num_tasks <PARSER_NUM_TASKS>   Parser Num Tasks
-pp,--parser_p <PARSER_PARALLELISM>          Parser Parallelism
-snt,--spout_num_tasks <NUM_TASKS>           Spout Num Tasks
-sp,--spout_p <SPOUT_PARALLELISM>            Spout Parallelism
```

james-sirota:
+1.  The only thing I would suggest would be to name the variable parallelism hint and not just parallelism because that name has meaning for someone who would be tuning storm.

cestella:
good catch, I'll make those changes now.

PR #142, opened 2016-06-03
METRON-204: Field Transformation Domain Specific Language

cestella:
Similar to the domain specific query language, it would be nice to have a domain specific language for transformations which is used as an optional FieldTransformation implementation.

* A fixed set of transformation functions:
   * `TO_LOWER(string)` : Transforms the first argument to a lowercase string
   * `TO_UPPER(string)` : Transforms the first argument to an uppercase string
   * `TO_STRING(string)` : Transforms the first argument to a string
   * `TO_INTEGER(x)` : Transforms the first argument to an integer
   * `TO_DOUBLE(x)` : Transforms the first argument to a double
   * `TRIM(string)` : Trims whitespace from both sides of a string.
   * `JOIN(list, delim)` : Joins the components of the list with the specified delimiter
   * `SPLIT(string, delim)` : Splits the string by the delimiter.  Returns a list.
   * `GET_FIRST(list)` : Returns the first element of the list
   * `GET_LAST(list)` : Returns the last element of the list
   * `GET(list, i)` : Returns the i'th element of the list (i is 0-based).
   * `MAP_GET(key, map, default)` : Returns the value associated with the key in the map.  If the key does not exist, the default will   be returned.  If the default is unspecified, then null will be returned.
   * `DOMAIN_TO_TLD(domain)` : Returns the TLD of the domain.
   * `DOMAIN_REMOVE_TLD(domain)` : Remove the TLD of the domain.
   * `REMOVE_TLD(domain)` : Removes the TLD from the domain.
   * `URL_TO_HOST(url)` : Returns the host from a URL
   * `URL_TO_PROTOCOL(url)` : Returns the protocol from a URL
   * `URL_TO_PORT(url)` : Returns the port from a URL
   * `URL_TO_PATH(url)` : Returns the path from a URL
   * `TO_EPOCH_TIMESTAMP(dateTime, format, timezone)` : Returns the epoch timestamp of the `dateTime` given the `format`.  If the format does not have a timestamp and you wish to assume a given timestamp, you may specify the `timezone` optionally.
* A FieldTransformer implementation, `MTL` which exposes the transformation language

Example MTL transformation:

Consider the following sensor parser config to add three new fields to a
message:
* `utc_timestamp` : The unix epoch timestamp based on the `timestamp` field, a `dc` field which is the data center the message comes     from and a `dc2tz` map mapping data centers to timezones
* `url_host` : The host associated with the url in the `url` field
* `url_protocol` : The protocol associated with the url in the `url` field

```
{
...
    "fieldTransformations" : [
          {
           "transformation" : "MTL"
          ,"output" : [ "utc_timestamp", "url_host", "url_protocol" ]
          ,"config" : {
            "utc_timestamp" : "TO_EPOCH_TIMESTAMP(timestamp, 'yyyy-MM-dd
HH:mm:ss', MAP_GET(dc, dc2tz, 'UTC') )"
           ,"url_host" : "URL_TO_HOST(url)"
           ,"url_protocol" : "URL_TO_PROTOCOL(url)"
                      }
          }
                      ]
   ,"parserConfig" : {
      "dc2tz" : {
                "nyc" : "EST"
               ,"la" : "PST"
               ,"london" : "UTC"
                }
    }
}
```

Note that the `dc2tz` map is in the parser config, so it is accessible
in the functions.


ddutta:
Nice contribution .... a DSL is very useful for setting rules/triggers etc 
* is there a doc in the JIRA for this?
* would be gr8 to discuss design on the dev list .... the dev list is quite silent!
* also for the community to participate it would be have been awesome to have smaller commits for easy code review (also true for many other awesome metron contributions!)

cestella:
I will definitely create a doc on the JIRA about the DSL and would be happy to discuss.  It's really just the extraction of the existing transformation functions from the query DSL into their own language.  It seemed like a useful abstraction now that we have field transformations.

cestella:
Also, it seems like a big PR, but honestly it's deceptive.  There was a lot of file moving and the generated Antlr code inflates the size.

cestella:
I uploaded a design doc to the JIRA.

nickwallen:
Love the functionality that you are creating with this.  I am curious if it would make more sense to provide this functionality via an existing JVM-capable language like Groovy, etc, rather than inventing our own DSL.  At first glance, it seems there are a lot of advantages in doing it this way.  What were the pros/cons when you thought through this?

cestella:
So, I think the main disadvantage of using groovy or any of the scripting
engine integrations is speed and expressibility.  Having a targeted domain
specific language is going to be a lot simpler to parse than something more
complex like groovy (or javascript for that matter).  It's hardly even a
DSL, frankly, just the ability to compose functions, which seems like a
sweet spot between the expressivity and complexity.  If we want to
implement a FieldTransformation which is a scripting language, there's
nothing keeping us from doing this.  The other reason I like a simple DSL
with just function composition is that I'd *really* prefer to not give
users enough rope to hang themselves with here if we can help it.  This DSL
really exposes only the things you can do in one line of code.  If it's
more than one line of code, we should be considering that a feature in the
language.

If we really want to integrate a scripting engine in the future, we can
create a separate FieldTransformer to do it in just one class.

metron-platform/metron-common/src/main/java/org/apache/metron/common/query/generated/PredicateLexer.java
```diff
@@ -114,59 +118,74 @@ public PredicateLexer(CharStream input) {
...
+		"\3\2\2\2\u00ba\u00bc\3\2\2\2\u00bb\u00b9\3\2\2\2\u00bc\u00be\7)\2\2\u00bd"+
+		"\u00ad\3\2\2\2\u00bd\u00b5\3\2\2\2\u00be\62\3\2\2\2\u00bf\u00c0\7=\2\2"+
+		"\u00c0\64\3\2\2\2\u00c1\u00c2\7\61\2\2\u00c2\u00c3\7\61\2\2\u00c3\u00c5"+
+		"\3\2\2\2\u00c4\u00c6\13\2\2\2\u00c5\u00c4\3\2\2\2\u00c6\u00c7\3\2\2\2"+
+		"\u00c7\u00c8\3\2\2\2\u00c7\u00c5\3\2\2\2\u00c8\u00ca\3\2\2\2\u00c9\u00cb"+
+		"\t\5\2\2\u00ca\u00c9\3\2\2\2\u00cb\u00cc\3\2\2\2\u00cc\u00cd\b\33\2\2"+
+		"\u00cd\66\3\2\2\2\u00ce\u00d0\t\6\2\2\u00cf\u00ce\3\2\2\2\u00d0\u00d1"+
+		"\3\2\2\2\u00d1\u00cf\3\2\2\2\u00d1\u00d2\3\2\2\2\u00d2\u00d3\3\2\2\2\u00d3"+
+		"\u00d4\b\34\2\2\u00d48\3\2\2\2\22\2AIQ[g\u0097\u009c\u00a2\u00a8\u00b1"+
+		"\u00b9\u00bd\u00c7\u00ca\u00d1\3\b\2\2";
```
nickwallen: My understanding is that this is generated code; no?  Why do we have to commit generated code to source control?

metron-platform/metron-common/src/main/java/org/apache/metron/common/aggregator/Aggregators.java
```diff
@@ -18,18 +18,21 @@
...
-  ,MIN( (numbers, config) -> accumulate(0d, (x,y) -> Math.min(x.doubleValue(),y.doubleValue()), numbers))
-  ,SUM( (numbers, config) -> accumulate(0d, (x,y) -> x.doubleValue() + y.doubleValue(), numbers))
+   MAX( (numbers, config) -> accumulate(0d, (x,y) -> Math.max(x.doubleValue(),y.doubleValue()), numbers, config))
+  ,MIN( (numbers, config) -> accumulate(0d, (x,y) -> Math.min(x.doubleValue(),y.doubleValue()), numbers, config))
+  ,SUM( (numbers, config) -> accumulate(0d, (x,y) -> x.doubleValue() + y.doubleValue(), numbers, config))
   ,MEAN( (numbers, config) -> scale(SUM.aggregate(numbers, config), numbers, n -> true))
-  ,POSITIVE_MEAN( (numbers, config) -> scale(SUM.aggregate(numbers, config), numbers, n -> n.doubleValue() > 0))
+  ,POSITIVE_MEAN( (numbers, config) -> positiveMean(numbers, config))
   ;
+  public static String NEGATIVE_VALUES_TRUMP_CONF = "negativeValuesTrump";
```
nickwallen: It is not intuitive to me what this means.  NEGATIVE_VALUES_TRUMP_CONF?

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/MTLTransformationTest.java
```diff
@@ -0,0 +1,143 @@
...
+                      }
+          }
+                      ]
+   }
+   */
+  @Multiline
+  public static String mtlConfig;
+
+  @Test
+  public void testMTL() throws Exception {
```
nickwallen: I would love to see descriptive comments or more descriptive names.  I have to think too hard to understand what these tests are doing.

metron-platform/metron-common/src/main/java/org/apache/metron/common/query/generated/PredicateLexer.java
cestella: Because this is generated code, people just cloning the repo and spinning it up in an IDE would have to either configure their IDE to run the maven generate-sources target or externally run maven to get the code to compile.  I think that generated code that is rarely touched existing in the SCM is generally better than instructing people to run maven before they start their IDEs or, worse, describe how to set up their IDEs to generate the sources.

I'm open to a discussion, obviously.

metron-platform/metron-common/src/main/java/org/apache/metron/common/aggregator/Aggregators.java
cestella: I noticed an issue with threat triage in-that we have no ability to cause a particular rule to trump the rest.  For instance, if you have rules 1 through 5, but if you hit rule 5, you want to make sure that the triage level is very high.  For that use-case, if you associate a negative value with a triage rule, the aggregators will return +Infinity.

metron-platform/metron-common/src/main/java/org/apache/metron/common/query/generated/PredicateLexer.java
cestella: I'll add, though, that consensus is not necessarily a reason to do something, it is quite common to commit generated code within the hadoop ecosystem projects:
* [Hive](https://github.com/apache/hive/tree/master/ql/src/gen/thrift)
* [Accumulo](https://github.com/apache/accumulo/tree/5cb5b9372103761c829403c03007b9f53241400f/core/src/main/java/org/apache/accumulo/core/tabletserver/thrift)

cestella:
I have modified the squid parser to use the MTL transformer to extract the subdomain-less hostname rather than using grok.  So, in order to test this, you should be able to 
* create the `squid` kafka topic
* run the squid parser
* pump data from the squid logs through using the console producer
* Look at the `enrichment` kafka topic and verify the messages coming out have
   * `url` : the full URL
   * `full_hostname` : the full hostname of the URL
   * `domain_without_subdomains` : the domain without subdomains

This is an improvement of the original functionality since it only stripped off the `www` if it exists.

dlyle65535:
+1, really great stuff!

merrimanr:
Awesome! +1

PR #144, opened 2016-06-06
METRON-211 Enable creation of new Kafka topics during deployment

nickwallen:
### [METRON-211](https://issues.apache.org/jira/browse/METRON-211)

During Metron deployment a fixed number of Kafka topics are created to support the input streams supported by Metron "out of the box" including Bro, YAF, and Snort. As new parsers are created, new Kafka topics will be needed to support those parsers that go beyond this default set of topics.

Use cases for this functionality would include the following.

- A developer needs to define new Kafka topics associated with parsers that he or she creates. A developer should also be able to define the replication factor, partition count, and retention size associated with those topics.

- An operator needs to define the Kafka topics that get created when deploying new parsers to a Metron environment. The operator needs to define a replication factor, partition count and retention size that is appropriate for the target environment.

An example showing the definition of two topics that will be created, each with different partitions, replication and retention settings.

```
topics_to_create:
  - { topic: "bluecoat", num_partitions: 10, replication_factor: 3, retention_gb: 10 }
  - { topic: "enrichments", num_partitions: 21, replication_factor: 3, retention_gb: 120 }
```

#### Changes
- Removed `hadoop_setup` role
- Kafka topics needed by Metron are now created in a new `metron_kafka_topics` role
  - The Kafka topics that are created are defined via properties.
  - Each Kafka topic can have unique retention, replication, and partition settings.
  - Using the `metron_kafka_topics` host group to target this role at hosts.
- HBase tables needed by Metron are now created in a new `metron_hbase_tables` role
  - The HBase tables that are created are defined via properties.
  - Using the `metron_hbase_tables` host groups to target this role at hosts.
- Using separate host groups to target each role. This is more flexible than before as there may not be a single host with both Kafka and HBase clients installed.


metron-deployment/roles/metron_streaming/tasks/main.yml
```diff
@@ -15,25 +15,18 @@
...
+      - lib
+      - bin
+      - config
 
 - name: Create Source Config Directory
-  file:
-    path: "{{ zookeeper_config_path }}"
-    state: directory
-    mode: 0755
+  file: path={{ zookeeper_config_path }} state=directory mode=0755
```
dlyle65535: I've noticed you're reverting tasks back to old-style parameter naming. Any reason for this?

nickwallen: More compact.  To my eyes it reads better with the other single-line code lines around it.  Purely aesthetic.  Can undo if there is an objection.

dlyle65535: Not a strong objection, but I do find the vertical presentation easier to read. I think it's because the yaml file decorator does a better job calling out the parameters in the vertical orientation.

dlyle65535:
👍 on this. Looks great to me.

metron-deployment/roles/metron_streaming/tasks/main.yml
nickwallen: Reverted to original vertical presentation. Not a good enough reason for me to change it.

cestella:
+1 this looks good

PR #150, opened 2016-06-14
METRON-222: Address ConcurrentModificationException in BulkMessageWriter

cestella:
In the situation where multiple consecutive bolts exist in different executors in the same JVM, the passed tuples are not serialized or deserialized for performance in Storm.  As such, we were emitting a message and then immediately mutating it.  Meanwhile, in a different thread (executor) the message, a HashMap, is being iterated on in the next bolt.  This causes a CME.

### DIFF
--- a/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/BulkMessageWriterBolt.java
+++ b/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/BulkMessageWriterBolt.java
@@ -61,26 +61,11 @@ public class BulkMessageWriterBolt extends ConfiguredEnrichmentBolt {
     }
   }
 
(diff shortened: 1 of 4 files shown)


dlyle65535:
Good deal on tracking this one down. Think it's worth switching back to the .clone method in the BulkMessageWriterBolt?

cestella:
I don't see a difference in the clone method vs the method currently implemented.  I think we should evaluate in a follow-on JIRA whether any of the clones are needed.  It seems that none of them should be needed.

dlyle65535:
💯  - really solid work, Casey. Looks like it was a bear to track down.

I ran this in EC2 with 
  2 workers and 15 executors for 16 minutes - no error.
  1 worker and 15 executors for 9 minutes and counting - no error.

Way to quick turn a really difficult bug.

nickwallen:
+1

PR #152, opened 2016-06-15
METRON-228: Fixing NPE when enrichment config does not exist.

cestella:
Because we are pulling the batch size from the enrichment sensor config, if it does not exist then it will throw a NPE.


dlyle65535:
+1. Tested on EC2. Ran squid parser without enrichment configuration. Data made it to the index and there was no NPE. Nice work.

merrimanr:
+1

PR #154, opened 2016-06-16
METRON-230: Bro parser should throw exception

cestella:
Right now, if an invalid message comes to the bro parser, it returns null, which is interpreted as no messages from the parser.  INstead, we should throw an exception so the message can be routed to the error queue.


dlyle65535:
+1 works great in quick-dev-vagrant.

nickwallen:
👍

PR #155, opened 2016-06-16
METRON-231: Snort parser should throw exception

cestella:
Snort parser returns null when we should be throwing an exception in the event of an improperly formatted message.  An exception will result in the message being sent to the error queue.


dlyle65535:
+1 - ran on quick-dev-vagrant, worked like a champ.

nickwallen:
+1 Looks good

PR #156, opened 2016-06-16
METRON-235 Expose filtering capability for PCAP via CLI tool

mmiklavc:
In the process of testing with Vagrant, but wanted to get this in front of people for review.

Relevant Jira:
https://issues.apache.org/jira/browse/METRON-235

In the process of upgrading to Kibana 4, we lost our about to query/filter pcap results. This PR exposes the 2 methods for filtering PCAP data, fixed parameters and the "Stellar" query language, via a command line tool. The tool is executed via ${metron_home}/bin/pcap_query.sh

**Note**: I also tweaked the mem settings for Ambari mapreduce to get around persistent OOM errors - mapreduce container sizes (mapreduce.[map | reduce].memory.mb) have been increased to 1.2 GiB, mapreduce.[map | reduce].java.opts have been increased to 1 GiB.

# PR Testing Instructions:

Get PCAP data into Metron: Install and setup pycapa - the instructions below reference/mirror those in [PR-93](https://github.com/apache/incubator-metron/pull/93#issue-151705836)

1. Install the pycapa library & utility ```$ cd /opt/pycapa/pycapa && pip install -r requirements.txt && python setup.py install```
2. (if using singlenode vagrant) Kill the enrichment and sensor topologies via for i in bro enrichment yaf snort;do storm kill $i;done
3. Start the pcap topology via /usr/metron/0.1BETA/bin/start_pcap_topology.sh
4. Start the pycapa packet capture producer on eth1 via /usr/bin/pycapa --producer --topic pcap -i eth1 -k node1:6667
5. Watch the topology in the Storm UI and kill the packet capture utility from before when the number of packets ingested is over 1k.
6. Ensure that at at least 2 files exist on HDFS by running hadoop fs -ls /apps/metron/pcap
7. Choose a file (denoted by $FILE) and dump a few of the contents using the pcap_inspector utility via /usr/metron/0.1BETA/bin/pcap_inspector.sh -i $FILE -n 5
8. Choose one of the lines and note the source ip/port and dest ip/port as well as the protocol.
9. Note that when you run the commands below, the resulting file will be placed in the execution directory where you kicked off the job from.

### Fixed filter

1. Run a fixed filter query by executing the following command with the values noted above (match your start_time format to the date format provided - default is to use millis since epoch)
2. `/usr/metron/0.1BETA/bin/pcap_query.sh fixed -st <start_time> -df "yyyyMMdd" -sa <address> -da <address> -sp <port> -dp <port> -p <protocol_num>`
3. Verify the MR job finishes successfully. Upon completion, you should see a file named with the current datestamp in your current directory, e.g. pcap-data-20160617160549737+0000.pcap
4. Copy the file to your local machine and verify you can open it in Wireshark

### Query filter

1. Run a Stellar query filter query by executing a command similar to the following, with the values noted above (match your start_time format to the date format provided - default is to use millis since epoch)
2. `/usr/metron/0.1BETA/bin/pcap_query.sh query -st "20160617" -df "yyyyMMdd" -query "ip_src_addr == '192.168.1.1' and ip_src_port == '49197' and ip_dst_addr == '192.168.1.2' and ip_dst_port == '80' and protocol == '6'"`
3. Verify the MR job finishes successfully. Upon completion, you should see a file named with the current datestamp in your current directory, e.g. pcap-data-20160617160549737+0000.pcap
4. Copy the file to your local machine and verify you can open it in Wireshark

**References:**
- https://github.com/apache/incubator-metron/pull/93#issue-151705836
- https://github.com/apache/incubator-metron/tree/master/metron-sensors/pycapa

### DIFF
--- a/metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/CliParser.java
+++ b/metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/CliParser.java
@@ -1,0 +1,83 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.metron.pcap.query;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+
+/**
+ * Provides commmon required fields for the PCAP filter jobs
+ */
+public class CliParser {
+
+  public Options buildOptions() {
+    Options options = new Options();
+    options.addOption(newOption("h", false, "Display help"));
+    options.addOption(newOption("basePath", true, String.format("Base PCAP data path. Default is '%s'", CliConfig.BASE_PATH_DEFAULT)));
+    options.addOption(newOption("baseOutputPath", true, String.format("Query result output path. Default is '%s'", CliConfig.BASE_OUTPUT_PATH_DEFAULT)));
+    options.addOption(newOption("startTime", true, "Packet start time range. Default is '0'"));
+    options.addOption(newOption("endTime", true, "Packet end time range. Default is current system time."));
+    return options;
+  }
+
+  protected Option newOption(String opt, boolean hasArg, String desc) {
+    return newOption(opt, hasArg, desc, false);
+  }
+
+  protected Option newOption(String opt, boolean hasArg, String desc, boolean required) {
+    Option option = new Option(opt, hasArg, desc);
+    option.setRequired(required);
+    return option;
+  }
+
+  public void parse(CommandLine commandLine, CliConfig config) {
+    if (commandLine.hasOption("h")) {
+      config.setShowHelp(true);
+    }
+    if (commandLine.hasOption("basePath")) {
+      config.setBasePath(commandLine.getOptionValue("basePath"));
+    }
+    if (commandLine.hasOption("baseOutputPath")) {
+      config.setBaseOutputPath(commandLine.getOptionValue("baseOutputPath"));
+    }
+    if (commandLine.hasOption("startTime")) {
+      try {
+        long startTime = Long.parseLong(commandLine.getOptionValue("startTime"));
+        config.setStartTime(startTime);
+      } catch (NumberFormatException nfe) {
+        //no-op
+      }
+    }
+    if (commandLine.hasOption("endTime")) {
+      try {
+        long endTime = Long.parseLong(commandLine.getOptionValue("endTime"));
+        config.setEndTime(endTime);
+      } catch (NumberFormatException nfe) {
+        //no-op
+      }
+    }
+  }
+
+  public void printHelp(String msg, Options opts) {
+    new HelpFormatter().printHelp(msg, opts);
+  }
+
+}
--- a/metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/CliParser.java
+++ b/metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/CliParser.java
@@ -18,58 +18,74 @@
 
 package org.apache.metron.pcap.query;
 
-import org.apache.commons.cli.CommandLine;
-import org.apache.commons.cli.HelpFormatter;
-import org.apache.commons.cli.Option;
-import org.apache.commons.cli.Options;
+import org.apache.commons.cli.*;
 
 /**
  * Provides commmon required fields for the PCAP filter jobs
  */
 public class CliParser {
+  private CommandLineParser parser;
+
+  public CliParser() {
+    parser = new PosixParser();
+  }
 
   public Options buildOptions() {
     Options options = new Options();
-    options.addOption(newOption("h", false, "Display help"));
-    options.addOption(newOption("basePath", true, String.format("Base PCAP data path. Default is '%s'", CliConfig.BASE_PATH_DEFAULT)));
-    options.addOption(newOption("baseOutputPath", true, String.format("Query result output path. Default is '%s'", CliConfig.BASE_OUTPUT_PATH_DEFAULT)));
-    options.addOption(newOption("startTime", true, "Packet start time range. Default is '0'"));
-    options.addOption(newOption("endTime", true, "Packet end time range. Default is current system time."));
+    options.addOption(newOption("h", "help", false, "Display help"));
+    options.addOption(newOption("bp", "base_path", true, String.format("Base PCAP data path. Default is '%s'", CliConfig.BASE_PATH_DEFAULT)));
+    options.addOption(newOption("bop", "base_output_path", true, String.format("Query result output path. Default is '%s'", CliConfig.BASE_OUTPUT_PATH_DEFAULT)));
+    options.addOption(newOption("st", "start_time", true, "(required) Packet start time range.", true));
+    options.addOption(newOption("et", "end_time", true, "Packet end time range. Default is current system time."));
+    options.addOption(newOption("df", "date_format", true, "Date format to use for parsing start_time and end_time. Default is to use time in millis since the epoch."));
     return options;
   }
 
-  protected Option newOption(String opt, boolean hasArg, String desc) {
-    return newOption(opt, hasArg, desc, false);
+  protected Option newOption(String opt, String longOpt, boolean hasArg, String desc) {
+    return newOption(opt, longOpt, hasArg, desc, false);
   }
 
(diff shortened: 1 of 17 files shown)


metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/CliParser.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Provides commmon required fields for the PCAP filter jobs
+ */
+public class CliParser {
+
+  public Options buildOptions() {
+    Options options = new Options();
+    options.addOption(newOption("h", false, "Display help"));
+    options.addOption(newOption("basePath", true, String.format("Base PCAP data path. Default is '%s'", CliConfig.BASE_PATH_DEFAULT)));
```
cestella: Is this a short option as well as a long option?  I tend to like having both at our disposal.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  public PcapCli(PcapJob jobRunner, ResultsWriter resultsWriter, Clock clock) {
+    this.jobRunner = jobRunner;
+    this.resultsWriter = resultsWriter;
+    this.clock = clock;
+  }
+
+  public int run(String[] args) {
+    if (args.length < 1) {
```
cestella: Don't we need to use GenericOptionsParser(conf, args).getRemainingArgs()?  Otherwise we're going to get all the hadoop args if we need to specify specific `-D` stuff for the MR job.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
```diff
@@ -0,0 +1,168 @@
...
+      printBasicHelp();
+      return -1;
+    }
+    String jobType = args[0];
+    List<byte[]> results = new ArrayList<>();
+    if ("fixed".equals(jobType)) {
+      FixedCliParser fixedParser = new FixedCliParser();
+      FixedCliConfig config = null;
+      try {
+        config = fixedParser.parse(Arrays.copyOfRange(args, 1, args.length));
```
cestella: Why are we starting at 1 here?  Is it to skip over some hadoop configs?  If so, see above comment about GenericOptionsParser for a more general approach.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/QueryCliParser.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Parses query pcap filter options and required parameters common to all filter types.
+   *
+   * @param args command line arguments to parse
+   * @return Configuration tailored to query pcap queries
+   * @throws ParseException
+   */
+  public QueryCliConfig parse(String[] args) throws ParseException {
+    CommandLineParser parser = new BasicParser();
```
cestella: All the existing utilities use PosixParser, can we move to that to conform?

metron-platform/metron-pcap-backend/src/main/scripts/pcap_query.sh
```diff
@@ -0,0 +1,34 @@
...
+# Autodetect JAVA_HOME if not defined
+if [ -e /usr/libexec/bigtop-detect-javahome ]; then
+  . /usr/libexec/bigtop-detect-javahome
+elif [ -e /usr/lib/bigtop-utils/bigtop-detect-javahome ]; then
+  . /usr/lib/bigtop-utils/bigtop-detect-javahome
+fi
+
+export METRON_VERSION=${project.version}
+export METRON_HOME=/usr/metron/$METRON_VERSION
+export API_JAR=${project.artifactId}-$METRON_VERSION.jar
```
cestella: Can we call the `API_JAR` variable something else since it's coming from the metron-pcap-backend project?  Perhaps `PCAP_BACKEND_JAR`?

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
```diff
@@ -0,0 +1,168 @@
...
+      } catch (ParseException e) {
+        System.out.println(e.getMessage());
```
cestella: We should probably print to stderr here.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
```diff
@@ -0,0 +1,168 @@
...
+        fixedParser.printHelp();
```
cestella: Also, would be nice for printHelp to take the stream as an arg

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/CliParser.java
```diff
@@ -0,0 +1,83 @@
...
+    options.addOption(newOption("baseOutputPath", true, String.format("Query result output path. Default is '%s'", CliConfig.BASE_OUTPUT_PATH_DEFAULT)));
+    options.addOption(newOption("startTime", true, "Packet start time range. Default is '0'"));
```
cestella: start time should be required.  As it stands, the defaults, if unspecified is to return everything.  That is likely not what we want and will result in a VERY big MR job in some of the installations.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
mmiklavc: I did that for Context-sensitive arguments, e.g.
$ ./pcap_query.sh fixed -srcAddr 1 -destAddr 2 -srcPort 8080 -destPort 8081...
and
$ ./pcap_query.sh query -query "my query string"...

Is there a better way to handle this in Commons CLI? groupoptions sounds correct, but it's a set of mutually exclusive options.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/CliParser.java
cestella: Can we specify a date format to use to parse the start and end times?  I suspect that's a useful piece of sugar as it would require people to not have to look up the epoch time translation every time they want to run this for a time range.  I'd just pass an arg with the date format to use.  If it's not specified, then assume epoch time.

metron-platform/metron-pcap-backend/src/main/scripts/pcap_query.sh
mmiklavc: Good catch - originally had this in the API project.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
cestella: Nope, this is fine by me.  I just wanted to make sure it wasn't to skip over hadoop options because we weren't using GenericOptionsParser.  I retract the comment. :)

cestella:
Overall outside of the things I mentioned, this looks really great.  Solid contribution @mmiklavc.  The only other thing that I would add is to please document this as part of the README.md in pcap-backend  A simple description and the possible options for the CLI in both modes would be sufficient.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/FixedCliParser.java
```diff
@@ -0,0 +1,66 @@
...
+public class FixedCliParser extends CliParser {
+  private Options fixedOptions;
+
+  public FixedCliParser() {
+    fixedOptions = buildFixedOptions();
+  }
+
+  private Options buildFixedOptions() {
+    Options options = buildOptions();
+    options.addOption(newOption("srcAddr", true, "Source IP address"));
```
cestella: Can we conform to the naming convention around src/dst addr/port and change:
* `srcAddr` to `ip_src_addr` as the long option
* `dstAddr` to `ip_dst_addr` as the long option
* `srcPort` to `ip_src_port` as the long option
* `dstPort` to `ip_dst_port` as the long option

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
mmiklavc: I'm going to leave this as is for now - attempted to swap this out in my tests (passed a printwriter to the commons help method) and it just spun indefinitely.

james-sirota:
I agree with Casey.  We need more docs on the CLI.  What can I query for?  Which commands should I run to validate this?

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
cestella: This is addressed

metron-platform/metron-pcap-backend/src/test/java/org/apache/metron/pcap/query/PcapCliTest.java
```diff
@@ -0,0 +1,275 @@
...
+  @Before
+  public void setup() {
+    MockitoAnnotations.initMocks(this);
+  }
+
+  @Test
+  public void runs_fixed_pcap_filter_job_with_default_argument_list() throws Exception {
+    String[] args = {
+            "fixed",
+            "-start_time", "500",
```
cestella: I thought these were long args now, how does - work as opposed to --?

cestella:
+1 from me

dlyle65535:
+1, looks good!

PR #157, opened 2016-06-16
METRON-224 Metron should build from top-level directory

dlyle65535:
NOTE - During testing of this PR, I found an issue with Apache Rat on Master. The pom in metron-platform is not currently being checked. I'd like to address that as a separate issue.


nickwallen:
+1 I was able to package from the top-level and see that Travis still runs the integration tests.

james-sirota:
+ 1 was able to build from the top level pom.  passing tests.  thanks for your contribution, dave

PR #159, opened 2016-06-17
METRON-237 Remove metron-ui from the code base

mmiklavc:
Removing metron-ui from code base as it is no longer used.

Related Jira - https://issues.apache.org/jira/browse/METRON-237


cestella:
If we're going to remove metron-ui, which I completely support btw, we are going to need to trim down the  sections in the LICENSE around metron-ui components.

dlyle65535:
Good deal @mmiklavc - always happy to see dead code go away. Did you run this on Vagrant and/or EC2?

mmiklavc:
@dlyle65535 Ran on Vagrant quick dev - verified Kibana UI comes up and Storm topologies are running.

dlyle65535:
+1!

cestella:
+1, looks good

PR #160, opened 2016-06-17
METRON-233: Expose Kafka producer and consumer configs for parser topologies

cestella:
We need the ability to tune KafkaSpout and KafkaWriter properties in a more granular way.  Currently we do not expose more than a few parameters when creating the KafkaProducer in the KafkaWriter and do not expose almost any of the KafkaSpout configurations.

This PR should allow for the KafkaWriter to have arbitrary KafkaProducer configs exposed in the zookeeper config and have the full range of KafkaSpout configs be exposed from the CLI used to start parser topologies.


james-sirota:
+1 from me. these features are needed to aid in tuning Metron

merrimanr:
+1

PR #161, opened 2016-06-20
Develop

ajayydv:


cestella:
So, a couple of things.  Please refactor this into separate PRs associated with JIRAs.  As of now, it's just impossible to review and we don't really know what we're getting.  It looks like a bunch of parsers, so maybe one JIRA per parser?  Also, name the PRs with the JIRA name as the first part of the name.  For instance, if your JIRA is METRON-123, then PR should be METRON-123: description.

Thanks for the contribution, though!

ajayydv:
by mistake

PR #162, opened 2016-06-20
METRON-239: NOOP Bulk Message Writer

cestella:
We should have a NOOP Writer that will sleep at random. This will help differentiate write slowdowns between index misconfiguration and storm/kafka misconfiguration.

It should be configurable from the object as well as from zookeeper.
* In zookeeper, the "noopLatency" field should be a string defining the latency.
* On the object, there should be a "withLatency" function which allows you to define the latency.

The latency is defined via string in one of two modes:
* An integer which provides a fixed latency
* A comma separated range (i.e. "10, 20") which allows for a random latency between the bounds.


james-sirota:
+1 from me.  this feature is needed to aid in debugging

dlyle65535:
💯  - worked like a champ on live system.

PR #165, opened 2016-06-21
METRON-245 mysql_server role needs to explicitly install defaults

dlyle65535:
Tested on quick-dev-platform and bare metal install.


james-sirota:
+1 Ran it up on AWS

nickwallen:
+1 Validated with Quick Dev

PR #166, opened 2016-06-21
METRON-247 Deployment fails on hosts with no 'eth0' network interface

nickwallen:
Snort was previously using the IPv4 address of the `eth0` network interface as its HOME_NET.  This is a problem, of course, if there is no such interface.  The deployment was failing when run on hosts that do not have an `eth0` network interface.

This fix uses the IPv4 address of the `sniff_interface` as a sensible default.  This works fine for most development or demo builds of Metron.  For other deployments, the user can override `snort_home_net` to define this value appropriately for their environment.

In the case of using canned or replayed pcap data for testing, this value also needs to be adjusted based on the pcap data.  For this scenario, I altered 'sensor-test-mode' to use a `HOME_NET` of `any` so that it correctly alerts on every packet seen.

### DIFF
--- a/metron-deployment/amazon-ec2/playbook.yml
+++ b/metron-deployment/amazon-ec2/playbook.yml
@@ -62,6 +62,7 @@
     - include: tasks/check-volume.yml vol_name=xvda vol_src=/dev/xvda vol_size={{ xvda_vol_size }}
   tags:
     - ec2
+    - ec2-mount
 
 #
 # build the metron cluster
--- a/metron-deployment/amazon-ec2/playbook.yml
+++ b/metron-deployment/amazon-ec2/playbook.yml
@@ -62,7 +62,6 @@
     - include: tasks/check-volume.yml vol_name=xvda vol_src=/dev/xvda vol_size={{ xvda_vol_size }}
   tags:
     - ec2
-    - ec2-mount
 
 #
 # build the metron cluster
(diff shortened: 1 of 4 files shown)


metron-deployment/amazon-ec2/playbook.yml
```diff
@@ -62,6 +62,7 @@
     - include: tasks/check-volume.yml vol_name=xvda vol_src=/dev/xvda vol_size={{ xvda_vol_size }}
   tags:
     - ec2
+    - ec2-mount
```
dlyle65535: Did this come in by mistake?

dlyle65535:
Thanks for taking this on. I assume this was Vagranted, did you do any other testing?

metron-deployment/amazon-ec2/playbook.yml
nickwallen: No.  This allows us to skip the timely disk mount step.

nickwallen:
Tested on Vagrant and AWS

metron-deployment/amazon-ec2/playbook.yml
dlyle65535: I think the timely part is the wait, so --skip-tags="wait" usually works for me, but if you really want this, couldn't it be part of a JIRA that addresses performance rather than the snort deploy bug?

james-sirota:
+1 Ran it up on AWS

cestella:
+1

PR #169, opened 2016-06-22
METRON-252 Metron Dashboard Can Be Broken By Adding Index

nickwallen:
[METRON-252](https://issues.apache.org/jira/browse/METRON-252)

#### Problem

There are some visuals in the default dashboard that cross multiple indices.  The only way to do this via Kibana's web interface, is to create a "*" or wildcard index.  Unfortunately, this can also pull in other unrelated indices that could potentially impact the functioning of the dashboard.

For example, in this specific scenario an additional index was added that had the same field name, but a different data type than Snort, YAF, and Bro.  This thoroughly confused Kibana and caused it to report the following error when viewing Metron's Dashboard.

```
Error: [unsupported_operation_exception] custom format isn't supported
```

#### Solution

If you modify the JSON used to create the .kibana index manually, there is a way to specify multiple indexes for a visualization instead of using a wildcard index pattern.  Previously, there were multiple visuals with index references as follows.

```
{\"index\":\"*\"
```

These have all been changed to specifically reference the 3 default sensors.

```
{\"index\":[\"yaf*\", \"bro*\", \"snort*\"]
```

This ensures that no other indices can impact the default dashboard.

This was tested by deploying to an AWS cluster.  The dashboard was able to function in the presence of the same index that had previously broken it.


james-sirota:
+ 1

cestella:
+1, spun up in vagrant

PR #170, opened 2016-06-22
METRON-253: Reloading parser configurations from zookeeper doesn't call init()

cestella:
This results in a NPE if you do not specify an output column (this output column gets inferred as part of init())

To validate this:
* Get a running instance of metron
* Run `/usr/metron/0.2.0BETA/bin/zk_load_configs.sh -m PUSH -i /usr/metron/0.2.0BETA/config/zookeeper -z node1:2181`
* Wait a full 2 minutes
* Check the `parser_error` topic via the kafka console consumer and ensure you don't see any weird NPEs

The problem here is that init() is not getting run on the sensor parser config, so some syntactic sugar in the FieldTransformer isn't getting updated.  Specifically, if you only specify `input` and not `output`, then it will NPE.


merrimanr:
+1

nickwallen:
+1

PR #171, opened 2016-06-23
METRON-254 pcap inspector emits fields that are named based on the enum, rather than the standardized field names

cestella:
PcapInspector utility prints field names that are enum names (i.e. SRC_ADDR) as opposed to the standard field names (ip_src_addr)

Check this by running some pcap data through (follow steps 1-7 from https://github.com/apache/incubator-metron/pull/156)


james-sirota:
+1

nickwallen:
+1 Looks solid

PR #172, opened 2016-06-23
METRON-255: Expose the IPProtocolTransformer as a function in the Stellar transformation language

cestella:
We have a TransformationFunction to map IANA protocol numbers to normalized strings. This should be exposed in the query language.


james-sirota:
+1

nickwallen:
+1 Looks good

PR #173, opened 2016-06-23
METRON-256: Make the PCap values returned ordered by timestamp

cestella:
Right now we are not specifying the partitioner for the MR job to filter PCaps as part of PCap retrieval, so we are getting back packets in the wrong order.


james-sirota:
+1

nickwallen:
+1 pending once the transient issues in the CI build is resolved

PR #174, opened 2016-06-23
METRON-249: Field Transformation functions fail to handle invalid user inputs

cestella:
The field transformation functions fail to handle invalid user input.  Specifically it could not handle negative numbers, so `URL_TO_HOST(-123)` fails with a parser exception rather than handling this gracefully.


james-sirota:
+1

PR #176, opened 2016-06-24
METRON-259 Using 'any' for Snort's HOME_NETWORK

nickwallen:
In some cases, Ansible is not yet aware of the 'tap0' interface when the Snort deployment begins.  This causes the deployment to fail with this error.

```
ERROR! ERROR! 'dict object' has no attribute u'ansible_tap0
```

It is not really necessary to set the HOME_NETWORK of Snort to the IP address of the sniff interface.  This needs to be customized for Snort on a case-by-case basis for production deployments based on the target environment.  For demo and development purposes, `any` works just fine and is much simpler.

### DIFF
--- a/metron-deployment/roles/snort/defaults/main.yml
+++ b/metron-deployment/roles/snort/defaults/main.yml
@@ -23,4 +23,4 @@ snort_src_url: "https://snort.org/downloads/archive/snort/snort-{{ snort_version
 snort_community_rules_url: "https://www.snort.org/downloads/community/community-rules.tar.gz"
 dag_src_url: "https://snort.org/downloads/snort/daq-{{ daq_version }}.src.rpm"
 sniff_interface: eth0
-snort_home_net: "{{ hostvars[inventory_hostname]['ansible_' + sniff_interface]['ipv4']['address'] }}"
+snort_home_net: any


dlyle65535:
I don't agree that using "any" works fine for either general or demonstration/development purposes.  If we don't sniff the same interface with all the sensors, we get events that are uncorrelated with the rest of the sensor stack. Additionally, we'll see events that have no pcap backup. 

Kind of related - there is a a production use case for the tap0 interface. You can direct all interfaces of interest to to the tap interface and sniff that.

Is there any reason to not investigate making this work using the sniff interface before falling back to "any"?

nickwallen:
AFAIK, the HOME_NETWORK is used primarily to designate which traffic should be validated against the rule set.  If Snort sees traffic where the source or destination is NOT associated with the HOME_NETWORK then it will not alert.  

For replaying raw pcap in our demo environment, I want all traffic to be run against the rule set.  Setting HOME_NETWORK to 'any' achieves this.  What do you think?

nickwallen:
The current code base is trying to set the HOME_NETWORK to the IP of the sniff interface.  I can't really think of any conditions under which you would want the HOME_NETWORK set to the IP of the sniff interface.  Maybe I'm wrong though.

dlyle65535:
Ah, that's a good clarification- I guess what we need is two things:

1) HOME_NETWORK to any (which you have done).
2) Configure the snort IFACE to be the sniff_interface. 

So I had the corrective action wrong in METRON-247. If I'm not mistaken, after this PR, Snort will still listen on the first interface by default, so we won't see the traffic we expect in many situations. Having looked closer, I think the above 2 would have been the actual fix. 

Thoughts?

james-sirota:
+1

nickwallen:
@dlyle65535 the code already configures Snort to listen on the `sniff_interface`. See bottom of `metron-deployment/roles/snort/snort.yml`

dlyle65535:
@nickwallen - so it does. Good deal. I'm +1.

Thanks!

PR #180, opened 2016-07-06
METRON-281 Simplify deployment to alternative AWS region

nickwallen:
[METRON-281](https://issues.apache.org/jira/browse/METRON-281)

To use an alternative AWS region, the user has to manually define the new AWS region along with lookup and redefine the Centos 6 AMI identifier that applies for their chosen region. The AMI identifiers can be difficult to find as they are not on the primary page for the CentOS 6 image.

This was simplified so that a user only has to specify the region they wish to deploy Metron to. The user no longer has to lookup and redefine the AMI identifier.  At the same time, the user can choose to override the `image` setting to override and use a specific AMI identifier, if desired.

### DIFF
--- a/metron-deployment/amazon-ec2/conf/defaults.yml
+++ b/metron-deployment/amazon-ec2/conf/defaults.yml
@@ -31,16 +31,30 @@ services_to_start:
   - snort-parser
   - enrichment
 
+# the ami for cent6 by region
+amis_by_region:
+  us-east-1:        ami-1c221e76    # US East (N. Virginia)
+  us-west-2:        ami-05cf2265    # US West (Oregon)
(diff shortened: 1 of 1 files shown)


dlyle65535:
+1 glad to have this. Puzzled why west stopped working though. I'll try to remember to dig into that. Think it's worth a Jira?

nickwallen:
Thanks. I wasn't sure if it deserved a JIRA or not.  Since it works in all AWS regions except us-west-2, I was assuming that it is more likely to be something about my AWS environment and not something that is Metron specific.  On the other hand, I don't really know.  It could be a Metron problem.

cestella:
+1 by inspection

PR #181, opened 2016-07-06
METRON-279 Removed duplicative roles; metron-kafka and metron-hbase

nickwallen:
### [METRON-279](https://issues.apache.org/jira/browse/METRON-279)

* Removed duplicative roles; metron-kafka and metron-hbase.
* Added creation of `parser_error` and `parser_invalid` topics as part of standard deployment.  These are needed per METRON-223.


dlyle65535:
So, if I'm not mistaken, METRON-279 was reported because the parser_invalid and parser_error topics were not created during deployment. The root cause was the duplicated roles causing confusion. So, unless I'm mistaken, METRON-279 won't be complete unless those topics are created by the deployment scripts. It's kind of buried in the commentary, but perhaps Neha can confirm/deny that.

nickwallen:
Added creation of parser_error and parser_invalid topics as part of standard deployment. These are needed per METRON-223.

dlyle65535:
+1, looks good!

cestella:
+1, by inspection

PR #184, opened 2016-07-07
METRON-284: SensorParserConfig object cannot be serialized with the default Jackson ObjectMapper	

cestella:
The SensorParserConfig object cannot be serialized without a custom Jackson ObjectMapper. Specifically, this is caused by the way the FieldTransformer is implemented. A call to SensorParserConfig.toJSON() will fail with this message if a transformer is configured:
No serializer found for class org.apache.metron.common.field.transformation.MTLTransformation and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) )


cestella:
I took the liberty to also add a unit test to ensure that the *enrichment* configs also serialize and deserialize properly.

merrimanr:
+1

dlyle65535:
+1 all tests pass.

PR #186, opened 2016-07-11
METRON-298 Remove the effective_tld_names.dat files

cestella:
Remove these files because they are mozilla licensed and seem to cause problems during the releases. We have a different way of doing this now and the adapter using them should be mothballed.
Be sure to remove the MPL reference in the LICENSE file too


james-sirota:
+1

james-sirota:
I ran this up in AWS and it worked ok.  Did anyone else check?

cestella:
Just for posterity, I ran it up in vagrant

dlyle65535:
I've been trying, but it's AWS is fighting me.

PR #187, opened 2016-07-13
Metron-321 Small update to and cleanup of .gitignore

mmiklavc:
https://issues.apache.org/jira/browse/METRON-321

Small update to and cleanup of .gitignore


cestella:
+1.  In the future can you please name the PR starting with the full JIRA name followed by the description, like so: "METRON-321: Add items to .gitignore".  The full JIRA name (with the dash) triggers comments to be copied from the PR to the JIRA and the full description helps me know what the heck it is without going to the JIRA. ;)

dlyle65535:
+1, looks good.

mmiklavc:
Apologies about the titles missing text - did github defaults change bc I don't recall having to set that before?

PR #188, opened 2016-07-13
METRON-227 (Time based flushing)

ajayydv:
METRON-227 (Time based flushing)

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterComponent.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterComponent.java
@@ -34,21 +34,37 @@ import java.util.*;
 import java.util.function.Function;
 
 public class BulkWriterComponent<MESSAGE_T> {
-  public static final Logger LOG = LoggerFactory
-            .getLogger(BulkWriterComponent.class);
+  public static final Logger LOG = LoggerFactory.getLogger(BulkWriterComponent.class);
   private Map<String, Collection<Tuple>> sensorTupleMap = new HashMap<>();
   private Map<String, List<MESSAGE_T>> sensorMessageMap = new HashMap<>();
   private OutputCollector collector;
   private boolean handleCommit = true;
   private boolean handleError = true;
+  private Long lastFlushTime;
+  private Long flushIntervalInMs;
+  private boolean flush = false;
+
+
   public BulkWriterComponent(OutputCollector collector) {
     this.collector = collector;
+    this.lastFlushTime = System.currentTimeMillis();
+    this.flush = false;
   }
 
   public BulkWriterComponent(OutputCollector collector, boolean handleCommit, boolean handleError) {
     this(collector);
     this.handleCommit = handleCommit;
     this.handleError = handleError;
+    this.lastFlushTime = System.currentTimeMillis();
+    this.flush = false;
+  }
+
+  public void setFlush(boolean flush) {
+    this.flush = flush;
+  }
+
+  public void setFlushIntervalInMs(Long flushIntervalInMs) {
+    this.flushIntervalInMs = flushIntervalInMs;
   }
 
   public void commit(Iterable<Tuple> tuples) {
@@ -84,12 +100,8 @@ public class BulkWriterComponent<MESSAGE_T> {
     sensorTupleMap.remove(sensorType);
     sensorMessageMap.remove(sensorType);
   }
-  public void write( String sensorType
-                   , Tuple tuple
-                   , MESSAGE_T message
-                   , BulkMessageWriter<MESSAGE_T> bulkMessageWriter
-                   , WriterConfiguration configurations
-                   ) throws Exception
+
+  public void write( String sensorType, Tuple tuple, MESSAGE_T message, BulkMessageWriter<MESSAGE_T> bulkMessageWriter, WriterConfiguration configurations) throws Exception
   {
     int batchSize = configurations.getBatchSize(sensorType);
     Collection<Tuple> tupleList = sensorTupleMap.get(sensorType);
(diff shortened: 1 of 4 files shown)


metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterComponent.java
```diff
@@ -34,21 +34,37 @@
...
 
   public BulkWriterComponent(OutputCollector collector, boolean handleCommit, boolean handleError) {
     this(collector);
     this.handleCommit = handleCommit;
     this.handleError = handleError;
+    this.lastFlushTime = System.currentTimeMillis();
+    this.flush = false;
+  }
+
+  public void setFlush(boolean flush) {
```
cestella: Since you're deciding whether to flush based on variables in the global config, we don't need member variables and setters for `flush` and `flushInterval`.  I don't like setting a member variable in the class every time we call write and we don't need it, they can be local variables to write

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterComponent.java
```diff
@@ -34,21 +34,37 @@
...
     this.collector = collector;
+    this.lastFlushTime = System.currentTimeMillis();
+    this.flush = false;
   }
 
   public BulkWriterComponent(OutputCollector collector, boolean handleCommit, boolean handleError) {
     this(collector);
     this.handleCommit = handleCommit;
     this.handleError = handleError;
+    this.lastFlushTime = System.currentTimeMillis();
```
cestella: These guys aren't necessary because you're calling the constructor on line 55 which sets them.

metron-platform/metron-integration-test/src/main/config/zookeeper/global.json
```diff
@@ -7,10 +7,12 @@
   "solr.collection": "metron",
   "solr.numShards": 1,
   "solr.replicationFactor": 1,
+  "flushIntervalInMs": "20000",
+  "flushOnTimeout": "false",
```
cestella: Do we need both parameters?  If we have a flush interval set, we should infer that we want to do a time-based flush, right?

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterComponent.java
```diff
@@ -103,10 +115,21 @@ public void write( String sensorType
     }
     messageList.add(message);
 
-    if (tupleList.size() < batchSize) {
-      sensorTupleMap.put(sensorType, tupleList);
-      sensorMessageMap.put(sensorType, messageList);
-    } else {
+    if(configurations.getGlobalConfig()!=null&&configurations.getGlobalConfig().get(Constants.TIME_FLUSH_FLAG)!=null)
```
cestella: I'd recommend saving off the global config in a member variable and reusing that reference.  It's possible the configuration will get updated between calls to `configurations.getGlobalConfig()` in zookeeper and therefore you'll get two separate configs.

cestella:
Great contribution, thanks!  I had a few comments, but good job. :)

cestella:
Whoops, forgot, we should probably have a unit test addition given this functionality.

metron-platform/metron-integration-test/src/main/config/zookeeper/global.json
ajayydv: Agree that flushIntervalInMs is dependent on other flag but if we think from user perspective both have different meaning.

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterComponent.java
ajayydv: agree

metron-platform/metron-integration-test/src/main/config/zookeeper/global.json
cestella: What would you do if the flush flag were set to true and the flushIntervalInMs flag is unset?

ajayydv:
added test case

mmiklavc:
How about using Storm's built-in tick tuples for this?

https://groups.google.com/forum/#!msg/storm-user/8addaQm3OT4/0OQfSgQkRwEJ

cestella:
@mmiklavc That is absolutely what the tick tuple was designed for. What are the pro/con versus the current proposed approach?

mmiklavc:
It is a much simpler design and it leverages Storm's built-in mechanisms for performing actions like DB flushes. Also, you don't have to worry about multi-threading in your bolt if you want to flush prior to your buffer being full when you haven't received a tuple in a while. e.g. I have 99 tuples and need 100 for a flush, but don't receive a tuple for n minutes - without using tick tuples, you either just sit and wait, add a thread that needs to check internally, or add another external bolt with a clock. 

I'm not aware of any cons to using the TT approach. This is what tick tuples were built for.

cestella:
There is one downside that I see, but I think it's acceptable.  We currently keep configuration around the indexing in zookeeper so that the topology configuration can be adjusted on the fly in a running topology.  This would externalize that and keep the tick frequency in flux, so to change it would require a topology restart.

I think that this is acceptable, but it's worth pointing out.

mmiklavc:
Any new comments or actions for this PR?

mattf-horton:
Michael, I'm in the process of reviewing this, and also pinging Ajay Yadav, in the context of METRON-322, to confirm he is not still active on this issue.  Assuming he isn't, I will change this implementation to use Storm Tick Tuples, as this seems clearly the right path to follow.
Thanks.

mattf-horton:
-1.  Upon closer review, I find that there is no timer thread or similar mechanism implemented in this patch.  Thus, the patch only works if there is a steady trickle of new tuples to trigger the time comparison.  But if a burst of tuples are enqueued, and then no more tuples happen for a long time, the queue will languish without flushing until another tuple triggers the time comparison, potentially much later than the desired timeout.
@mmiklavc 's suggestion to use Tick Tuples instead is precisely the correct answer, as it avoids the considerable complexity of implementing a (robust) timer thread.
Since this PR has been dormant since July, and the contributor has not responded to queries on the Jira, I will take over the ticket, unless a more senior member of the community wishes to do so.
I recommend that this PR be closed without committing, and I will do the related work in the context of METRON-322, which I am also assigning to.  Thanks.

james-sirota:
I agree on the -1.  There are additionally merge conflicts now with this pull request and I don't believe it's worthwhile to do the merge and change the implementation. It's easier to start over.  Matt, the Jira as been re-assigned to you.  Thank you for taking this over.   Lets let this sit for 72 hours and if no one objects we will cancel this pull request.

nickwallen:
Opened https://issues.apache.org/jira/browse/INFRA-12959 for Apache Infra to close this PR.

PR #191, opened 2016-07-14
Metron-214 Build RPM Packages for Deployment

mmiklavc:
This PR includes multiple sub-tasks as well.
https://issues.apache.org/jira/browse/METRON-214

METRON-299 Build Parsers Deployment RPM
METRON-300 Build Enrichment Deployment RPM
METRON-301 Build Elasticsearch Deployment RPM
METRON-302 Build Common Deployment RPM
METRON-303 Build Dataloads Deployment RPM
METRON-304 Build PCAP Deployment RPM
METRON-305 Build Solr Deployment RPM

**Instructions for Testing**

- Run the metron build from incubutor-metron (project root)

`mvn clean install`

- Copy the build artifacts to the rpm SOURCES directory

`cd incubator-metron/metron-deployment/packaging/docker/rpm-docker && ./copysources.sh`

- Install Docker on your host machine - [Docker Install Guide](https://docs.docker.com/engine/installation/)
- Run Docker image and mount the rpm-docker directory

`docker run -d -P --name rpm-docker -v <absolute-path>/incubator-metron/metron-deployment/packaging/docker/rpm-docker:/root/rpmbuild -it dlyle65535/rpm-docker`

- ...If you need to re-run this for any reason, you can delete the docker instance by running

`docker rm rpm-docker`

- Login to the container

`docker exec -it rpm-docker /bin/bash`

- Execute the rpm build - this cleans the rpm dirs, runs the build, and executes verification via rpmlint. Verify no Exceptions occur during the build, and that rpmlint also completes without exceptions (you'll see some warnings, but these are OK)

```
cd rpmbuild
./build.sh
```

- Install the rpm packages

`for i in rpmbuild/RPMS/noarch/*.rpm; do rpm -ivh $i; done`

- Check the installation - verify expected files appear in the metron install directory

`find /usr/metron`

- Uninstall and verify all files removed

`for i in $(rpm -qa|grep metron); do rpm -e $i; done`

- `/usr/metron` should no longer exist


cestella:
Why do we not have a separate PR per JIRA?

It appears that the various tasks have various authors, wouldn't it be better to have each author submit their own JIRA so that they can be listed on the commit?

Also, this isn't the complete set of tasks for METRON-214, so when we close this JIRA, it won't be the complete body of work for METRON-214, just a portion.

dlyle65535:
That's a good question. If we don't squash this pr prior to committing, will that give the authors the proper credit?

On the second, this is the completion of METRON-214.

cestella:
Ok, after some out of band discussion, this is fine..lots of small subtasks that had a lot of interdependency.  I'm +1 upon inspection.

james-sirota:
+1

PR #196, opened 2016-07-19
METRON-334 Travis cache Maven dependencies

justinleet:
Adding .m2 directory to caching, so that it is reused on subsequent builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ jdk:
 script:
   - |
     mvn -q integration-test install
+cache:
+  directories:
+  - $HOME/.m2


justinleet:
Initial Travis build, from before the cache is populated, downloads everything
https://travis-ci.org/justinleet/incubator-metron/builds/145862346#L174

Afterwards, on a whitespace commit no longer downloads the dependencies
https://travis-ci.org/justinleet/incubator-metron/builds/145870931#L171

mmiklavc:
+1 (non-binding)
28 min vs 27:40... I had hoped this would have made a bigger impact

justinleet:
The timings are really inconsistent in general. Some builds (essentially the same) earlier were taking anywhere as low as 17 minutes.  It makes it hard to tell how worthwhile timings are, even in close proximity.

dlyle65535:
+1, worked and created a 975MB cache.

cestella:
+1 via inspection.  Great job, Justin.

PR #198, opened 2016-07-20
METRON-336 Update README.md to use brew instead of pip for ansible 2.0.0.2 install

jfrazee:
This is the same as #197 but rebased correctly.

### DIFF
--- a/metron-deployment/amazon-ec2/README.md
+++ b/metron-deployment/amazon-ec2/README.md
@@ -27,7 +27,9 @@ Any platform that supports these tools is suitable, but the following instructio
 2. With Homebrew installed, run the following command in a terminal to install all of the required tools.
 
   ```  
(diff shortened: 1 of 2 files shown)


dlyle65535:
Hi,

Thanks for the contribution. Is there a reason you prefer brew over pip?

jfrazee:
@dlyle65535 It's mostly about consistency I think, but it should also be easier (not impossible) to manage the multiple ansible versions inside homebrew than with pip. Include both sets of instructions?

dlyle65535:
I don't really have a strong preference, just curious.

cestella:
I get ya, yeah, we have half the instructions in brew and then the ansible switching in pip.    Sounds good, I'm +1.

nickwallen:
+1

PR #200, opened 2016-07-20
METRON-333 Add excludes to metron.spec for test.yaml files

mmiklavc:
This PR completes https://issues.apache.org/jira/browse/METRON-333

In order to test, follow same instructions from https://github.com/apache/incubator-metron/pull/191#issue-165652191, but simply verify that no "test.yaml" files show up in the flux directory after installing the rpm's.


cestella:
Honestly, we don't want ANY of those flux topologies for the parsers anymore.  They should be removed from the RPM as well as from the underlying project.  Is there a JIRA to complete the removal of these?

mmiklavc:
@cestella That's an excellent point. I saw the test.yaml files didn't belong, but in reality we're not using any of this. I'll modify the Jira and PR to reflect this.

PR #201, opened 2016-07-21
METRON-338 Add indexing rpm. Remove dead flux code from metron-parsers

mmiklavc:
Addresses https://issues.apache.org/jira/browse/METRON-338

**To test:**

There are a few things to check

1. You can check that the parser rpm does not include the flux files any longer - https://github.com/apache/incubator-metron/pull/191#issue-165652191
2. Also check that an indexing rpm is created
3. Spin up a vagrant instance and verify parser -> enrichment -> indexing still works as intended.


dlyle65535:
+1 looks good

metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
```diff
@@ -47,7 +47,8 @@ Source2:        metron-elasticsearch-%{full_version}-archive.tar.gz
 Source3:        metron-data-management-%{full_version}-archive.tar.gz
 Source4:        metron-solr-%{full_version}-archive.tar.gz
 Source5:        metron-enrichment-%{full_version}-archive.tar.gz
-Source6:        metron-pcap-backend-%{full_version}-archive.tar.gz
+Source6:        metron-indexing-%{full_version}-archive.tar.gz
```
cestella: So, is the Source${x} naming convention imposed by spec?  Seems like we might have the opportunity to name things more descriptively.  I don't mind it for this PR and it's not a big thing..I'm just curious.

cestella:
+1, ship it!

metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
mmiklavc: Yeah, I believe it is - https://docs.fedoraproject.org/en-US/Fedora_Draft_Documentation/0.1/html-single/RPM_Guide/#id566895

PR #207, opened 2016-08-08
METRON-362: Unify the Stellar Languages

cestella:
At the moment, stellar has some architectural issues:
* The query and transformation languages are distinct despite sharing the same transformation functions
* The transformation language does not have any notion of the comparison operations or the logical (aka boolean) functions from the query language
* Neither language has the ability to do arithmetic or even represent negative numbers as constants
* Neither language has the ability to do simple if/then/else constructs

We should unify the languages and correct the deficiencies as they have come up in multiple situations. This should be done in a backwards compatible way so that deployed stellar statements as part of field transformations or threat triage do not have to change.

This is completely tested via the unit test `StellarTest` in `metron-common` as well as the existing integration tests around FieldTransformations and Threat Triage.


merrimanr:
+1 by inspection

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
```diff
@@ -0,0 +1,422 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class StellarTest {
+  @Test
+  public void testIfThenElse() {
+    {
+      String query = "if 1 < 2 then 'one' else 'two'";
+      Assert.assertEquals("one", run(query, ImmutableMap.of("foo", "casey ")));
```
nickwallen: What does adding the variable foo='casey' have to do with these tests?

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
```diff
@@ -0,0 +1,422 @@
...
+    Assert.assertEquals("www", run("GET_FIRST(SPLIT(DOMAIN_REMOVE_TLD(foo), '.'))", variables));
+    Assert.assertEquals("www", run("GET(SPLIT(DOMAIN_REMOVE_TLD(foo), '.'), 0)", variables));
+    Assert.assertEquals("google", run("GET_LAST(SPLIT(DOMAIN_REMOVE_TLD(foo), '.'))", variables));
+    Assert.assertEquals("google", run("GET(SPLIT(DOMAIN_REMOVE_TLD(foo), '.'), 1)", variables));
+  }
+  private static Object run(String rule, Map<String, Object> variables) {
+    StellarProcessor processor = new StellarProcessor();
+    Assert.assertTrue(rule + " not valid.", processor.validate(rule));
+    return processor.parse(rule, x -> variables.get(x));
+  }@Test
```
nickwallen: Easy enough to have Intellij fix the spacing?  Sometimes we have spaces between methods, sometimes not, here there is no line break at all.  All simple, but add up to making it more difficult to read than it might need to be.

nickwallen:
What is the reasoning to have Stellar not share a common parent package?  We have org.apache.metron.common.dsl, also org.apache.metron.common.query, org.apache.metron.common.transformation.  Aren't all these packages focused on Stellar?  Maybe org.apache.metron.common.stellar?

nickwallen:
Why is there still both a PredicateProcessor and a TransformationProcessor?  Can't we unify these?

cestella:
@nickwallen There is no longer an org.apache.metron.common.{transformation,query} and there is, as part of this PR an org.apache.metron.common.stellar.  Not sure where you're seeing these packages continuing to exist.

nickwallen:
Oops, maybe i checked out the wrong PR.  That would explain a lot.

cestella:
@nickwallen There is still a PredicateProcessor because having a stellar function which acts as a predicate is a sensible differentiation from a stellar function which returns an arbitrary function.  Just as there is a java.util.function.Function and a java.util.function.Predicate and they are distinct.  They both reuse a base class, however.

cestella:
I will defend the org.apache.metron.common.dsl package.  That is intended to be utilities for domain specific languages (like variable resolvers, etc).  My intention was to open up the possibility to reuse across *other* DSLs that we might need going forward.

nickwallen:
Understand your reasoning on DSL package. Personally, I would collapse 'org.apache.metron.common.dsl' under the stellar package.  If we're going to use it a different way in the future then we can make that change in the future.  But right now, it just confuses the simplicity of having a single parent stellar package.

nickwallen:
It would be good to add Javadocs to the StellarProcessor and related classes that describe your reasoning.  When am I supposed to use one or the other?  Right now there are no comments.

metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/StellarCompiler.java
```diff
@@ -16,276 +16,224 @@
...
-        b = false;
-        break;
-      default:
-        throw new ParseException("Unable to process " + ctx.getText() + " as a boolean constant");
-    }
-    tokenStack.push(new Token<>(b, Boolean.class));
+
+
+  /*@Override
+  public void enterArithExpr_plus(StellarParser.ArithExpr_plusContext ctx) {
```
nickwallen: Assuming we don't need this since it is commented out?

dlyle65535:
Ran it up in quick dev. Worked as expected. +1.

james-sirota:
+1 via inspection

PR #208, opened 2016-08-10
METRON-309 Create a normalcy profiler

nickwallen:
### [METRON-309](https://issues.apache.org/jira/browse/METRON-309)

Created a normalcy profiler, otherwise known as the Metron Profiler.   

The Profiler is a feature extraction mechanism that can generate a profile describing the behavior of an entity on a network. An entity might be a server, user, subnet or application. Once a profile has been generated defining what normal behavior looks-like, it can be used to build models that identify anomalous behavior.

This is achieved by summarizing the streaming telemetry data consumed by Metron over sliding windows. A summary statistic is applied to the data received within a given window. Collecting this summary across many windows results in a time series that is useful for analysis.

### Changes

This PR contains the following high-level changes to Metron.
* A Storm topology that generates Profiles. `metron-analytics/metron-profiler`
* Changes to Ansible to deploy the Profiler topology. `metron-deployment` 
* A new bolt and associated tooling to perform HBase writes from a Storm topology. `metron-platform/metron-hbase`
* Additions to the Stellar language to provide arithmetic and control flow logic required to generate Profiles. `metron-platform/metron-common`

### Testing

Follow the instructions contained in the [README](https://github.com/nickwallen/incubator-metron/tree/METRON-309/metron-analytics/metron-profiler#getting-started) to get your very first Profile running.

### Documentation

* Documentation for the Profiler is contained within the [README](https://github.com/nickwallen/incubator-metron/blob/METRON-309/metron-analytics/metron-profiler/README.md).
* The code itself contains plenty of comments and Javadocs.  
* The design document is attached to the [JIRA](https://issues.apache.org/jira/browse/METRON-309).


metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -0,0 +1,265 @@
...
+   * and emits the ProfileMeasurement.  Clears all state in preparation for
+   * the next window period.
+   */
+  private void flush(Tuple tickTuple) throws IOException {
+    LOG.info(String.format("Flushing profile: profile=%s, entity=%s, start=%d",
+            measurement.getProfileName(), measurement.getEntity(), measurement.getStart()));
+
+    // execute the 'result' expression
+    String resultExpr = profileConfig.getResult();
+    Long result = executor.execute(resultExpr, new JSONObject(), Long.class);
```
nickwallen: I intended for this to be a Double, not a Long.  We expect the profiler to produce a time series, so a large floating point value is what I would expect to be most useful to users.

On the other hand, do I even need knowledge of what type the resulting value is?  It seems more user-friendly If I just treat it as an Object, and then allow the HBaseBolt to serialize it to HBase.  Then each profile can produce a time series of whatever values they prefer.

The only ugly part is that I'd have to an "if-else instanceof" chain to cast it to the appropriate type in ProfileHBaseMapper.columns(Tuple).  I'd have to do this because Java will not be able to resolve Bytes.toBytes(Object).

nickwallen:
@mmiklavc I ran into some issues talking to HBase with the new Profiler topology in the current Quick Dev environment.  Once I did the old-school injection of the site resources everything worked fine.

```
cd /etc/hbase/conf  && jar -uf /usr/metron/0.2.0BETA/lib/metron-profiler-0.2.0BETA.jar hbase-site.xml
cd /etc/hadoop/conf && jar -uf /usr/metron/0.2.0BETA/lib/metron-profiler-0.2.0BETA.jar core-site.xml
cd /etc/hadoop/conf && jar -uf /usr/metron/0.2.0BETA/lib/metron-profiler-0.2.0BETA.jar hdfs-site.xml
```

I am wondering if this is because of [METRON-356](https://github.com/nickwallen/incubator-metron/commit/b9282b438422d56fac23301dc854a39ae7d83a83).  Could the current version of Quick Dev be broken because we removed the site injection from Ansible, but have not updated the Vagrant box for Quick Dev out in Atlas?  

If so, it is not really a problem.  I just want to make sure that I fully understand the root cause for the issue that I am seeing.

metron-analytics/metron-profiler/README.md
```diff
@@ -0,0 +1,229 @@
...
+## Usage
+
+Any field contained within a message can be used to generate a profile.  A profile can even be produced from combining fields that originate in different data sources.  A user has considerable power to transform the data used in a profile by leveraging the Stellar language. A user only need configure the desired profiles in Zookeeper and ensure that the Profiler topology is running.
+
+### Configuration
+
+The Profiler configuration requires a JSON-formatted set of elements, many of which can contain Stellar code.  The configuration contains the following elements.
+
+* `profile` A unique name identifying the profile.
+* `foreach` A separate profile is maintained for each of these.  This is effectively the entity that the profile is describing.  For example, if `ip_src_addr` then a separate profile would be maintained for each unique IP source address.
```
cestella: Should we indicate that these are stellar statements and link to the language ref in metron-common's README?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -0,0 +1,281 @@
...
+
+  @Override
+  public void execute(Tuple input) {
+    try {
+      doExecute(input);
+      collector.ack(input);
+
+    } catch (Throwable e) {
+      LOG.error("exception processing tuple: " + input, e);
+      collector.reportError(e);
```
cestella: Should we be acking here?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileHBaseMapper.java
```diff
@@ -0,0 +1,136 @@
...
+   */
+  private int saltDivisor = 10;
+
+  /**
+   * The name of the column family.
+   */
+  private String columnFamily;
+
+  public ProfileHBaseMapper() {
+    setColumnFamily("cfProfile");
```
cestella: Generally people try to keep the CF name super short (often 1 character) because it's replicated with every data point (see http://hbase.apache.org/0.94/book/rowkey.design.html).  Can we call it 'P' for profile?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileSplitterBolt.java
```diff
@@ -0,0 +1,158 @@
...
+    this.parser = new JSONParser();
+  }
+
+  @Override
+  public void execute(Tuple input) {
+    try {
+      doExecute(input);
+
+    } catch (IllegalArgumentException | ParseException | UnsupportedEncodingException e) {
+      LOG.error("Unexpected exception", e);
```
cestella: Can you bubble the exception message up like `"Unexpected exception: e.getMessage()`?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/stellar/DefaultStellarExecutor.java
```diff
@@ -0,0 +1,116 @@
...
+  @Override
+  public <T> T execute(String expr, JSONObject message, Class<T> clazz) {
+    Object result = execute(expr, message);
+
+    // ensure the result type is as expected
+    if (clazz.isAssignableFrom(result.getClass())) {
+      return (T) result;
+
+    } else {
+      throw new RuntimeException(String.format("Unexpected type: expected=%s, actual=%s, expression=%s",
```
cestella: If you wanted to be extra adaptable, you might consider using the ConversionUtils class to try to convert to the target type if it's not assignable and *then* (if null is returned) throw an exception.  Just a thought.

metron-analytics/metron-profiler/src/test/config/zookeeper/readme-example-1/profiler.json
```diff
@@ -0,0 +1,17 @@
+{
+  "inputTopic": "indexing",
+  "profiles": [
+    {
+      "profile": "example1",
+      "foreach": "ip_src_addr",
+      "onlyif": "protocol == 'HTTP'",
+      "init": {
+        "total_bytes": 0.0
```
cestella: Can init take stellar statements?  There will be situations where we'll want to init via stellar (e.g. bloom filter with specific parameters).  If not, we should make it accept it.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
```diff
@@ -0,0 +1,226 @@
...
+    @Override
+    public HTableInterface getTable(Configuration config, String tableName) throws IOException {
+      return provider.getTable(config, tableName);
+    }
+  }
+
+  public void setup(String pathToConfig) throws Exception {
+
+    // create input messages for the profiler to consume
+    input = Stream.of(message1, message2, message3)
```
cestella: Dig.  +1!

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
```diff
@@ -247,19 +301,24 @@ public static void uploadConfigsToZookeeper(String globalConfigPath, String pars
   }
 
   public static void visitConfigs(CuratorFramework client, ConfigurationVisitor callback) throws Exception {
-    visitConfigs(client, callback, ConfigurationType.GLOBAL);
-    visitConfigs(client, callback, ConfigurationType.PARSER);
-    visitConfigs(client, callback, ConfigurationType.ENRICHMENT);
+    visitConfigs(client, callback, GLOBAL);
```
cestella: Not a big deal, but maybe we could make visitConfigs take a vararg of ConfigurationType and then pass in `GLOBAL`, `PARSER`, `ENRICHMENT`, `PROFILER` in as args

cestella:
So I think this looks good, the one thing that I'd like to see that I didn't is a set of classes around reading the data from the HBase table populated by the profiler.  Things like reading the output of a profile for a given time (or if not specified, then the latest).

metron-analytics/metron-profiler/src/test/config/zookeeper/readme-example-1/profiler.json
nickwallen: Yes, the values within the 'init' block are executed as Stellar expressions.  See ProfileBuilderBolt.init(...).

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/stellar/DefaultStellarExecutor.java
nickwallen: Yes, I think that is a good idea.  If I am adding two integers and I am expecting a float, it would be useful to have that just work.

metron-analytics/metron-profiler/src/test/config/zookeeper/readme-example-1/profiler.json
cestella: awesome, thanks

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileHBaseMapper.java
nickwallen: Yes, makes sense.  Done.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
nickwallen: Yes, it should be acked whether there is an exception or not.  The main failure scenario is if someone creates a profile with a bad Stellar expression.  Retrying the tuple won't fix that.

nickwallen:
Yeah, I can add some read functionality as part of a separate PR, if you think that makes sense.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileSplitterBolt.java
```diff
@@ -0,0 +1,161 @@
...
+
+    // is this message needed by this profile?
+    String onlyIf = profile.getOnlyif();
+    if (executor.execute(onlyIf, message, Boolean.class)) {
+
+      // what is the name of the entity in this message?
+      String entity = executor.execute(profile.getForeach(), message, String.class);
+
+      // emit a message for the bolt responsible for building this profile
+      collector.emit(input, new Values(entity, profile, copyMessage(input)));
```
cestella: Are you doing a deep copy here because you're caching off the tuple and modifying it after it's emitted?

nickwallen: No.  I don't think I need a deep copy there.  I started with that early on in the build, but it is not needed now.

metron-analytics/metron-profiler/src/main/config/profiler.properties
```diff
@@ -0,0 +1,36 @@
...
+#
+#
+
+##### Storm #####
+
+profiler.workers=1
+profiler.executors=0
+profiler.input.topic=indexing
+profiler.flush.interval.seconds=900
+profiler.hbase.salt.divisor=10
```
cestella: Can we have a larger salt divisor?  Shouldn't it be bigger than the size of the cluster (in nodes) or am I thinking about this wrong?  We might want some rough guidance around choosing the right one in the docs too.

nickwallen: I am thinking that we have a lot of performance testing to do before we can intelligently set a default value.  

There is some guidance on setting this value in ProfileHBaseMapper.  I can certainly add some additional docs here too.

cestella:
So, regarding the salt divisor, I completely agree that we're going to need to performance test.  Maybe a blurb in the docs around what its intent is.

nickwallen:
I just added some docs in the README on the topology configuration settings.  This includes the salt divisor.  Let me know if that makes sense.  Its a table so you'll need to view the rendered markdown.

metron-analytics/metron-profiler/src/test/config/zookeeper/readme-example-1/profiler.json
nickwallen: I also added a better description of each of the fields in the README.  That should help with this type of question.

cestella:
+1, great job!

nickwallen:
Changes to the [project's by-laws](https://cwiki.apache.org/confluence/display/METRON/Apache+Metron+Bylaws) recently passed.  Only lazy consensus needed for pull requests.  This PR has already reached that threshold.  About to merge.

PR #211, opened 2016-08-15
METRON-365: Load demo dashboard on Kibana Service install

dlyle65535:
The next iteration of the Kibana service definition. Will now add the demo dashboard to the .kibana index on the specified ES instance.


cestella:
+1 on observation

PR #212, opened 2016-08-15
METRON-367 Enhance Profiler to Support Multiple Numeric Types

nickwallen:
### [METRON-367](https://issues.apache.org/jira/browse/METRON-367)

Note: This change depends on #208 .  The diff will be easier to grok once that PR has been merged.

### Changes

The Profiler currently only allows a profile to produce a Double value as a result. The Profiler should be enhanced so that a Profile can write any common numeric type including Double, Long, Short, Integer, and Float.  The type written depends on the Stellar code that is written as part of the profile definition.


cestella:
+1 by inspection

PR #213, opened 2016-08-15
METRON-368 Simplify Profile Configuration with Sensible Defaults

nickwallen:
### [METRON-368](https://issues.apache.org/jira/browse/METRON-368)

Note: This change depends on #208 #212.  The diff will be easier to grok once those PRs have been merged.

### Changes
The user should not need to define every configuration field.  This will help simplify the configuration for many types of profiles.

* If the user does not define 'onlyif', then all messages should be included in a profile.  
* If the user does not define 'init' then no vars should be initialized.
* If the user does not define 'update' then no vars should be updated.

### Testing
* Create a profile with no 'onlyif' field.  All messages should be applied to the profile.  
* Create a profile with no 'init' field. The profiler should not error.
* Create a profile with no 'update' field.  The profiler should not error

### DIFF
--- a/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileSplitterBolt.java
+++ b/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileSplitterBolt.java
@@ -26,6 +26,7 @@ import backtype.storm.topology.OutputFieldsDeclarer;
 import backtype.storm.tuple.Fields;
 import backtype.storm.tuple.Tuple;
 import backtype.storm.tuple.Values;
+import org.apache.commons.lang.StringUtils;
 import org.apache.metron.common.bolt.ConfiguredProfilerBolt;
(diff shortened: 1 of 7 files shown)


cestella:
+1 by inspection

nickwallen:
In some cases, the HBaseTestingUtility does not work unless you specify it this way.  Need to update.

```
Configuration config = HBaseConfiguration.create();
        config.set("hbase.master.hostname", "localhost");
        config.set("hbase.regionserver.hostname", "localhost");
        HBaseTestingUtility testUtil = new HBaseTestingUtility(config);
```

nickwallen:
@cestella Give this one more look before we merge it.  I added the additional configuration elements for `HBaseTestingUtility` and also suppressed excessive logging coming out of the `metron-hbase` unit tests.

cestella:
Yep I took a look last night and just did again.  Looks good

nickwallen:
About to get my merge on.

PR #214, opened 2016-08-16
METRON-372 Enhance Statistical Operations Available for Use with the Profiler

nickwallen:
### [METRON-372](https://issues.apache.org/jira/browse/METRON-372)

Note: This change depends on #208, #212, & #213.  The diff will be easier to grok once those PRs are merged.

### Changes

Only basic math functions are currently available in Stellar for use with the Profiler.  This makes life difficult for users to create even basic profiles like a running average.  

This can be seen in the following example where the average must be calculated manually in Stellar.  A variable `sum` and `cnt` must be maintained and then used to calculate the average.

```
{
  "profile": "example3",
  "foreach": "ip_src_addr",
  "init": {
    "sum": 0.0,
    "cnt": 0.0
  },
  "update": {
    "sum": "sum + resp_body_len",
    "cnt": "cnt + 1"
  },
  "result": "sum / cnt"
}
```

This change introduces a series of summary functions that make creating profiles much simpler for the user.  Instead of re-implementing the calculation of an average in Stellar, this leverages Commons Math to perform all the heavy lifting.

The example above for calculating an average can be re-defined as follows.

```
{
  "profile": "example3",
  "foreach": "ip_src_addr",
  "init":   { "s": "STATS_INIT(0)" },
  "update": { "_": "STATS_ADD(length, s)" },
  "result": "STATS_MEAN(s)"
}
```

The following summary functions are supported.  These are all statistics that can be calculated in a single pass.  This means that none of the values being summarized are stored in memory.
* count
* mean
* geometric mean
* max
* min
* sum
* population variance
* variance
* second moment
* quadratic mean
* standard deviation
* sum of logs
* sum of squares.


nickwallen:
Closing in favor of #225

PR #215, opened 2016-08-17
METRON-377 Enable Profiles that Use Non-Single Pass Summary Functions

nickwallen:
Note: This change depends on #208, #212, #213, #214 . The diff will be easier to grok once those PRs are merged.

### [METRON-377](https://issues.apache.org/jira/browse/METRON-377)

As of METRON-372 and #214 , Profiles can be built using many statistical summaries that only require a single-pass over the data.  This is less memory intensive and more scalable for high volume loads.

Unfortunately, not all functions can be calculated in a single pass.  In particular, the skewness, ketosis and percentiles require all data to be stored in memory for the calculation to occur.  The platform was enhanced so that a user can leverage skewness, ketosis and percentiles. 

### Changes

The `STATS_INIT` function was enhanced to accept a `window_size`.  This defines the number of input data elements that are maintained in memory.  

If the `window_size` is greater than 0, a rolling window of the most recent `window_size` elements is maintained in memory.  The skewness, ketosis and percentiles are calculated over this rolling window.  The `window_size` must be >0 otherwise these values cannot be calculated.

If a user supplies a `window_size` equal to 0 then the more efficient implementation that does not maintain a rolling window in memory is used.  Of course, in this case, the skewness, ketosis, and percentiles cannot be calculated.

The following additional functions were also added.
* STATS_KURTOSIS
* STATS_SKEWNESS
* STATS_PERCENTILES

Another integration test was added to validate this end-to-end also.


cestella:
did you mean Kurtosis?  Ketosis is a metabolic process, not a stats function. ;)

nickwallen:
Ha. Yes.  Fortunately, the code is right.  Just my description in the PR is wrong.

nickwallen:
Replaces by #226

PR #217, opened 2016-08-19
METRON-381: Add support for multiple reducers in pcap_query.sh

cestella:
As a stopgap, the pcap_query.sh MapReduce job uses a single reducer to ensure that the packets captured come in the correct order. This has issues on large queries because it forces all of the data to go through one reducer. Instead, we should have a custom partitioner that will allow us to use multiple reducers.


dlyle65535:
+1, looks good!

PR #221, opened 2016-08-22
METRON-384 Allow elasticsearch to bind to loopback addresses

2xyo:
This PR addresses https://issues.apache.org/jira/browse/METRON-384

### DIFF
--- a/metron-deployment/roles/elasticsearch/tasks/elasticsearch.yml
+++ b/metron-deployment/roles/elasticsearch/tasks/elasticsearch.yml
@@ -66,8 +66,8 @@
     state=present
   with_items:
     - { regexp: '^# *cluster\.name:', line: 'cluster.name: metron' }
-    - { regexp: '^# *network\.host:', line: 'network.host: _{{
-    elasticsearch_network_interface  }}:ipv4_' }
+    - { regexp: '^# *network\.host:', line: 'network.host: ["_{{
+    elasticsearch_network_interface  }}:ipv4_","_local:ipv4_"]' }
     - { regexp: '^# *discovery\.zen\.ping\.unicast\.hosts:',
     line: 'discovery.zen.ping.unicast.hosts: [ {{ es_hosts }} ]'}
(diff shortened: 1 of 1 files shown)


dlyle65535:
I like what you did here, but could you expose it as a parameter and default it to ipv4 instead?

2xyo:
Thanks @dlyle65535  for the feedback.

Would you prefer a configuration like this?

Old `full-dev-platform/group_vars/all`:
```
elasticsearch_network_interface: eth0
```

New  `full-dev-platform/group_vars/all`:
```
elasticsearch_network_interface:
- eth0
- local
```

If yes, I dont' know how to implement a loop with the `lineinfile` module :/

Could you also elaborate the "ipv4 instead"? (As the result is already ipv4 only "`network.host: ["_eth0:ipv4_","_local:ipv4_"]`" )

dlyle65535:
Sorry, I meant eth0 only. I'd try adding network_host: _{{
    elasticsearch_network_interface  }}:ipv4_' } to the role's defaults and then you can override it in the group vars to whatever you'd like rather than mess with looping.

2xyo:
So, you would recommend something like this:

New  `elasticsearch.yml`:
```
  - { regexp: '^# *network\.host:', line: 'network.host: {{elasticsearch_network_interface  }}' }
```

New  `full-dev-platform/group_vars/all`, `multinode-vagrant/group_vars/all`:
```
elasticsearch_network_interface: _eth1:ipv4_
```
And new  `metron_example/group_vars/all `, `amazon-ec2/conf/defaults.yml`:
```
elasticsearch_network_interface: _eth0:ipv4_
```

But, in this case, elasticsearch will not listening to localhost by default (and that's why I opened this issue). IMHO, the default value should be instead:
```
elasticsearch_network_interface: ["_eth1:ipv4_","_local:ipv4_"]
```

dlyle65535:
I'm +1 on this, thanks for the contribution.

dlyle65535:
Could you post an email address to use for the commit? Thanks!

2xyo:
Thanks! Here is my email: yohann@lepage.info

PR #222, opened 2016-08-22
METRON-385 Create Ambari Service Definition for Indexing

justinleet:
https://issues.apache.org/jira/browse/METRON-385

Opening this PR for community review and feedback.

Prereq:
* Ambari 2.4

Testing this PR:
* The RPMs in metron-deployment should be built.  They should either be mounted or copied to /localrepo in the node to be used as the indexing master node.
* Copy or mount the services and stack contents into the appropriate places on the ambari-server node.  Afterwards, restart the Ambari server if it is running: ```service ambari-server restart```. For example:
  * incubator-metron/metron-deployment/packaging/ambari/src/main/resources/stacks/HDP/2.4/services/INDEXING -> /var/lib/ambari-server/resources/stacks/HDP/2.4/services/INDEXING
  * incubator-metron/metron-deployment/packaging/ambari/src/main/resources/common-services/INDEXING -> /var/lib/ambari-server/resources/common-services/INDEXING
* Install the master to a node with the Kafka Broker.

Limitations / Notes
* Service advisor for ensuring that the service is colocated appropriately doesn't appear to work. It's currently in the PR as-is.  It can easily be removed if we want to leave it out for now.
* Because the RPMs are currently not published to a yum repo, it's necessary to do the copy / mount step. Assuming the RPMs find a more permanent home, the repolist for the used stack should be updated and preferred.
* Elasticsearch.properties had a couple properties added to reflect what was in the Ansible scripts, because the current file was missing properties necessary to run the script outside of that context.
* Ambari currently provides no hooks for performing actions when a service is deleted.  To ensure cleanup please run

```
#!/bin/bash

# kafka
/usr/hdp/current/kafka-broker/bin/kafka-topics.sh \
        --zookeeper $ZK_QUORUM \
        --delete \
        --topic indexing
done

# storm kill
storm kill indexing

# yum delete
yum -y erase metron-indexing metron-common

# final file delete
rm -rf /usr/metron
```


metron-platform/metron-elasticsearch/src/main/config/elasticsearch.properties
```diff
@@ -57,10 +57,13 @@ org.apache.metron.metrics.TelemetryParserBolt.fails=true
 
 bolt.hdfs.batch.size=5000
 bolt.hdfs.field.delimiter=|
+bolt.hdfs.rotation.policy=org.apache.storm.hdfs.bolt.rotation.TimedRotationPolicy
+bolt.hdfs.rotation.policy.units=DAYS
+bolt.hdfs.rotation.policy.count=1
+
 bolt.hdfs.file.rotation.size.in.mb=5
```
dlyle65535: I'm a little confused by this as part of this PR. Can you tell me a bit about it?

justinleet: Those lines are actually added by Ansible in the original path (topologies.yml).  They don't exist with any properties in elasticsearch.properties (which gets bundled up as part of the RPM build).  Because the properties don't exist, the topology will fail.

I just used what Ansible defined as the configs, but they're easy to change if we want to (or if there's some difference between environments I missed).

Otherwise if we want to leave them out of elasticsearch.properties, we'll probably want to do something a little more clever (e.g. make elasticsearch.properties a config file handled through Ambari and use that instead of what the RPM has).  I'm unsure if that file gets used in other manners in other deploys, so I was hesitant to make that change.

dlyle65535: Oh, I get it. Yeah, we'll likely need to expose it as configs, but for now it makes sense.

dlyle65535:
+1 looks good!

justinleet:
One issue discovered, which is that, to make things compatible with earlier Kafka versions that don't handle topics already existing that functionality was dropped. This PR needs to be updated to handle restarts properly without that.

james-sirota:
+1

justinleet:
Added a change to ensure that this works with centos 6's default Python version (2.6.x)

metron-deployment/packaging/ambari/src/main/resources/common-services/INDEXING/0.2.0BETA/service_advisor.py
```diff
@@ -0,0 +1,65 @@
...
+
+try:
+    with open(PARENT_FILE, 'rb') as fp:
+        service_advisor = imp.load_module('service_advisor', fp, PARENT_FILE, ('.py', 'rb', imp.PY_SOURCE))
+except Exception as e:
+    traceback.print_exc()
+    print("Failed to load parent service_advisor file '{}'".format(PARENT_FILE))
+
+
+class INDEXING020BETAServiceAdvisor(service_advisor.ServiceAdvisor):
```
nickwallen: The Metron version is embedded in this class name.  Also, the path for the Indexing `metron-deployment/packaging/ambari/src/main/resources/common-services/INDEXING/0.2.0BETA/...` includes the Metron version.  

How will we handle this when we bump Metron versions?  Is there no better way to handle this?  I know we have to play within the sandbox provided by Ambari.  Is this an Ambari-imposed pain point?

Everything else looks great BTW.  Very easy to read code.

justinleet: It's an Ambari imposed pain point. From what I can tell, it just looks for that class name via magic. If this actually existed in it's own stack, we could use the Stack advisor. Service advisor is more for adding custom services to another stack, so this is probably closer to what we can do until everything coalesces a little more nicely.

The better way is probably to have this handled by an Ambari mpack (which bundles everything up on it's own), but I'm not sure how nicely Service and Stack advisors play, particularly when an mpack is installed alongside an existing stack.  It's something I don't know enough about to say.

Relevant portion of the [Ambari wiki](https://cwiki.apache.org/confluence/display/AMBARI/How-To+Define+Stacks+and+Services#How-ToDefineStacksandServices-ServiceAdvisor):


"Unlike the Stack-advisor scripts, the service-advisor scripts do not automatically extend the parent service's service-advisor scripts. The service-advisor script needs to explicitly extend their parent's service service-advisor script."

I wouldn't be opposed to dropping the service-advisor for now (especially given the note in the original description).  Especially since I'd like to drop this method anyway in the future, assuming that's feasible.

nickwallen: If the Service Advisor can be dropped and doesn't affect the core functioning of your PR, then it would be nice to take it out. Less code is more.

It sounds like you have a go-forward plan to resolve the 'version embedded in the path' issue with Ambari mpacks.  I'm good with that plan.

justinleet: Removed the service advisor and updated PR

nickwallen:
+1

PR #223, opened 2016-08-22
METRON-386 Create Ambari Service Definition for Elasticsearch

justinleet:
https://issues.apache.org/jira/browse/METRON-386

Opening the PR to get community feedback.

Prereq:

Ambari 2.4
Testing this PR:

* Copy or mount the services and stack contents into the appropriate places on the ambari-server node. Afterwards, restart the Ambari server if it is running: ```service ambari-server restart```. For example:
  * incubator-metron/metron-deployment/packaging/ambari/src/main/resources/stacks/HDP/2.4/services/ELASTICSEARCH -> /var/lib/ambari-server/resources/stacks/HDP/2.4/services/ELASTICSEARCH
  * incubator-metron/metron-deployment/packaging/ambari/src/main/resources/common-services/ELASTICSEARCH -> /var/lib/ambari-server/resources/common-services/ELASTICSEARCH
* ES Master eligible nodes will not contain data. Don't colocate master eligible nodes and data nodes.
* It's necessary to provide the ES zen_discovery_ping_unicast_hosts.

Limitations / Notes
* This is based off of Symantec's work at: https://github.com/Symantec/ambari-elasticsearch-service.  Should I add this into the NOTICE that includes similar contributions?
* Most configs are used from Symantec's build.  If there are more sane configs, we should change the defaults (I'm not an ES expert by any means).
* It would be really nice if zen_discovery_ping_unicast_hosts could be automatically populated with something sane.  I'm unsure of how to accomplish this with Ambari.
* Master-eligible nodes are non-data holding nodes that are eligible to be elected master. Data nodes are, not surprisingly, nodes that only hold data, but are not master-eligible. It would be nice if thing could be modified to allow master eligible nodes to also hold data cleanly, but this is left as a further enhancement if needed.
* A service advisor definition could probably be added to force Ambari to avoid placing master-eligible and data nodes onto the same node.

### DIFF
--- a/metron-deployment/packaging/ambari/src/main/resources/common-services/ELASTICSEARCH/2.3.3/configuration/elastic-env.xml
+++ b/metron-deployment/packaging/ambari/src/main/resources/common-services/ELASTICSEARCH/2.3.3/configuration/elastic-env.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0"?>
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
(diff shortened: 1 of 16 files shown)


dlyle65535:
+1 thanks!

PR #225, opened 2016-08-23
METRON-372 Enhance Statistical Operations Available for Use with the Profiler

nickwallen:
### [METRON-372](https://issues.apache.org/jira/browse/METRON-372)

This PR serves as a replacement for #214 .

### Changes

Only basic math functions are currently available in Stellar for use with the Profiler.  This makes life difficult for users to create even basic profiles like a running average.  

This can be seen in the following example where the average must be calculated manually in Stellar.  A variable `sum` and `cnt` must be maintained and then used to calculate the average.

```
{
  "profile": "example3",
  "foreach": "ip_src_addr",
  "init": {
    "sum": 0.0,
    "cnt": 0.0
  },
  "update": {
    "sum": "sum + resp_body_len",
    "cnt": "cnt + 1"
  },
  "result": "sum / cnt"
}
```

This change introduces a series of summary functions that make creating profiles much simpler for the user.  Instead of re-implementing the calculation of an average in Stellar, this leverages Commons Math to perform all the heavy lifting.

The example above for calculating an average can be re-defined as follows.

```
{
  "profile": "example3",
  "foreach": "ip_src_addr",
  "init":   { "s": "STATS_INIT()" },
  "update": { "_": "STATS_ADD(length, s)" },
  "result": "STATS_MEAN(s)"
}
```

The following summary functions are supported.  These are all statistics that can be calculated in a single pass.  This means that none of the values being summarized are stored in memory.
* count
* mean
* geometric mean
* max
* min
* sum
* population variance
* variance
* second moment
* quadratic mean
* standard deviation
* sum of logs
* sum of squares.


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/SummaryStatisticsFunctions.java
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * Add an input value to those that are used to calculate the summary statistics.
+   *
+   *  STATS_ADD (value, stats)
+   */
+  public static class Add extends BaseStellarFunction {
+    @Override
+    public Object apply(List<Object> args) {
+      double value = convert(args.get(0), Double.class);
```
cestella: Should we swap the args so the SummaryStatistics arg comes first and allow users to pass as many numbers as they want to STATS_ADD(...)?  Such as `STATS_ADD(stats, 1, 2, 3, 4)`

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/SummaryStatisticsFunctions.java
```diff
@@ -0,0 +1,198 @@
...
+      SummaryStatistics stats = convert(args.get(1), SummaryStatistics.class);
+      stats.addValue(value);
+      return null;
```
cestella: Should we return the SummaryStatistics class here?  That way we can chain them?

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarStatisticalFunctionsTest.java
```diff
@@ -0,0 +1,146 @@
...
+ */
+public class StellarStatisticalFunctionsTest {
+
+  private List<Double> values;
+  private Map<String, Object> variables;
+  private SummaryStatistics stats;
+
+  private static Object run(String rule, Map<String, Object> variables) {
+    StellarProcessor processor = new StellarProcessor();
+    return processor.parse(rule, x -> variables.get(x), StellarFunctions.FUNCTION_RESOLVER(), Context.EMPTY_CONTEXT());
```
cestella: Can we check that stellar's`validate()` works here too?

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/SummaryStatisticsFunctions.java
nickwallen: How would a user use this?  Is there a 'prettier' way that I am not thinking about?  This looks absolutely nasty.

```
STATS_ADD(STATS_ADD(STATS_ADD(s, 1), 2), 3)
```

cestella: Yeah, let's not do that.  Instead, let's make it so that the stats are first and the user can pass as many nums as they like, so that would be `STATS_ADD(s, 1, 2, 3)`

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarStatisticalFunctionsTest.java
nickwallen: A call to validate has been added.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/SummaryStatisticsFunctions.java
nickwallen: This is not supported in the latest patch.

cestella:
+1

PR #226, opened 2016-08-23
METRON-377 Enable Profiles that Use Non-Single Pass Summary Functions

nickwallen:
### [METRON-377](https://issues.apache.org/jira/browse/METRON-377)

This pull request replaces #215 .

As of METRON-372 , Profiles can be built using many statistical summaries that only require a single-pass over the data.  This is less memory intensive and more scalable for high volume loads.

Unfortunately, not all functions can be calculated in a single pass.  In particular, the skewness, ketosis and percentiles require all data to be stored in memory for the calculation to occur.  The platform was enhanced so that a user can leverage skewness, ketosis and percentiles. 

### Changes

The `STATS_INIT` function was enhanced to accept a `window_size`.  This defines the number of input data elements that are maintained in memory.  

If the `window_size` is greater than 0, a rolling window of the most recent `window_size` elements is maintained in memory.  The skewness, ketosis and percentiles are calculated over this rolling window.  The `window_size` must be >0 otherwise these values cannot be calculated.

If a user supplies a `window_size` equal to 0 then the more efficient implementation that does not maintain a rolling window in memory is used.  Of course, in this case, the skewness, ketosis, and percentiles cannot be calculated.

The following additional functions were also added.
* STATS_KURTOSIS
* STATS_SKEWNESS
* STATS_PERCENTILES

Another integration test was added to validate this end-to-end also.


metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarStatisticsFunctionsTest.java
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  /**
+   * Runs a Stellar expression.
+   * @param expr The expression to run.
+   * @param variables The variables available to the expression.
+   */
+  private static Object run(String expr, Map<String, Object> variables) {
+    StellarProcessor processor = new StellarProcessor();
+    return processor.parse(expr, x -> variables.get(x), StellarFunctions.FUNCTION_RESOLVER(), Context.EMPTY_CONTEXT());
```
cestella: Can we make sure that we run Stellar's `validate()` function on the expressions passed in.

nickwallen: Yes.  Will be fixed once I merge in all those changes from METRON-372.  Almost there

nickwallen:
The PR has been merged and re-tested.  Ready for a look-see.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StellarStatistics.java
```diff
@@ -0,0 +1,207 @@
...
+   * This implementation is used if the windowSize == 0.
+   */
+  private SummaryStatistics summStats;
+
+  /**
+   * @param windowSize The number of input data elements to maintain in memory.  If
+   *                   windowSize == 0, then no data elements will be maintained in
+   *                   memory.
+   */
+  public StellarStatistics(int windowSize) {
```
cestella: Can we move the functions for StellarStatistics into their own interface and have two implementations, one backed by SummaryStatistics and one backed by DescriptiveStatistics?  We will likely have a third implementation, ProbabalisticStatistics that supports approximate distributional stats while not requiring a window size and I'd like it to be as simple as implementing an interface, not adding to a branch at every method.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StellarStatistics.java
```diff
@@ -0,0 +1,207 @@
...
+    if(descStats != null) {
+      return descStats.getPopulationVariance();
+    } else {
+      return summStats.getPopulationVariance();
+    }
+  }
+
+  public double getSecondMoment() {
+    if(descStats != null) {
+      throw new NotImplementedException("second moment not available if 'windowSize' > 0");
```
cestella: Any reason why we're not implementing the second moment for SummaryStatistics?  It should be the square of the standard deviation, right?

cestella:
Looks good, +1 pending green build.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StellarStatistics.java
nickwallen: @cestella pointed out that second moment is the same as variance.  The naming between DescriptiveStatistics and SummaryStatistics just differs.  Just added a commit that removes `STATS_SECOND_MOMENT` as it is redundant.  Thanks!

PR #227, opened 2016-08-23
METRON-388: Stellar - Unable to Operate on Result of If-Then-Else

cestella:
As per [METROn-388](https://issues.apache.org/jira/browse/METRON-388), we added a regression to Stellar around conditional expressions.  This is corrected in this PR and the test cases suggested are adopted.


nickwallen:
+1 Looks solid

PR #228, opened 2016-08-24
METRON-390: Stellar functions should initialize upon use, rather than all at once

cestella:
To test this, you should just spin up the environment in vagrant and everything should work as expected.

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
@@ -193,6 +193,7 @@ public class MaaSFunctions {
         LOG.error(ex.getMessage(), ex);
       }
       finally {
+        //We always want to set initialize to true because we don't want to keep trying to initialize over and over
         isInitialized = true;
       }
     }
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import org.apache.curator.framework.CuratorFramework;
+import org.apache.hadoop.security.authorize.Service;
 import org.apache.metron.common.dsl.Context;
 import org.apache.metron.common.dsl.ParseException;
 import org.apache.metron.common.dsl.StellarFunction;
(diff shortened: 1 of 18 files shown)


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
```diff
@@ -188,15 +189,27 @@ public void initialize(Context context) {
       catch(Exception ex) {
         LOG.error(ex.getMessage(), ex);
       }
+      finally {
+        isInitialized = true;
```
nickwallen: If initialization fails, do we will want to set `isInitialized` to true?

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
```diff
@@ -234,25 +248,36 @@ public Object apply(List<Object> args, Context context) throws ParseException {
...
+          return;
+        }
+        try {
+          MaaSConfig config = ConfigUtil.INSTANCE.read(client, "/metron/maas/config", new MaaSConfig(), MaaSConfig.class);
+          discoverer = new ServiceDiscoverer(client, config.getServiceRoot());
+          discoverer.start();
+          context.addCapability(Context.Capabilities.SERVICE_DISCOVERER, () -> discoverer);
+          isValidState = true;
+        } catch (Exception e) {
+          LOG.error(e.getMessage(), e);
```
nickwallen: If initialization fails, we're just logging the error.  Does it not make sense to fail fast or make some noise?

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
cestella: I'd prefer for initialization to not continue to retry for every message, but rather the initialization routine should throw an exception.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
cestella: Yeah, agreed.  Throwing an exception here.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/MaaSFunctions.java
nickwallen: That makes sense.  Maybe a little comment to that effect, will prevent me from asking you this same question in a few months.  Ha.

nickwallen:
+1 looks good

PR #229, opened 2016-08-24
METRON-374: Add appropriate bundled 3rd party licenses to NOTICE and LICENSE where appropriate

cestella:
This PR ensures:
* Every place where we create a shaded jar, we must add appropriate links to a binary distribution LICENSE file in the META-INF directory to indicate we are bundling
* When new components are added, we must ensure that we have a handle on licensing to ensure they are not prohibited.

From here on, if you add a dependency, you must add a line to the `dependencies_with_url.csv`.  Travis will take the transitive dependency list and check against this file to ensure that it's listed.  This will make sure we track dependencies. 

Also, `dependencies_with_url.csv` can be used with the `build_utils/create_bundled_licenses.sh` to generate the binary licenses that live inside of the shaded jars.


nickwallen:
It would be good to add a quick README.md under `build_utils` that outlines what the scripts do and how to use them.  It would also be a good place to add the guidance that you provide in this PR itself.

james-sirota:
+1 by inspection

cestella:
Thanks @nickwallen, I added appropriate documentation and updated to current master.

PR #230, opened 2016-08-25
METRON-392 Allow User to Define Custom 'Group By' for a Profile

nickwallen:
### [METRON-392](https://issues.apache.org/jira/browse/METRON-392)

Allows a user to optionally define a custom set of 'groupBy' expressions that controls how the data is persisted.  This is intended to allow for contiguous scans when training on subsets of the data. 

The 'groupBy' expressions can refer to any field within a `ProfileMeasurement`.  This includes the following fields: 
  * `profileName`: The name of the profile.
  * `entity`: The name of the entity being profiled.
  * `start`: The window start time in milliseconds from the epoch.
  * `end`: The window end time in milliseconds from the epoch.
  * `value`: The summary value calculated over the window period.
  * `groupBy`: The set of 'groupBy' expressions; not the result of those expressions.

A common use case would be grouping the data by day of week.  This would allow a contiguous scan to access all profile data for Mondays only.  The Stellar expression `DAY_OF_WEEK(start)` would achieve this. 

*NOTE*: A series of date functions will be added to Stellar in a follow-on PR to enhance the types of groups that can be created.

### Example
```
{
  "inputTopic": "indexing",
  "profiles": [
    {
      "profile": "example3",
      "foreach": "ip_src_addr",
      "groupBy": "DAY_OF_WEEK(start)",
      "update": { "s": "STATS_ADD(s, length)" },
      "result": "STATS_MEAN(s)"
    }
  ]
}
```

### Testing
To test this change do the following. 
* Create a profile and do not define a 'groupBy' expression.  Prior to this change the row key would include the day of week, week of month, etc which altered how the data was sorted on disk.  After this change, these fields will not be included in the row key.
* Create a profile and define a 'groupBy' expression.  The result of this expression will be embedded in the row key.


nickwallen:
#231 includes the necessary Stellar date functions to handle the common uses cases of grouping by day of week, etc.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileHBaseMapper.java
```diff
@@ -49,7 +65,7 @@
 
   public ProfileHBaseMapper() {
     setColumnFamily("P");
-    setSaltDivisor(1000);
+    setSaltDivisor(0);
```
cestella: Why are you setting the salt divisor to 0?  Under what situations would you not need a salt?  I'm especially concerned that this appears to be the default behavior.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/profiler/ProfileConfig.java
```diff
@@ -27,36 +29,57 @@
...
+   * expression.  The map can include 0 or more variables/expressions.
    */
-  private Map<String, String> init = new HashMap<>();
+  private Map<String, String> update = new HashMap<>();
 
   /**
-   * Defines how the state is updated when a new message is received.
+   * A list of Stellar expressions that is executed in order and used to group the
+   * resulting profile data.
+   * TODO
```
cestella: What is the TODO there for?

nickwallen: Fixed this in the latest commit. Thanks

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileHBaseMapper.java
nickwallen: Restored default to 1000.  Thanks

nickwallen:
@cestella Updated the PR based on your feedback.  Let me know what you think.

cestella:
+1, looks good post master merge

PR #231, opened 2016-08-25
METRON-394 Create Stellar Date Functions to Use with Profile 'Group By'

nickwallen:
### [METRON-394](https://issues.apache.org/jira/browse/METRON-394)

[METRON-392](https://issues.apache.org/jira/browse/METRON-392) and #230  Allows a user to optionally define a custom set of 'groupBy' expressions that controls how Profile data is persisted. This is intended to allow for contiguous scans when training on subsets of the data.

A common use case would be grouping the data by some calendar period, like day of week. This would allow a contiguous scan to access all profile data, for example, on Mondays only. The Stellar expression DAY_OF_WEEK(start) would achieve this.

The following functions have been added for this purpose.
* WEEK_OF_MONTH
* WEEK_OF_YEAR
* DAY_OF_WEEK
* DAY_OF_MONTH
* DAY_OF_YEAR
* MONTH
* YEAR

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/DateFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/DateFunctions.java
@@ -22,20 +22,26 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import org.apache.metron.common.dsl.BaseStellarFunction;
+import org.apache.metron.common.utils.ConversionUtils;
 
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.Calendar;
 import java.util.List;
 import java.util.Optional;
 import java.util.TimeZone;
 import java.util.concurrent.ExecutionException;
-import java.util.function.Function;
 
+/**
+ * Stellar data functions.
+ */
 public class DateFunctions {
 
   private static class TimezonedFormat {
+
     private String format;
     private Optional<String> timezone;
+
     public TimezonedFormat(String format, String timezone) {
       this.format = format;
       this.timezone = Optional.of(timezone);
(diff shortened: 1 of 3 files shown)


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/DateFunctions.java
```diff
@@ -114,17 +122,176 @@ public Object apply(List<Object> objects) {
...
+   * Stellar Function: DAY_OF_WEEK
+   *
+   * The numbered day within the week.  The first day of the week, Sunday, has a value of 1.
+   */
+  public static class DayOfWeek extends BaseStellarFunction {
+    @Override
+    public Object apply(List<Object> args) {
+
+      // expect epoch milliseconds
+      Long epochMillis = ConversionUtils.convert(args.get(0), Long.class);
```
cestella: can you do an args.get(0) null check before this?

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/DateFunctions.java
```diff
@@ -114,17 +122,176 @@ public Object apply(List<Object> objects) {
...
+   * Stellar Function: DAY_OF_MONTH
+   *
+   * The day within the month.  The first day within the month has a value of 1.
+   */
+  public static class DayOfMonth extends BaseStellarFunction {
+    @Override
+    public Object apply(List<Object> args) {
+
+      // expect epoch milliseconds
+      Long epochMillis = ConversionUtils.convert(args.get(0), Long.class);
```
cestella: args.get(0) null check, please.

cestella:
Looks good, +1

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/DateFunctions.java
nickwallen: No worries.  Thanks for the review.

PR #232, opened 2016-08-25
METRON-395 Fix Metron Bro parser not parsing some timestamp values

mmiklavc:
This PR addresses https://issues.apache.org/jira/browse/METRON-395

This fix corrects the BasicBroParser so that it will properly handle Double timestamp values that have less than 3 digits, e.g. 123.0, 123.1, 123.11. In the previous examples, the current code would throw an index out of bounds exception.

Verified in full-dev-platform.

**Testing:**

- Deploy the bro parser topology.
- Add a record to dns.log or http.log with a ts value that has less than 3 significant decimal places.
- Confirm the record(s) make it into indexing.


metron-platform/metron-integration-test/src/main/sample/data/bro/raw/BroExampleOutput
```diff
@@ -1,10 +1,10 @@
...
-{"http":{"ts":1402307733473,"uid":"KIRAN","id.orig_h":"10.122.196.204","id.orig_p":58808,"id.resp_h":"72.163.4.161","id.resp_p":80,"trans_depth":1,"method":"GET","host":"www.cisco.com","uri":"/","user_agent":"curl/7.22.0 (x86_64-pc-linux-gnu) libcurl/7.22.0 OpenSSL/1.0.1 zlib/1.2.3.4 libidn/1.23 librtmp/2.3","request_body_len":0,"response_body_len":25523,"status_code":200,"status_msg":"OK","tags":[],"resp_fuids":["FJDyMC15lxUn5ngPfd"],"resp_mime_types":["text/html"]}}
+{"http":{"ts":1402307733.473,"uid":"CTo78A11g7CYbbOHvj","id.orig_h":"192.249.113.37","id.orig_p":58808,"id.resp_h":"72.163.4.161","id.resp_p":80,"trans_depth":1,"method":"GET","host":"www.cisco.com","uri":"/","user_agent":"curl/7.22.0 (x86_64-pc-linux-gnu) libcurl/7.22.0 OpenSSL/1.0.1 zlib/1.2.3.4 libidn/1.23 librtmp/2.3","request_body_len":0,"response_body_len":25523,"status_code":200,"status_msg":"OK","tags":[],"resp_fuids":["FJDyMC15lxUn5ngPfd"],"resp_mime_types":["text/html"]}}
```
cestella: I'm confused, did the bro plugin change its format for `ts` or was this reference data always wrong?

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/bro/BasicBroParserTest.java
```diff
@@ -46,14 +46,37 @@ public BasicBroParserTest() throws Exception {
...
+						Pair.of(new Double(12345678.0), "12345678.0"),
+						Pair.of(new Double(12345678.1), "12345678.1"),
+						Pair.of(new Double(12345678.11), "12345678.11"),
+						Pair.of(new Double(12345678.111), "12345678.111"),
+						Pair.of(new Double(12345678.1111), "12345678.1111"),
+						Pair.of(new Double(12345678.11111), "12345678.11111"),
+						Pair.of(new Double(12345678.111111), "12345678.111111")
+		};
+		for (Pair pair : pairs) {
+			assertEquals("Format did not match", pair.getRight(), new DecimalFormat("0.0#####").format(pair.getLeft()));
```
cestella: Can we refer to BasicBroParser.DECIMAL_FORMAT rather than explicitly creating the decimal format.  This way if the decimal format changes, we're aware of it in the unit test and the gut-check you just did can be redone for the new implementer.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bro/BasicBroParser.java
```diff
@@ -34,6 +35,7 @@
 
   protected static final Logger _LOG = LoggerFactory
           .getLogger(BasicBroParser.class);
+  public static final DecimalFormat DECIMAL_FORMAT = new DecimalFormat("0.0#####");
```
cestella: DecimalFormat is not threadsafe, can you please make this a `ThreadLocal<DecimalFormat>`?

metron-platform/metron-integration-test/src/main/sample/data/bro/parsed/BroExampleParsed
```diff
@@ -1,10 +1,10 @@
...
\ No newline at end of file
+{"bro_timestamp":"1402307733.473","status_code":200,"method":"GET","ip_dst_port":80,"request_body_len":0,"uri":"\/","tags":[],"source.type":"bro","uid":"CTo78A11g7CYbbOHvj","resp_mime_types":["text\/html"],"trans_depth":1,"protocol":"http","original_string":"HTTP | id.orig_p:58808 status_code:200 method:GET request_body_len:0 id.resp_p:80 uri:\/ tags:[] uid:CTo78A11g7CYbbOHvj resp_mime_types:[\"text\\\/html\"] trans_depth:1 host:www.cisco.com status_msg:OK id.orig_h:192.249.113.37 response_body_len:25523 user_agent:curl\/7.22.0 (x86_64-pc-linux-gnu) libcurl\/7.22.0 OpenSSL\/1.0.1 zlib\/1.2.3.4 libidn\/1.23 librtmp\/2.3 ts:1402307733.473 id.resp_h:72.163.4.161 resp_fuids:[\"FJDyMC15lxUn5ngPfd\"]","ip_dst_addr":"72.163.4.161","ip_src_port":58808,"host":"www.cisco.com","status_msg":"OK","response_body_len":25523,"ip_src_addr":"192.249.113.37","user_agent":"curl\/7.22.0 (x86_64-pc-linux-gnu) libcurl\/7.22.0 OpenSSL\/1.0.1 zlib\/1.2.3.4 libidn\/1.23 librtmp\/2.3","resp_fuids":["FJDyMC15lxUn5ngPfd"],"timestamp":1402307733473}
```
cestella: It appears that bro_timestamp is changing format, which will have an effect on the kibana dashboard that we ship.  Have you investigated if we rely on `bro_timestamp` at all in that dashboard and, if so, what the impact is?  Wouldn't want to inadvertently break our dashboard.

metron-platform/metron-integration-test/src/main/sample/data/bro/raw/BroExampleOutput
mmiklavc: It appears the ref data was always incorrect. I did a data dump from Kafka and the values are in seconds as a Double, not long millis.

```
{"http": {"ts":1471985653.216189,"uid":"C4OfBp49ZhOKrYwSca",...
{"http": {"ts":1471985653.921796,"uid":"CS9CEh4RXdS66P9CB9"
```

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/bro/BasicBroParserTest.java
mmiklavc: That's a good suggestion

cestella:
+1, this looks good.

cestella:
So, @mmiklavc this is failing under travis, but the failure that I saw generally is due to travis being overwhelmed and one of the integration test components (kafka in this case, I believe) timing out.  I would suggest closing and reopening or doing a whitespace commit to get a rerun of travis.

metron-platform/metron-integration-test/src/main/sample/data/bro/raw/BroExampleOutput
dlyle65535: Sorry I'm late to the party, those deleted lines also came out of Kafka, I think we need to handle both, don't we? The last test case in the deleted list, for example, was added because it failed on a running system.

mmiklavc: @dlyle65535 Can you clarify which test case you're referring to?

> The last test case in the deleted list, for example, was added because it failed on a running system.

PR #233, opened 2016-08-26
METRON-396: Make Stellar function resolution happen via an annotation and classpath search

cestella:
At the moment, the only way to add a stellar function is to implement StellarFunction and add the instance to the StellarFunctions enum. This is deficient in a few ways:
* It requires you to modify Metron code to add a stellar capability
* It requires you to make metron-common depend on your project
* There is no way to enforce documentation of the function

Instead, we should create an annotation which will define:
* function name
* namespace
* description
* List of parameters
* return documentation


cestella:
Ok, this should be a transparent change to the topology.  Things should continue to function as expected.  I also verified that this did not break pcap's `query` filter.

As such, a reasonable test plan is to spin the topologies up, ensure that data flows through.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/FunctionResolverSingleton.java
```diff
@@ -0,0 +1,206 @@
...
+    if(ret == null) {
+      throw new IllegalStateException("Unable to resolve function " + s);
+    }
+    return ret.getFunction();
+  }
+
+  private void loadFunctions(final Map<String, StellarFunctionInfo> ret) {
+    try {
+      ClassLoader classLoader = getClass().getClassLoader();
+      Reflections reflections = new Reflections(new ConfigurationBuilder().setUrls(effectiveClassPathUrls(classLoader)));
```
nickwallen: Would it make sense to introduce some kind of 'import' like functionality?  Instead of searching the entire classpath, the user would specifically define the locations of Stellar functions that they want to use.

I am curious if a search of the entire classpath will be tenable in all situations.  In terms of performance, its border line acceptable now, but how might that behave differently in other environments; could it be even slower?  

A full classpath search may also increase the chance of unintended naming collisions.  And in the case of a collision, the definition that wins is non-deterministic. That would cause a major migraine to troubleshoot.

cestella: So the class path search happens once in the prepare phase of storm and
name collision results in an initialization failure.

I think it is quite tenable now, but we can consider an import statement as
a follow-on.

nickwallen: Yes, I think it would be a good follow-on.  I really just bring it up as a discussion point to think about.  Good idea, bad idea?

nickwallen: It just pains me to see us spending so much time searching jars in the classpath that we know will never contain any Stellar functions. :)  It currently searches all of the dependencies like HBase, Storm, Mockito, Jackson, etc.  It is all of those that give us the 5-8 second delay.

An easy way to do this would be to leverage the `Context` functionality that already exists.  

1. Create `Capabilities.STELLAR_PATH` that is either a colon-separated list of paths to search (like the Java CLASSPATH) or a regular expression of paths to search (ex; `.*metron-common.*`).

2. Update `FunctionResolverSingleton.initialize(Context)` to extract the new capability, if it exists.  If the capability does not exist, or `initialize` is not called as part of lazy initialization, then choose a sensible default, which could be either:
   * Include All - The current behavior now which searches the entire classpath
   * Metron Only - Include only Metron jars known to have Stellar functions; `.*-metron.*`
   * Something in between like "anything that looks like `.*-stellar-.*` or `.*-metron.*`

3. Update `FunctionResolverSingleton.loadFunctions` so that it only searches the paths defined by the variable from step 2.

Even if we default to "Include All" this approach at least lets me update unit and integration tests to confine the breadth of the classpath search so that we don't introduce unnecessary delay running them.

cestella: I think that's a fine addition as a separate JIRA.  At the moment, this is executed once per topology start, so I'm ok with this default implementation.  Also, keep in mind that there are edge cases here, specifically around jars with other jars referenced in their manifest that the Reflections API is handling under the hood.  It's a worthy task, but one that should be a follow-on to this task.

nickwallen:
FYI - Your branch needs merged with master before we can push this one along any further.  Not sure if you've missed that.

cestella:
Yeah getting to that; in a bit of a Internet connection blackhole

cestella:
Ok, merged master in.

dlyle65535:
+1, good deal.

nickwallen:
I have been tracing an issue caused by static initialization.  I have a Stellar function that uses values within the Context to perform initialization.  I have multiple tests in the same class; test1(), test2(), etc.  In each of these tests, I need to initialize the function with different values in the Context.  

If I run each test individually, everything works just fine, but if I run all the tests together the tests fail. The problem is that the functions don't actually re-initialize in test2() after they have already been initialized in test1().  

I think I need a way to manually clear the functions cached in `FunctionResolverSingleton._getFunctions()` that I would call before running each unit test.

nickwallen:
Due to the static initialization and use of a singleton for the function resolver, this was a bit difficult to trace.  We should think about making this easier to grok as a follow-on to this PR.  This would be painful for a new user creating their own functions.  I don't personally have a solution right now, but need to noodle on it.

cestella:
Yeah, I see what you're saying; that's a good catch.  You are doing some testing for initialization and you want to uninitialize.  One thing we could do is provide a function in the FunctionResolver to retrieve the StellarFunction given a name and you could set your initialized boolean to false when you wanted to selectively reset.

nickwallen:
I had to address this issue as part of some other work.  It is included in #242.

PR #236, opened 2016-08-30
METRON-389 Create Java API to Read Profile Data During Model Scoring

nickwallen:
[METRON-389](https://issues.apache.org/jira/browse/METRON-389)

### Changes
* Created a Java API for reading profile data written by the Profiler.  This is contained within a new project `metron-analytics/metron-profiler-client`.
* Altered the Profiler to use deterministic row keys so that multiple Gets can fetch the data.  This should scale much better at high volumes.

### Points of Interest
* `ProfileMeasurement` - Represents a single data point within a Profile.  A Profile is effectively a time series.  To this end a Profile is composed of many ProfileMeasurement values which in aggregate form a time series.  
* `ProfilePeriod` - The Profiler captures one `ProfileMeasurement` each `ProfilePeriod`.  A `ProfilePeriod` will occur at fixed, deterministic points in time.  This allows for efficient retrieval of profile data.
* `RowKeyBuilder` - Builds row keys that can be used to read or write profile data to HBase.
* `ColumnBuilder` - Defines the columns of data stored with a profile measurement.
* `ProfileHBaseMapper` - Defines for the `HBaseBolt` how profile measurements are stored in HBase.  This class leverages a `RowKeyBuilder` and `ColumnBuilder`.

### Testing
Spin up a Metron install and write some Profile data to HBase.  Read the profile data using `metron-profiler-client`.


metron-analytics/metron-profiler/src/main/flux/profiler/remote.yaml
```diff
@@ -25,11 +25,28 @@ components:
...
+
+    -   id: "hbaseMapper"
+        className: "org.apache.metron.profiler.bolt.ProfileHBaseMapper"
+        properties:
+            - name: "rowKeyBuilder"
+              ref: "rowKeyBuilder"
+            - name: "columnBuilder"
+              ref: "columnBuilder"
+            - name: "executor"
+              ref: "defaultExecutor"
```
nickwallen: I have multiple components that each need their own `StellarExecutor`. Does Flux treat the `defaultExecutor` reference as a singleton and so every reference to `defaultExecutor` is pointing to the same object?  Or is treated more like a Spring Bean's 'Prototype' scope?  I would think it would have to be treated as a prototype, but I'd like to make sure.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -100,7 +101,11 @@ public ProfileBuilderBolt(String zookeeperUrl) {
   @Override
   public Map<String, Object> getComponentConfiguration() {
     Config conf = new Config();
-    conf.put(Config.TOPOLOGY_TICK_TUPLE_FREQ_SECS, getFlushFrequency());
+
+    // how frequently should the bold receive tick tuples?
```
cestella: s/bold/bolt

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -194,8 +194,8 @@ private void init(Tuple input) {
       expressions.forEach((var, expr) -> executor.assign(var, expr, message));
 
     } catch(ParseException e) {
-      String msg = format("Bad 'init' expression: %s, profile=%s, entity=%s, start=%d",
-              e.getMessage(), measurement.getProfileName(), measurement.getEntity(), measurement.getStart());
+      String msg = format("Bad 'init' expression: %s, profile=%s, entity=%s",
+              e.getMessage(), measurement.getProfileName(), measurement.getEntity());
```
cestella: Can we LOG.error these here?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/hbase/SaltyRowKeyBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+   *
+   * @param profile The name of the profile.
+   * @param entity The name of the entity.
+   * @param groups The group(s) used to sort the profile data.
+   * @param durationAgo How long ago?
+   * @param unit The time units of how long ago.
+   * @return All of the row keys necessary to retrieve the profile measurements.
+   */
+  @Override
+  public List<byte[]> rowKeys(String profile, String entity, List<Object> groups, long durationAgo, TimeUnit unit) {
```
cestella: Is there a reason why we do not have a method on RowKeyBuilder which takes an arbitrary start and end time?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/hbase/Serializer.java
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Serialize a profile measurement's value.
+   *
+   * The value produced by a Profile definition can be any numeric data type.  The data
+   * type depends on how the profile is defined by the user.  The user should be able to
+   * choose the data type that is most suitable for their use case.
+   *
+   * @param value The value to serialize.
+   */
+  public static byte[] toBytes(Object value) {
```
cestella: If we are restricting to numeric data, why not have toBytes take a Number argument as opposed to Object?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
nickwallen: Yes, they already are LOG.error'd inside `execute`.  

The point of this catch is to add the extra note that the failure came from the 'init' expression, rather than one of the other expressions involved in creating a profile.  I wanted to make it blatantly obvious to the user which expression failed.

cestella: Ok, retracted

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/hbase/SaltyRowKeyBuilder.java
nickwallen: Just because it matches the `ProfilerClient` interface.  Could easily be changed to start/end.  What are you thinking the advantages of doing so are?  Easier to test?

cestella: Nah, I was thinking that we might compare output of a given profile against a lookback not including the current range for outlier analysis (i.e. comparing the current profile period against the profile period of the last month trailing the current profile period).

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/hbase/Serializer.java
nickwallen: Yes, I like that.  I went down this path before, but ran into some unrelated issues and ended up unwinding the change.  There are at least 4 touch points that would change based on this; all good changes.

Wondering if I should just submit this change as a separate PR to make my life easier with the other dependent changes I have in the pipeline.  Looking at the code now  to see how painful that might be.

cestella: I don't see a strong reason to stress about it as long as it gets done in a follow-on PR at the end of the pipeline.  Do me a favor and evaluate the impact, though.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/hbase/SaltyRowKeyBuilder.java
nickwallen: Then we would also want to change the ProfilerClient interface to let you do this.  I think its a reasonable change.  I'd prefer to do this one as a separate PR as it would definitely impact the PRs in the pipeline.

cestella:
Ok, I think I'm +1 on this pending the two JIRAs mentioned from the review:
* Adjust `ProfilerClient` to add a method that takes a start/end time method. 
* Change the `Serializer.toBytes(Object)` to `Serializer.toBytes(Number)`

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/hbase/Serializer.java
nickwallen: Yep, the suggested change looks good.  Makes the code more clear in a few places too.  I'll do a follow-on PR.  Created [METRON-412](https://issues.apache.org/jira/browse/METRON-412) to track this.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/hbase/SaltyRowKeyBuilder.java
nickwallen: Added [METRON-413](https://issues.apache.org/jira/browse/METRON-413) to track this change.

nickwallen:
Created METRON-412 and METRON-413 to track those changes.  Will merge once the final CI build completes.

PR #237, opened 2016-08-30
METRON-399 Stellar Date Functions Should Default to Current Time

nickwallen:
### [METRON-399](https://issues.apache.org/jira/browse/METRON-399)

The Stellar date functions expect an argument of a Long containing the epoch time in milliseconds.  If no argument is supplied, a parse exception is thrown.

```
WEEK_OF_YEAR(timestamp)
```

This should be changed so that if no argument is supplied, the function assumes the current time.

```
WEEK_OF_YEAR()
```

If an argument is passed that is not valid, the function should continue to return null.

This updates the functioning of all date functions included in #231 .


cestella:
+1, good stuff

PR #239, opened 2016-08-31
METRON-401 ParseException Thrown by ThreatIntelJoinBolt

nickwallen:
If the configuration for the ThreatIntelJoinBolt used any Stellar code that called a function, the following error occured.

```
org.apache.metron.common.dsl.ParseException: Unable to find string function IN_SUBNET.  Valid functions are...
```

The `FunctionResolver` used by the `ThreatIntelJoinBolt` was not being initialized.  If the configuration included any function calls, the functions could not be resolved and a ParseException was thrown.  

Configuration containing Stellar code that does not call any functions was not impacted.  This is why the existing unit and functional tests did not catch this.  Updated the unit tests to validate the fix.


cestella:
+1 pending travis checks

dlyle65535:
+1 Ran this up on Quick Dev. Thanks for the quick turn!

PR #240, opened 2016-09-02
METRON-404 Stellar Compiler Hides Function Initialization Errors

nickwallen:
The Stellar compiler is swallowing exceptions thrown during function initialization. The compiler then reports that the function does not exist when in fact the function exists, but initialization failed. The actual error during initialization is lost and not reported anywhere.

For example, I have a function `PROFILE_GET` that is defined, but is set to fail on initialization.  Prior to this change, an error would be reported that is misleading and hides the actual failure during initialization.
```
org.apache.metron.common.dsl.ParseException: Unable to find string function PROFILE_GET.  Valid functions are IS_DATE,IS_DOMAIN,MONTH,...
	at org.apache.metron.common.stellar.StellarCompiler.exitTransformationFunc(StellarCompiler.java:249)
	at org.apache.metron.common.stellar.generated.StellarParser$TransformationFuncContext.exitRule(StellarParser.java:1616)

```

After this change, the error looks like the following.
```
org.apache.metron.common.dsl.ParseException: Unable to initialize function 'PROFILE_GET'

	at org.apache.metron.common.stellar.StellarCompiler.initializeFunction(StellarCompiler.java:286)
...
Caused by: java.lang.IllegalStateException: missing required context: PROFILER_HBASE_TABLE, PROFILER_ROW_KEY_BUILDER, PROFILER_COLUMN_BUILDER
	at org.apache.metron.profiler.client.stellar.GetProfile.validateContext(GetProfile.java:121)
	at org.apache.metron.profiler.client.stellar.GetProfile.initialize(GetProfile.java:68)
	at org.apache.metron.common.stellar.StellarCompiler.initializeFunction(StellarCompiler.java:282)
	... 38 more
```

This PR is dependent on #233 and METRON-396.

PR #242, opened 2016-09-06
METRON-391 Create Stellar Function to Read Profile Data for Model Scoring

nickwallen:
### [METRON-391](https://issues.apache.org/jira/browse/METRON-391)

A Java API was created as part of #236  to allow Profile data to be read during model scoring. A Stellar function(s) was created that uses this API to read the profile data. This would be used in conjunction with other Stellar functions like MODEL_APPLY.

#### Dependencies
This PR depends on a few other existing PRs that should be merged prior to this one.
- [x] #233 
- [x] #240
- [x] #236 
- [x] #237 

#### Examples

Retrieve all values for 'entity1' from 'profile1' over the past 4 hours.
```
PROFILE_GET('profile1', 'entity1', 4, 'HOURS')
```

Retrieve all values for 'entity1' from 'profile1' over the past 2 days.
```
PROFILE_GET('profile1', 'entity1', 2, 'DAYS')
```

Retrieve all values for 'entity1' from 'profile1' that occurred on 'weekdays' over the past month.
```
PROFILE_GET('profile1', 'entity1', 1, 'MONTHS', 'weekdays')
```

### DIFF
--- a/metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
+++ b/metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
@@ -21,27 +21,32 @@
 package org.apache.metron.profiler.client.stellar;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.hbase.HBaseConfiguration;
 import org.apache.hadoop.hbase.client.HTableInterface;
 import org.apache.metron.common.dsl.Context;
 import org.apache.metron.common.dsl.ParseException;
 import org.apache.metron.common.dsl.Stellar;
 import org.apache.metron.common.dsl.StellarFunction;
 import org.apache.metron.common.utils.ConversionUtils;
+import org.apache.metron.hbase.HTableProvider;
+import org.apache.metron.hbase.TableProvider;
 import org.apache.metron.profiler.client.HBaseProfilerClient;
 import org.apache.metron.profiler.client.ProfilerClient;
 import org.apache.metron.profiler.hbase.ColumnBuilder;
 import org.apache.metron.profiler.hbase.RowKeyBuilder;
+import org.apache.metron.profiler.hbase.SaltyRowKeyBuilder;
+import org.apache.metron.profiler.hbase.ValueOnlyColumnBuilder;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static java.lang.String.format;
-import static org.apache.metron.common.dsl.Context.Capabilities.PROFILER_COLUMN_BUILDER;
-import static org.apache.metron.common.dsl.Context.Capabilities.PROFILER_HBASE_TABLE;
-import static org.apache.metron.common.dsl.Context.Capabilities.PROFILER_ROW_KEY_BUILDER;
+import static org.apache.metron.common.dsl.Context.Capabilities.GLOBAL_CONFIG;
 
 /**
  * A Stellar function that can retrieve data contained within a Profile.
@@ -76,6 +81,24 @@ import static org.apache.metron.common.dsl.Context.Capabilities.PROFILER_ROW_KEY
 )
 public class GetProfile implements StellarFunction {
 
+  /**
+   * A global property that defines the name of the HBase table storing profile definitions.
+   */
+  public static final String PROFILER_HBASE_TABLE = "profiler.hbase.table";
+
+  /**
+   * A global property that defines the name of the column family used to store profile data.
+   */
+  public static final String PROFILER_COLUMN_FAMILY = "profiler.column.family";
+
+  /**
+   * A global property that defines the name of the HBaseTableProvider implementation class.
+   */
+  public static final String PROFILER_HBASE_TABLE_PROVIDER = "profiler.hbase.table.provider";
+
+  /**
+   * A client that can retrieve profile values.
+   */
   private ProfilerClient client;
 
   /**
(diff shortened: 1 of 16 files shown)


nickwallen:
Rewrite the commit history to match master.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
```diff
@@ -0,0 +1,179 @@
...
+
+  private ProfilerClient client;
+
+  /**
+   * Initialization.
+   */
+  @Override
+  public void initialize(Context context) {
+
+    Context.Capabilities[] required = { PROFILER_HBASE_TABLE, PROFILER_ROW_KEY_BUILDER, PROFILER_COLUMN_BUILDER };
```
cestella: If these capabilities are not specified, please pull them from the global config as backup.

cestella:
It appears that these functions require a number of capabilities (profiler hbase table, etc) that are not setup by default in the parser or enrichment topology.  Have you tested calling this function from a field transformation in the parser or as part of a triage rule?

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
nickwallen: Not sure what you mean.  Can you point me to some code?

cestella: Sure, You notice [here](https://github.com/cestella/incubator-metron/blob/cd6348279744ba521d99f1dbe71030fafad6cd1a/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/stellar/SimpleHBaseEnrichmentFunctions.java#L91) that I'm doing something similar except instead of insisting that the capability be there prior to the function, I look for the configuration information (i.e. the table provider and such) in the global config.  This way, we don't have to add new capabilities to every single topology (parser, enrichment and now profiler) for every stellar function.  Instead, they can pull their configs from the global config.

cestella: also, I will note that it tees us up to add support for reinitialization when configs change.  That way the functions can adjust their configs at runtime based on a zookeeper update.

nickwallen:
No, the new capabilities are not defined in those topologies.  I wasn't sure how we wanted to handle that.  

It seems a little icky to define these capabilities in our existing topologies.  When we continue to add new Stellar functions, are we also going to always add the necessary Capabilities in the topology initialization?  This gets messy quick.

Or If a user adds a Stellar function in their own JAR, how are they going to use it if it requires additional Capabilities in the Context which is hard-coded in Metron? 

I feel like we need to externalize this somehow.  The current mechanism might not scale too well.  What do you think?

cestella:
@nickwallen Yeah, that's why I don't want to treat these as capabilities, but rather pull them from the global configs.  The function should configure itself from the global config, IMO.  That way the only capability that we need for config is `GLOBAL_CONFIG` and maybe `SENSOR_CONFIG`.  I added and used the `GLOBAL_CONFIG` capability in my PR for the HBase enrichment functions [here](https://github.com/apache/incubator-metron/pull/234)

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
```diff
@@ -0,0 +1,179 @@
...
+  public Object apply(List<Object> args, Context context) throws ParseException {
+
+    String profile = getArg(0, String.class, args);
+    String entity = getArg(1, String.class, args);
+    long durationAgo = getArg(2, Long.class, args);
+    String unitsName = getArg(3, String.class, args);
+    TimeUnit units = TimeUnit.valueOf(unitsName);
+    List<Object> groups = getGroupsArg(4, args);
+
+    return client.fetch(profile, entity, durationAgo, units, Integer.class, groups);
```
nickwallen: I did not mean to leave this hard-coded as an integer.  Trying to think of a good way to specify this easily from Stellar.  I am not sure how yet.  Hmm.

cestella: What is hard coded as an integer?  The argument index (e.g. `getArg(i, ...)`)?

cestella: nvm, you mean the return for the client.fetch, right?

nickwallen: Yes, exactly.  Specifying the type of numbers you expect to get back from the profile. 

Could accept a string argument, "integer", "double", etc, and then manually convert to the correct class; Integer.class, Double.class. But I really don't want to do it that way.

cestella: Now that I think about it, I'm not sure why you're assuming that the result is always a Number?  You could pass back a bloom filter from the profiler just as easily or a biased sample or whatever, right?

nickwallen: Yes, I think that would be a good enhancement.  But right now it assumes Numeric.  Mainly because of serialization.  Opened METRON-415 to track the enhancement of supporting a variety of types as profile values.

nickwallen:
Ok.  We probably need to document the distinction between a Capability and a Global config.  It was not clear to me.

nickwallen:
What I mean is, add some docs describing when it is appropriate to use a Capability versus the global configuration.  Having multiple ways of configuring things is a little icky.  I'm sure there is a good reason why, but I have no idea because there are no comments/docs in the code.

cestella:
Agreed, however in this case and in every case I can think of as a stellar function writer, you should use the configs from zookeeper to configure your function.

nickwallen:
The Stellar function is now configured via global properties instead of Capabilities.

cestella:
+1, looks good.  I'll tackle METRON-415 to expand this to non-numeric data.

PR #243, opened 2016-09-06
METRON-406: Stellar variable resolution does not resolve variables with ':' in them

cestella:
This is entirely tested via unit tests at the moment.
I will spin up a vagrant box to ensure no obvious regressions were created.


merrimanr:
+1

PR #244, opened 2016-09-06
METRON-407: We currently do not provide defaults if the Stix Observable does not specify a condition

cestella:
When manually creating indicators in Soltra, the condition for the observable is not specifiable via the Soltra UI. As such, Soltra does not specify this condition explicitly, expecting the default to be "equals" if unspecified.
We should, if the condition is unspecified, assume equality.

This is currently tested via unit tests.  I will spin it up in vagrant and update the PR.


cestella:
Just for posterity, from [here](http://stixproject.github.io/documentation/suggested-practices/) justifying considering an absence of a condition to be an equality condition:

> When creating observables for use as patterns within indicators, you should always set the condition attribute on all possible fields to an appropriate value, even if that value is equals. Leaving off the condition attribute implies that the observable is an instance rather than a pattern.

merrimanr:
+1

PR #245, opened 2016-09-07
METRON-408 Intermittent Failures of Profile Integration Tests

nickwallen:
[METRON-408](https://issues.apache.org/jira/browse/METRON-408)

The profiler configuration was not being updated from Zookeeper correctly in all cases.  After some period of time the configurations would be updated and the tests would work.  But if this period of time was greater than 90 seconds, the amount of time the profile integration tests wait, then the tests would fail.  

In all cases, messages were being sent through Storm prior to the bolt actually receiving the profiler configuration, which would cause the `ProfileSplitterBolt` to throw an exception indicating that the configuration could not be found. This message `Fatal: Unable to find valid profiler definition` was a symptom of the problem and occurred much more frequently than the tests actually failed.

Since this issue was intermittent, I cannot be sure it is completely gone, but I have been unable to replicate the problem after these changes.  I also am unable to replicate the `Fatal: Unable to find valid profiler definition` error after this fix.


cestella:
Looks good, +1

PR #246, opened 2016-09-10
METRON-411 Support Greater Range of Profile Periods 

nickwallen:
### METRON-411

Prior to this change the user defines the profile period duration as the number of 'periods per hour'.  For a 15 minute period, the user would define 4 periods per hour.  With this change, the user instead defines the duration of each profile period; '15 minutes' or '4 hours'.
 
### Problem

The current Profiler's deterministic row key design limits the range of profile period durations available to a user.  The original goal was to support 15 minute periods which it does, along with 5, 20, 30 and 60 minute periods.  Different types of profiles and input data will require a range of periods and the Profiler should be flexible in this regard.

Currently, a profile period must start on each and every hour.  While 1, 2, 3, 4, 6, or 240 periods per hour is acceptable, 5 periods per hour is not.  The number of periods per hour must be either a multiple or divisor of 60.  Also, a period cannot be any longer than an hour.  In some cases 2, 4, or 8 hour periods might be useful.

### Solution

This change uses a much simpler method of calculating row keys and allows the user to set the  specific duration of each profile period.  Based on this duration, each period starting at the epoch, is assigned a monotonically increasing value which is used as part of the row key.  Periods can span multiple hours and there is no arbitrary durations that are unacceptable, like 5 periods per hour.

Depends on existing pull requests.  The diff might look a little wonky until these PRs are merged into master.
- [x] #236
- [x] #237 
- [x] #242

### DIFF
--- a/metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/GetProfileTest.java
+++ b/metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/GetProfileTest.java
@@ -114,6 +114,8 @@ public class GetProfileTest {
    */
   @Test
   public void testWithNoGroups() {
+    final long periodDuration = 15;
+    final TimeUnit periodUnits = TimeUnit.MINUTES;
     final int periodsPerHour = 4;
     final int expectedValue = 2302;
     final int hours = 2;
@@ -122,7 +124,7 @@ public class GetProfileTest {
 
     // setup - write some measurements to be read later
     final int count = hours * periodsPerHour;
-    ProfileMeasurement m = new ProfileMeasurement("profile1", "entity1", startTime, periodsPerHour);
+    ProfileMeasurement m = new ProfileMeasurement("profile1", "entity1", startTime, periodDuration, periodUnits);
     profileWriter.write(m, count, group, val -> expectedValue);
 
     // execute - read the profile values - no groups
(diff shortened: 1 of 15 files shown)


nickwallen:
Rewrote history to match master and #242 on which this is dependent.

nickwallen:
Reconciled with master

cestella:
Could you please walk us through a test plan of how this feature should be exercised in quickdev?

nickwallen:
Create a profile using the profiler just as you would before.  Instead of defining the number of 'periods per hour' just define the duration of each period '20 minutes' or '4 hours'.  The behavior of the profiler should be exactly the same.

All of the unit and integration tests have been updated to reflect this change and should pass.  The integration test itself uses a '5 second' profile period.

cestella:
Cool, looks great.  Much cleaner. :)

+1 if you tested in vagrant already.  If not, then I'll hold the +1 til you let me know.

cestella:
I'm specifically interested in making sure that you can call the stellar functions around the profiler from within the enrichment topology on vagrant, actually.  I may be missing it, but I don't see a dependency from the enrichment project to the profiler client.

nickwallen:
There may be some dependencies we need to change to make the Stellar client functions visible out-of-the-box in the Enrichment topology.  This PR doesn't touch the Profiler client side of things though. That seems unrelated in my mind.  Might make sense to work that separately.

cestella:
That sounds fair.  +1 as long as in that follow-on PR gets created and prioritized.

nickwallen:
Created METRON-420 for that.  Will take a run at it.

PR #247, opened 2016-09-12
METRON-412 Limit Profiler to Numeric Types Via Type System

nickwallen:
[METRON-412](https://issues.apache.org/jira/browse/METRON-412)

A `ProfileMeasurement` value must now be a `Number` not just any `Object`.  This more accurately reflects the requirements of a value, which must be numeric.  This was previously only checked at run-time, but with these changes it is now enforced at compile-time.

In addition, changed the `Serializer` from a simple container of two static methods, to an "interface plus implementation" that can be optionally swapped-out with a different implementation, should that be needed.  I imagine that someone may want to support BigInteger or something else like that in the future. 

Depends on existing pull requests. 
- [ ] #242


nickwallen:
Opened [METRON-415](https://issues.apache.org/jira/browse/METRON-415) in lieu of this PR.  This will provide greater flexibility for users of the Profiler.

PR #248, opened 2016-09-12
METRON-414: Kibana Ansible Install Fails with SSL Error

dlyle65535:
It looks as if nodesource introduced SNI over the weekend. Curl 7.18.1 supports this, so I've switched to curl for the nodesource repo download.

Tested on Quickdev Vagrant.


cestella:
+1, pending travis build

PR #249, opened 2016-09-13
METRON-413 Allow Start/End Time Range Search in Profiler Client API

nickwallen:
The current Profiler client API introduced in METRON-389 has an interface to search X seconds/minutes/days ago.  A start/end time range will give the user more flexibility in extracting the data they need.


cestella:
+1

PR #250, opened 2016-09-13
METRON-416: Provide the ability to store mergeable data structures for summarizing data on-line

cestella:
With the currently worked on profiler, we it will be advantageous for us to be able to store more than just numeric data as a profile. In particular, adding some common data structures that can be merged will allow us to store data at a fixed tick-rate, but merge the results from hbase across multiple ticks.

The following structures should be supported:
* An implementation of the `STATS` stellar functions which is backed by an online and mergeable class using distributional sketches for querying distribution.
* An implementation of bloom filters so that probabalistic existence queries can be made

This should facilitate simple statistical outlier analysis.


metron-platform/metron-common/src/main/java/org/apache/metron/common/math/stats/OnlineStatisticsProvider.java
```diff
@@ -0,0 +1,243 @@
...
+   * @param value
+   */
+  @Override
+  public void addValue(double value) {
+    long n1 = n;
+    min = min == null?value:Math.min(min, value);
+    max = max == null?value:Math.max(max, value);
+    sum += value;
+    sumOfLogs += Math.log(value);
+    sumOfSquares += value*value;
```
nickwallen: Do we need to check for overflow?  Does it make sense to accumulate doubles in a double?  Shouldn't we use something bigger than a double, like a BigDecimal?

cestella: Yeah, I'd prefer to avoid `BigDecimal` if at all possible due to performance concerns.  Also, I will point out that Commons Math also [uses](https://github.com/apache/commons-math/blob/050dfa6f0850174fdf958ce6751d2f06900201f8/src/main/java/org/apache/commons/math4/stat/descriptive/summary/SumOfSquares.java) `double` for its accumulation, so I think we're pretty safe for some large amount of headroom.  That being said, I will do the math ops via `ExactMath` so it throws an exception on overflow.  Good catch.

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarStatisticsFunctionsTest.java
```diff
@@ -215,10 +261,11 @@ public void testSumSquares() throws Exception {
...
   }
 
-  @Test(expected = ParseException.class)
+  @Test
   public void testKurtosisNoWindow() throws Exception {
     statsInit(0);
-    run("STATS_KURTOSIS(stats)", variables);
+    Object actual = run("STATS_KURTOSIS(stats)", variables);
+    assertEquals(stats.getKurtosis(), (Double) actual, 0.1);
   }
```
nickwallen: I think you can get rid of all the "test*[No]Window" tests (for skewness, percentile, and kurtosis) and replace them with a single test just like the others (max, min, count, etc) that test both scenarios.  Those were created only for tests where there was a behavioral difference between the 'windowed' versus 'non-windowed' implementation.

cestella:
Done, thanks for the comments.

nickwallen:
I don't see any unit tests that check for correct handling of overflow or underflow.  Am I missing them?

I modified the tests to see what happens and when you actually hit an overflow condition.  The code correctly detects it from what I can tell, but there is no indication that it occurred.  Similar to METRON-404, but not during function initialization.  The exception gets swallowed and all you see is "ParseException: Unable to pop an empty stack".  Could be addressed as a separate PR, but worthy of a fixin.

cestella:
Agree, both of them are irritating and now fixed.

nickwallen:
Nice +1 pending CI build

cestella:
Validated on Vagrant.

PR #251, opened 2016-09-14
METRON-418 Set TTL on HBase Puts

nickwallen:
For METRON-417, the goal is to use HBase's TTL mechanism to expire or age-out Profile data.  This PR provides the mechanism to set TTL values on HBase Puts.  

The HBaseClient is used to write ProfileMeasurement values to HBase. The client was updated so that a TTL can be attached to each Put. This will allow us to attach a TTL to each ProfileMeasurement written by the Profiler.

Tests were added to validate successful reads/write to HBase with a TTL, along with successful expiration of data based on a TTL.

Depends on the following existing PRs...
- [x] #246

PR #252, opened 2016-09-14
METRON-419 Update Tuple to HBase Mapper/Bolt to Set TTL

nickwallen:
For METRON-417, the goal is to use HBase's TTL mechanism to expire or age-out Profile data. This PR is another step toward that goal.

This PR alters the mechanism that maps tuples to HBase so that a TTL can be specified in a generic way. Each tuple is capable of being written to HBase with its own unique TTL value.

The dependency on the storm-hbase project, which is part of Apache Storm, has been removed.  The storm-hbase code could not be enhanced to support a TTL without changing the underlying implementation.  The HBaseMapper interface was altered so that implementation can specify the TTL value for each tuple.

Unit tests have been added to test writes with and without a TTL to HBase.

Depends on the following existing PRs...
- [x] #246 
- [x] #251


nickwallen:
merged master

cestella:
+1

PR #253, opened 2016-09-15
METRON-415: Allow a Profile to Store Any Type as its Value

cestella:
The profiler currently only supports profile's with one of the following numeric types; Integer, Short, Long, Double, or Float. There will be cases where a user will want to store non-numeric Objects like distributional statistics or a bloom filters. Enhance the Profiler to allow storage of any Object.


nickwallen:
+1

PR #255, opened 2016-09-15
METRON-422: Remove bluecoat.json

cestella:
title says it all.


dlyle65535:
+1. RPMs built and quick-dev ran.

PR #257, opened 2016-09-16
METRON-426: Stellar does not support scientific notation as a literal

cestella:
Stellar does not support scientific notation for expressing literal doubles. It should.

This is exercised in the unit tests for `StellarStatisticsTest` and `StellarTest` directly


cestella:
The issue around the `StellarScientificTest` change is that merging t-digests is not precisely the same as adding the same data to one t-digest.  Whether this is a bug in the t-digest code or whether it's an artifact of the theory, I don't know.  Because this test is using low amounts of data, it's prone to a bit of jitter it seems and the compression level specified before isn't sufficient to ensure that the epsilon chosen (1e-2) is sufficient for repeated runs.  I tightened up the compression level to 150 (which is still quite sensible) to make sure we can support at least 1e-2 accuracy for normally distributed data across 100 runs of 1000 elements.

metron-platform/metron-common/src/main/java/org/apache/metron/common/math/stats/OnlineStatisticsProvider.java
```diff
@@ -43,7 +43,7 @@
    * 100 is a sensible default and the number of centroids retained (to construct the sketch)
    * is usually a smallish (usually < 10) multiple of the compression.
    */
-  public static final int COMPRESSION = 100;
+  public static final int COMPRESSION = 150;
```
justinleet: Could you update the comment to reflect the compression default change?

cestella: yep

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
```diff
@@ -173,6 +173,22 @@ public void testMapConstant() {
...
+      Assert.assertEquals(3, ((Number)run(query, ImmutableMap.of("casey", "casey"))).doubleValue(), 1e-3);
+    }
+    {
+      String query = "1.2 + 2";
+      Assert.assertEquals(3.2, ((Number)run(query, ImmutableMap.of("casey", "casey"))).doubleValue(), 1e-3);
+    }
+    {
+      String query = "1.2e-3 + 2";
+      Assert.assertEquals(1.2e-3 + 2, ((Number)run(query, ImmutableMap.of("casey", "casey"))).doubleValue(), 1e-3);
+    }
```
nickwallen: The `ImmutableMap.of("casey","casey")`(besides promoting your global brand ;) ) do nothing for these tests, right?  It'd be more clear, if we removed those.  Everything else looks good.  I'm a +1 if we can take care of this small issue.

cestella: You got it. :)

nickwallen:
+1 Less ego-driven. Ha.

nickwallen:
@justinleet I think everything you mentioned has been addressed.  Are you good with this PR?

PR #258, opened 2016-09-16
METRON-428: Allow a kafka offset to be passed to the ParserTopology CLI

cestella:
Currently the parser start utility does not support passing whether we want to start the topology at the beginning or where they left off.


dlyle65535:
+1 by inspection. Didn't run it up.

PR #259, opened 2016-09-16
METRON-429 Profiler Missing Dependencies When Uber Jar Deployed

nickwallen:
When the Profiler is packaged and deployed on a Metron environment, the topology does not start because the uber jar is missing HBase as a dependency.  Also, corrected the 'Getting Started' instructions in the README.

### DIFF
--- a/metron-analytics/metron-profiler-client/pom.xml
+++ b/metron-analytics/metron-profiler-client/pom.xml
@@ -44,6 +44,12 @@
         </dependency>
         <dependency>
             <groupId>org.apache.metron</groupId>
+            <artifactId>metron-integration-test</artifactId>
+            <version>${project.parent.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.metron</groupId>
             <artifactId>metron-test-utilities</artifactId>
             <version>${project.parent.version}</version>
             <scope>test</scope>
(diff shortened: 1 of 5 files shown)


cestella:
Great job, this looks good.  Could you walk us through a testing scenario in the quickdev environment with sample configs for the profiler and sample enrichment configs, please?  Similar to [here](https://github.com/apache/incubator-metron/pull/210#issuecomment-239825715).

nickwallen:
The scope of this PR is to fix a bug that causes the Profiler topology to fail during deployment due to missing dependencies.  This is supposed to fix an issue that you saw last week.  (Although, this seems to have broken the CI build so looks like I'll need to chase that now.)

I think you are thinking of METRON-421 which involves getting the Profiler Client API working in the Enrichment topology.  That work continues. Our lack of documentation around Enrichment is making METRON-421 quite painful.  But the link you sent above might help.  Thanks.

cestella:
I understand; I'll defer the request until METRON-421, then.  Sorry enrichment is biting you.

nickwallen:
Agreed. I'll have a working step-by-step set of instructions for METRON-421.  That's a good thing to have.  Maybe I'll add it as a 'Getting Started' type section to the README in `metron-profiler-client`.

cestella:
Yep, I think it's probably due for some more up-to-date instructions on at least the stellar enrichment adapter to make some of this easier.  If you have other scenarios that you're trying to do, I might be able to point you to testing instructions from other PRs that are similar enrichment-wise, too.  Just let me know.

cestella:
+1 pending CI build

PR #261, opened 2016-09-19
METRON-434: JSON Parser

cestella:
There are some situations where your data is already in JSON form and parsing should be as simple as passing the data through, adding `timestamp` and `original_message` along the way.  This JIRA will create a parser which will take a JSON map in.

Since we have standardized on one-dimensional maps, we must handle data which contains inner maps.  Inner maps should be handled pluggably by doing one of the following (configurably):
* doing nothing
* unfolding maps (e.g. `{ "foo" : { "bar" : 7 } }` is transformed into `{"foo.bar" : 7 }`)
* throwing an exception
* dropping the inner map

### DIFF
--- a/metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/json/JSONMapParserTest.java
+++ b/metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/json/JSONMapParserTest.java
@@ -77,15 +77,15 @@ public class JSONMapParserTest {
   @Test(expected=IllegalStateException.class)
   public void testCollectionHandlingError() {
     JSONMapParser parser = new JSONMapParser();
-    parser.configure(ImmutableMap.of("mapStrategy", "ERROR"));
-    List<JSONObject> output = parser.parse(collectionHandlingJSON.getBytes());
+    parser.configure(ImmutableMap.of(JSONMapParser.MAP_STRATEGY_CONFIG, JSONMapParser.MapStrategy.ERROR.name()));
+    parser.parse(collectionHandlingJSON.getBytes());
   }
 
 
   @Test
   public void testCollectionHandlingAllow() {
     JSONMapParser parser = new JSONMapParser();
-    parser.configure(ImmutableMap.of("mapStrategy", "ALLOW"));
+    parser.configure(ImmutableMap.of(JSONMapParser.MAP_STRATEGY_CONFIG, JSONMapParser.MapStrategy.ALLOW.name()));
     List<JSONObject> output = parser.parse(collectionHandlingJSON.getBytes());
     Assert.assertEquals(output.size(), 1);
     //don't forget the timestamp field!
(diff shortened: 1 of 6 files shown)


metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/json/JSONMapParserTest.java
```diff
@@ -0,0 +1,112 @@
...
+    JSONMapParser parser = new JSONMapParser();
+    parser.configure(ImmutableMap.of("mapStrategy", "ERROR"));
+    List<JSONObject> output = parser.parse(collectionHandlingJSON.getBytes());
+  }
+
+
+  @Test
+  public void testCollectionHandlingAllow() {
+    JSONMapParser parser = new JSONMapParser();
+    parser.configure(ImmutableMap.of("mapStrategy", "ALLOW"));
```
mmiklavc: Could strings instead reference the enums directly? e.g. MapStrategy.ALLOW.name()

cestella: Agreed

ottobackwards:
I have a question -

Shouldn't the addition of the timestamp and the original string be something that is handled in Stellar?

I am looking at running something through this with a field that I would want to transform with stellar into the timestamp.  Incoming data from other systems ( SIEMs etc ) may already be normalized and have a raw log data field as well.  I would think that Stellar would be the better place to these things.

ottobackwards:
Is it correct that multiple separate instances of this parser ( or any parser ) could be configured by adding multiple zookeeper jsons with separate kafka topics?

cestella:
@ottobackwards That's correct.  This is a general purpose parser and you could have multiple instances of it.

cestella:
@ottobackwards Regarding your other question, we mandated that every message capture the original message and a timestamp as required fields.  If the message doesn't have that, then it fails validation and an error is thrown for that message.

Certainly after it's parsed, you can apply a stellar transformation and replace the timestamp field with another one using Stellar.

There is currently a bug where we apply stellar transformations AFTER that validation happens, so we don't have the ability to specify a timestamp via stellar if it is not there already.  If we had that fixed, then we could apply the timestamp to the message in stellar, for sure.

ottobackwards:
Casey, would you feel about adding in recursion, such that nested maps will be unfolded?
I have that working.

ottobackwards:
Should I open a new jira and do a PR?

cestella:
@ottobackwards I think that's great work!  Why don't you submit it as a PR after this one gets through?

ottobackwards:
No problem - do you require a jira?

cestella:
Yep, make a JIRA first and then submit the PR with the JIRA name as the first part, like this one is.

ottobackwards:
METRON-449

mmiklavc:
Looks good, +1

PR #264, opened 2016-09-20
METRON-437 The Profile Definition's inputTopic field is Extraneous

nickwallen:
The `inputTopic` field that is part of the Profiler definition (aka `ProfilerConfig`) is not used.  The input topic is defined by the topology configuration contained within the static `profiler.properties` file.  The `inputTopic` field should be removed from the Profiler definition as it is misleading.

All integration tests have been updated to reflect and test this change.


cestella:
+1 by inspection

PR #265, opened 2016-09-21
METRON-438: Back the Stellar REPL with a readline implementation

cestella:
The REPL as it currently stands is great, but it could be closer in-line with the more popular REPL's by being backed by a readline implementation.
Specifically, we should:
* add variable assignment
* add loading a JSON document with variables from disk
* add the ability to do reverse searches in the history
* add the ability to use the arrow keys to recover previous inputs
* add the ability to adjust input via `~/.inputrc`
* add autocomplete for function names and variable names


nickwallen:
Love it.  

One other thing I was noticing.  Not a deal breaker, but want to get your feedback on it.  If I auto-complete a function, it adds a space after the function name.  If I type "WEEK_OF_YE" then tab, it auto-completes to "WEEK_OF_YEAR " with a trailing space.  

If I then immediately hit ENTER, Stellar executes it, interprets it as a variable name, sees no value for the variable and does nothing.  Would it be possible to auto-complete a function with a trailing '(' so that I am forced to provide arguments?  Or maybe there is a better way to handle this situation.

cestella:
@nickwallen I think I can do something about that, yep.  Let me noodle on just the right way to do it.

nickwallen:
@cestella Great.  Remember auto-complete for doc strings would need to continue auto-completing with a trailing space.

metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/shell/StellarShell.java
```diff
@@ -244,6 +304,55 @@ private void write(String out) {
...
+    String expression = output.getBuffer().trim();
+    if(StringUtils.isNotBlank(expression)) {
+      if(isMagic(expression)) {
+        handleMagic( expression);
+
+      } else if(isDoc(expression)) {
+        handleDoc(expression);
+
+      } else if (expression.equals("quit")) {
+        try {
```
nickwallen: Does this prevent the user from referring to a variable called `quit`?  If so, maybe `%quit`?

nickwallen:
+1 Works great.  I'm a go as-is, pending CI build.  Anything else you can do with auto-complete earns a gold star.

PR #270, opened 2016-09-24
METRON-449 - JSON Parser should unfold nested maps

ottobackwards:
For greater flexibility and utility the JSONMapParser should support unfolding maps to arbitrary depths.

This change introduces a new function called recursively to handle this, because the enum lambdas cannot reference themselves.


ottobackwards:
I don't think I should be carrying the other commits.  I didn't rebase.  I'll redo

PR #271, opened 2016-09-24
METRON-449 JSONMapParser should unfold maps to arbitrary depths

ottobackwards:
The JSON Parser currently only unfolds maps at the root level.  The parser would support a wider range of sources is it supported unfolding maps at any depth of the document.

METRON-449 adds a recursive function to unfold Maps as they are encountered when parsing.  A new function was added because the enum lambdas cannot reference themselves.

### DIFF
--- a/metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/json/JSONMapParserTest.java
+++ b/metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/json/JSONMapParserTest.java
@@ -56,7 +56,7 @@ public class JSONMapParserTest {
 
   /**
    {
-    "collection" : { "blah" : 7, "blah2" : "foo" }
+    "collection" : { "blah" : 7, "blah2" : "foo", "bigblah" : { "innerBlah" : "baz", "reallyInnerBlah" : { "color" : "grey" }}}
    }
    */
    @Multiline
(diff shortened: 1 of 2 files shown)


metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/json/JSONMapParserTest.java
```diff
@@ -56,7 +56,7 @@ public void testHappyPath() {
 
   /**
    {
-    "collection" : { "blah" : 7, "blah2" : "foo" }
+    "collection" : { "blah" : 7, "blah2" : "foo", "bigblah" : { "innerBlah" : "baz", "reallyInnerBlah" : { "color" : "grey" }}}
```
cestella: I like it!  Would you make one more test case that involves just a single map?

cestella: like 
```
{ 
  "collection" : {
    "key" : "value"
  },
"key" : "value"
}
```

ottobackwards: Sure, I'll do that

ottobackwards:
@cestella , from looking at the ES templates etc, it seems that ':' is used as the name separator and not '.'.  Should the the unfolding not use the same?

cestella:
No, ES uses ':' and modifies the keys upon write converting '.' to ':'.
Before that, '.' is the preferred source.  I'm hoping we can go back to '.'
eventually.  Also, non-ES indices are perfectly fine with '.'.

ottobackwards:
Never mind - just found the field name converter

cestella:
+1, tested in single node vagrant.  

Did the following:
* Created a config called `custom` at `$METRON_HOME/config/zookeeper/parsers/custom.json`:
```
{
  "sensorTopic":"custom",
  "parserConfig" : {
    "mapStrategy" : "UNFOLD"
  }
}
```
* Created kafka topic for `custom` via `/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper node1:2181 --create --topic custom --partitions 1 --replication-factor 1`
* Loaded the configs into zookeeper via `$METRON_HOME/bin/zk_load_configs.sh --mode PUSH -i /usr/metron/0.2.0BETA/config/zookeeper -z node1:2181`
* Started the topology via `/usr/metron/0.2.0BETA/bin/start_parser_topology.sh -k node1:6667 -z node1:2181 -s custom`
* Created some sample data in `~/test_data.dat`:
```
{ "foo" : "bar" }
{ "map" : { "blah" : 1 } }
{ "foo" : "grok", "map" : { "blah" : { "tim" : "sort" } } }
```
* Loaded the sample data into the kafka queue via `cat ~/test_data.dat | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic custom`
* Checked in ES Head that I had a message with a field called `map:blah:tim`

PR #269, opened 2016-09-24
METRON-452: Add rudimentary configuration management functions to Stellar

cestella:
Currently, the only way to add enrichments is via editing JSON files and pushing them to zookeeper. Rather than that, because we have a REPL, we should create management functions around:
* Shell functions - Functions surrounding interacting with the shell in either a nicer way or a more functional way.
* Configuration functions - Functions surrounding pulling and pushing configs from zookeeper
* Parser functions - Functions surrounding adding, viewing, and removing Parser functions.
* Enrichment functions - Functions surrounding adding, viewing and removing Stellar enrichments as well as managing batch size and index names for the enrichment topology configuration
* Threat Triage functions - Functions surrounding adding, viewing and removing threat triage functions.

PR #272, opened 2016-09-24
METRON-453: Add a stellar shell function to open an external editor and return the editor's contents

cestella:
In order to create multiline strings (e.g. JSON or grok statements), we should have a stellar shell function which can open up an editor (specified by the `EDITOR` or `VISUAL` environment variable) optionally initialized by the contents of a variable and return the output of the editor when saved and closed.

This work builds on METRON-452 (#269), so it will be hard to review until that is merged.


justinleet:
+1 tested on quickdev and was able to use SHELL_EDIT with and without the optional param and with various input and on both vi and nano as editors.

PR #275, opened 2016-09-26
METRON-459 Bad file location for org.apache.metron.spout.pcap.Endianness.java

danieljue:
Class had the wrong package, causing compilation error.


cestella:
I'm +1 on the change; good catch.  Could you please make a JIRA for this and rename this PR to ${JIRA}: ${JIRA DESCRIPTION}.  For instance, if the JIRA is METRON-411 adn the description is "Fixed package error", could you call the title "METRON-411: Fixed package error"?

nickwallen:
Hi @danieljue - Can you send in the compile error that you are seeing?  I just want to make sure I can replicate it.  Thanks!

nickwallen:
For me, master builds and all integration tests pass.  Also, I don't see any failures in Travis CI.  But when I specifically open up the file in Intellij, it does show an error indicating that the source file location does not match the package name.

Would be just super, if CI broke when our code is broke. :)

cestella:
maven also builds it..that was the baffling thing to me.

nickwallen:
Seeing [unsubstantiated](http://stackoverflow.com/questions/9416017/package-structure-should-be-same-as-directory-structure-in-java-always) [references](http://stackoverflow.com/questions/8395916/package-name-is-different-than-the-folder-structure-but-still-java-code-compiles) on the interwebs that enforcement of source code location to package name is optional.  Fun fact of the day.

danieljue:
https://issues.apache.org/jira/browse/METRON-459

danieljue:
So basically went from metron.pcap to metron.spout.pcap

dlyle65535:
+1 - Thanks!

nickwallen:
+1 Hoping that if we add more strict checking of our source code via checkstyle or something similar, we might catch this type of issue in the future.  But that can be left for another day.

danieljue:
Sorry, while fixing 451 I neglected to put 451's change in it's own branch.  Reverted the 451 fix from this PR and then cherry picked 451's fix into it's own branch.

nickwallen:
This one is good to go.  Thank you for the contribution! 👍

PR #280, opened 2016-09-27
Metron 398: Bump release version to 0.2.1BETA in master

mmiklavc:
Completes https://issues.apache.org/jira/browse/METRON-398

Tested this on quick-dev

I first cleared my .m2 cache just to be sure all dependencies are resolved using the newest version *only*.
`~/.m2$ rm -rf ./org/apache/metron`

**Testing done**

Full dev build
`incubator-metron$ mvn clean install`
rpm build
`incubator-metron/metron-deployment$ mvn clean install -Pbuild-rpms`
quick-dev build
`incubator-metron/metron-deployment/vagrant/quick-dev-platform$ ./run.sh`


cestella:
+1 by inspection

nickwallen:
Do we need to change something like the following?  There may be others.  This is the new Ambari installer stuff, which I don't know that well.

```
metron-deployment/packaging/ambari/src/main/resources/common-services/INDEXING/0.2.0BETA

cestella:
Oh, agreed, we should make sure the ambari stuff continues to work on this branch.  My +1 is pending that check.

mmiklavc:
@nickwallen Where is it that you're seeing that path? I don't believe that exists any longer.

nickwallen:
@mmiklavc You are right.  I was comparing Github's diff against the wrong branch.  
+1

mmiklavc:
Spun up the Ambari mpack installation and am able to run data through to indexing

PR #277, opened 2016-09-27
METRON-455: Create stellar management functions to put, get and list data in HDFS

cestella:
Create functions around reading, writing and removing local and HDFS files.

Functions added:
* Interacting with HDFS
  * `HDFS_LS`
  * `HDFS_RM`
  * `HDFS_READ`
  * `HDFS_WRITE`
  * `HDFS_READ_LINES`
* Interacting with the local filesystem
  * `LOCAL_LS`
  * `LOCAL_RM`
  * `LOCAL_READ`
  * `LOCAL_WRITE`
  * `LOCAL_READ_LINES`

You can test these by deploying the management jar as per the instructions in `metron-management` and try them out in the Stellar REPL.

PR #281, opened 2016-09-28
METRON-456: Create stellar management functions to test grok statements

cestella:
Create two stellar management functions:
* GROK_EVAL - Test grok statements
* GROK_PREDICT - Given a string, predict a grok statement that matches it.

A section in the `metron-management` README.md has been added to show these functions' as well as the functions provided by #277 to debug and create grok statements.


justinleet:
+1.

Ran up in quick dev and both functions look good.

Two notes to anyone testing this:
1. it's actually GROK_EVAL, not GROK_APPLY like in the PR description.
2. The example depends on METRON-456, but the functions work independently of that being pulled in.  I just used a couple examples on the fly, and used the standard squid pattern + a couple entries from the log for GROK_EVAL.  GROK_PREDICT will take anything.

PR #282, opened 2016-09-28
Metron 430: Ability to load specific GeoIP data to mysql server

nsinha17:
With the current implementation, the latest geoip data gets loaded to MySQl db as the url in geoip.yml is made to point to :- http://geolite.maxmind.com/download/geoip/database/GeoLiteCity_CSV/GeoLiteCity-latest.tar.xz
```
name: Download GeoIP databases
unarchive:
src: http://geolite.maxmind.com/download/geoip/database/GeoLiteCity_CSV/GeoLiteCity-latest.tar.xz
dest: /tmp/geoip
copy: no
creates: /tmp/geoip/*/GeoLiteCity-Blocks.csv
```
However, we would like to provide the user with the ability to specify the geodata that he wishes to load to the mysql and build metron project.
The user can specify the URL or the filename directly that should be picked up for loading the geo data.

**Test Scenario 1 - Specifying URL hosting Geolite DB - Build Defaults**
1. Edit `roles/mysql_server/defaults/main.yml` file and specify the Geolite DB URL for  `geodata_file` parameter. Note that this points to the Maxmind Geolite DB by default. You may leave this parameter unedited if a latest copy of the Maxmind Geolite DB is desired.
2. Now run the deployment code and the specified geodata would be populated in the mysql server.

**Test Scenario 2 - Specifying Geolite DB file manually**
1. Copy the required Geolite db file to `roles/mysql_server/files/`
2. Edit the `roles/mysql_server/defaults/main.yml` file and provide the name of the file to the parameter `geodata_file`
3. Now run the deployment code and the specified geodata would be populated in the mysql server.
Note :- A sample Geolite DB file called `geodata.tar.xz` is included in the  `roles/mysql_server/files/` directory along with this pull request for test purposes.

**Testing Done**
- Tested with single and multi node deployments on openstack
- Tested both scenarios 1 and 2 above.


cestella:
+1 by inspection.  Thanks for the contribution!

metron-deployment/roles/mysql_server/tasks/geoip.yml
```diff
@@ -15,12 +15,21 @@
...
 ---
-- name: Download GeoIP databases
+- name: Download GeoIP databases from specified URL
   unarchive:
-    src:  http://geolite.maxmind.com/download/geoip/database/GeoLiteCity_CSV/GeoLiteCity-latest.tar.xz
+    src: "{{ geodata_file }}"
     dest: /tmp/geoip
     copy: no
     creates: /tmp/geoip/*/GeoLiteCity-Blocks.csv
+  when: '"{{ geodata_file }}".find("http://") != -1'
```
nickwallen: A better way to do this check might be the following.  This approach is a little simpler and doesn't preclude the use of FTP or HTTPS.

```
when: geodata_file is defined
```

metron-deployment/roles/mysql_server/tasks/geoip.yml
```diff
@@ -15,12 +15,21 @@
...
+
+- name: Copy GeoIP database file from files directory
+  unarchive:
+    src: "{{ geodata_file }}"
+    dest: /tmp/geoip
+    copy: yes
+    creates: /tmp/geoip/*/GeoLiteCity-Blocks.csv
+  when: '"{{ geodata_file }}".find("http://") == -1'
```
nickwallen: Same here.

nickwallen:
Neha and I just talked through this.  The other issue that we discovered is that the PR has a canned tarball containing geoip data from Maxmind.  Since this is proprietary data we cannot host it in Metron.

We agreed that we just want to cover two use cases:
1. The user wants to download the latest data from MaxMind.  This is the default behavior.
2. The user wants to use data that is hosted at an alternative URL.  The alternative URL could be HTTP, HTTPS, or FTP.  This might occur when running Metron on an isolated/secure network or a test environment.

To cover these two use cases, we won't need the `metron-deployment/roles/mysql_server/files/geodata.tar.xz` tarball.

nickwallen:
+1 looks good pending travis build

PR #283, opened 2016-09-29
METRON-472: add runAgain.sh to handle a proper vagrant provision

ottobackwards:
When you fail after running run.sh in quickdev you will be told to run vagrant provision... but what you really want to do is run vagrant provision with the same parameters that you used for vagrant up inside run.sh.

runAgain.sh has that.

### DIFF
--- a/metron-deployment/vagrant/quick-dev-platform/runAgain.sh
+++ b/metron-deployment/vagrant/quick-dev-platform/runAgain.sh
@@ -17,9 +17,7 @@
 # limitations under the License.
 #
 
-echo "If run.sh fails, instead of just running vagrant provision, you should run ./runAgan.sh"
-
 vagrant \
   --ansible-tags="hdp-deploy,metron" \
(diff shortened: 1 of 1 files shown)


justinleet:
Could we just make this a flag in run.sh?

Or (possibly) better, is there a way to make vagrant run the provisioner anyway during subsequent up calls?  At that point, it'd be transparent if we can just rerun the script on failure and have it do the right thing without the user having to do anything.

nickwallen:
I think everyone is in agreement that we are going with METRON-474, instead of this one.  Correct me if I am wrong @ottobackwards

nickwallen:
FYI - The discussion is attached to [METRON-472](https://issues.apache.org/jira/browse/METRON-472)

justinleet:
@nickwallen I like it. In that case, I'd say we probably just have this PR get closed?

ottobackwards:
No you are right
I thought I closed the pr, I guess I did not.
Sorry

PR #286, opened 2016-09-30
METRON-326 Error Handling in ElasticsearchWriter

justinleet:
To facilitate partial successes writing in Elasticsearch (i.e. subset of Tuples succeeds and another subset fails), the BulkMessageWriter.write() now returns a BulkWriterResponse object which contains both the successful and failed Tuples (So that we know which ones to ack successfully and which to send to the error stream).

ElasticsearchWriter has been modified to be able to take advantage of partial writes.  The other implementations of the BulkWriter have only been affected to produce total success at the end of their methods (which is the same functionality previously implied by not hitting an Exception.

Unit tests have been added for the ElasticsearchWriter writers implementation of the partial successes and failures, along with modifying existing tests elsewhere to account for the BulkWriterResponse.

One question I'd like resolved in review, is where to actually place the BulkWriterResponse in the code hierarchy.  It's obviously closely tied to the BulkMessageWriter (really only existing to add functionality to it), but that lives in an interface package and BWR is a concrete class. If there are opinions on where that should live, I'm happy to move it and update the PR.


cestella:
I'd vote for one of two options:
* Investigate how if we can move those interfaces away from `metron-common` and into `metron-writer` into the `org.apache.metron.writer` package.
* If you can't do above, I'd move `BulkMessageWriter`, `MessageWriter` and the new response object you've created into `org.apache.metron.common.writer`

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
```diff
@@ -166,14 +166,37 @@ public void write(String sensorType, WriterConfiguration configurations, Iterabl
...
+
+        if (item.isFailed()) {
+          writerResponse.addError(item.getFailure().getCause(), tuple);
+        } else {
+          writerResponse.addSuccess(tuple);
+        }
+
+        // Should never happen, so fail the entire batch if it does.
+        if (respIter.hasNext() != tupleIter.hasNext()) {
+          throw new Exception(bulkResponse.buildFailureMessage());
```
cestella: Perhaps a named exception?  `IllegalStateException`?

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/WriterToBulkWriter.java
```diff
@@ -44,11 +45,15 @@ public void init(Map stormConf, WriterConfiguration config) throws Exception {
...
 
   @Override
-  public void write(String sensorType, WriterConfiguration configurations, Iterable<Tuple> tuples, List<MESSAGE_T> messages) throws Exception {
+  public BulkWriterResponse write(String sensorType, WriterConfiguration configurations, Iterable<Tuple> tuples, List<MESSAGE_T> messages) throws Exception {
     if(messages.size() > 1) {
       throw new IllegalStateException("WriterToBulkWriter expects a batch of exactly 1");
     }
     messageWriter.write(sensorType, configurations, Iterables.getFirst(tuples, null), Iterables.getFirst(messages, null));
+
+    BulkWriterResponse response = new BulkWriterResponse();
```
cestella: What if there's an error writing?  Do we still want to add the tuples to the response as successes?

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/HdfsWriter.java
```diff
@@ -69,14 +66,19 @@ public void init(Map stormConfig, WriterConfiguration configurations) {
...
+  public BulkWriterResponse write(String sourceType
                    , WriterConfiguration configurations
                    , Iterable<Tuple> tuples
                    , List<JSONObject> messages
                    ) throws Exception
   {
     SourceHandler handler = getSourceHandler(sourceType);
     handler.handle(messages);
+
+    // HDFS SourceHanlder manages writing and rotating files.  Will throw its own Exceptions.
```
cestella: typo on `SourceHandler`.

Also, if the `SourceHandler` manages writing and will throw exceptions, do we want to catch the exception, add the tuples as errors to the response and rethrow?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
justinleet: Good catch, I'll make that change.  IllegalStateException seems pretty reasonable.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/WriterToBulkWriter.java
justinleet: Right now, the Exception handling path is slightly different from the main path, and would just treat everything as failures, regardless. The BulkWriterComponent catches exceptions and uses its complete list to mark everything as failed regardless of what the response is.  

The write() method interface does note in the comments that an exception should be treated as a full batch failure, but it could pretty easily be modified to catch the exception, mark failures and kick it back up that way. I left it as-is, because that matches current behavior exactly.  At that point, I'd want to do the same thing for all the implementations (hbase, etc.)

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/HdfsWriter.java
justinleet: Will fix typo.  Same deal as previous comment where it's pretty easy to change and manage that directly.

justinleet:
Moving the interfaces `metron-writer` should be pretty trivial.  The only things that depend on them, unless I missed something, also depend on either `metron-writer` directly or `metron-enrichment` (which itself depends on `metron-writer`.  I'll move stuff, make sure nothing implodes and update accordingly.

justinleet:
Updated with moving the files (did need to make metron-hbase depend on metron-writer), fixed the typo, and added the named exception.

Have not modified the writers to catch the exceptions and pass the errors up that way.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriterTest.java
```diff
@@ -0,0 +1,190 @@
...
+        BulkItemResponse itemResponse = buildBulkItemFailure(e);
+        when(response.iterator()).thenReturn(ImmutableList.of(itemResponse).iterator());
+
+        BulkWriterResponse expected = new BulkWriterResponse();
+        expected.addError(e, tuple1);
+
+        ElasticsearchWriter esWriter = new ElasticsearchWriter();
+        BulkWriterResponse actual = esWriter.buildWriteReponse(ImmutableList.of(tuple1), response);
+
+        assertEquals("Response should have no errors and two successes", expected, actual);
```
mmiklavc: I believe this assert message should read "Response should have one error and zero successes"

justinleet: You are correct, will update

justinleet:
Moving everything to `metron-writer` seems to cause random build failures (e.g. the one above), based on how things get spun up.  It is not deterministic.  I'm going to try using option 2, and try testing that and seeing if it's consistent.

justinleet:
Moving things using option 2 (writer package in common) seems stable, after running locally a few times.  Travis also passed.  Not sure why moving them to different module causes such issues.

cestella:
+1 by inspection

PR #288, opened 2016-10-03
METRON-480: Kibana 4.5+ Requires http.cors.enabled set to false on ES

dlyle65535:
Tested with simulated Docker cluster.

Installed Elasticsearch and Kibana services together post-install and verified Kibana could initialize.

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/configuration/elastic-site.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/configuration/elastic-site.xml
@@ -47,7 +47,15 @@
         <name>path_data</name>
         <value>"/opt/lmm/es_data"</value>
         <description>Path to directory where to store index data allocated for this node. e.g. "/mnt/first", "/mnt/second"</description>
-    </property>    
+    </property>
+    <property>
+        <name>http_cors_enabled</name>
+        <value>"false"</value>
+        <description>Enable or disable cross-origin resource sharing, i.e. whether a browser on another origin can do requests to Elasticsearch. Defaults to false.</description>
+        <value-attributes>
+            <type>string</type>
+        </value-attributes>
+    </property>
     <!--  Discovery -->
     <property>
         <name>transport_tcp_port</name>
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/package/scripts/params.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/package/scripts/params.py
@@ -45,6 +45,7 @@ cluster_name = config['configurations']['elastic-site']['cluster_name']
 zen_discovery_ping_unicast_hosts = config['configurations']['elastic-site']['zen_discovery_ping_unicast_hosts']
 
 path_data = config['configurations']['elastic-site']['path_data']
+http_cors_enabled = config['configurations']['elastic-site']['http_cors_enabled']
 http_port = config['configurations']['elastic-site']['http_port']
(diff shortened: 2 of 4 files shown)


anandsubbu:
I have verified this fix on a 12-node openstack cluster. 

Generated a fresh set of RPMs with the fix provided in this pull request, and setup the cluster afresh. After the installation completes, the Kibana dashboard comes up fine.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/configuration/elastic-site.xml
```diff
@@ -47,7 +47,15 @@
         <name>path_data</name>
         <value>"/opt/lmm/es_data"</value>
         <description>Path to directory where to store index data allocated for this node. e.g. "/mnt/first", "/mnt/second"</description>
-    </property>    
+    </property>
+    <property>
+        <name>http_cors_enabled</name>
+        <value>"false"</value>
```
justinleet: Assuming that the quotes don't get removed in the .j2 files, can we move the quotes into those, then just leave this as a plain string without quotes? I'm not sure if using the quotes is required or not, but if they are then users might not remember to add them.

If we do want to change that, given that path_data does this even in the Github snippet, I'm sure this occurs more frequently and we may want to create a separate task for cleaning all of them up in one fell swoop.

nickwallen:
Was this intended to fix the following error thrown by Kibana when it tries to talk to Elasticsearch?

```
Courier Fetch Error: unhandled courier request error: Authorization Exception
```

nickwallen:
Ah, chased down the JIRA.  Looks like it is.

nickwallen:
To fix this using Ambari on an existing ES cluster would I just go to...
  Elasticsearch -> Configs -> Custom elastic_site -> Add Property -> key=http_cors_enabled, value=false?

dlyle65535:
No, it defaults to false, so no user action is required.

dlyle65535:
Darn. Can't edit on the phone.  Are you seeing this on an existing cluster? A default ES instance should default to false.

nickwallen:
I am seeing this on a cluster that I deployed with the mpack from 'master'.  ES is happy and excited to play.  When I hit up Kibana I get this error.

```
Courier Fetch Error: unhandled courier request error: Authorization Exception
Version: 4.5.1
Build: 9892
```

In researching the error, I ran across this on [Github/Kibana](https://github.com/elastic/kibana/issues/6719) which refers to the `http.cors.enabled` setting.  I remembered you opening this PR and thought it might connect.  

Now I'm confused though because in this PR you set a `http_cors_enabled` variable which uses underlines versus `http.cors.enabled` which is mentioned in the link above.

nickwallen:
If I manually login to each ES master and remove the `http.cors.enabled = true` from `/etc/elasticsearch/elasticsearch.yml` then restart ES, I am able to connect to ES with Kibana.

Of course, once Ambari gets involved it undoes my manual edit.  I have not yet figured out how to 'unset' this through Ambari yet.

nickwallen:
+1 This PR did fix the issue that I was seeing.  Tested on a live cluster.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/configuration/elastic-site.xml
justinleet: +1, I'm fine with it being merged as-is.

dlyle65535: We have to have the quotes, otherwise Ambari will change treat it as a boolean which will be written to False in elasticsearch.yml and Elasticsearch doesn't recognize that as False for this parameter only.
Please don't make them all the same in a clean-up pr, it will break things that work.

PR #289, opened 2016-10-03
METRON-461: Install Metron Data Management tools

dlyle65535:
We omitted metron-data-management in our first iteration of the MPack. This corrects that.

Tested on simulated distributed cluster in Docker.


justinleet:
+1, repeated test and files show up and yum reports the Data Management RPM is installed.

PR #290, opened 2016-10-03
METRON-421 Make Stellar Profiler Client API Accessible in Parser and Enrichment Topologies

nickwallen:
[METRON-421](https://issues.apache.org/jira/browse/METRON-421)

### Changes

These changes were required to make the Stellar Profiler Client API accessible from the Parser and Enrichment topologies.

* Refactored the Profiler into three projects.  
   * This dramatically helps with the clashing dependencies between Storm and HBase when the Profiler Client API is added to Enrichment.
   * `metron-profiler-common` - Contains all code shared between the Profiler topology and client
   * `metron-profiler` - Contains the topology used to build Profiles.
   * `metron-profiler-client` - The client API used to access Profile data.
* Added the `metron-profiler-client` as a dependency to `metron-parser` and `metron-enrichment`.  This allows Profile data to be accessed from both the Parser and Enrichment topologies.
* Added a README for the `metron-profiler-client` to describe its usage.
* Pushing the `metron-profiler-client` archive to the Metron host to make the library accessible to the Stellar Shell.
* Fixed an issue with potentially unsafe String to Byte conversion that may cause issues on different platforms.
* All "uber JARs" now use the naming convention "[project]-[version]-uber.jar".  This included updates to Parsers, Elasticsearch, and Solr JARs.
   * These projects did not have <shadedClassifierName>uber</shadedClassifierName> set. The dependencies coming from those projects would not show-up when running `mvn dependency:tree`. Once I turned that on so that I could actually see the dependencies being pulled in, the non-uber jars would be deployed and so nothing would actually work. These changes ensure that when troubleshooting you can actually see the dependencies and the right jar actually gets deployed.

### Testing

Beyond the unit and integration tests, these changes were tested in the 'Quick Dev' environment.
 * [x] Deployed, stopped, started, and validated the Parser topology using Snort and Bro
 * [x] Deployed, stopped, started, and validated the Enrichment topology
 * [x] Deployed, stopped, started, and validated the Indexing topology
 * [x] Deployed, stopped, started, and validated the Profiler topology
 * [x] Used the Profiler Client API to retrieve Profile data
 * [x] Validated full end-to-end automated deployment with 'Quick Dev'
 * [x] Build and Install Metron with RPMs 
 * [x] Build Metron Ambari MPack

These changes can be manually tested by following the 'Getting Started' instructions included in both the `metron-profiler/README.md` and `metron-profiler-client/README.md`.


metron-analytics/metron-profiler/src/main/assembly/assembly.xml
```diff
@@ -63,7 +63,7 @@
         <fileSet>
             <directory>${project.basedir}/target</directory>
             <includes>
-                <include>${project.artifactId}-${project.version}.jar</include>
+                <include>${project.artifactId}-${project.version}-uber.jar</include>
             </includes>
```
dlyle65535: Is it important to change the jar names? If so, could you check for ripple changes to the RPM build scripts and the Ambari scripts?

nickwallen: I'd be glad to check.  If you can point me to a README or specifics on what to check that would be appreciated.

nickwallen: Think I've got it.  Added some additional tests to the list in the PR description.  Will have another commit coming that updates the RPM spec file.

nickwallen:
I can build and install Metron with the RPMs.  I can build the Mpack.  But I am not able to easily test installation of the Mpack.  (It might make life easier to have Vagrant/Ansible spin-up with Ambari 2.4 so I could easily test this.) 

I did a diff of /usr/metron as installed via Vagrant/Ansible to what is installed with the RPMs.  The only notable gap that I would not expect was that the RPMs did not install `config/zookeeper/global.json`.

The other differences that are likely expected (need confirmation) is that the following files were installed via Vagrant/Ansible, but not by the RPM.
 * `lib/metron-api-0.2.1BETA.jar`
 * `lib/metron-maas-service-0.2.1BETA-uber.jar`
 * `lib/metron-profiler-0.2.1BETA-uber.jar` and `lib/metron-profiler-client-0.2.1BETA.jar` as the Profiler is not installed via the RPMs currently.

justinleet:
@nickwallen Looking over the spec change, since it's just a slight rename of things in the /lib folder (not adding a bunch of new files, or anything.), making sure the RPMs build, deploy, and the changed jars have correct perms should be good given the other testing of things.

nickwallen:
Would love a review on this.  It's getting aged.

cestella:
+1 on inspection

PR #292, opened 2016-10-05
METRON-171 add .class to .gitignore

ottobackwards:
Add .class files to .gitignore


james-sirota:
+1 by inspection. thanks for catching this

PR #293, opened 2016-10-05
METRON-473 Add LENGTH() To Stellar

ottobackwards:
Add LENGTH to Stellar string functions and unit tests.


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -222,4 +222,16 @@ public Object apply(List<Object> args) {
       return null;
     }
   }
+
+  @Stellar( name="LENGTH"
+          , description = "Returns the length of a string"
+          , params = { "input - String" }
+          , returns = "String"
```
nickwallen: Probably need to change`returns = "String"` to indicate that it returns an Integer which is the length of the String.  However you want to say that is fine, but it does not actually return a String.

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
```diff
@@ -285,6 +285,18 @@ public void testHappyPath() {
   }
 
   @Test
```
nickwallen: I would add one more test that validates the behavior when you call LENGTH on an undefined variable; `LENGTH(foo)` when there is no variable `foo`.

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
```diff
@@ -285,6 +285,18 @@ public void testHappyPath() {
...
+  public void testLength(){
+    String query = "LENGTH(foo)";
+    Assert.assertEquals(5, run(query,ImmutableMap.of("foo","abcde")));
+  }
+
+  @Test
```
nickwallen: If I mistakenly call `LENGTH()` with no argument, then I get an `IndexOutOfBoundsException`.  And this is not unique to the `LENGTH` function.  There are a number of other Stellar functions that do the same.

This may be confusing to a user of Stellar who is not familiar with the underlying implementation.  "IndexOutOfBounds? What the heck does that mean?"  

I wonder if it makes sense to handle this in a standard way within Stellar.  Ideally we would get something along the lines of "LENGTH: Missing first argument 'input'.  Expected 'String'".  

I bring this up only as a discussion point for future work.  This PR does not necessarily have to address this situation.

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
ottobackwards: So, I was going to make a test passing in NULL, but the Immutable map will not take nulls, can you clarify how you would test this?

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
ottobackwards: I will fix - see my comment above about how to test

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
nickwallen: Yes, empty map.  If a variable does not exist, Stellar will pass in a null as the variable.

cestella:
This is cool, but what do you think about making `LENGTH` handle different types other than just string?  For instance, if you pass a `Collection`, it'll return the `size` and if a string, the length?

This is currently how `IS_EMPTY` works

ottobackwards:
@cestella I can do that.  Should I move LENGTH to DataStructureFunctions then?
Also - I notice that IS_EMPTY throws IllegalStateExceptions instead of returning null, is that how no-var should be handled?  In the String functions they return null

cestella:
@ottobackwards I'd vote for it to be in DataStructureFunctions and it's an interesting question about that.  Should the length of `null` be an exception, null or `0`?

I, personally, vote for `0`, but I might be persuaded otherwise.

ottobackwards:
OK I have it in.  Here is the thing that I found,  I know there is a pull about fixing IS_EMPTY from throwing exceptions - and I had to take that same return 0 approach instead of throwing.  IS_EMPTY did NOT have a test in StellarTests, so it was never being tested through the stellar variable resolver, validation, parser etc.  If it had had a test there ( using the run() as implemented in the test ) it never would have been committed, because it NEVER works, no matter what you pass it as originally implemented with the IllegalStateExceptions.  This is because the validate calls down to the apply with a specifically null argument which triggers the exception.
So, a couple of things:

- every stellar function needs to be tested through the whole stellar framework and not just on top of StellarFunction.apply()
- why would validate call parse which calls apply?  Is that the way it is at runtime?  Does apply always get called twice?  Is the test harness wrong?

cestella:
So, yes, a couple of things need to be ensured for stellar functions:
* The output of the call is serializable via kryo (the profiler stores the output of stellar statements in HBase)
* It passes the `validate` call

Validate is intended to check syntactic errors and it does so by passing a VariableResolver that returns null for every variable.  I will admit that I've been thinking this should be rethought and we should do this another way:
* Have a special implementation of StellarBaseListener that barfs if function resolution happens, but doesn't actually apply functions
* Uses the normal Stellar Lexer to ensure syntactic errors
This should ensure that the stellar statement is valid syntactically IMO.

Thoughts?

mmiklavc:
@ottobackwards A unit test has been added as part of PR#296 -> https://github.com/mmiklavc/incubator-metron/blob/bd6e1a5cc48962af36b131e189ffb461e836cd20/metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/DataStructureFunctionsTest.java

ottobackwards:
I took this jira to get some start of an idea about Stellar, and after debugging through it to track down to exitVariable to find the argument resolution and then back up to the validate x->null I would say I got what I bargained for.

I am not sure how I would answer your question on validation, I don't know Stellar well enough.  My experience and intuition tells me that running the 'execute' and using an exception or error case as validation is not very efficient at runtime, if in fact this is what happens at runtime as well.  Better to have real metadata on the function and just validating the metadata and they syntax of the query I would think.   *A logical/runtime error does not mean an invalid expression*.  This method is in effect equating them which I think is incorrect ( unless I am mistaking the intent of validate()).

ottobackwards:
@mmiklavc I had seen that.  The tests in StellarTests test the compiler and the variable resolution etc as well is my point.  It is worth having both.  If both were done from the start, maybe  METRON-439 doesn't get through

cestella:
Sorry was responding on the dev list, not this PR.  The purpose of validate is to ensure syntactic correctness from a stellar point of view.  This is not called per message, but only when stellar statements change (e.g. At config push to zookeeper time)

ottobackwards:
ok, as above, I'm not sure about x->null to do that, but I am the noob here

cestella:
Yeah agreed.  I don't think we should be executing the function as part of validate.  Perhaps just ensuring the function resolves.  Maybe expanding the interface for stellar function to provide a validate method so it's pluggable by function.

ottobackwards:
Well....  If you have the annotation, the method object and the expression - you *could* do it all from that.  You may need to have the annotation have more formal input definitions though

cestella:
The annotation would have to be made more complex.  We can accept varargs as part of the function, for instance.  I think it's possible, but perhaps we would need to iterate a bit on the annotation.

ottobackwards:
It would make a nice little POC -

cestella:
+1 by inspection.  Great work, otto.

PR #294, opened 2016-10-06
METRON-487: Correct the license in the StixExtractorTest

cestella:
The StixExtractorTest has an example taken from the Stix project which blanket licenses its examples as BSD. We need to pull it out of the test and note it in the LICENSE file


cestella:
Note, this also adds a license to the unit test REST service

james-sirota:
+1 by inspection. thanks for pulling it out

PR #295, opened 2016-10-06
METRON-371: Changing logging level to INFO when there's not a config.

cestella:


cestella:
Testing Steps
1. Ensure squid topology is up.
2.  Inject the following message to the kafka-producer to ingest
```
1461576382.642    161 127.0.0.1 TCP_MISS/200 103701 GET http://www.abc.com/ - DIRECT/199.27.79.73 text/html
```
3. Wait for the enrichment and index to be generated. 
4. Review the enrichment kafkaspout log file and ensure that the error cannot be seen.

james-sirota:
+1 by inspection

PR #296, opened 2016-10-06
METRON-439: Stellar : IS_EMPTY(host) throws exception

mmiklavc:
Fix Stellar IS_EMPTY validation to handle empty and null

Addresses https://issues.apache.org/jira/browse/METRON-439

Things have changed a bit since the original Jira was filed. Most notably, this error appears while dumping config from Zookeeper after it has already been loaded. The current functionality/validation would not have allowed the load to succeed in the first place. Even so, empty and null checks could be handled more gracefully. This PR changes IS_EMPTY to return true on null or empty string rather than throw an exception.

Added new unit tests and validated on quick-dev with multiple bro messages - normal host, empty host string, and null/non-existent host string.

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/DataStructureFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/DataStructureFunctions.java
@@ -146,8 +146,8 @@ public class DataStructureFunctions {
 
     @Override
     public Object apply(List<Object> list) {
-      if(list.size() == 0) {
-        throw new IllegalStateException("IS_EMPTY expects one string arg");
+      if(null == list || list.size() == 0) {
+        return true;
       }
       Object o = list.get(0);
       if(o instanceof Collection) {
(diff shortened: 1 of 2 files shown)


ottobackwards:
I'm working on making LENGTH work like IS_EMPTY.  I can't get either one to work as both are failing validation.  There is no StellarTest for IS_EMPTY ( I have added one to check if it is just my code that is failing ).  Can you one and try it there an make sure it works all the way through the Stellar parse and variable resolve?
ParseException: Unable to parse IS_EMPTY(foo): Unable to execute: IS_EMPTY expects a collection or a string

This is when calling with
`String query = "IS_EMPTY(foo);
 Asser.assertEquals(true,run(query,ImmutableMap.of("foo",""));
`

ottobackwards:
OK, if I change my functions to not throw then it works.  This doesn't seem right to me, but it may be something to do with why the tests are run that way.

mmiklavc:
@ottobackwards The Stellar validation currently runs by passing an empty member to the apply function. The original reason for the validation was to help keep bad configs from ending up in zookeeper. Now with the Stellar REPL, users should be able to more easily do their own client-side validations.

cestella:
+1 by inspection

PR #297, opened 2016-10-06
METRON-488: Snort should use a proper CSV implementation

cestella:
Right now if you have a custom snort rule (e.g. alert tcp any any -> any any (msg:'snort alert message having a ,(comma) to check csv parsing'; sid:999158; ) ) the snort parser will fail to parse because it's splitting on the comma naively.
It should use the existing CSV parsing infrastructure that we have and that is used in the CSVParser.


james-sirota:
Does the same apply to other parsers as well? or just snort?

cestella:
@james-sirota it depends on the parser.  It applies to the CSVParser, for sure, but this isn't a generic fix (not sure how you'd go about making a generic fix here, in fact).

mmiklavc:
+1 by inspection

PR #298, opened 2016-10-06
METRON-432: Fix pcap field resolver to return object instead of string value

mmiklavc:
Addresses https://issues.apache.org/jira/browse/METRON-432

Passes PcapTopologyIntegrationTest. Running manual tests now in quick-dev, but wanted to get this up for review asap.


james-sirota:
this is a fairly minor change, but makes sense.  +1 builds, passes tests

cestella:
+1 by inspection

mmiklavc:
FYI, verified this properly queries and filters based on numeric types in quick-dev

PR #299, opened 2016-10-06
METRON-425 Stellar transformation fails to handle special characters

ottobackwards:
The issue is that per our grammar the arithmetic and comparison operators need to be escaped to be used as literals.

I have added a unit test to the system for this case ( for field transformations ).
I have also taken a stab at updating the readme.md.  I do not thing the readme is correct though.  It may need to be written out more better by someone who writes good.

### DIFF
--- a/metron-platform/metron-common/README.md
+++ b/metron-platform/metron-common/README.md
@@ -16,6 +16,8 @@ The query language supports the following:
 * The ability to have parenthesis to make order of operations explicit
 * User defined functions
 
+####Note: string values containing arithmetic operations and comparison operators used as literals such as: "foo : <ok>" require escaping such as "foo: \"<ok>\""
+
 The following functions are supported:
 
 * `BLOOM_ADD`
--- a/metron-platform/metron-common/README.md
+++ b/metron-platform/metron-common/README.md
@@ -18,7 +18,8 @@ The query language supports the following:
 
 ##Stellar Language Keywords
 The following keywords need to be single quote escaped in order to be used in Stellar expressions:
-|               |               |             |
+
+|              |              |           |
 |   :---:       |     :---:     |   :---:     | 
 | not| else | exists | 
 | if | then | and |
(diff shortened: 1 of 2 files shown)


justinleet:
Good catch.  I've been looking at this, and ended up looking into red herring.  A further note on top of this is that (unsurprisingly in light of the cause) other reserved keywords tend to cause problems.  Somewhat inconsistently, too. E.g. 
`"newStellarField" : "not"` fails, but
`"newStellarField" : "false"` does not.

I'd have to dig into it more, but I assume it's because not and the other comparison operators in the grammar expect to be followed by more things, and don't end up being string here.

I'm open to suggestions on what the appropriate behavior is, but I really don't like that it's inconsistent on what gets rejected.

ottobackwards:
I think that we need to expand the readme, properly.  Or add a separate language guide?  The not, false thing is troubling.  I'd have to go through the grammar more closely.

ottobackwards:
So - based on 
`identifier_operand : (TRUE | FALSE) # LogicalConst
                   | arithmetic_expr #ArithmeticOperands
                   | STRING_LITERAL # StringLiteral
                   | list_entity #List
                   | map_entity #MapConst
                   | NULL #NullConst
                   | EXISTS LPAREN IDENTIFIER RPAREN #ExistsFunc
                   | LPAREN conditional_expr RPAREN#condExpr_paren
                   ;`

Anything that can be completely expressed from this list will work.  TRUE and FALSE can, but NOT um, can not.

It is not random.

cestella:
It's not inconsistent; it's just a programming language ;)  `"newStellarField" : "not"` is invalid because `not` is a reserved word.  If you want the string literal, it'd be `"newStellarField" : "'not'"` which should resolve fine.  I'd agree that a section about reserved words in the metron-common readme section for stellar could be useful.

cestella:
Oh, I also wanted to point out that stellar accepts both `"` and `'` as string bookends.  So, you didn't have to escape your double quotes.  You could have just used single quotes.

justinleet:
To validate Casey's point, I just ran a quickly modified version of your unit test with things single quotes, and things seem to line up quite nicely at that point on special chars and not.  I think we just update the README with the list of keywords and set a unit test up with the single quotes.  At that point, I'm good.

justinleet:
+1, I'm satisfied updating the docs and the unit test addresses the original concern seen.

PR #300, opened 2016-10-07
METRON-489: RemoveSubdomains Stellar Function behaves incorrectly for some domains

cestella:
com.com throws an exception
www.subdomain.com.com returns subdomain.com
Unsure if other standard weirdness with TLDs get handled like this (e.g. net.net, co.uk.co.uk)

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
@@ -78,13 +78,16 @@ public class NetworkFunctions {
 
     @Override
     public Object apply(List<Object> objects) {
+      if(objects.isEmpty()) {
+        return null;
+      }
       Object dnObj = objects.get(0);
       InternetDomainName idn = toDomainName(dnObj);
       if(idn != null) {
         String dn = dnObj.toString();
         String tld = idn.publicSuffix().toString();
-        String suffix = Iterables.getFirst(Splitter.on(tld).split(dn), null);
-        if(suffix != null)
+        String suffix = dn.substring(0, dn.length() - tld.length());
+        if(suffix != null )
         {
           String hostnameWithoutTLD = suffix.substring(0, suffix.length() - 1);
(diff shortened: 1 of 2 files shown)


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
```diff
@@ -78,13 +78,16 @@ public Object apply(List<Object> list) {
...
+      }
       Object dnObj = objects.get(0);
       InternetDomainName idn = toDomainName(dnObj);
       if(idn != null) {
         String dn = dnObj.toString();
         String tld = idn.publicSuffix().toString();
-        String suffix = Iterables.getFirst(Splitter.on(tld).split(dn), null);
-        if(suffix != null)
+        String suffix = dn.substring(0, dn.length() - tld.length());
+        if(suffix != null )
```
justinleet: Could you just drop this if-else?  Substring never returns null, so it can be dropped entirely.

james-sirota: www.subdomain.com.com is not a valid TLD.  for a list of valid domains see here: https://publicsuffix.org/list/effective_tld_names.dat

If the system sees this kind of domain this should immediately be flagged as alert and triaged with a very high score

justinleet: @james-sirota Can you expand on that?  Here's my understanding, and please correct me if I'm wrong.

www.subdomain.com.com is a valid address.  When we apply the function, we should lose the 'www.subdomain' and leave the domain + '.' + TLD.  In this case the domain is 'com' and the TLD is (coincidentally) 'com', leaving us with 'com.com'.  The unit tests shows this behavior at
` Assert.assertEquals("com.com", removeSubdomains.apply(ImmutableList.of("www.subdomain.com.com")));`

justinleet:
I'm +1 on this. Although I agree that this type of domain should trigger an alert at another point, the stellar function itself should still handle the input correctly and gracefully.

cestella:
well, this is at least as graceful as a baby giraffe, so done and done. ;)

PR #301, opened 2016-10-07
METRON-490 Stellar Validation of Required Parameters

nickwallen:
### [METRON-490](https://issues.apache.org/jira/browse/METRON-490)

Currently, each Stellar function handles validation of required function parameters in its own way.  In some cases, we have functions that throw an IndexOutOfBoundsException, in other cases the function will produce its own error message and exception.

There needs to be a standard mechanism to validate required function parameters.  This mechanism should handle missing parameters gracefully and provide an error message that makes sense to a user.  The handling should be accessible across all functions.
### Changes
- Added an additional Stellar annotation called `requiredParams`.  
- Added a method to `BaseStellarFunction.validate` to validate the `requiredParams`.
- Use of the annotation field and validation mechanism is optional.
- Modified the Stellar STATS_\* functions to use this functionality.
- If an incorrect type is specified, an IllegalArgumentException is thrown with a message like 
  `
  SOME_FUNCTION: unexpected parameter [2]: expected class java.lang.Integer, actual class java.lang.String`
  `
- If a required parameter is missing, an IllegalArgumentException is thrown with a message like:
  `
  SOME_FUNCTION: parameter missing [2]: expected class java.lang.Integer
  `

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/BaseStellarFunction.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/BaseStellarFunction.java
@@ -17,11 +17,67 @@
  */
 package org.apache.metron.common.dsl;
 
+import org.apache.commons.lang.StringUtils;
+
 import java.util.List;
-import java.util.Map;
+
+import static java.lang.String.format;
 
 public abstract class BaseStellarFunction implements StellarFunction {
+
   public abstract Object apply(List<Object> args);
+
+  /**
+   * Validates that the parameters received by a function match the parameters
+   * that have been declared as required.
+   *
+   * @param args The actual parameters.
+   */
+  public void validate(List<Object> args) {
+
+    Stellar annotation = this.getClass().getDeclaredAnnotation(Stellar.class);
+    if(annotation != null) {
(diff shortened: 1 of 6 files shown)


ottobackwards:
Nick - I have a pr on your branch with an attempt to fix some of these, although it may be a couple of days out of date

nickwallen:
Oh, I did not see that, thanks. I will resolve merge conflicts.

nickwallen:
I totally forgot about this.  Looking over this again, I am a little ambivalent about this PR.  

I did as an experiment. I thought it would be cool, if parameter handling could be taken care of auto-magically.  I wanted to just specify the parameters that I need via the annotation and have some shared logic perform parameter handling and validation, along with enhancing the docs.  Seems more DRY-ish and all that.

As this stands, I don't think it is very useful.  Different functions want to handle required params in different ways.  Sometimes an exception is thrown, sometime a value is returned like NaN, etc.  This doesn't give you any flexibility in how to respond to missing params.   There are probably a few other things that need to be covered before this is close to being useful.

I would be interested in opinions on this though.  Maybe with input from others this could become useful.  But right now, I am a -1 on this.

PR #302, opened 2016-10-08
METRON-492 Run metron_common build check as local_action

kylerichardson:
When testing various deployment options, noticed the logic for the metron_common role didn't work as expected. Updated to check for metron jar locally (where the playbook was initiated) and fail when it doesn't exist.

Tested successfully in single node vm.


dlyle65535:
+1, this is great. It'll be good to have this working.

PR #303, opened 2016-10-11
METRON-424 ability to validate ip addresses against both IPV4 and IPV6

ottobackwards:
Allow the specification of multiple validation types for IP
` 
   {
     "fieldValidations" : [
     {
       "input" : [ "field1", "field2" ]
       ,"validation" : "IP"
       ,"config" : {
       "type" : ["IPV4","IPV6"]
     }
   }
   ]
   }`

The question with this PR is if this is the correct approach.  Is this the most consistent approach?  Is it the least surprising?  Would it be better to add a ALL keyword such that instead of specifying both types you can just do

"type": "ALL"?

I did it this way to start, but I'm open to change it based on review.

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/configuration/elastic-site.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/configuration/elastic-site.xml
@@ -47,7 +47,15 @@
         <name>path_data</name>
         <value>"/opt/lmm/es_data"</value>
         <description>Path to directory where to store index data allocated for this node. e.g. "/mnt/first", "/mnt/second"</description>
-    </property>    
+    </property>
+    <property>
+        <name>http_cors_enabled</name>
+        <value>"false"</value>
+        <description>Enable or disable cross-origin resource sharing, i.e. whether a browser on another origin can do requests to Elasticsearch. Defaults to false.</description>
+        <value-attributes>
+            <type>string</type>
+        </value-attributes>
+    </property>
     <!--  Discovery -->
     <property>
         <name>transport_tcp_port</name>
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/package/scripts/params.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/package/scripts/params.py
@@ -45,6 +45,7 @@ cluster_name = config['configurations']['elastic-site']['cluster_name']
 zen_discovery_ping_unicast_hosts = config['configurations']['elastic-site']['zen_discovery_ping_unicast_hosts']
 
 path_data = config['configurations']['elastic-site']['path_data']
+http_cors_enabled = config['configurations']['elastic-site']['http_cors_enabled']
 http_port = config['configurations']['elastic-site']['http_port']
 transport_tcp_port = config['configurations']['elastic-site']['transport_tcp_port']
 
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/package/templates/elasticsearch.master.yaml.j2
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/package/templates/elasticsearch.master.yaml.j2
@@ -39,7 +39,7 @@ node:
 path:
   data: {{path_data}}
 
-http.cors.enabled: true
+http.cors.enabled: {{http_cors_enabled}}
 
 port: {{http_port}}
 
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/package/templates/elasticsearch.slave.yaml.j2
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/ELASTICSEARCH/2.3.3/package/templates/elasticsearch.slave.yaml.j2
@@ -39,7 +39,7 @@ node:
 path:
   data: {{path_data}}
 
-http.cors.enabled: true
+http.cors.enabled: {{http_cors_enabled}}
 
 port: {{http_port}}
 
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.2.1BETA/metainfo.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.2.1BETA/metainfo.xml
@@ -198,6 +198,9 @@
               <name>metron-common</name>
             </package>
             <package>
+              <name>metron-data-management</name>
+            </package>
+            <package>
               <name>metron-parsers</name>
             </package>
             <package>
@@ -239,7 +242,6 @@
           </packages>
         </osSpecific>
       </osSpecifics>
-
       <commandScript>
         <script>scripts/service_check.py</script>
         <scriptType>PYTHON</scriptType>
--- a/metron-platform/metron-common/README.md
+++ b/metron-platform/metron-common/README.md
@@ -595,7 +595,7 @@ structured like so:
 * `config` : A String to Object map for validation configuration.  This is optional if the validation function requires no configuration.
 * `validation` : The validation function to be used.  This is one of
    * `STELLAR` : Execute a Stellar Language statement.  Expects the query string in the `condition` field of the config.
-   * `IP` : Validates that the input fields are an IP address.  By default, if no configuration is set, it assumes `IPV4`, but you can specify the type by passing in the config by passing in `type` with either `IPV6` or `IPV4`.
+   * `IP` : Validates that the input fields are an IP address.  By default, if no configuration is set, it assumes `IPV4`, but you can specify the type by passing in the config by passing in `type` with either `IPV6` or `IPV4` or by passing in a list [`IPV4`,`IPV6`] in which case the input(s) will be validated against both.
    * `DOMAIN` : Validates that the fields are all domains.
    * `EMAIL` : Validates that the fields are all email addresses
    * `URL` : Validates that the fields are all URLs
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/validation/network/IPValidation.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/validation/network/IPValidation.java
@@ -117,6 +117,7 @@ public class IPValidation implements FieldValidation, Predicate<List<Object>> {
     }
     return type.isValid(ip.toString());
   }
+
   @Override
   public boolean isValid( Map<String, Object> input
                         , Map<String, Object> validationConfig
@@ -128,7 +129,7 @@ public class IPValidation implements FieldValidation, Predicate<List<Object>> {
     for(Object typeObject : types) {
       IPType type = IPType.get(typeObject.toString());
       for (Object o : input.values()) {
-        if(o != null && type.isValid(o.toString())) {
+        if(o == null || type.isValid(o.toString())) {
           return true;
         }
       }
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/validation/network/IPValidation.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/validation/network/IPValidation.java
@@ -24,8 +24,7 @@ import org.apache.metron.common.dsl.Predicate2StellarFunction;
 import org.apache.metron.common.dsl.Stellar;
 import org.apache.metron.common.field.validation.FieldValidation;
 
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.function.Predicate;
 
 public class IPValidation implements FieldValidation, Predicate<List<Object>> {
@@ -34,7 +33,7 @@ public class IPValidation implements FieldValidation, Predicate<List<Object>> {
           , description = "Determine if an string is an IP or not."
           , params = {
               "ip - An object which we wish to test is an ip"
-             ,"type (optional) - one of IPV4 or IPV6.  The default is IPV4."
+             ,"type (optional) - Object of string or collection type (e.g. list) one of IPV4 or IPV6 or both.  The default is IPV4."
                      }
           , returns = "True if the string is an IP and false otherwise.")
   public static class IS_IP extends Predicate2StellarFunction {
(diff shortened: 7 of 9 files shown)


JonZeolla:
I would personally vote for `"type" : ["IPV4","IPV6"]`.  I usually push for explicit and obvious so something like `retrieve_json.py | grep IPV4` would work.

cestella:
I tend to agree, I prefer `["IPV4", "IPV6"] where we just do an `or` between the two.

nickwallen:
Jon described it well.  I'd prefer "explicit and obvious."

ottobackwards:
OK, good then.  I don't have to change anything.
I have no idea why the travis build is failing - it doesn't look todo with my changes ( btw there are always so many errors it is hard to look at the log ).  How does on kick off a new build in the event of a transient error?  Can anyone help with what is going on in the log?

cestella:
So, the problem with the tests is this:
```
test(org.apache.metron.parsers.integration.YafIntegrationTest)  Time elapsed: 58.442 sec  <<< ERROR!
java.lang.RuntimeException: Too many retries: 11
	at org.apache.metron.integration.ComponentRunner.process(ComponentRunner.java:140)
	at org.apache.metron.parsers.integration.ParserIntegrationTest.test(ParserIntegrationTest.java:72)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at org.junit.internal.runners.TestMethod.invoke(TestMethod.java:59)
	at org.junit.internal.runners.MethodRoadie.runTestMethod(MethodRoadie.java:98)
	at org.junit.internal.runners.MethodRoadie$2.run(MethodRoadie.java:79)
	at org.junit.internal.runners.MethodRoadie.runBeforesThenTestThenAfters(MethodRoadie.java:87)
	at org.junit.internal.runners.MethodRoadie.runTest(MethodRoadie.java:77)
	at org.junit.internal.runners.MethodRoadie.run(MethodRoadie.java:42)
	at org.junit.internal.runners.JUnit4ClassRunner.invokeTestMethod(JUnit4ClassRunner.java:88)
	at org.junit.internal.runners.JUnit4ClassRunner.runMethods(JUnit4ClassRunner.java:51)
	at org.junit.internal.runners.JUnit4ClassRunner$1.run(JUnit4ClassRunner.java:44)
	at org.junit.internal.runners.ClassRoadie.runUnprotected(ClassRoadie.java:27)
	at org.junit.internal.runners.ClassRoadie.runProtected(ClassRoadie.java:37)
	at org.junit.internal.runners.JUnit4ClassRunner.run(JUnit4ClassRunner.java:42)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:283)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:173)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:153)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:128)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:203)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:155)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:103)
```

This test failed because of a timeout (probably because the box running it is overloaded.  The stream of failures after that point is due to the Parser Integration tests not shutting down the infrastructure it spins up (e.g. storm, etc.) when the tests fail.  I'd close and immediately reopen this PR and it should trigger a test rerun.  It is probably worth while making a JIRA to correct the integration tests to be smarter and not spew horrible things.

justinleet:
I pulled the PR down locally, and ran into the same integration test issue.  I also ran on master locally and it succeeded.  I'm not sure what the root issue is if that's (consistently) the case.

ottobackwards:
Justin, how do I run the integration tests?

justinleet:
I just outright did a mvn clean install for the entire project and let it run while I was doing other stuff.  but integration-test target I believe also hits them.

nickwallen:
To run the integration tests... Run `mvn integration-test` from Metron's root directory.

I do not believe that the Integration tests are run with an `install` or `package`, those just run the unit tests.

On a side note, you can see what Travis is doing by looking at the `.travisyml` file that is in Metron's root directory.

nickwallen:
FWIW - I am also getting the integration test failures for this PR when I run them locally.  Not sure what the problem is though.  I think an exception is being swallowed somewhere.

ottobackwards:
Ok - I'm running the integration-test now.  Is there a way to debug it?

nickwallen:
I would run one of the tests and put a break point at `ParserIntegrationTest.test()` and start walking through it from there.

ottobackwards:
I'm running as in https://maven.apache.org/surefire/maven-surefire-plugin/examples/debugging.html and attaching with intellij in test, as you describe actually.  It was how to kick off the test in debug that got me

nickwallen:
If that works for you, then great.  

I usually use Maven when I need to run all the tests before a merge or PR.  When I want to dig into one test and debug something, I just open up Intellij.  That makes setting the breakpoint and triggering a single test run very easy.  

You just have to watch out because sometimes (rarely) something will work in Intellij, but not in Maven or vice versa.  Each builds the code slightly differently.

cestella:
I figured out what was going on with this PR and it's kinda interesting, so forgive the long-winded explanation.

@ottobackwards is indeed right, it is not apparent why his PR about a FieldTransformation and Stellar Function would affect the `YafIntegrationTest`.  Unfortunately, it became apparent when it consistently failed on this PR in both travis and locally that something was happening.

It turns out the problem is [here](https://github.com/apache/incubator-metron/blob/master/metron-platform/metron-integration-test/src/main/config/zookeeper/global.json#L14).  The keen observer will notice in the global config we are ensuring that the `src_ip_addr` and `dst_ip_addr` are valid IP addresses for the integration tests.  This, of course, is a typo (that I made :P ).  The correct field names are `ip_src_addr` and `ip_dst_addr`.  
Now, you may ask:
* How has this ever worked?
* Given that it did work before, why is it broken now?

Well, for the first question, it was never correct in the integration tests, but since we did not exercise the situation where we send in invalid IP addresses.  Furthermore, the default behavior for fields which do not exist is to return `true` (probably wrong).

For the second question, @ottobackwards has changed the default behavior for fields which do not resolve is to return `false` (almost assuredly right).

For clarity to show how I go about debugging these problems, I'll be rather explicit in how I figured out what was happening:
* In my git branch, I added @ottobackwards 's repo as a remote repo and checked out the branch:
  * `git remote add otto https://github.com/ottobackwards/incubator-metron.git`
  * `git checkout otto/METRON-424`
  * `git branch METRON-424-debug`
  * `git checkout METRON-424-debug`
* In Intellij, I ran the YafIntegrationTest to verify that it continued to fail
* I set breakpoints at the exit points of the `ParserBolt`:
  * [Line 135](https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java#L135)
  * [Line 147](https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java#L147)
  * [Line 131](https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java#L131)
* It became apparent that we were exiting because the message was not valid.  I then set breakpoints inside of @ottobackwards 's code and validated that it was doing the right thing, but the fields were coming in as `null`, which means the fields were not in the message.
* I tracked the `global.json` down and looked at the field names and there the bug became apparent.

I submitted a PR to your PR branch, @ottobackwards, so if you merge that in, the build *should* work (fingers crossed ;)

nickwallen:
Interesting. Is there anything we can do to make integration test failures like this more transparent in the future?

cestella:
@nickwallen Funny you mention that; yes, I think we can.  We should modify `ParserIntegrationTest` [here](https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/integration/ParserIntegrationTest.java#L77) to check the invalid and error queues and dump their contents and fail immediately.

ottobackwards:
Ok - from what I can see things are written to the kafka topic, but the bolt is never called.  I don't see an crashes in the field validation code at this time.  Can anyone make sense of these exceptions I get ( I am only running the BRO Integration test )


[brointegrationexception.txt](https://github.com/apache/incubator-metron/files/527323/brointegrationexception.txt)

ottobackwards:
That is awesome.  Thank you.
I have been looking at the bro test and getting the failures, but I haven’t
gotten hits inside the IP Validator, so i’m a few steps behind.
I will take care of your pull request right way.

The good news, is in true ‘learning by carnage’ fashion, I know know a bit
about the integration tests ;)

On October 13, 2016 at 09:43:23, Casey Stella (notifications@github.com)
wrote:

I figured out what was going on with this PR and it's kinda interesting, so
forgive the long-winded explanation.

@ottobackwards <https://github.com/ottobackwards> is indeed right, it is
not apparent why his PR about a FieldTransformation and Stellar Function
would affect the YafIntegrationTest. Unfortunately, it became apparent when
it consistently failed on this PR in both travis and locally that something
was happening.

It turns out the problem is here
<https://github.com/apache/incubator-metron/blob/master/metron-platform/metron-integration-test/src/main/config/zookeeper/global.json#L14>.
The keen observer will notice in the global config we are ensuring that the
src_ip_addr and dst_ip_addr are valid IP addresses for the integration
tests. This, of course, is a typo (that I made :P ). The correct field
names are ip_src_addr and ip_dst_addr.

Now, you may ask:

   - How has this ever worked?
   - Given that it did work before, why is it broken now?

Well, for the first question, it was never correct in the integration
tests, but since we did not exercise the situation where we send in invalid
IP addresses. Furthermore, the default behavior for fields which do not
exist is to return true (probably wrong).

For the second question, @ottobackwards <https://github.com/ottobackwards>
has changed the default behavior for fields which do not resolve is to
return false (almost assuredly right).

For clarity to show how I go about debugging these problems, I'll be rather
explicit in how I figured out what was happening:

   - In my git branch, I added @ottobackwards
   <https://github.com/ottobackwards> 's repo as a remote repo and checked
   out the branch:
      - git remote add otto
      https://github.com/ottobackwards/incubator-metron.git
      - git checkout otto/METRON-424
      - git branch METRON-424-debug
      - git checkout METRON-424-debug
   - In Intellij, I ran the YafIntegrationTest to verify that it continued
   to fail
   - I set breakpoints at the exit points of the ParserBolt:
      - Line 135
      <https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java#L135>
      - Line 147
      <https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java#L147>
      - Line 131
      <https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java#L131>
   - It became apparent that we were exiting because the message was not
   valid. I then set breakpoints inside of @ottobackwards
   <https://github.com/ottobackwards> 's code and validated that it was
   doing the right thing, but the fields were coming in as null, which
   means the fields were not in the message.
   - I tracked the global.json down and looked at the field names and there
   the bug became apparent.

I submitted a PR to your PR branch, @ottobackwards
<https://github.com/ottobackwards>, so if you merge that in, the build
*should* work (fingers crossed ;)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/apache/incubator-metron/pull/303#issuecomment-253516607>,
or mute the thread

ottobackwards:
Are you going to open a jira for that?

-- 

On October 13, 2016 at 10:16:13, Casey Stella (notifications@github.com)
wrote:

@nickwallen <https://github.com/nickwallen> Funny you mention that; yes, I
think we can. We should modify ParserIntegrationTest here
<https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/integration/ParserIntegrationTest.java#L77>
to check the invalid and error queues and dump their contents and fail
immediately.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/apache/incubator-metron/pull/303#issuecomment-253525933>,
or mute the thread

cestella:
Yep, I will. :)

cestella:
Ok, so there is still an issue here.  Some of the other parsers do not create an `ip_src_addr` and `ip_dst_addr` field in their test data (e.g. `JSONMapParser`).  In those cases, the `global.json` used fails every message as invalid because the default behavior changed for the IP validation code.

As I see it, we have 2 choices:
1. Make the behavior of the IP validator return valid if the field doesn't exist.
2. Remove the IP validation check from the integration test global config
3. Force the various parsers to provide those fields.

I suggest 1 upon further consideration.  Global validation across ALL parsers is going to have to be tolerant of missing values as not all parsers are going to provide every fields.  I think it should be interpreted, "Are these fields globally valid if they exist?" rather than "are these fields globally valid?"

That being said, I think that the accommodation for this should exist in the FieldValidation function, not in the individual validator functions.  I'll make a PR to your PR to this effect, but I wanted to make sure everyone was on board with that and gather thoughts.

cestella:
Actually, upon even further consideration, I think we should revert the default behavior of the IP validator to return `true` if the field doesn't exist.  Handling it in the `FieldValidator` means we cannot express a validation function that ensures the existence of a field across all parsers (e.g. `timestamp`).

Thoughts?

ottobackwards:
I would prefer changing the IPValidator.  As an aside to this, if someone were to want to add in new models to the system I would think Metron would need to have the validation mapped to model type and not so hard coded into the system.  Some of this would fall out of that effort, unless I'm mistaken

cestella:
Yeah, I agree.  I'd like to have a sensor-specific validator as well as a global validator.  Right now we have a `MessageFilter`, but it's slightly different in intent.  It is intended to pick out which messages we want to pass through the parser and drop those that don't meet the criteria.  I'd like to see all of these concepts merged into a mapping between stellar statements and actions (i.e. DROP, REDIRECT_TO_INVALID, REDIRECT_TO_ERROR) and have sensor specific as well as global versions of this.

cestella:
That JIRA about making the ParserIntegrationTest better is at https://issues.apache.org/jira/browse/METRON-502

ottobackwards:
I would like to get the discussion on that going, as adding in new models is something near and dear.

ottobackwards:
I can't be assigned jiras but I'll take a stab at METRON-502, unless you have already completed it ;)

cestella:
@ottobackwards Sure, why don't you start a DISCUSS thread and lay out how you envision that looking?

cestella:
Go for it, man.  It's all yours. :)

cestella:
+1 on inspection

PR #304, opened 2016-10-11
METRON-496: Field transformations are applied after parser validation

cestella:
Field transformations are applied after validation, which means that the validation cannot be affected by the transformations. Consider a situation where you get a timestamp field in as a string and the parser validation expects a long. Conversion could be done as part of a field transformation, whereas now it would fail validation.


ottobackwards:
two things, if I may comment
* did you mean to bring in the CSV parser change?
* isn't the benefit of this that you can _remove_ the timestamp injection from the generic parsers and do it in stellar?  Why add it into the csv?

cestella:
Haha, definitely @ottobackwards comment!  Thanks for the feedback. :)

* I found a potential situation with the CSVParser wrt timestamps that would prevent users from retaining the timestamps passed in.  If you pass in a timestamp and do NOT configure the CSVParser to parse the timestamp (via the `timestampFormat` config option), then it'll overwrite your timestamp field with null, which would mean you have no ability to parse the timestamp in stellar downstream. 

Consider the following config:
```
 {
     ,"parserConfig": {
         "columns" : {
                "data" : 0
               ,"timestamp" : 1
                 }
     }
      ,"fieldTransformations" : [
          {
           "transformation" : "STELLAR"
          ,"output" : "timestamp"
          ,"config" : {
            "timestamp" : "TO_EPOCH_TIMESTAMP(timestamp, 'yyyy-MM-dd HH:mm:ss', 'UTC')"
                      }
          }
                               ]
   }
```
with the following data, `foo,2016-01-05 17:02:30`
As it stands, the CSVParser would return:
```
{
  "original_string" : "foo,2016-01-05 17:02:30"
  ,"data" : "foo"
  ,"timestamp" : null
```
thereby dropping the timestamp field instead of passing it through if you choose to NOT have the parser manage the timestamps.

* The benefit is that you have the choice now to do it in Stellar, but I would prefer to stay backwards compatible.  Since the CSVParser supported timestamp parsing before this PR, I left it in (and fixed a bug that would have made it impossible to pass timestamps through).

ottobackwards:
So, for the JSONMapParser - we could remove it though?

cestella:
I'd prefer to keep the functionality for the CSVParser the way it is in this JIRA.  I just wanted to make it so that the CSVParser's `parse` method wouldn't actively stomp on the data passed through if you don't choose to use the timestamp conversion functionality.  If you want to add another JIRA around removing handling timestamps in parsers, then we can.

Ultimately though, I am not super offended by the capability remaining as long as the user has the choice to not use it.

justinleet:
I'm +1 by inspection. I'm pretty agnostic on the question of leaving the capability in and would be inclined to do what Casey said and add another jira if we are interesting removing timestamp handling.

PR #308, opened 2016-10-12
Metron-498 Grok patterns are now read from zookeeper parser config property "grokPattern"

merrimanr:
This PR moves Grok patterns from HDFS to a property in Zookeeper ("SensorParserConfig.parserConfig.grokPattern").  Most of the important changes were made to the GrokParser class and related tests.  There were 2 primary challenges with implementing this:
- A change to a Grok pattern in Zookeeper should take effect immediately without a topology restart.  Currently the MessageParser does not have access to the Zookeeper config.
- Grok patterns must now comply with JSON formatting standards (files had less restrictions)

The first challenge was solved by changing the MessageParser.parse method to include an additional "SensorParserConfig" parameter that the GrokParser can use to determine if the grok pattern has changed and reinit if necessary.  The ParserBolt now passes in the most recent SensorParserConfig to the MessageParser for every message processed.  This is similar to how the writer interfaces work.  I believe this approach is simpler than having the ParserBolt listen for changes and reconfigure/reinit the GrokParser.  Changing this interface makes this PR look much bigger than it actually is because all the parsers and parser tests had to be adjusted.  Most of files included are 1 line changes.

The second challenge was overcome by using '\n' to separate lines inside the grok pattern JSON property.  An extra '\' also had to be added to '\' escape characters in the grok pattern files.  The advantage of this is that no special serialization/deserialization is needed inside of the GrokParser code.  The disadvantage is that the patterns are not as human-readable.

All unit tests and integration tests are passing.  I added an extra unit test in SampleGrokParserTest to simulate and test a config being updated.  I also tested yaf and squid sensors in quick-dev by updating the grokPattern properties and verifying the new patterns were applied.


metron-platform/metron-parsers/src/main/config/zookeeper/parsers/websphere.json
```diff
@@ -3,7 +3,7 @@
...
...(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:)))(%.+)?\nIPV4 (?<![0-9])(?:(?:25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{1,2})[.](?:25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{1,2})[.](?:25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{1,2})[.](?:25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{1,2}))(?![0-9])\nIP (?:%{IPV6:UNWANTED}|%{IPV4:UNWANTED})\n# Message - the message body of the log\nMESSAGE .*\n# WebSphere - the entire log message\nWEBSPHERE %{LOGSTART:UNWANTED} %{LOGMIDDLE:UNWANTED} %{MESSAGE:message}",
```
cestella: Is this the only way to do multi line grok statements?  This looks super awkward.

merrimanr: I agree.  The websphere Grok pattern is awkward to begin with so it's even worse in JSON.  It's a tradeoff between simpler code and being easier to read.  I prefer simpler code but I'm probably biased because I was the one writing the code :)

merrimanr:
Also, should we remove these files?

- /incubator-metron/metron-platform/metron-parsers/src/main/resources/patterns/asa
- /incubator-metron/metron-platform/metron-parsers/src/main/resources/patterns/fireeye
- /incubator-metron/metron-platform/metron-parsers/src/main/resources/patterns/sourcefire

I don't think they are being used but wasn't sure.

metron-platform/metron-parsers/src/main/config/zookeeper/parsers/websphere.json
james-sirota: we need to test the heck out of this, guys.  i like the change because this makes things much simpler because there is no level of indirection anymore where you have to load the grok statement from a file system.  but i see potential impacts on vagrant and management pack because we are now changing the way a setting is being interpreted.  I didn't see any mods there and I would have expected that. this is one of those PRs we probably want 2 people to independently verify to make sure it's solid. i think there are two things that still need to be done.  first, i think the patterns do have to be removed from HDFS because they are now redundant.  second, ansible and message pack installers have to be modified to reflect the fact that the value is now in zookeeper.  i volunteer to do the first pass and run it up on AWS once this is done.  i also volunteer to clean up the docs and tutorials to account for this change.

merrimanr: That wasn't too bad.  The grokPattern property now accepts a string or list of strings.

ottobackwards: What is the benefit of embedding multiple grok patterns, and that number could be very large for some implementations, inside the configuration?

Why not have the grok patterns in ZK as named objects - their own tree, and then have parser implementations refer to a pattern or set of patterns.

These patterns could be shared between parser configurations, if you have multiple configurations for the same type.

You could also avoid the json formatting problem maybe?

This makes the configurations harder to manage in cases with large numbers of statements and makes re-use impossible.

merrimanr: I like that idea but it's a significant increase in scope for this PR.  When Grok patterns were kept in HDFS files you couldn't reuse or share them either so as far as I know there has not been a regression.  This should be a separate follow on Jira in my opinion.

ottobackwards: So - if the grok parser loaded all the grokPattern, and also all the grokPatternSets.. as the improvement?

merrimanr: I'm sorry I don't quite understand the question.  Do you mind expanding on that?

ottobackwards: I was cribbing how I would propose it would work, for the new jira.  Sorry.

ottobackwards: that sounds good

merrimanr:
@james-sirota You are correct about changes on the Ansible side.  There are steps where HDFS directories are created and grok pattern files written that are not necessary anymore.  I removed those steps from the scripts and tested it in quick-dev.  

Are there any other cleanup items for metron-deployment?  Fortunately no major changes were required for Ansible since the deployment process is decoupled from topology configuration (as long as they are Zookeeper configs).  Only the format of the json and the serde logic in the parser were changed.  I completely agree though, would be good to have a few people run this up and test it.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -117,7 +117,7 @@ public void execute(Tuple tuple) {
       int numWritten = 0;
       if(sensorParserConfig != null) {
         List<FieldValidator> fieldValidations = getConfigurations().getFieldValidations();
-        Optional<List<JSONObject>> messages = parser.parseOptional(originalMessage);
+        Optional<List<JSONObject>> messages = parser.parseOptional(originalMessage, sensorParserConfig);
```
mattf-horton: This patch is premised on the idea that changes to the sensorParserConfig will be recognized and passed in to the GrokParser from the ParserBolt.  Sorry I'm not familiar enough with the code, but I can't find where the sensorParserConfig is refreshed, or how often it happens.  Can you please point it out?  Thanks.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/GrokParser.java
```diff
@@ -132,8 +130,9 @@ public void init() {
 
   @SuppressWarnings("unchecked")
   @Override
-  public List<JSONObject> parse(byte[] rawMessage) {
-    if (grok == null) {
+  public List<JSONObject> parse(byte[] rawMessage, SensorParserConfig sensorParserConfig) {
+    if (grok == null || isGrokPatternUpdated(sensorParserConfig) || isPatternLabelUpdated(sensorParserConfig)) {
+      configure(sensorParserConfig.getParserConfig());
```
merrimanr: The timestampField is not what you think it is.  It represents the primary timestamp field of the message that's formatted downstream in the parser, not the last time the pattern was changed.  I would like to see it go away eventually, since you can now do the same thing with Stellar in the transformation phase.

I agree with you though, since it's called for every message that check needs to be as fast as possible (although you wouldn't use a GrokParser for a high velocity sensor anyways).  I think a hash of grokPattern + patternLabel will do the trick.  I will add that in.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
merrimanr: Line 112.  The ConfiguredParserBolt classes ensures a call to getSensorParserConfig() always returns the latest config.  Check out that class and ConfiguredBolt, they leverage Apache Curator to listen for Zookeeper changes.

merrimanr: The sensorParserConfig refresh happens asynchronously.

mattf-horton: Ah, I see.  ConfiguredBolt:prepare() installs a Curator event handler which can call updateConfig() asynchronously.  Thanks for the pointer.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
```diff
@@ -34,15 +34,15 @@
    * @param rawMessage
    * @return If null is returned, this is treated as an empty list.
    */
-  List<T> parse(byte[] rawMessage);
+  List<T> parse(byte[] rawMessage, SensorParserConfig sensorParserConfig);
```
merrimanr: I like this idea but I'm a little confused on the implementation.  Wouldn't just adding this do the trick:  "BasicParser would provide an overridable implementation of the two-argument form as a trivial call to the one-argument form"?  Why would we want both methods in the interface?  Am I missing something?

mattf-horton: Actually, since sensorParserConfig is so nicely cached as a POJO, why pass it through arguments at all?  GrokParser can just fetch it again, for the comparison in GrokParser:134.  Then the signature of parse() can remain the same and none of the client parsers or their unit tests need to change; just the unit test for overall GrokParser.  Am I missing something?

mattf-horton: The reason I had previously suggested polymorphism is so that the children of BasicParser and the children of GrokParser would not have apparently different interfaces.  Granted making both implement both, when each only wants one, is less than ideal.  But I think the suggestion I just sent is best: no need to change the interface.

nickwallen: I agree with @mattf-horton 's concerns here.  It is not ideal to change the interface.  

The GrokParser is a Configurable and so it has a configure method.  Can we use the configure method to pass through the necessary grok configuration?

The ParserBolt already calls parser.configure(...) in ParserBolt.prepare.  Can we have the ParserBolt.execute call parser.configure(...) anytime the configuration changes?  Or even each time, if you like?

It seems the Configurable interface was created to pass through configuration values, so it might be a bit cleaner this way.

mattf-horton: I really like this idea architecturally, if it is assured that the call to configure() only happens between messages and not asynchronously during a parse.  If so, ParserBolt can call a parser-specific routine "parser.checkParserConfigForChange(parserConfig)" to see if the changes are of interest to the particular parser being run, which would be cheaper than checking for ANY changes in the whole config.  If the answer is yes, a call to parser.configure() is exactly what the current patch does.  But we only want to do this for parsers that care about config, so maybe we should distinguish between MessageParsers and ConfiguredMessageParsers, and do the check in ParserBolt only for ConfiguredMessageParsers.

Otherwise, I did figure out how to give GrokParser access to the cached value of grokPattern in a clean and efficient way.  But I like @nickwallen 's idea more, if we can agree on it.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/GrokParser.java
mattf-horton: Agree, on both counts.  Thanks for the correction.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
mattf-horton: If we only want to check for changes every N minutes, instead of on every message, this approach gives us a central location to implement that policy in ParserBolt too, without every parser having to worry about it.

cestella: I believe that we should use polymorphism to provide `parse(byte[], SensorParserConfig)` with a default implementation for `parse(byte[])`.  I think that should be done in the `MessageParser` interface using the `default` keyword.  We have done precisely this in that interface for `parseOptional`.  This would provide people the ability to override either one without negatively affecting existing parsers.

As for calling configure per message, this would change the semantics of the configure method for existing parsers (i.e. shifting from configure being a one-time thing to a per-message thing might change how people use the method regarding possibly computationally intensive setup).  

As a side-note, I think we should dump the `init()` method from that interface and just adopt `configure`.  It's confusing having two setup methods with precisely the same semantic characteristics.

cestella: @mattf-horton IMO I do not think we want to check for changes every N minutes considering we get push notifications from Zookeeper on configuration change.

cestella: Now that I think about it, you could actually do the following:
* Provide a `configurationChanged` method in `MessageParser` with a default noop implementation
* Change the argument to `parseOptionally` to take a Config and in the default implementation of `parseOptionally` call `configurationChanged` upon configuration change.  

This would leave `parse` alone and provide the capability to reconfigure on configuration changes seamlessly to parsers going forward.

cestella: On second thought, maybe instead of changing `parseOptionally`, you should just call `configurationChanged` from the bolt itself upon configuration change.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
ottobackwards: Couldn't we have a Configuration Management object - + interfaces that was the 'source of truth' for the configuration?  A default implementation that handled the ZK/Curator changes hand handed out the most recent configuration when asked could then be created.  A different implementation could be from file or from ????? etc.

We could have the access to that object be part of the interface that the parser could use, instead of implicitly passing it around and cluttering the api

mmiklavc:
Has this been tested with the rpm and mpack stuff? Pretty sure this will break those items, so there are probably changes that would need to be made as part of this PR to keep from having regressions.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
merrimanr: Before we go any further I think we need to decide if changing the MessageParser interface is acceptable or not.  If I remember correctly the change to the interface is what prompted this whole conversation.

My opinion is that we quit worrying so much about the interface changing and instead focus on getting it right.  Thoughts?

cestella: @nickwallen The differences between `configurationChanged` and `configure` is that `configure` has an existing intent and this would change it, we would need to ensure every parser's configure implementation is compatible with this change in intent and doesn't cause regression.

Regarding the similarity in names, I obviously don't think they're too similar (otherwise I'd not have suggested it ;).  I think `configurationChanged` sounds like a hook that is called when configurations change, which is exactly what it would be.  `configure` has an existing understanding, which is a one-time configuration.

I've expressed a desire to get rid of one of `init` or `configure`, but I don't really care which.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/GrokParser.java
```diff
@@ -150,8 +149,8 @@ public void init() {
 
       if (message.size() == 0)
         throw new RuntimeException("Grok statement produced a null message. Original message was: "
-                + originalMessage + " and the parsed message was: " + message + " . Check the pattern at: "
-                + grokPath);
+                + originalMessage + " , parsed message was: " + message + " , pattern was: "
+                + grokPattern);
```
merrimanr: You have a point here.  The whole pattern shouldn't be logged on every failure, especially since a bad grok pattern will get logged for every message.  However there is a reason I did it that way.  As a developer, I prefer to have everything I need to troubleshoot something in the same log line.  If we log a pattern only when it changes it will be hard to tell which grokPattern caused a failure.  You have the message but you'll have to go digging in the logs to find the grokPattern that was applied.  Can you think of a way to log efficiently but still provide the context needed to troubleshoot an issue?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
nickwallen: > Regarding the similarity in names, I obviously don't think they're too similar (otherwise I'd not have suggested it ;).

Sorry, I don't know what is obvious to you and what is not.  It was not obvious to me that you had thought about that.

cestella: `parseOptional` is the only one called by the parseBolt.  This just allows the parser creator to choose between whether they want to use java 8's optional to indicate no data is returned or just return null.  For a similar reason to why people had issue changing the interface during this PR, it wasn't ideal to change the interface when we moved to java 8 back when this was introduced.

cestella: @nickwallen Apologies for the snark, it was unintended.  I had considered it and I think it's not similar in meaning.

nickwallen: @merrimanr  Whichever way we go, the interface has to change. I think we just need to decide the best way to do it.

I totally get the logic in adding `parseOptional` in that you don't want to break existing code.  But on the other hand, the project is young and we control all of the parsers. We can update and change them however we need to. 

When we just append methods to the interface (like `parseOptional`), we get interfaces that are not concise and clear.  To me that is a bigger problem that will only get worse with time.  We should bite the bullet now and make the interface change to something that is clear and concise.

cestella: @nickwallen we don't actually control all of the parsers.  There are Metron installations in existence that will break and we would need to make that breaking change on a non-point release IMO to do it responsibly.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/GrokParser.java
mattf-horton: Instead of "+ grokPattern", would you consider:
+ (LOG.isDebugEnabled() ? grokPattern : (patternLabel + " (Turn on DEBUG logging to see pattern text.)"))

But if you think it's really important to log the full pattern regardless of DEBUG setting, I can accept that.  It's just that under load in the field you might run 100,000's of messages through a faulty grokPattern before realizing the error, and blow out your log storage.  Which irritates users :-)

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
mattf-horton: A lively discussion :-)  There seems to be a proposal to limit the scope of this PR to moving the grokPatterns into configuration, and open a discussion on dev as to how to implement automated parser configuration update.  I am +1 on that.  I am (non-binding) -1 on changing the parse() and parseOptional() interfaces to take a ParseConfiguration on every message.  Configuration and Parsing are currently clearly separated in the structure of the code, and I think it should stay that way.  With your agreement, I'll open the dev discussion with that reasoning.

cestella: @mattf-horton +1 to that.  I am also -1 to changing `parse()` and `parseOptional()`.  Whatever change is made here, it should be backwards compatible with the existing parsers IMO.

kylerichardson:
The ASA patterns file `/incubator-metron/metron-platform/metron-parsers/src/main/resources/patterns/asa` is being used (or soon will be) by PR #276.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/GrokParser.java
merrimanr: This is exactly what I had in mind.  Thanks!

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
mattf-horton: Are there any user-developed parsers out there that don't subclass either BasicParser or GrokParser?  If so, they'll break on the new call to the new method in MessageParser.  If this is a concern, it can be avoided by making "ConfiguredMessageParser extends MessageParser" and putting the new configurationUpdate() method there.  GrokParser would be the only current implementation of ConfiguredMessageParser, and the call to parser.configurationUpdate() in ParserBolt would be made conditional on "(parser instanceof ConfiguredMessageParser)"

mattf-horton: Hah, didn't think about Java 8 default methods in interfaces.  Of course, you're right.

Your above last set of changes looks excellent, but in GrokParser the methods configurationUpdate() and init() have to become synchronized methods.  Then I think the code as written allows the configurationUpdated flag to not need further locking, altho different ordering of things might require synchronized get/set/check methods.

merrimanr: I'll be honest, the Java 8 defaults weren't obvious to me either until I looked closer at the MessageParser interface :)

All this synchronization talk makes me nervous.  Would it make things simpler if we detected the config change in ParserBolt.execute and re-initialized the grok object from there?  We could save a cached copy of the config in the ParserBolt on each call to execute and detect changes before parse is called using the hashing approach you suggested earlier.  This is the reason I initially changed the interface parse method to include the most recent config, to avoid multi-threading complexity.  This is approach is very similar, we're just using a different method to pass this config in and detecting a change in ParserBolt instead of GrokParser.  The parse method is synchronous by nature anyways.  It doesn't need to be aware of a config change immediately, only when the parse method is called.

mattf-horton: My inclination (and my original suggestion) is to keep all the asynchronous behavior entirely in the ParserBolt (where it already is) and only call parser.configurationUpdate() synchronously in the execute method.  So yes to that.  Then no synch at all needed in GrokParser.

However, your understanding that you could use ParserBolt#updateConfig() to capture notification of changes WITHOUT expensive checks, is brilliant, and exactly what updateConfig() is for.  If we do it in ParserBolt, all you need is one synchronized routine:
private boolean configUpdatedFlag = false;
synchronized public boolean atomicCheckAndSet(boolean newValue) {
boolean oldValue = configUpdatedFlag; configUpdatedFlag = newValue; return oldValue;}

Trivial, safe as can be, and ultra cheap.  No way to get deadlock, altho if paranoid, one could give it its own lock object instead of using implicit locking on the ParserBolt instance.

mattf-horton: Sorry, that may not have been real clear.  To complete the thought, 

method added to ParserBolt:

@Override
public void updateConfig(String path, byte[] data) throws IOException {
super.updateConfig(path, data);
if (path.startsWith(ConfigurationType.PARSER.getZookeeperRoot() + "/" + getSensorType())) {
atomicCheckAndSet(true);
}}

and in execute() method, just before invoking parser.parseOptional(), do:

if (atomicCheckAndSet(false)) { parser.configurationUpdated(getSensorParserConfig().getParserConfig());}

mattf-horton: Of course one really should use AtomicBoolean.  Then, again, no locking.  Sigh.

merrimanr: I like it.  Will add these changes in and put it up for review.  Thanks @mattf-horton!

merrimanr:
Added a commit that addresses the concerns about changing the MessageParser interface.  All non-Grok related parser files have been reverted and are no longer part of this PR.  Passing unit/integration tests and tested on quick-dev.  In addition to any feedback people might have, there are 2 issues still pending:  testing with rpm/mpack and adjusting code to accommodate PR #276 once it is merged.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -116,6 +122,9 @@ public void execute(Tuple tuple) {
       boolean ackTuple = !writer.handleAck();
       int numWritten = 0;
       if(sensorParserConfig != null) {
+        if (configUpdatedFlag.getAndSet(false)) {
+          parser.configurationUpdated(getSensorParserConfig().getParserConfig());
```
mattf-horton: nice catch on the need to re-call getSensorParserConfig().  But I think it needs to be captured in the sensorParserConfig variable for use in lines 132 and 134, eg:
        if (configUpdatedFlag.getAndSet(false)) {
          sensorParserConfig = getSensorParserConfig();  //re-fetch in case it's changed
          parser.configurationUpdated(sensorParserConfig.getParserConfig());
        }

merrimanr: You're right that's definitely incorrect.  I think it would be safer if we use the config object fetched in line 118 though.  Is there a benefit to updating the config in the middle of the execute method?  We're already delaying the change in the parser until the next execute.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -168,4 +177,12 @@ public void declareOutputFields(OutputFieldsDeclarer declarer) {
     declarer.declareStream(Constants.INVALID_STREAM, new Fields("message"));
     declarer.declareStream(Constants.ERROR_STREAM, new Fields("message"));
   }
+
+  @Override
+  public void updateConfig(String path, byte[] data) throws IOException {
+    super.updateConfig(path, data);
+    if (path.startsWith(ConfigurationType.PARSER.getZookeeperRoot() + "/" + getSensorType())) {
```
merrimanr: Yes good catch.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
mattf-horton: It is necessary to re-call getSensorParserConfig(), because of the tiny chance that the change in config was interjected between the "execute()" thread doing line 118 and doing line 125.  If this happened, we would send the OLD config (from sensorParserConfig) to parser.configurationUpdated(), then clear the flag, resulting in not picking up the config change until...the next config change.  Bad.  Unless I misunderstand the TreeCache, which may be.
However, if we do this correctly, we are not delaying the change to the parser in a significant way, because the ParserBolt calls parser.configurationUpdated() before calling any other parser.* methods.

merrimanr: Sorry I wasn't completely clear.  Line 125 would change to this:
parser.configurationUpdated(sensorParserConfig.getParserConfig());

So config is fetched in line 118 and used everywhere after that until the next execute.

mattf-horton: Ah.  All that's needed is to fetch sensorParserConfig AFTER doing configUpdatedFlag.getAndSet(false).  So right before line 118 do

  boolean mustUpdateConfig = configUpdatedFlag.getAndSet(false); //must check before fetching sensorParserConfig

then in the current line 125 use mustUpdateConfig instead of configUpdatedFlag.getAndSet(false).

merrimanr: Sounds good!

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -181,7 +185,8 @@ public void declareOutputFields(OutputFieldsDeclarer declarer) {
   @Override
   public void updateConfig(String path, byte[] data) throws IOException {
     super.updateConfig(path, data);
-    if (path.startsWith(ConfigurationType.PARSER.getZookeeperRoot() + "/" + getSensorType())) {
+    String pathWithoutTrailingSlash = path.replaceAll("/+$", "");
+    if (pathWithoutTrailingSlash.equals(ConfigurationType.PARSER.getZookeeperRoot() + "/" + getSensorType())) {
```
mattf-horton: I'm concerned that there may be at least one segment of structured name between sensorType and a value that may have changed, thus still requiring a "starts with"-like calculation.  How about replace both lines 188 and 189 with:
if (path.matches("^" + ConfigurationType.PARSER.getZookeeperRoot() + "/" + getSensorType() + "(/|$)")) {

mattf-horton:
And one last, doubtless unwelcome :-) thought --
You mentioned regarding line 68 in GrokParser (referring to parserConfig.get("timestampField")), that the timestampField is in the configuration, but "represents the primary timestamp field of the message that's formatted downstream in the parser".

So, does the timestampField in the parserConfig get changed with every message?  If so, will we end up re-loading the grokPattern on every message?
I hope not, obviously, but thought I'd better ask.  Thanks.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
merrimanr: I'm sorry I don't follow.  ConfigurationType.PARSER.getZookeeperRoot() is constant and sensorType is a leaf node in Zookeeper.  Maybe an example will help.

merrimanr:
No problem, that's a good question.  The "timestampField" property only changes when you change it in Zookeeper by uploading a config.  Let's say we're parsing a sensor and the parsed message keeps it's timestamp value in a field named "timestamp_field".   The configuration would be "timestampField=timestamp_field".  If for some reason the structure of the message changed and the field were now named different, then you would change that setting to the new name.

mattf-horton:
Ah, that makes sense -- although I would have called it "timestampFieldName" :-)

+1.  I think it's ready to go, and a great piece of work!
Thanks for all your patience during the review process.

merrimanr:
Can't argue with you there, that is a better name.  Thanks for all the great suggestions @mattf-horton.

nickwallen:
@merrimanr Can you resolve the branch conflicts so we can get this merged into master?  Maybe you didn't notice.

PR #309, opened 2016-10-14
METRON-441: Aggregator function "MIN" does not work for threat triage

cestella:
If you are aggregating a series of strictly positive numbers, `MIN` does not function properly (returns `0` instead of the real minimum).


james-sirota:
+1

PR #311, opened 2016-10-16
METRON-502 Make the ParserIntegrationTest more clear on errors

ottobackwards:
This is an attempt to enhance the Integration Test framework to support exposing processor errors and invalid messages.

There are two goals here:

1. Expose errors that are in present but not exposed because the kafka queues or other error sources are not check or exposed through the Processor run
2. Allow the integration tests to fail quicker as opposed to timing out ( if for example there are invalid messages or errors in the kafka queue such that the message in never == message out the test will keep re-trying until timeout ).

A new type ProcessorResult<T> has been added to the system as the return from running a Processor.  This type encapsulates the results, errors and invalids from any given run.

All integration tests with master run.

As always any feedback is appreciated.  Specifically in two areas:

- Is there a better way, or a way more consistent with the system than introducing the ProcessorResult type?
- The integration project does not have any tests, should it ( unit tests?)
- What is the best way to test this?  Should each individual test have a 'failure mode' test as well?

### DIFF
--- a/metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
+++ b/metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
@@ -95,7 +95,7 @@ public class EnrichmentIntegrationTest extends BaseIntegrationTest {
               "{\"ip\":\"10.1.128.237\", \"local\":\"UNKNOWN\", \"type\":\"unknown\", \"asset_value\" : \"important\"},\n" +
               "{\"ip\":\"10.60.10.254\", \"local\":\"YES\", \"type\":\"printer\", \"asset_value\" : \"important\"},\n" +
               "{\"ip\":\"10.0.2.15\", \"local\":\"YES\", \"type\":\"printer\", \"asset_value\" : \"important\"}]");
-      setProperty("hbase.provider.impl","" + Provider.class.getName());
+      setProperty("hbase.provider.impl", "" + Provider.class.getName());
       setProperty("threat.intel.tracker.table", trackerHBaseTableName);
       setProperty("threat.intel.tracker.cf", cf);
       setProperty("threat.intel.simple.hbase.table", threatIntelTableName);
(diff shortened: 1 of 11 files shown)


metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
```diff
@@ -160,16 +159,36 @@ public void test() throws Exception {
...
-      Assert.assertEquals(inputMessages.size(), docs.size());
-      List<Map<String, Object>> cleanedDocs = docs;
-      validateAll(cleanedDocs);
-    }
-    finally {
+      ProcessorResult<List<Map<String, Object>>> result = runner.process(getProcessor(inputMessages));
+      List<Map<String, Object>> docs = result.getResult();
+      if (result.failed()){
+        result.printBadResults();
+        System.out.println(String.format("%d Valid Messages Processed", docs.size()));
```
cestella: maybe System.err or accumulate a string with the full dump and then do a `Assertion.fail` with the reason so it bubbles right up to the top?

metron-platform/metron-integration-test/src/main/java/org/apache/metron/integration/ProcessorResult.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public List<byte[]> getProcessInvalids(){
+        return processInvalids;
+    }
+
+    public boolean failed(){
+        return processErrors.size() > 0 || processInvalids.size() > 0;
+    }
+
+    public void printBadResults(){
```
cestella: Maybe pass a StringBuffer here and have it accumulate, so the caller can decide what to do with the bad results?

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
ottobackwards: So, print to err and not out in the printBadResults()?  I discuss the accumulating function below

cestella:
@ottobackwards I was suggesting that `printBadResults` take a StringBuffer where what would be printed is accumulated and then passed back.  Then in the individual tests, you could decide what to do with the StringBuffer.  I'd suggest passing it to `Assertion.fail` so it bubbles to the top of the test.

If you do change `printBadResults` to do that, you might consider changing it to `getBadResults` or something that doesn't imply printing.

ottobackwards:
Thanks @cestella I see what you mean now

cestella:
+1 by inspection

PR #312, opened 2016-10-17
METRON-505: Add environment variable and system property functions to the Stellar language

mmiklavc:
Tested in quick-dev from the Stellar REPL.

Build the platform, as normal. The new functions will get picked up automatically on the classpath via the annotations.

**Examples:**
```
SYSTEM_ENV_GET for exposing environment variables

[root@node1 bin]# export MIKE_VAR="blah blah blah"
[root@node1 bin]# echo $MIKE_VAR
blah blah blah

[Stellar]>>> mikevar := SYSTEM_ENV_GET('MIKE_VAR')
[Stellar]>>> mikevar
blah blah blah

SYSTEM_PROPERTY_GET for exposing system properties

[Stellar]>>> prop := SYSTEM_PROPERTY_GET( 'os.name' )
[Stellar]>>> prop
Linux
[Stellar]>>> prop := SYSTEM_PROPERTY_GET( 'user.name' )
[Stellar]>>> prop
root
```

EDIT: Also please note that I have added these functions to the metron-common project as this functionality seems like it would be useful not just in the REPL, but also within Storm topologies.

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/SystemFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/SystemFunctions.java
@@ -22,6 +22,7 @@ import org.apache.metron.common.dsl.Stellar;
 import org.apache.metron.common.system.Environment;
 
 import java.util.List;
+import java.util.function.Function;
 
 public class SystemFunctions {
 
@@ -46,11 +47,26 @@ public class SystemFunctions {
 
     @Override
     public Object apply(List<Object> args) {
-      if (args.size() == 0) {
-        return null;
-      } else {
-        return env.get((String) args.get(0));
-      }
+      return extractTypeChecked(args, 0, String.class, x -> env.get((String) x.get(0)));
+    }
+  }
+
+  /**
+   * Extract type-checked value from an argument list using the specified type check and extraction function
+   *
+   * @param args Arguments to check
+   * @param i Index of argument to extract
+   * @param clazz Object type to verify
+   * @param extractFunc Function applied to extract the value from args
+   * @return value from args if passes type checks, null otherwise
+   */
+  public static Object extractTypeChecked(List<Object> args, int i, Class clazz, Function<List<Object>, Object> extractFunc) {
+    if (args.size() < i + 1) {
+      return null;
+    } else if (clazz.isInstance(args.get(i))) {
+      return extractFunc.apply(args);
+    } else {
+      return null;
     }
   }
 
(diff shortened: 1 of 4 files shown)


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/BaseStellarFunction.java
```diff
@@ -18,8 +18,10 @@
 package org.apache.metron.common.dsl;
 
 import java.util.List;
-import java.util.Map;
 
+/**
```
cestella: Nice! 👍

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/SystemFunctions.java
```diff
@@ -0,0 +1,75 @@
...
+    public EnvGet(Environment env) {
+      this.env = env;
+    }
+
+    @Override
+    public Object apply(List<Object> args) {
+      if (args.size() == 0) {
+        return null;
+      } else {
+        return env.get((String) args.get(0));
```
cestella: I know we don't always do this in stellar functions, but it'd be nice to check `if( args.get(0) instanceof String)` so we don't get weird `ClassCastException`s on weird input.

ottobackwards: Metron should have a standard stellar verification function set so that it is more consistant

cestella: That being said, we could probably do better across the board.  Maybe some functions that consistently coerce types (or not, in the case of strings) that people can use.

cestella: Well, if we were going to coerce strings, I'd suggest using the `ConversionUtils` stuff just for consistency.  I had thought of suggesting coercing the argument, but I couldn't imagine any situation where the coerced string would be what I would want.  Like, if you passed in a list or a number accidentally, would you rather coercion or just return null?

I guess the argument for coercion is it does no harm.  If there's no environment variable, then fine, then it'll return null anyway.  If there is, then you have done good.  I don't know, I am fine with either way.

mmiklavc: Yeah, I'm not strongly opinionated on this one. I think I prefer the explicit check also.

cestella:
+1

PR #313, opened 2016-10-17
METRON-500: fix assembly id

2xyo:
The build on current master fails because <id></id> is empty.

In previous versions of maven (before 2.2 final), leaving off the assembly id and leaving the classifier unconfigured resulted in the assembly being used as the project's main artifact. With the 2.2 release, this configuration results in a validation error.


mmiklavc:
+1 on inspection.

Just to clarify for the community, this is specifically an issue with the Maven assembly plugin version, not Maven core. Here's the reference doc - http://maven.apache.org/plugins/maven-assembly-plugin/faq.html#required-classifiers

We might want to consider specifying plugin versions explicitly to avoid this type of problem with other plugins in the future.

cestella:
Agreed, while I think the rest of the assembly manifests do specify an ID, it's a good practice to specify the version explicitly to avoid this kind of issue in the future.

PR #315, opened 2016-10-19
METRON-464: Force co-location of all Metron components

dlyle65535:
Now the Ambari Add Service will show an error in the case that:

a) The Metron Parsers service is not deployed with the Storm Supervisor and Kafka Broker.
b) Any of the other Metron services aren't installed to the host containing Metron Parsers.

Tested on Docker cluster following procedure here:[ Metron MPack](https://www.evernote.com/l/AhLFVR-9CsFIYYnOnF43BlxSsT4F856qwaY)

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.2.1BETA/service_advisor.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.2.1BETA/service_advisor.py
@@ -54,11 +54,11 @@ class METRON021BETAServiceAdvisor(service_advisor.ServiceAdvisor):
         #Metron Must Co-locate with KAFKA_BROKER and STORM_SUPERVISOR
         if metronParsersHost not in kafkaBrokers:
             message = "Metron must be colocated with an instance of KAFKA BROKER"
-            items.append({ "type": 'host-component', "level": 'ERROR', "message": message, "component-name": 'METRON_PARSERS', "host": metronParsersHost })
+            items.append({ "type": 'host-component', "level": 'WARN', "message": message, "component-name": 'METRON_PARSERS', "host": metronParsersHost })
 
         if metronParsersHost not in stormSupervisors:
             message = "Metron must be colocated with an instance of STORM SUPERVISOR"
-            items.append({ "type": 'host-component', "level": 'ERROR', "message": message, "component-name": 'METRON_PARSERS', "host": metronParsersHost })
+            items.append({ "type": 'host-component', "level": 'WARN', "message": message, "component-name": 'METRON_PARSERS', "host": metronParsersHost })
 
         if metronParsersHost != metronEnrichmentMaster:
             message = "Metron Enrichment Master must be co-located with Metron Parsers on {0}".format(metronParsersHost)
@@ -69,7 +69,7 @@ class METRON021BETAServiceAdvisor(service_advisor.ServiceAdvisor):
             items.append({ "type": 'host-component', "level": 'ERROR', "message": message, "component-name": 'METRON_INDEXING', "host": metronIndexingHost })
 
         if metronParsersHost != metronEnrichmentMysqlServer:
-            message = "Metron Enrichment Master must be co-located with Metron Parsers on {0}".format(metronParsersHost)
+            message = "Metron MySQL Server must be co-located with Metron Parsers on {0}".format(metronParsersHost)
             items.append({ "type": 'host-component', "level": 'ERROR', "message": message, "component-name": 'METRON_ENRICHMENT_MYSQL_SERVER', "host": metronEnrichmentMysqlServer })
 
         return items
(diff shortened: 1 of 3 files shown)


dlyle65535:
Bump?

anandsubbu:
I did a run with of this PR on a 12-node openstack setup and here are some observations. Note that this was on a fresh deployment and _not_ one with HDP already present.

At the outset, the co-location error messages are seen against the Metron nodes. After a few moments, there is an exclamation against a faulty host with the error message seen on mouse-over.

1) Against the MySQL server, the error text says "Metron Enrichment Master must be co-located...". I understand that the primary usage of this node is for the enrichment, yes. But wanted to know if this text is intentional. Or should we rename the error to indicate the MySQL node?

mysql-server-error-colocation


2) At the very first time, Metron Parsers node displays the error about KAFKA_BROKER co-location. Subsequently, once the KAFKA node is added, this error changes to STORM SUPERVISOR co-location. This is seen to work as expected. 

kafka-colocation

storm-colocation


3) I hit a weird issue where the whole install screen froze when I tried the following:
a. With the defaults, i.e. Metron Parsers showing error about KAFKA colocation
b. Added a KAFKA host to the same as Metron Parsers -> Now, the error shows STORM SUPERVISOR colocation
c. Removed the KAFKA host that was added in Step 'a' above.
d. And the Ambari installer froze, unable to click the "Next" button.

dlyle65535:
HI @anandsubbu - thanks for taking a peek.

1) is a copy/paste bug, I'll fix and push it up.
2) agreed. That was the desired behavior, but now that you have me thinking about it, perhaps it shouldn't be. I think it'll introduce a problem in installation from scratch, because the user hasn't yet selected the slave components. Unless anyone has an objection, I'll change that from ERROR to WARN.
3) That should be fixed with [PR-322](https://github.com/apache/incubator-metron/pull/322)

Thanks for the feedback, good catches all.

dlyle65535:
In addressing the merge conflicts caused by [PR-314](https://github.com/apache/incubator-metron/pull/314), I noticed that [PR-313](https://github.com/apache/incubator-metron/pull/313) had inadvertantly changed the name of the mpack. I also noticed that we were creating an jar file for the mpack which is unnecessary.

In addition to merging, the latest commit addresses the two above items. I retested per the original procedure.

justinleet:
I'm +1 (by inspection in conjunction with Anand's testing and the most recent fixes).  Great work figuring this out.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.2.1BETA/service_advisor.py
```diff
@@ -54,11 +54,11 @@ def getServiceComponentLayoutValidations(self, services, hosts):
         #Metron Must Co-locate with KAFKA_BROKER and STORM_SUPERVISOR
         if metronParsersHost not in kafkaBrokers:
             message = "Metron must be colocated with an instance of KAFKA BROKER"
-            items.append({ "type": 'host-component', "level": 'ERROR', "message": message, "component-name": 'METRON_PARSERS', "host": metronParsersHost })
+            items.append({ "type": 'host-component', "level": 'WARN', "message": message, "component-name": 'METRON_PARSERS', "host": metronParsersHost })
```
dlyle65535: Excellent point, I'll make that change.

anandsubbu:
Thank you @dlyle65535 for a nice fix.

I'm +1 (non-binding)

I did tests with the modified commit and here are my observations:

1) The metron_mpack jar does not get generated during ``mvn clean package`` now.

2) Fresh deploy of HDP + Metron at once
- In the beginning the Metron Parser error about Kafka appears, and then changes to a warn message about Storm Supervisor. This is working as expected. 
- The co-location error for the MySQL Server, Metron Enrichment and Metron Indexing is shown when any of these are not on the same host as Metron Parsers
- In the event that a user _still_ hits **Next** button when any of the above Error exists, the Ambari wizard pops up a **Validation Issues** window that allows the users to Continue Anyway. This is the way Ambari is designed to work.
- However that said, the **Validation Issues** windows does appear even when the WARN message about Storm Supervisor is present. One would have to "Continue Anyway" to proceed. I'm not sure if there is anything that can be done for this scenario.
- Clicked through the wizard and was able to bring up the cluster fine without issues.

3) Deploy with a existing HDP cluster - Adding Metron services subsequently
- In the **Assign Masters** page, the ERROR and WARN messages are gone when all the colocation criteria are met.
- In my case, I did have an issue where the host where I was installing did not have Supervisors running. I had to go back to Ambari and use the Host->Actions screen to add a storm supervisor to the node. Likewise, I had to also use the Host->MetronHost->Add Clients option to install HBASE client, which is required during service startup, since the MetronHost did not have the clients installed during the initial HDP deployment.
- Clicked through the wizard and was able to bring up the cluster fine without issues

dlyle65535:
reopening for travis error.

dlyle65535:
@justinleet - if you're still good with my latest change, I'll go ahead and merge this.

justinleet:
@dlyle65535 I'm good with the latest change.  Still +1.  Go ahead and merge it in.

PR #320, opened 2016-10-24
Metron-402: Snort timestamp field shows up wrong value

mmiklavc:
Need to manually test in full-dev, but wanted to get this up for review asap.

Resolves https://issues.apache.org/jira/browse/METRON-402

UPDATE 10/25/16 - Tested in full-dev. Tested with incorrect/unparsable timezone and format. Verified configuration logging statements also.

```
{
  "parserClassName":"org.apache.metron.parsers.snort.BasicSnortParser",
  "sensorTopic":"snort",
  "parserConfig": {
      "dateFormat" : "MM/dd/yy-HH:mm:ss.SSSSSS",
      "timeZone" : "America/New_York"
  }
}
```

**Changes:**
- Change Snort configuration to include date in the timestamp output
- Fix BasicSnortParser to handle microseconds properly - switched to Java 8's java.time API.
- Added the ability to specify timezone and dateformat configuration to the snort parser. Defaults to system default zone for the ZoneId, MM/dd/yy-HH:mm:ss.SSSSSS for the dateformat. Note the addition of "yy" to the dateformat.

**Testing:**
Can pass in different dateformat and timezone configuration and note the different behavior.
Options are "timeZone" and "dateFormat". Valid timezones are per ZoneId.getAvailableZoneIds(). DateFormats should be valid per options here - https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html


cestella:
What happens if the user does not configure snort to include the year in the timestamp?  Do we have an error or do we assume the current year (the previous behavior)?

cestella:
Can you please rename this PR to "METRON-402: Snort timestamp field shows up wrong value"?  I don't think that the infra bot will register the change, but it'll at least ensure the commit is consistent.

mmiklavc:
It would fail to parse unless they pass in a custom format. Are we sure we want to assume the year? I'd prefer to document setting up Snort to include the year in the timestamp, however maybe it's not good to assume this can be changed.

cestella:
I think that's fine, honestly.  It does change the functionality as it stands here and any docs that were created around snort (which none come to mind) suggesting snort configuration should be adjusted.

+1 by inspection.

mmiklavc:
Had to adjust the junit version back to the global variable, per the work in https://github.com/apache/incubator-metron/pull/325

mmiklavc:
Go Travis, Go!

mmiklavc:
Just confirming I still have a +1 with the JUnit version change added after the fact

cestella:
confirmed

PR #322, opened 2016-10-25
METRON-513: Ambari MP Metainfo should not advertise version

dlyle65535:
Set versionAdvertised to false in metainfo.xml for the MPack.

Prior to this change, the cluster would report the Metron host as "OUT_OF_SYNC".

Tested on Docker cluster following procedure here: [Metron MPack](https://www.evernote.com/l/AhLFVR-9CsFIYYnOnF43BlxSsT4F856qwaY)

No longer reports "OUT_OF_SYNC"
> curl -H "X-Requested-By: ambari" -X GET -u admin:admin 
http://localhost:8080/api/v1/clusters/docker-metron/stack_versions/1
> {
>  "ClusterStackVersions" : {
>     "cluster_name" : "docker-metron",
>     "id" : 1,
>     "repository_version" : 1,
>     "stack" : "HDP",
>     "state" : "CURRENT",
>     "version" : "2.4",
>     "host_states" : {
>       "CURRENT" : [
>         "amb1.service.consul",
>         "amb2.service.consul",
>         "amb3.service.consul",
>         "amb4.service.consul"
>       ],
>       "INIT" : [ ],
>       "INSTALLED" : [ ],
>       "INSTALLING" : [ ],
>       "INSTALL_FAILED" : [ ],
>       "NOT_REQUIRED" : [ ],
>       "OUT_OF_SYNC" : [ ]
>     }

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.2.1BETA/metainfo.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.2.1BETA/metainfo.xml
@@ -30,7 +30,7 @@
           <displayName>Metron Parsers</displayName>
           <category>MASTER</category>
           <cardinality>1</cardinality>
-          <versionAdvertised>true</versionAdvertised>
+          <versionAdvertised>false</versionAdvertised>
           <reassignAllowed>false</reassignAllowed>
           <clientsToUpdateConfigs></clientsToUpdateConfigs>
           <dependencies>
@@ -77,7 +77,7 @@
           <displayName>Metron Enrichment</displayName>
           <category>MASTER</category>
           <cardinality>1</cardinality>
-          <versionAdvertised>true</versionAdvertised>
+          <versionAdvertised>false</versionAdvertised>
           <dependencies>
             <dependency>
               <name>HDFS/HDFS_CLIENT</name>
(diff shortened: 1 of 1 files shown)


ottobackwards:
What does it report then?

dlyle65535:
Hi @ottobackwards , it correctly reports the cluster is CURRENT unless one of the other stack components aren't. That logic doesn't apply to the Metron service because it is an add-on and not directly part of the HDP stack. So when I originally had set versionAdvertised to true, it was a bug.

ottobackwards:
Great - that is what we need to verify

dlyle65535:
So, prior to this patch, the OUT_OF_SYNC array in the stack_versions report would have been populated with the host where Metron was installed. That would make the cluster un-upgradeable and also has some interactions with the Add Service Wizard that are problematic. This will take care of that.

nsinha17:
+1 (non-binding) 
I tested this out in the openstack environment and the fix looks good.
Steps followed for testing:-
1. Built the Mpack and RPMs by pulling out the code from dlyle65535:METRON-513
2. Installed Metron and related services on the 12 node openstack cluster.
3. Navigated to "Admin->Stacks and versions" page in Ambari UI and the services' status showed up as expected.
Attached is the image :-
screen shot 2016-10-26 at 2 13 21 pm

screen shot 2016-10-26 at 2 14 08 pm

cestella:
+1 by inspection, great job @dlyle65535

PR #323, opened 2016-10-25
METRON-514: Snort kafka topic fails to get any data on a setup running for several days

merrimanr:
This PR attempts to solve the issue of snort data not flowing after an extended amount of time by removing Flume as a dependency and simplifying the process.  Flume is replaced by a combination of the "tail" command and kafka-console-producer, making it much easier to troubleshoot.

The implementation was inspired by the yaf sensor.  The one significant change is in the "metron-deployment/roles/snort/templates/snort-producer" init script.  The yaf init script stops the child processes by name with "killall" which doesn't work in this case because "snort-producer" is not in the first "tail" command.  Instead the process group id is retrieved in line 64 and passed to the "kill" command in line 67.

This has been tested in quick-dev, including managing the sensor through Monit.  I searched for other Flume dependencies but could not find any, therefore Flume was removed completely.


metron-deployment/roles/snort/tasks/producer.yml
```diff
@@ -15,5 +15,18 @@
...
+
+- name: Install Snort producer start script
+  template: src=start-snort-producer.sh dest={{ snort_producer_home }}/start-snort-producer.sh mode=0755
+
+- name: Install init.d service script
+  template: src=snort-producer dest=/etc/init.d/snort-producer mode=0755
+
+- name: Register the service with systemd
+  shell: systemctl enable snort-producer
+  when: ansible_distribution == "CentOS" and ansible_distribution_major_version == "7"
```
merrimanr: No problem.  Should we remove it from "/incubator-metron/metron-deployment/roles/yaf/tasks/yaf.yml" as well?

ottobackwards:
Is there a jira for this?

metron-deployment/roles/snort/tasks/producer.yml
merrimanr: Okay removed those lines from producer.yml and yaf.yml

dlyle65535:
Thanks @merrimanr, based on your testing I'm +1 by inspection (travis pending)

merrimanr:
https://issues.apache.org/jira/browse/METRON-514

metron-deployment/roles/yaf/tasks/yaf.yml
```diff
@@ -48,7 +48,3 @@
 
 - name: Install init.d service script
   template: src=yaf dest=/etc/init.d/yaf mode=0755
-
```
nickwallen: Did you intend to remove these lines related to YAF and CentOS 7?

merrimanr: I removed it based on a comment in this PR from @dlyle65535 .  Happy to put it back in if it makes sense.

dlyle65535:
@merrimanr - I was hoping to merge this in today, but it's conflicted. Think you'd have time to take a peek?

merrimanr:
@dlyle65535 conflict is resolved

PR #324, opened 2016-10-25
METRON-515: Stellar IS_EMPTY() function does not work as expected

merrimanr:
The "IS_EMPTY" Stellar function is always returning true if the field value is not a String type.  This PR changes the function to only return true when a value is missing, an empty string, or has a null value.  

I also modified the unit test to include a test for an Integer type and removed the test that asserts true when the value is a newly created object.  In my opinion that is not a valid test case.

### DIFF
--- a/metron-platform/metron-common/README.md
+++ b/metron-platform/metron-common/README.md
@@ -146,9 +146,9 @@ The following functions are supported:
     * address - The String to test
   * Returns: True if the string is a valid email address and false otherwise.
 * `IS_EMPTY`
-  * Description: Returns true if string or collection is empty and false otherwise
+  * Description: Returns true if value is null or if string or collection is empty
   * Input:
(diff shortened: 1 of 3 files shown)


mmiklavc:
For context, the original purpose and implementation of IS_EMPTY was exclusively collections and string types. And given the current validation approach, true is returned when an invalid type is passed through. The approach for a non-string type would have been IS_EMPTY(TO_STRING(rcode)). I'm not opposed to this change, but docs should be updated with this as well.

cestella:
Nice 1-liner.  +1 pending a doc change as @mmiklavc suggests to the Stellar annotation and the `README.md` for `metron-common`

merrimanr:
Docs have been updated

mmiklavc:
+1, pending Travis cooperating.

PR #325, opened 2016-10-26
METRON-512 up default junit to 4.12

ottobackwards:
"Global default JUnit version is 4.4. It would be nice to update this to 4.12 and expose newer JUnit featues. Existing pom files should be updated - some are pointing to the version in an ad-hoc fashion rather than using the global version."

full build and tests passing

### DIFF
--- a/metron-analytics/pom.xml
+++ b/metron-analytics/pom.xml
@@ -50,7 +50,7 @@
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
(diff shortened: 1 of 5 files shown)


kylerichardson:
Great idea! +1 (non-binding)

cestella:
I can dig it.  Let's leave this for a couple days for more conversation, but it has my +1 by inspection.  Thanks @ottobackwards

mmiklavc:
Full build ran for me, +1

PR #326, opened 2016-10-26
METRON-510: Update elasticsearch bro templates for *_body_len

JonZeolla:
## Problem

[METRON-510](https://issues.apache.org/jira/browse/METRON-510)

The bro *_body_len fields in [HTTP::Info](https://www.bro.org/sphinx/scripts/base/protocols/http/main.bro.html#type-HTTP::Info) can exceed the range of an int, and so writing to ElasticSearch fails with the following exception:

```
MapperParsingException[failed to parse [response_body_len]]; nested: JsonParseException[Numeric value (9876543210) out of range of int
```
## Solution

I updated the bro_index elasticsearch template to use a datatype of `long` for {request,response}_body_len, as opposed to an `integer`.

### DIFF
--- a/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template
+++ b/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template
@@ -127,7 +127,7 @@
           "index": "not_analyzed"
         },
         "request_body_len": {
-          "type": "integer"
+          "type": "long"
         },
         "uri": {
           "type": "string",
@@ -153,7 +153,7 @@
           "index": "not_analyzed"
         },
         "response_body_len": {
-          "type": "integer"
+          "type": "long"
         },
         "user_agent": {
           "type": "string"


dlyle65535:
Hi @JonZeolla - this looks like a good change to me, thanks! Can you tell me how you tested it?

JonZeolla:
Not the best, but from my install node, inside the docker container I used to install, I inserted the updated template to my ES cluster via:
`curl --data "@/root/incubator-metron/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template" -XPUT server5:9200/_template/bro_index`

Then I put a message on the indexing topic with a response_body_len of 4261543936 (a value I saw cause this issue in the past):
`echo ${jsonBlob} | /usr/hdp/2.4.0.0-169/kafka/bin/kafka-console-producer.sh --topic indexing --broker-list server1:6667,server2:6667,server3:6667,server4:6667`

And watched the indexing topic, monitoring for the value in the response_body_len field:
`/usr/hdp/2.4.0.0-169/kafka/bin/kafka-console-consumer.sh --zookeeper $zk --topic indexing | grep 4261543936`

Finally, I successfully searched for a response_body_len of 4261543936 in kibana, as well as a more general response_body_len:>2147483647, which provided valid responses.


I also pursued testing this in quick-dev but it knocked over my laptop, so I'll be setting up a VM to do that in the future...

dlyle65535:
@JonZeolla - rgr that. Good deal. I'll try to give it a spin in quick dev today. Thanks again!

dlyle65535:
+1 ran it up on quick dev, looks good!

PR #328, opened 2016-10-27
METRON-440 - handle empty strings in DOMAIN_REMOVE_SUBDOMAINS

ottobackwards:
DOMAIN_REMOVE_SUBDOMAINS would throw an exception if passed an empty string.  Changed to handle same as being passed null ( return null )


cestella:
+1 by inspection

PR #330, opened 2016-10-28
Metron 506 - Add Otto Fowler to committers

ottobackwards:
Add Otto Fowler to committers list on main site


ottobackwards:
Kicking travis

james-sirota:
+1

PR #332, opened 2016-10-31
METRON-527: Connection pool shut down seen when running threatintel_t…

justinleet:
…axii_load.sh script

### DIFF
--- a/metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/nonbulk/taxii/TaxiiHandler.java
+++ b/metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/nonbulk/taxii/TaxiiHandler.java
@@ -66,7 +66,6 @@ import javax.xml.transform.stream.StreamResult;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URL;
 import java.text.SimpleDateFormat;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


justinleet:
The Taxii HttpClient closes connections in callTaxiiService as seen here:
https://github.com/TAXIIProject/java-taxii/blob/master/src/main/java/org/mitre/taxii/client/HttpClient.java#L335

Because TaxiiHandler attempted to reuse a closed connection, the error was seen.  The fix is just to ensure the connection is created fresh before each call.  Integration test is updated to just call the run() method twice without caring about results to ensure that the connection doesn't explode.

In addition, this was tested in quick-dev using a slightly modified version (to work with Vagrant) of https://community.hortonworks.com/articles/59698/pushing-stixtaxii-feeds-from-opentaxii-server-into.html, where the playbook was run with
`ansible-playbook --private-key=./vagrant/quick-dev-platform/.vagrant/machines/node1/virtualbox/private_key -i 192.168.66.121, playbooks/install-opentaxii.yml -e ansible_python_interpreter=python -e ansible_user=vagrant -vvvv`
with the IP being quick-dev's IP and the install-opentaxii script modified to reflect this.

After the change, polling continues until ended and hbase dumps grow as new records are added.

cestella:
+1 by inspection

mmiklavc:
+1 by inspection

justinleet:
Please hold off on merging this a bit.  Anand and I are looking into an issue that is almost certainly separate, but doing due diligence.

justinleet:
I'm satisfied with my PR after further investigation.

PR #333, opened 2016-10-31
METRON-249: Field Transformation functions fail to handle invalid user inputs

cestella:
The field transformation functions fail to handle invalid user input in a way that is apparent to the user what went on.  As discussed in the dev-list discussion, a better error message with the actual stellar statement at fault should be displayed rather than an opaque, general purpose antlr exception.

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
@@ -224,7 +224,12 @@ public class NetworkFunctions {
     if(dnObj != null) {
       if(dnObj instanceof String) {
         String dn = dnObj.toString();
-        return InternetDomainName.from(dn);
+        try {
+          return InternetDomainName.from(dn);
+        }
+        catch(IllegalArgumentException iae) {
+          return null;
+        }
       }
       else {
         throw new IllegalArgumentException(dnObj + " is not a string and therefore also not a domain.");
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
@@ -222,8 +222,13 @@ public class NetworkFunctions {
 
   private static InternetDomainName toDomainName(Object dnObj) {
     if(dnObj != null) {
-      String dn = dnObj.toString();
-      return InternetDomainName.from(dn);
+      if(dnObj instanceof String) {
+        String dn = dnObj.toString();
+        return InternetDomainName.from(dn);
+      }
+      else {
+        throw new IllegalArgumentException(dnObj + " is not a string and therefore also not a domain.");
+      }
     }
     return null;
   }
@@ -232,10 +237,15 @@ public class NetworkFunctions {
     if(urlObj == null) {
       return null;
     }
-    try {
-      return new URL(urlObj.toString());
-    } catch (MalformedURLException e) {
-      return null;
+    if(urlObj instanceof String) {
+      try {
+        return new URL(urlObj.toString());
+      } catch (MalformedURLException e) {
+        return null;
+      }
+    }
+    else {
+      throw new IllegalArgumentException(urlObj + " is not a string and therefore also not a URL.");
     }
   }
 }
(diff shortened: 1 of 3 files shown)


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
```diff
@@ -232,10 +237,15 @@ private static URL toUrl(Object urlObj) {
...
       return null;
     }
-    try {
-      return new URL(urlObj.toString());
-    } catch (MalformedURLException e) {
-      return null;
+    if(urlObj instanceof String) {
+      try {
+        return new URL(urlObj.toString());
+      } catch (MalformedURLException e) {
```
ottobackwards: Why would Malformed Url not be an exception as well?   It would be much better to get a detailed error than null.

cestella: I'm treating a distinction here between a type or syntactic error which is likely a configuration error and should be immediately percolated up versus a semantic error (the data is malformed, but still of the proper type) which should be allowed to pass.  It's not necessarily an error.  You could have a source which contains a `url` field which is not populated or populated with a default value, so you'd expect the processed fields to also be null.

ottobackwards:
This works really well, I also tested sending down nulls and sending down an invalid domain name.  As per my comment I would ask if we should include the malformed url in the invalid state to get the better message.

cestella:
@ottobackwards Thanks so much for the review.  

Yeah, it's a general question about stellar functions about when to error versus when to tolerate.  We have traditionally tried to be as permissive as possible about errors that are associated with missing data.  I think you can make an argument for malformed URLs to be a proper error, but it has historically not been mostly because it was considered a semantic error.  The original concern was you might want to transform fields which were inconsistently populated or populated with default values some of the time.  In that case, if the field is malformed then the URL that it produced should be `null`.  You still have the original field and can make decisions downstream in enrichment or threat triage.

Contrasting that would be syntactic or systemic errors, like passing in a number when you intended a string.  These would indicate a high likelihood of misconfiguration, so should be errors.  This was the original thinking behind when stellar functions error and when they tolerate variable input.

I think it is high time to provide an `EXCEPTION` function so you can choose to throw an exception under certain circumstances if you want a more conservative interpretation of error.

cestella:
I wanted to add, I'm just giving historical context on why decisions were made because recent discussions have been around what to do around error states in stellar functions.  Not saying that what we came up with before is right or wrong, just trying to provide a model for the thinking.

ottobackwards:
I think what we are seeing with the question here, the handling of null's in the arithmetic functions and other recent work here is that it is confusing and possibly inconsistent. Maybe it can explained better ( is there a Stellar Language Guide to how parameters are handled?  I don't think the doc we have is enough ), or maybe it needs some more ... coordinated work as opposed to a bunch of bugs getting fixed separately.

ottobackwards:
That having been said, and being an issue that is greater than this pr, I'm +1 on this.

cestella:
@ottobackwards Yeah, I think we need better documentation and perhaps a rethinking of how we handle errors in stellar functions.  That is a broader discussion, agreed.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
ottobackwards: so - do you want to handle InvalidArgumentException for InternetDomainName.from() which throws for invalid domain names?  It should return null as well then?

cestella: Yeah, that's a good catch.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
```diff
@@ -224,7 +224,12 @@ private static InternetDomainName toDomainName(Object dnObj) {
     if(dnObj != null) {
       if(dnObj instanceof String) {
         String dn = dnObj.toString();
-        return InternetDomainName.from(dn);
+        try {
+          return InternetDomainName.from(dn);
```
ottobackwards: The modified configuration from your test that would test this is:

` /**
   {
    "fieldTransformations" : [
     {
       "transformation" : "STELLAR"
      ,"output" : [ "full_hostname", "domain_without_subdomains" ]
      ,"config" : {
         "full_hostname" : "URL_TO_HOST('http://1234567890123456789012345678901234567890123456789012345678901234567890/index.html')"
        ,"domain_without_subdomains" : "DOMAIN_REMOVE_SUBDOMAINS(full_hostname)"
                  }
     }
                      ]
   }
   */`

cestella:
Ran this up in SNV (with storm 1.0.x even! ;) and it works.  See image below for the error message.

proof

PR #334, opened 2016-11-01
METRON-529 Allow Configurable Periods for Profiler Client

nickwallen:
## [METRON-529](https://issues.apache.org/jira/browse/METRON-529)

By default, the Profiler creates Profiles with a period duration of 15 minutes. This means that data is accumulated, summarized and flushed every 15 minutes. The Client API must also have knowledge of this duration to correctly retrieve the profile data. If the client API is expecting 15 minute periods, it will not be able to read data generated by a Profiler that has been configured with a 1 hour period.

The period duration can be configured in the Profiler by altering the Profiler topology's static properties file. The Stellar Client API currently provides no means to configure the period duration and defaults also to 15 minutes. This means that the Stellar Client API can only read profiles with a period duration of 15 minutes. 

This PR addresses this limitation so that the Profiler Client can be configured to read profile data no matter what configuration settings were used to generate that data.

### Testing

I tested this change in the "Quick Dev" environment.  I used the Profiler to create a Profile using a 1 minute period.  I then confirmed that I was unable to read the data using the Profiler client with the default client configuration.  I then changed the client configuration to a 1 minute duration.  I was then able to read the profile data as expected.

#### Steps

1. Ensure the Snort sensor and the Snort parser are running.  Stop all other topologies and sensors.

2. Edit `/usr/metron/0.2.1BETA/config/profiler.properties` and change the period duration to 1 minute.  Also, point the Profiler at the `enrichments` topic to read data directly from the Snort Parser.
  ```
  profiler.input.topic=enrichments
  profiler.period.duration=1
  ```
3. Follow the Profiler's [Getting Started](https://github.com/apache/incubator-metron/tree/master/metron-analytics/metron-profiler#getting-started) instructions to start generating profile data.  

4. Configure the client to read data with a 1 minute period.  To do so, edit `/usr/metron/0.2.1BETA/config/zookeeper/global.properties`.
  ```
  {
    "es.clustername": "metron",
    "es.ip": "node1",
    "es.port": "9300",

    "profiler.client.period.duration": "1",
    "profiler.client.period.duration.units": "MINUTES"
  }
  ```

5.  Push the previous global configuration change to Zookeeper.
    ```
    bin/zk_load_configs.sh -m PUSH -i config/zookeeper/ -z node1:2181
    ```

6. Open up the Stellar shell and use the Profile client to read the data.
  ```
  [root@node1 0.2.1BETA]# bin/stellar -z node1:2181
  Stellar, Go!
  {es.clustername=metron, es.ip=node1, es.port=9300, es.date.format=yyyy.MM.dd.HH, profiler.client.hbase.table=profiler, profiler.client.column.family=P, profiler.client.hbase.table.provider=org.apache.metron.hbase.HTableProvider, profiler.client.period.duration=1, profiler.client.period.duration.units=MINUTES, profiler.client.salt.divisor=1000}
  [Stellar]>>> PROFILE_GET('test','192.168.138.158', 2, 'DAYS')
  [161.0]
```

### CODE REVIEW

**Comment by nickwallen** [comment]
Go Travis, Go!

File: metron-analytics/metron-profiler-client/README.md [hunk 1]
```diff
@@ -41,13 +41,20 @@ Retrieve all values of 'snort-alerts' from '10.0.0.1' that occurred on 'weekdays
...
 
-The period duration can be configured in the Profiler by altering the Profiler topology's static properties file.  The Stellar Client API currently provides no means to configure the period duration and defaults also to 15 minutes.  This means that the Stellar Client API can only read profiles with a period duration of 15 minutes.  This is a known limitation that has not yet been addressed.
+| Key                                   | Description                                                                                                                        | Required | Default  |
+| ------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- | -------- | -------- |
+| profiler.client.hbase.table           | The name of the HBase table used to store profile data.                                                                            | Optional | profiler |
+| profiler.client.column.family         | The name of the HBase column family used to store profile data.                                                                    | Optional | P        |
```
mattf-horton: The hbase column family is not documented as a config option in profiler.properties, https://github.com/apache/incubator-metron/tree/master/metron-analytics/metron-profiler#user-content-topology-configuration
Whether that's a docs bug or a profiler implementation limitation, would this be a good time for that fix or enhancement too?
Is a variable TableProvider also an issue for profiler?

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
```diff
@@ -226,14 +262,23 @@ private ColumnBuilder getColumnBuilder(Map<String, Object> global) {
...
+    String configuredUnits = (String) global.getOrDefault(PROFILER_PERIOD_UNITS, PROFILER_PERIOD_UNITS_DEFAULT);
+    TimeUnit units = TimeUnit.valueOf(configuredUnits);
+    LOG.debug("profiler client: {}={}", PROFILER_PERIOD_UNITS, units);
+
+    // what is the salt divisor?
+    String configuredSaltDivisor = (String) global.getOrDefault(PROFILER_SALT_DIVISOR, PROFILER_SALT_DIVISOR_DEFAULT);
+    int saltDivisor = Integer.parseInt(configuredSaltDivisor);
+    LOG.debug("profiler client: {}={}", PROFILER_SALT_DIVISOR, saltDivisor);
+
+    return new SaltyRowKeyBuilder(saltDivisor, duration, units);
```
mattf-horton: 1. If GetProfile attempts to read a profile with the wrong window duration, will the user get a reasonable error message?  I was unable to find this in GetProfile.java or GetProfileTest.java.

2. Unlike the profiler, it seems to me that there may be need for multiple simultaneous clients for different users, some of which may be consuming historical profile data that was taken with different window duration (and other settings) than is currently in use at the time the client is run.  Therefore, I suggest that GetProfile should also take an optional override set of config arguments (any or all of profiler.client.* , possibly excepting table.provider) so that clients accessing different data won't stomp on each other.

metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/ProfileWriter.java
```diff
@@ -95,4 +102,22 @@ private void write(ProfileMeasurement m, List<Object> groups) {
...
+    Configuration config = HBaseConfiguration.create();
+    config.set("hbase.master.hostname", "node1");
+    config.set("hbase.regionserver.hostname", "node1");
+    config.set("hbase.zookeeper.quorum", "node1");
+
+    HTableProvider provider = new HTableProvider();
+    HTableInterface table = provider.getTable(config, "profiler");
+
+    long when = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(2);
+    ProfileMeasurement measure = new ProfileMeasurement("profile1", "192.168.66.121", when, 15, TimeUnit.MINUTES);
```
mattf-horton: Running tests with interval set to 15 minutes will make unit testing and integration testing unreasonably long.  Please use small numbers of seconds as duration setting for all tests.

metron-analytics/metron-profiler-client/README.md
```diff
@@ -41,13 +41,20 @@ Retrieve all values of 'snort-alerts' from '10.0.0.1' that occurred on 'weekdays
...
 ```
 
-### Period Duration
+### Configuration
 
-*WARNING*
+By default, the Profiler creates Profiles with a period duration of 15 minutes. This means that data is accumulated, summarized and flushed every 15 minutes. The Client API must also have knowledge of this duration to correctly retrieve the profile data. If the client API is expecting 15 minute periods, it will not be able to read data generated by a Profiler that has been configured with a 1 hour period.
 
-By default, the Profiler creates Profiles with a period duration of 15 minutes.  This means that data is accumulated, summarized and flushed every 15 minutes.  The Client API must also have knowledge of this duration to correctly retrieve the profile data.  If the client API is expected 15 minute periods, it will not be able to read data generated by a Profiler that has been configured with a 1 hour period.
+The period duration can be configured in the Profiler by altering the Profiler topology's static properties file. The Stellar Client API can be configured by setting the following properties in Metron's global configuration.
```
nickwallen: Sure.  Makes sense.

metron-analytics/metron-profiler-client/README.md
nickwallen: I will update the Profiler docs.  Thanks

metron-analytics/metron-profiler-client/README.md
```diff
@@ -41,13 +41,20 @@ Retrieve all values of 'snort-alerts' from '10.0.0.1' that occurred on 'weekdays
...
+| ------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- | -------- | -------- |
+| profiler.client.hbase.table           | The name of the HBase table used to store profile data.                                                                            | Optional | profiler |
+| profiler.client.column.family         | The name of the HBase column family used to store profile data.                                                                    | Optional | P        |
+| profiler.client.hbase.table.provider  | The name of the HBaseTableProvider implementation class.                                                                           | Optional |          |
+| profiler.client.period.duration       | The duration of each profile period.  This value should be defined along with `profiler.client.period.duration.units`.             | Optional | 15       |
+| profiler.client.period.duration.units | The units used to specify the profile period duration.  This value should be defined along with `profiler.client.period.duration`. | Optional | MINUTES  |
```
nickwallen: In actuality, the Profiler does match the same scheme as the client (using duration and units).  The docs for the Profiler were not updated.  I will fix that now.  Good catch.

metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/ProfileWriter.java
nickwallen: This doesn't actually run as part of the tests and does not consume 15 minutes.  So many times I've had to load up profile data for demo purposes.  This main() function just provides a simple means to load up a bunch of profile data in a single shot.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
nickwallen: 2.  That's a good thought.  I'm not ignoring it.  I just need to noodle on that a bit.  Hmm.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
```diff
@@ -82,19 +84,53 @@
...
    * A global property that defines the name of the column family used to store profile data.
    */
-  public static final String PROFILER_COLUMN_FAMILY = "profiler.column.family";
+  public static final String PROFILER_COLUMN_FAMILY = "profiler.client.hbase.column.family";
 
   /**
    * A global property that defines the name of the HBaseTableProvider implementation class.
    */
-  public static final String PROFILER_HBASE_TABLE_PROVIDER = "profiler.hbase.table.provider";
+  public static final String PROFILER_HBASE_TABLE_PROVIDER = "hbase.provider.impl";
```
nickwallen: I changed this property to `hbase.provider.impl` to match what is defined in other places like Enrichment.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
mattf-horton: (1) "wrong" == GetProfile attempts to read a profile using a window duration different from what the profile was written with.
So if the profile was written using 1-hr intervals, and the client attempts to read it using 15-minute intervals, does the user get a reasonable error message?

metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/ProfileWriter.java
mattf-horton: Got it. The client tests run almost instantly. This writes something for the client to read WITHOUT having to wait for several sampling intervals. Thanks.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
```diff
@@ -91,12 +91,12 @@
...
    * A global property that defines the name of the column family used to store profile data.
    */
-  public static final String PROFILER_COLUMN_FAMILY = "profiler.client.column.family";
+  public static final String PROFILER_COLUMN_FAMILY = "profiler.client.hbase.column.family";
 
   /**
    * A global property that defines the name of the HBaseTableProvider implementation class.
    */
-  public static final String PROFILER_HBASE_TABLE_PROVIDER = "profiler.client.hbase.table.provider";
+  public static final String PROFILER_HBASE_TABLE_PROVIDER = "hbase.provider.impl";
```
mattf-horton: This isn't an objection, just a question: I'm not yet real familiar with the structure of the configuration tree in ZK.  Is this "hbase.provider.impl" parameter stored in a part of the tree where it clearly ONLY applies to profiler?  Otherwise I would prefer to leave it with a "profiler." prefix.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
mattf-horton: This isn't an objection, just a question: I'm not yet real familiar with the structure of the configuration tree in ZK. Is this "hbase.provider.impl" parameter stored in a part of the tree where it clearly ONLY applies to profiler? Otherwise I would prefer to leave it with a "profiler." prefix.

nickwallen: For example, EnrichmentIntegrationTest.java:98 or ProfilerIntegrationTest.java:266.  Open to suggestions

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
nickwallen: Unfortunately, the user does not get a reasonable error message.  It is a little bit of a sticky wicket.  I am sure its doable, but I just haven't thought of a way that I really like.

A ProfilerClient uses a RowKeyBuilder along with the profile name, entity, period duration, and date range to construct a set of row keys that meet the user's query (see HBaseProfilerClient:100).  

If there are no 'hits' for that set of row keys, then the user gets nothing back.  I'd need some way to distinguish between "there is no data" versus "there is data, but it is written with a different period duration".  

I think it'd be a good bit of work to catch all the corner cases.  Although, I'm totally open to suggestions.  Maybe there is an easy way that I am not thinking about.

nickwallen: I had previously opened [METRON-450](https://issues.apache.org/jira/browse/METRON-450) to help the user with this kind of problem.  With METRON-450, I could query the types of profiles that I have and see that profile P has a duration of D.  This would help the user understand that they are using the wrong period duration, among other things.

METRON-450 is not exactly what you're looking for, but I thought I would mention it.

mattf-horton: Okay, I see where #1 is a hard problem.  Thanks for the ref to METRON-450, which is certainly related.  Let's drop it for a future enhancement.

What about #2 ? :-)  I think that one's easier.  And probably more important.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
mattf-horton: Got it.  No problem.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
nickwallen: Right now the entire Profiler runs with one period duration.  If the Profiler is generating multiple profiles, they will each have the same profile period.  I don't foresee multiple clients trying to read profiles with different periods at the same time.  At least, not right now.

In the future though, I think it would be very useful to enhance the Profiler so that each profile can define its own period; 15 minutes, 4 hours, etc.  The period that you want to use is going to depend on the data itself and how you want to use the data.  I just opened [METRON-530](https://issues.apache.org/jira/browse/METRON-530) to track this.  

When we complete METRON-530, then I think we definitely need to make this improvement that you are suggesting.  A Profiler Client will need to be able to read from many different profiles with unique durations.  But at the same time, we will need [METRON-450](https://issues.apache.org/jira/browse/METRON-450), and probably some other enhancements.

My gut tells me that we could defer this enhancement until we prioritize METRON-530.  What do you think?

mattf-horton: I had thought about the profiler having that limitation, I'm glad you opened the jira, and I definitely agree that can be deferred.  Regarding the client, though, I'm thinking about this scenario:
(a) There are one or two clients doing on-going routine analysis of up-to-date profiles, perhaps they are updaters for ML modules;
(b) Suddenly a security event requires analyzing some historical profile data from 2 months ago, that was done with different profiles -- different table, different period, something.
We shouldn't need to turn off (a) in order to do (b).

I won't insist on this, because certainly what you've already done is a great improvement.  But I'd still like to suggest that a small addition to the mods you've made in GetProfile would give this flexibility.  Do you mind if I just whomp it up?  Then you can look at it, and if you still don't like it, I'll shut up :-)

nickwallen: I think what you are saying makes total sense.  It is a good idea. We should have it.  I'd prefer to do it as a follow-on JIRA though.  Hey, maybe you could follow-on with your 'whomp'.  Ha.

The one problem in doing it is due to the signature of PROFILE_GET.  It would be ideal to add the period duration, etc as arguments towards the end of the function.  That way the user can decide to pass in those additional args like duration or not.  I'd hate to require the user to pass it in.

The problem is that the last argument is effectively a var arg of the group names.  That makes it difficult to do what you are asking.  How do you think we could work around that?

mattf-horton: Right.  Because of that we can't just generalize the PROFILE_GET argument list to include this.  I'm thinking a PROFILE_GET_CONF that has a single string arg before "groups", consisting of a comma-separated list of key=value pairs, each overriding the global client parameter of the same name.  

PROFILE_GET_CONF has the whole implementation, essentially same as current but with each of your new parameters possibly coming from the override string instead of from the global config.  PROFILE_GET becomes a stub, invoking PROFILE_GET_CONF with an empty override string.

mattf-horton: Okay, let's open a new jira for this too (separate from METRON-530, because they are completely separable and this is higher priority), and get on with this PR.

mattf-horton:
I'm now a strong +1 on this (by inspection), subject to the usual tests.  Really important piece of work, kudos to Nick!

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
nickwallen: Good idea.  I opened [METRON-532](https://issues.apache.org/jira/browse/METRON-532).  Feel free to comment on that issue, change the title, change the description, whatever you feel is appropriate.  Hopefully, I've captured the core idea.

cestella:
This looks good, @nickwallen.  +1 binding from me by inspection.

PR #337, opened 2016-11-02
METRON-295: Script parsing bolt

rkarthik29:
I have added support for creating parsers using js, python or groovy and using them in metron. This is for Jira Metron-295

### DIFF
--- a/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ScriptParser.java
+++ b/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ScriptParser.java
@@ -10,6 +10,7 @@ import java.util.Map;
 
 import javax.script.Invocable;
 import javax.script.ScriptEngine;
+import javax.script.ScriptEngineFactory;
 import javax.script.ScriptEngineManager;
 
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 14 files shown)


cestella:
Hi Karthik.  Can you close this and reopen with the title: "METRON-295: Script parsing bolt"?

rkarthik29:
i changed the title

ottobackwards:
don't forget the reopen part

ottobackwards:
This is interesting.  I would think we would want some tooling like https://github.com/YelpArchive/pyleus around building a jar that contains all the python dependencies for a script for people to do non-trivial things in python.

I am not sure how you would do the same for javascript or groovy.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ScriptParser.java
```diff
@@ -0,0 +1,175 @@
...
+		        if (fieldValue != null) {
+		          message.put(timeField, toEpoch(fieldValue));
+		        }
+		      }
+		      if (timestampField != null) {
+		        message.put(Constants.Fields.TIMESTAMP.getName(), formatTimestamp(message.get(timestampField)));
+		      }
+		      messages.add(message);
+		      if (LOG.isDebugEnabled()) {
+		        LOG.debug("Grok parser parsed message: " + message);
```
james-sirota: Looks like cut and paste error from the Grok code

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ScriptParser.java
```diff
@@ -0,0 +1,175 @@
...
+		      }
+		}catch (Exception e) {
+		    LOG.error(e.getMessage(), e);
+		    throw new IllegalStateException("Grok parser Error: " + e.getMessage() + " on " + originalMessage , e);
```
james-sirota: Here is another one

PR #341, opened 2016-11-03
METRON-535: Global Validations type still refers to MQL rather than Stellar

cestella:
The global validation processor that allows you to specify global validations via Stellar is still named "MQL" rather than "STELLAR". The documentation states otherwise. This is inconsistent and should have its name changed.


mmiklavc:
+1 by inspection

PR #342, opened 2016-11-03
METRON-536: Add ASA Parser Artifacts to RPM spec files

dlyle65535:
Tested on Docker cluster following procedure here: [Metron MPack](https://www.evernote.com/l/AhLFVR-9CsFIYYnOnF43BlxSsT4F856qwaY)

- RPMs build successfully
- MPack deployment successful. 
- Asa files are in correct locations.


anandsubbu:
+1 (non-binding)

I have verified this on a openstack VM. Before the fix, building metron RPMs using docker was failing complaining about the unpackaged ASA json files. Now with the fix applied, I am able to build the RPMs fine. Thank you!

dlyle65535:
Booting Travis. There is nothing in this PR that would cause the ElasticsearchWriter integration test to fail.

justinleet:
+1, by inspection.

PR #345, opened 2016-11-06
METRON-532 Define Profile Period When Calling PROFILE_GET

mattf-horton:
This PR builds on top of @nickwallen 's work in METRON-529, which has already been committed to master.  It adds an optional argument to PROFILE_GET, allowing run-time overrides of the profiler client global configuration parameters.  The primary use case is when historical profiles have been created with a different profile configuration than is currently configured, and the analyst needing to access them does not want to change the global client configuration so as not to disrupt the work of other analysts working with current profiles.

The proposed placement of the optional Map argument is before the group sequence, so as not to disrupt the current usage of a variable-length sequence of 'group' arguments.  Example:
```
// Retrieve all values for 'entity1' from 'profile1' that occurred on 'weekdays' over the past month,
// overriding the usual global client configuration parameters for window duration.
PROFILE_GET('profile1', 'entity1', 1, 'MONTHS', {'profiler.client.period.duration' : '2', 'profiler.client.period.duration.units' : 'MINUTES'}, 'weekdays')
```

An alternative would be to change the 'groups' argument into a List, rather than a variable-length sequence of Strings, then put the config_override Map after the groups List.  Let me know if you think this would be a better implementation, as I can change it very quickly.  Of course the original usage would be retained for backward compatibility, but perhaps deprecated.

Testing:  This change is restricted to the Stellar function implementation, so the expanded junit test is largely sufficient.  Also, a previous version of this code passed Travis in a preview PR against Nick's work branch.

### DIFF
--- a/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template
+++ b/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template
@@ -127,7 +127,7 @@
           "index": "not_analyzed"
         },
         "request_body_len": {
-          "type": "integer"
+          "type": "long"
         },
         "uri": {
           "type": "string",
@@ -153,7 +153,7 @@
           "index": "not_analyzed"
         },
         "response_body_len": {
-          "type": "integer"
+          "type": "long"
         },
         "user_agent": {
           "type": "string"


metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/GetProfile.java
```diff
@@ -132,35 +144,33 @@
...
     Context.Capabilities[] required = { GLOBAL_CONFIG };
     validateCapabilities(context, required);
-    Map<String, Object> global = (Map<String, Object>) context.getCapability(GLOBAL_CONFIG).get();
-
-    // create the profiler client
-    RowKeyBuilder rowKeyBuilder = getRowKeyBuilder(global);
-    ColumnBuilder columnBuilder = getColumnBuilder(global);
-    HTableInterface table = getTable(global);
-    client = new HBaseProfilerClient(table, rowKeyBuilder, columnBuilder);
+    global = (Map<String, Object>) context.getCapability(GLOBAL_CONFIG).get();
```
cestella: If the global config changes, you probably want to pick that up post-initialization.  I would calling `context.getCapability(GLOBAL_CONFIG).get()` from the apply rather than caching a copy.

mattf-horton:
Apologies, but I'm going to be offline for a couple days, due to a surgery.  I will fix this when I get back. --Matt

cestella:
@mattf-horton No worries, get better, man.

nickwallen:
@mattf-horton I like the approach.  I would agree with @cestella 's suggestion.  After that update, I think it will be a good improvement.

mattf-horton:
Thanks @cestella and @nickwallen .  Suggestion implemented.

cestella:
Awesome!  Did you run this up and test out the profiler in vagrant?

nickwallen:
Hey @mattf-horton, there is an outstanding question from @cestella on testing.  Once we get that addressed, I'd love to get your work merged in.  Thanks!

mattf-horton:
@nickwallen , I can't satisfactorily run the quick dev platform on my workstation, so I'm working on bringing up a 16GB test platform in the cloud.  Doing a brief side trip to enable the Mpack to do a single-node install, then will test this properly.

PR #344, opened 2016-11-03
METRON-538  Ensure proper shutdown, exceptions of integration test components

ottobackwards:
This is an effort to eliminate spam errors from our testing and hopefully improve our testing at the same time.

This PR will track the commits for this effort, and allow for timely comment and review.

So far the things we are looking for or seeing are

- problems with shutdown dependencies with the testing components like yarn and zk
- problems were we cannot completely mock out components.  Even if the tests pass, we are spammed with errors because of the attempts to connect to components that are part of the test because they are not needed.

Because of Apache Storm Issue STORM-1730
https://issues.apache.org/jira/browse/STORM-1730 it is not possible
to shutdown the LocalServer we use for storm without leaking threads.
These threads will output error messages when they 'wake up' and try to connect
with curator - to a zk embedded server that has already been shutdown.
This issue is fixed in Storm 1.0.3, but that is not released yet

Because of this unresolved issue  ZOOKEEPER-1237: https://issues.apache.org/jira/browse/ZOOKEEPER-1237 we also may see errors like
Running org.apache.metron.enrichment.integration.EnrichmentIntegrationTest

2016-11-09 22:25:50,293 ERROR [SyncThread:0] server.NIOServerCnxn (NIOServerCnxn.java:sendBuffer(178)) - Unexpected Exception: 

java.nio.channels.CancelledKeyException

	at sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73)

	at sun.nio.ch.SelectionKeyImpl.interestOps(SelectionKeyImpl.java:77)

	at org.apache.storm.shade.org.apache.zookeeper.server.NIOServerCnxn.sendBuffer(NIOServerCnxn.java:151)

	at org.apache.storm.shade.org.apache.zookeeper.server.NIOServerCnxn.sendResponse(NIOServerCnxn.java:1081)

	at 

My goal is to keep this PR mergable with every commit as I go.  And as a bonus, each travis build will hopefully show the progress.

PR #347, opened 2016-11-07
METRON-541: Set log levels for Metron components

merrimanr:
This PR accomplishes primarily 2 things:  sets the log level for all components (and their dependencies) to ERROR and suppresses ERROR log statements in tests for cases where an exception is expected.  Ideally a combination of this PR and #344 will make the output logs in Travis much cleaner.

The approach for setting the log level is to exclude all transitive SJF4J bindings and instead declare the log4j binding in the metron-platform and metron-analytics poms (would top level be better?).  There is some code that uses log4j instead SLF4J so this binding seemed appropriate.  Now there are no more SLF4J warnings and the log level for every component can be configured in test/resources/log4j.properties.  It is currently set to ERROR by default but do we want to add different log files for different profiles (dev, travis, prod, etc)?

The approach for suppressing expected ERROR log messages is to temporarily change the log level to SEVERE/FATAL during a test that generates an ERROR message and change it back right after the test (or line within the test) has executed.  A couple of methods to programmatically change the log level at runtime was added to the UnitTestHelper class.  

There were a few test cases where an exception was incorrectly thrown so I fixed those (BulkMessageWriterBoltTest.java and ParserBoltTest.java).

There are a lot of different ways to do logging so I'm expecting the community will have opinions on what the right approach is for us.  I have tested this with unit/integration tests and will test with vagrant once we get closer to a consensus.


cestella:
Did you check this on vagrant?  I seem to remember there being some drama when I tried to remove the slf4j dependencies around some of the data management functions.  Specifically, calling the `flatfile_loader.sh`.

mmiklavc:
@merrimanr Love this effort. We just went through something similar in getting the apache curator versions lined up for the Storm upgrade support ticket. Thinking more about it, I'm not sure we need to explicitly exclude all of these dependencies. Can we try just an explicit declaration at the global level?  Per "Dependency mediation" from the Maven docs:

> You can always guarantee a version by declaring it explicitly in your project's POM. Note that if two dependency versions are at the same depth in the dependency tree, until Maven 2.0.8 it was not defined which one would win, **but since Maven 2.0.9 it's the order in the declaration that counts: the first declaration wins**. 

I *think* our own explicit declarations should supersede the transitive dependencies, which should clean up our poms quite a bit if I'm right about that.

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies

merrimanr:
@cestella I don't think that will be an issue since the binding is now coming from the metron-platform pom but I will definitely test it in vagrant after we get some feedback.

ottobackwards:
Ok - you beat me in.  I think my PR should be dependent on this.  When this goes in, I'll sync up with master etc etc

merrimanr:
@mmiklavc Good question.  We definitely need to exclude the SLF4J bindings that we don't use.  The most common one is the Log4J2 binding that Storm brings along ("org.apache.logging.log4j:log4j-slf4j-impl").  

It could be argued that we don't need to exclude the binding we are using ("org.slf4j:slf4j-api") and control the version through dependency management as you suggest.  The downside I see there though is that if we ever did want to change the logging implementation we would have to go through all the poms and exclude that binding (what's already been done here).  As it stands, changing the logging implementation is as simple as changing the binding in the metron-platform/analytics pom files and adding logging configuration files.

merrimanr:
Resolved a merge conflict with SnortParserTest and changed Travis configuration to not run mvn install (since we're already doing it as part of our tests).  Check out those logs!  The stack traces should be gone after #344.

mmiklavc:
I'll miss those verbose logs. +1 pending the vagrant checks.

PR #350, opened 2016-11-10
METRON-547 Create integration components for Yarn and Zookeeper, use them in Maas Tests

ottobackwards:
Standardize the Maas Integration testing to use the Component/ComponentRunner system and create re-usable components for Yarn and Zookeeper

Test with mvn clean integration-test


metron-platform/metron-integration-test/src/main/java/org/apache/metron/integration/components/YarnComponent.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public String getAppMasterJar(){
+        return appmasterJar;
+    }
+
+    public YarnConfiguration getConfig(){
+        return conf;
+    }
+
```
ottobackwards: One question here is to expose things like the node manager like this, as needed or to just expose the cluster itself.  Thoughts?

cestella:
Could you please rename this preface METRON-547?

ottobackwards:
sorry

metron-platform/metron-integration-test/src/main/java/org/apache/metron/integration/components/YarnComponent.java
cestella: I'd just expose the `YarnCluster`, honestly.

cestella:
I dig it.  +1 by inspection.

ottobackwards:
changed based on review, thanks!

PR #351, opened 2016-11-10
METRON-561 the system property should override the sys env shell if set

ottobackwards:
When running mvn integration-test , after having set EDITOR=vim in my .zshrc I see the following:

Running org.apache.metron.management.ShellFunctionsTest

Vim: Warning: Output is not to a terminal
Vim: Warning: Input is not from a terminal

And the tests never complete.

The tests should run to completion and not count on system variables being configured the "right" way

---

If the System.Property for EDITOR is set, then let that override the sys env


cestella:
+1 by inspection

PR #352, opened 2016-11-10
METRON-562: Add rudimentary statistical outlier detection

cestella:
With the advent of the profiler, we can now capture state. Furthermore, with Stellar, we can capture statistical summaries. We should provide rudimentary outlier detection functionality in the form of Stellar functions that can operate on captured state from the profiler.
To begin, we should enable simple outlier tests using distance from a central measure such as Median Absolute Deviation (see http://www.itl.nist.gov/div898/handbook/eda/section3/eda35h.htm).


cestella:
Note: I'm still in the process of testing this on vagrant and I owe a README.md with an example.

cestella:
I want to also summarize a couple of my thoughts that I had in the course of creating this PR.  It appears to me that there are a number of these sorts of outlier analysis models that can be split into a common pattern:
* Gather state via the profiler.
* Operate on that state during the scoring mechanism

For these z-score based statistical outlier analysis techniques/models, that state comes in the form of a statistical summary or multiple statistical summaries in the case of median absolute deviation.  The scoring component is literally composing a modified z-score (scaled by MAD).  The question arises as to whether or not this model is better served in MaaS or as a stellar function.  Where I have come down in this PR is that I feel these very simple z-score or deviation-from-a-central-moment style outlier models fit better within Stellar due to their simplicity from a computational perspective.  Furthermore, it would seem to be less than ideal performance-wise to perform a network hop to a REST interface in MaaS to do simple subtraction and division.

An alternate architecture here would be to use MaaS and have the model pull the appropriate statistical context from the profiler's store in HBase.   I am sensitive to the possibility that the placement of model application within Metron may be a point of confusion and a consistent message is important.  I decided to go with simplicity and performance for the moment, but I'd love to hear if there is opposition in the audience.

cestella:
One more thing, I also added a new lifecycle hook for the profiler.  Now you can update the state on the tick.  This is needed for MAD because we will need to refresh the statistical summary of the lookback window that we are using to compute the MAD.  This should become more clear upon the creation of a README with a worked example, which I owe y'all.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -246,6 +247,8 @@ private void flush(Tuple tickTuple) {
     // clear the execution state to prepare for the next window
     executor.clearState();
 
+    profileConfig.getTickUpdate().forEach((var, expr) -> executor.assign(var, expr, new HashMap<>()));
```
cestella: @nickwallen My intent here is to add a hook in the profiler on tick.  Would you take a look and make sure I'm doing this right?  I'm interested specifically in the empty hashmap here.  I'd like to ensure that the current message is being passed in on tick, but I fear it won't be the case.

nickwallen: At this point in execution, during a flush operation, we don't have a telemetry message.  It is the tick tuple that drives the flush.  We won't have a telemetry message to offer to the 'tick' expressions.

With how you have it now, the 'tick' expressions will be able to reference all of the state being maintained by the 'init' and 'update' expressions.

The one missing piece that I can think of is that the 'tick' expressions cannot access the result of the 'result' expression.  It might be useful to offer that up to the 'tick' expressions.

cestella: I do not need to access the message, but rather just the internal state.  I have made a modification which does what I think it should do with some comments.  If you see issues, I'd love it if you could let me have it. :)

Taking a step back, the purpose of this is in situations where we need to do things like capture a trailing window of state.  In the instance of Median Absolute Deviation it is required because we need not just the distribution of the values, but also the distribution of the DEVIATION of the value from the median (which is the median of the values of the trailing window).  That window can be updated upon tick.  So, for instance, if we want 3 hours lookback, at hour 4, we'd have hours 1, 2 and 3.  At hour 5, we'd want hours 2, 3, and 4 and so on.  Constructing this trailing window lead me to this.  I could have captured this state internally in memory, but that had the issue that internal state for the window would be removed for the entire trailing window upon worker crash/restart.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -243,8 +244,18 @@ private void flush(Tuple tickTuple) {
     measurement.setValue(result);
     emit(measurement, tickTuple);
 
+    // Execute the update with the old state
+    Map<String, String> tickUpdate = profileConfig.getTickUpdate();
+    Map<String, Object> state = executor.getState();
+    if(tickUpdate != null) {
+      tickUpdate.forEach((var, expr) -> executor.assign(var, expr, executor.getState()));
```
nickwallen: The 3rd argument to executor.assign is called 'transientState'.  It exists only to make additional state available to the expression;  additional state that is not already maintained by the executor itself.  The normal use for it is when you want to make the fields in a telemetry message available to the expression.

Setting transientState=executor.getState() does not make sense here.  It is exactly the same as passing nothing at all.  The state in executor.getState() is already exposed to the expression.

Hopefully, I am describing this well enough.  As another feeble attempt at explanation, the following two expression are equivalent.

```
tickUpdate.forEach((var, expr) -> executor.assign(var, expr, executor.getState()));
```
```
tickUpdate.forEach((var, expr) -> executor.assign(var, expr, null));
```

cestella: Ok, I'm confused.  When we do the update on a per-tuple basis, `executor.assign` is called on every expression that will be updated in the config.  Let's say my expression is `"s" : "STATS_ADD(s, value)"`  Now, obviously, within context there is data from my message (`value`) and message from the existing state (the `s` variable).  In the case of a tick update, I'd like to have available just the variables from the existing state (e.g. the `s` variable in my example above).  How would I go about accomplishing that?  I assumed that `s` would be held in the `executor.getState()`, but maybe not.

nickwallen: What you had previously, before this latest commit, does exactly that.

nickwallen: This is what you had in a prior iteration that does what you are describing.

```
profileConfig.getTickUpdate().forEach((var, expr) -> executor.assign(var, expr, new HashMap<>()));
```

The only change that i suggested previously was to make the `result` available to the `tick` expressions, which I think could be generally useful.

```
profileConfig.getTickUpdate().forEach((var, expr) -> executor.assign(var, expr, Collections.singletonMap("result", result));
```

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -243,8 +244,18 @@ private void flush(Tuple tickTuple) {
...
+    }
     // clear the execution state to prepare for the next window
     executor.clearState();
+    //make sure that we bring along the update state
+    if(tickUpdate != null) {
+      tickUpdate.forEach((var, expr) -> executor.getState().put(var, state.get(var)));
```
nickwallen: Ok.  So we are taking on the risk of premature worker death.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
cestella: and this is after the state is cleared, right?

nickwallen: The code above only covers what you describe in this thread; executing 'tick' expressions.  

You still need additional logic to persist the 'tick' state beyond the flush.  And the way that you are doing that in the current code looks fine to me.

cestella: Awesome, updated.  Look ok now?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
cestella: We are taking on risk of premature worker death for one tick, just like we are currently in the profiler.  If the worker gets respawned, it will rerun init, which should initialize the state appropriately (with the lookback window intact, but with the current tick's work of data lost).  In the case of MAD, init and tickUpdate are identical sets of expressions.

We really should make sure we spell out in the docs that worker death will cause state that has not been persisted to be lost.  I'm not sure if it is or not, honestly.

cestella:
I worked out an example in vagrant.  Verified that the statistical and profiler functions can be called from all the relevant places:
* The REPL
* The parser topology
* The enrichment topology

To test this yourself, please check out the README under `metron-analytics/metron-statistics`

metron-analytics/metron-statistics/README.md
```diff
@@ -0,0 +1,346 @@
...
+{
+  "index": "mad",
+  "batchSize": 1,
+  "enrichment": {
+    "fieldMap": {
+      "stellar" : {
+        "config" : {
+          "parser_score" : "OUTLIER_MAD_SCORE(OUTLIER_MAD_STATE_MERGE(
+PROFILE_GET( 'sketchy_mad', 'global', 10, 'MINUTES') ), value)"
+         ,"is_alert" : "if parser_score > 3.5 then true else is_alert"
```
james-sirota: Can 3.5 be pulled out into a hyper parameter?

metron-analytics/metron-statistics/README.md
```diff
@@ -0,0 +1,346 @@
...
+      }
+    }
+  ,"fieldToTypeMap": { }
+  },
+  "threatIntel": {
+    "fieldMap": { },
+    "fieldToTypeMap": { },
+    "triageConfig" : {
+      "riskLevelRules" : {
+        "parser_score > 3.5" : 10
```
james-sirota: Slick integration of model and triage function

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/outlier/MedianAbsoluteDeviationFunctions.java
```diff
@@ -0,0 +1,193 @@
...
+           ,"scale? - Optionally the scale to use when computing the modified z-score.  Default is 0.6745, see the first page of http://web.ipac.caltech.edu/staff/fmasci/home/astro_refs/BetterThanMAD.pdf"
+            }
+          ,description="Get the modified z-score normalized by the MAD: scale * | x_i - median(X) | / MAD.  See the first page of http://web.ipac.caltech.edu/staff/fmasci/home/astro_refs/BetterThanMAD.pdf"
+          ,returns="The modified z-score."
+  )
+  public static class Score implements StellarFunction{
+
+    @Override
+    public Object apply(List<Object> args, Context context) throws ParseException {
+      double scale = 0.6745;
```
james-sirota: this is 1/b from the caltech paper

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/outlier/MedianAbsoluteDeviationFunctions.java
```diff
@@ -0,0 +1,193 @@
+/*
```
james-sirota: We need to file a follow-on Jira to add some caching to this

james-sirota:
+ 1. i got it to work.  one point of feedback, though.  a lot of telemetry data distributions tend to be heavy in the tails.  not all are normally distributed.  i would augment your Gaussian test data to also include something from a heavy-tailed distribution of your choice..... https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.standard_t.html maybe?

I know your method will work with both because I have experience using it in production with real data.  but i think the tests should reflect it as well.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/outlier/MedianAbsoluteDeviationFunctions.java
cestella: Agreed; given the structure, I think the profiler client itself should have a LRU cache that `PROFILE_GET` should pull from before hitting HBase.  I'll make the JIRA.

cestella:
@james-sirota Thanks for running it up!  Agreed about the long tailed distributions; MAD should have interpretations for non-normally distributed data, but it we could stand generating a t-distribution in the unit tests for sure.

metron-analytics/metron-statistics/README.md
cestella: Because the output of MAD might shift over time (if distribution changes), I thought it best to externalize it as a z-score rather than bake it into state capture since the two really should be independent.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/outlier/MedianAbsoluteDeviationFunctions.java
cestella: https://issues.apache.org/jira/browse/METRON-579

PR #354, opened 2016-11-10
METRON-557 Create Stellar Functions for Kafka

nickwallen:
## [METRON-557](https://issues.apache.org/jira/browse/METRON-557)

Add functions to read and write messages to Kafka topics.  This should enable a user to debug a Grok expression or Enrichment configuration inside the REPL. 

For example, I can pull a message off of the error topic to see why it failed to parse. Then fix the configuration and resubmit the message to try again. I also won’t have to go outside of the Stellar REPL to monitor topic activity.

### `KAFKA_GET`
 
Retrieves messages from a Kafka topic.  Subsequent calls will continue retrieving messages sequentially from the original offset.
 
Example: Retrieve one message from a topic.
```
KAFKA_GET('topic')
```
Example: Retrieve 10 messages from a topic.
```
 KAFKA_GET('topic', 10)
```
Example: Retrieve the first message in a topic.  This must be the first retrieval from the topic, otherwise the messages will be retrieved starting from the previously stored consumer offset.
```
KAFKA_GET('topic', 1, { "auto.offset.reset": "earliest" })
```

### `KAFKA_TAIL`

Retrieves messages from a Kafka topic always starting with the most recent message first.

Example: Retrieve the latest message from a topic.
 ```
KAFKA_TAIL('topic')
```
Example: Retrieve 10 messages from a topic starting with the latest.
 ```
KAFKA_TAIL('topic', 10)
```

### `KAFKA_PUT`

Sends messages to a Kafka topic.

Example: Put two messages on the topic 'topic'.
```
 KAFKA_PUT('topic', ["message1", "message2"])
```
Example: Put a message on a topic and also define an alternative Kafka broker.
```
KAFKA_PUT('topic', ["message1"], { "bootstrap.servers": "kafka-broker-1:6667" })
```

### `KAFKA_PROPS`

Retrieves the Kafka properties that are used by other KAFKA_* functions like KAFKA_GET and KAFKA_PUT.  The Kafka properties are compiled from a set of default properties, the global properties, and any overrides.

Example: Retrieve the current Kafka properties.
```
KAFKA_PROPS()
```
Example: Retrieve the current Kafka properties taking into account a set of overrides.
```
KAFKA_PROPS({ "max.poll.records": 1 })
```

## Testing

These were tested on a local Stellar Shell in addition to running within the "Quick Dev" environment.  Note that in the "Quick Dev" environment, you must manually copy the jar built within the Management project (`metron-platform/metron-management/target/...`) to `/usr/metron/0.2.1BETA/lib`.

### DIFF
--- a/metron-platform/metron-common/README.md
+++ b/metron-platform/metron-common/README.md
@@ -1,4 +1,15 @@
-#Stellar Language
+
+
+# Contents
+
+* [Stellar Language](#stellar-language)
+    * [Stellar Language Keywords](#stellar-language-keywords)
+    * [Stellar Core Functions](#stellar-core-functions)
+    * [Stellar Shell](#stellar-shell)
+* [Global Configuration](#global-configuration)
+* [Management Utility](#management-utility)
+
+# Stellar Language
 
 For a variety of components (threat intelligence triage and field
 transformations) we have the need to do simple computation and
@@ -16,7 +27,7 @@ The query language supports the following:
 * The ability to have parenthesis to make order of operations explicit
 * User defined functions
 
-##Stellar Language Keywords
+## Stellar Language Keywords
 The following keywords need to be single quote escaped in order to be used in Stellar expressions:
 
 |              |              |           |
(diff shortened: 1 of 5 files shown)


ottobackwards:
Is there something in STELLAR to write to disk?  
How about a KAFKA_TEE - output results + log to disk?
Come to think of it, maybe the REPL should have the ability to log everything to disk?

cestella:
There are functions to read and write from local disk and HDFS [here](https://github.com/apache/incubator-metron/tree/master/metron-platform/metron-management#file-functions).  The REPL does allow you to log to disk because it uses a proper GNU readline implementation.  You can set it (and a number of other options) by modifying your ~/.inputrc.  [Here](http://aeshell.github.io/docs/settings/index.html) are the full options supported.

cestella:
Yeah, I dig this a lot, @nickwallen +1 by inspection.

PR #355, opened 2016-11-11
METRON-564: Mpack shows UI error when changing configuration

justinleet:


justinleet:
Removed duplicate method and moved some storm relevant items into the 'if' block so that configs weren't used if they weren't available (i.e. during config changes rather than install).

Tested by running up per: https://www.evernote.com/shard/s530/sh/c5551fbd-0ac1-4861-89ce-9c5e37065c52/b13e05f39eaac1a6

After installation, I changed the parser list and the error no longer appears.  Correct parsers still spun up

dlyle65535:
+1, thanks @justinleet!

PR #356, opened 2016-11-11
METRON-563: Ambari Metron service uses incorrect port for installing Elasticsearch templates

justinleet:


justinleet:
PR splits the es_url into 3 fields: es_hosts, es_binary_port, and es_http_port.  The two port fields are defaulted to ES's defaults, so only hosts needs to be provided by the user on install.

As a note, only the first entry in es_hosts is used for the installed templates. It was already broken in the case of multiple entries anyway (because it tried to use the old es_url as a whole, but could be a comma delimited list), and I didn't want to put in logic around iterating through servers until one is found and/or retries etc.  It seems like a reasonable requirement to have working servers for something that should be done pretty soon after install.

To test this, I spun up an Ambari cluster as detailed https://www.evernote.com/shard/s530/sh/c5551fbd-0ac1-4861-89ce-9c5e37065c52/b13e05f39eaac1a6, along with a Docker instance of Elasticsearch to hit. The new configs all show up in the UI and the ports are populated correctly.  The service action to install templates works, and data passes through to indexing using the Bro topology.

dlyle65535:
+1 great work, @justinleet

PR #357, opened 2016-11-11
METRON-566: Add Casey's Key to the KEYS file

cestella:
As the release manager, Casey Stella's code signing key must be added to the KEYS file.


dlyle65535:
+1

PR #358, opened 2016-11-13
METRON-517: Update elasticsearch bro templates for uri

JonZeolla:
## Problem

[METRON-517](https://issues.apache.org/jira/browse/METRON-517)

The bro uri field in [HTTP::Info](https://www.bro.org/sphinx/scripts/base/protocols/http/main.bro.html#type-HTTP::Info) can grow to a size which fails to insert the message into Elasticsearch.  The related error message is:
`IllegalArgumentException[Document contains at least one immense term in field=\"uri\" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped...`

## Short-term Solution

Set the elasticsearch template to [ignore the URI field](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html) if it is greater than 10922 (32776 / 3) bytes, as [RFC 3986](https://tools.ietf.org/html/rfc3986) allows UTF-8 characters to be used in URIs.  32766 is a size limitation built into Lucene ([MAX_TERM_LENGTH](https://lucene.apache.org/core/6_2_1/core/constant-values.html#org.apache.lucene.index.IndexWriter.MAX_TERM_LENGTH)), and each UTF-8 character can be at most 4 bytes, but are predominantly [3 bytes](https://en.wikipedia.org/wiki/CJK_characters).

## Long-term Solution

Currently being tracked via [METRON-542](https://issues.apache.org/jira/browse/METRON-542)

## Testing

Inserted this into my bare metal cluster via `curl --data "@/root/incubator-metron/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template" -XPUT server5:9200/_template/bro_index`

Did some kibana queries and cluster monitoring on my bare metal cluster.

### DIFF
--- a/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template
+++ b/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template
@@ -132,7 +132,7 @@
         "uri": {
           "type": "string",
           "index": "not_analyzed",
-          "ignore_above": 10922
+          "ignore_above": 8191
         },
         "uid": {
           "type": "string",
--- a/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template
+++ b/metron-deployment/roles/metron_elasticsearch_templates/files/es_templates/bro_index.template
@@ -131,7 +131,8 @@
         },
         "uri": {
           "type": "string",
-          "index": "not_analyzed"
+          "index": "not_analyzed",
+          "ignore_above": 10922
         },
         "uid": {
           "type": "string",


cestella:
I am +1 here under a condition.  I would like to point out that despite what the ES documentation states, UTF-8 characters can occupy 4 bytes, so I'd ask that we move the limit to `32766/4 = 8191`  Looking at the lucene source, I do not think they're doing anything to limit the characters to 3 bytes.

JonZeolla:
Right, I did take a look to see how realistic it would be for 4 byte characters to be used in real world URIs and it didn't look very concerning, which is why I went with 10922 instead of 8191.  When I evaluated using 8191 I looked through 1 week of my http logs and found numerous legitimate URI fields which exceeded that, but none that exceeded 10922 (not that the latter doesn't happen, it's just much more rare).

Specifically I considered that legitimate URIs will usually contain CJK characters along with the standard URI delimiters.  Because CJK characters only use up to 3 bytes (afaict), non-CJK characters are only rarely used (i.e. if enough of them are used that this problem is encountered, it is probably nefarious, but definitely an extremely rare event), as well as the fact that 8191 seems to truncate some legitimate URI fields, I went with 10922.

Of course, if the project is looking to drop its BETA flag before [METRON-542](https://issues.apache.org/jira/browse/METRON-542) (which would truncate based on URI length upstream from indexing and have other mitigations in place) has been implemented, I would agree that this should be 8191 for now.

If you'd still prefer 8191 after considering my previous comments, I'd be more than happy to swap it out, but I wanted to make sure to note that it may prevent a higher percent of legitimate URI fields from being indexed.

cestella:
We are planning on dropping the BETA for 0.3.0 for the release being cut today.  I think I'm ok with defaulting to the safer number (8191) and letting the user adjust in the template if they choose to not want that.

PR #359, opened 2016-11-15
METRON-569: Change acking to prevent duplicate tuples in enrichment topology

DomenicPuzio:
Adding this ack statement prevents duplicate messages from being sent through the _enrichment_ topology. Previously, in the _JoinBolt_, when a message does not complete a join, it does not get acked - this cause the message to get re-sent through the topology, resulting in duplicate messages being indexed.

This is a show-stopping bug, impacting performance and resulting in duplicate data.

I have tested this in the Quick-Dev enrichment as well as in our production system at high velocity.

### DIFF
--- a/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/JoinBolt.java
+++ b/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/JoinBolt.java
@@ -118,6 +118,7 @@ public abstract class JoinBolt<V> extends ConfiguredEnrichmentBolt {
         LOG.trace("Emitted message for key: {}", key);
       } else {
         cache.put(key, streamMessageMap);
+        collector.ack(tuple);
         if(LOG.isDebugEnabled()) {
           LOG.debug(getClass().getSimpleName() + ": Missed joining portions for "+ key + ". Expected " + Joiner.on(",").join(streamIds)
                   + " != " + Joiner.on(",").join(streamMessageKeys)


cestella:
How does this result in duplicate data?  Either the message completes a join and it isn't failed or it does not complete the join, does not pass past the join bolt and is replayed from the last anchor (which I think is the split bolt).

The rationale for not acking there is to have the message replay under circumstances where one of the enrichments is experiencing transient failure.  A failure in the join would result in a replay of the message and a retry.

dlyle65535:
@DomenicPuzio - any chance you can introduce a unit or integration test that exhibits the behavior you're fixing here?

cestella:
I do want to point out that it's really appreciated that you submitted a PR.  I have seen potential bugs similar to this one that turned out to be duplicate data being sent to kafka.  I just want to rule that out.

DomenicPuzio:
Here's what I was seeing:

We were running only one Enrichment. So our EnrichmentSplitter sends _originalTuple_ to the Join and _originalTuple_ to the Enrichment. Enrichment then sends _enrichedTuple_ to Join. Join knows that there are two streams, so it waits for two tuples to join together.

When _originalTuple_ comes in first, we put it in the cache and do nothing. When _enrichedTuple_ comes in, we see that it has its mate in the cache, we join _originalTuple_ and _enrichedTuple_, and then we ack **the current tuple**, in this case, _enrichedTuple_. This means that we have never acked _originalTuple_, which caused Storm to re-send that tuple unnecessarily from the EnrichmentSplitter.

Since none of the cached messages are ever acked, Storm does not know that they have been correctly handled. Adding in the ack statement completely resolved this for me.

Does this clarify?

And @dlyle65535, I'm not sure on the best way to do that. Now that I've explained the problem, can you provide some guidance on what you'd like to see?

mmiklavc:
Looking into this further, but I imagine that we do probably want to ack all tuples in the join bolt regardless of them being anchored or not. From what you're describing it looks like this would be an issue even in happy path scenarios without enrichments or workers failing. It's not clear to me how this would be possible and not see this in our testing the other numerous streams.

cestella:
@mmiklavc I completely agree that we should ack every tuple that comes through in the `JoinBolt`.  I think, however, that they should be acked only on a successful join.  

The failure scenarios are currently handled thusly:
* A failure in an enrichment (e.g. exception, hbase or mysql going down, etc) is handled by the enrichment adapter being robust enough to send an empty message even on failure
* A failure in an enrichment bolt's infrastructure (e.g. the process dying or a machine failure) should be handled by the join failing and the original message replaying.  We anchor the message coming on the `message` channel in the `SplitBolt`, so it should get replayed.

If we ack every message as they come down the pipe, we could be in a situation where we've acked a message that never joins and never gets through.  Instead, I think we should ack every message on a successful join.

As of now, the tuple that sends the last enrichment segment in will get acked, which is definitely wrong.  What I would like to see is for the `streamMessageMap` to get augmented to track the tuples as well as the messages and for a successful join to result in all of the tuples in the map to get ack'd.  Possibly it's time for that map to actually be a proper class and it really only has to keep track of the tuple from the `message` stream.

Thoughts?

cestella:
Also, @DomenicPuzio would you mind renaming this PR to METRON-569: Change acking to prevent duplicate tuples in enrichment topology so that the PR comments get replicated properly?

DomenicPuzio:
@cestella, I made the change to the PR name; thanks for the tip there.

I completely agree that we don't want to miss out on catching failures due to an enrichment source (like MySQL) or in the enrichment infrastructure. However, after we have put a tuple into the JoinBolt's cache, isn't it already past the point where these pitfalls could occur? If there is a failure in an enrichment, then Storm will time out while waiting for an ack that never takes place, so this message will be re-sent; but if the message is already in the cache, isn't its journey complete?

My thought with placing the ack there was (1) at this stage of the topology, that tuple has been correctly processed, and (2) for simplicity's sake so that we wouldn't have to modify `streamMessageMap`. I do agree that acking everything after the join has taken place also makes a lot of sense, and I can work on that if you would like.

We saw the duplicate data while running this in our development environment in EC2. Perhaps this is due to different ack timeout settings in the Storm topologies? We've repeated the duplication of data many times on our end.

cestella:
Hey @DomenicPuzio thanks man :)

The problem with placing the ack there is that in the situation where the enrichment adapter worker gets killed, we would *like* the data to replay from the anchor point (the SplitBolt) because Storm will have respawned the worker elsewhere and we'd like that enrichment to be captured.  If you keep the ack where you have it, the message will never join if the enrichment worker gets killed, so the message will ultimately get dropped, which isn't ideal.

Ultimately, I think we can all agree that we need to ack at LEAST the tuple on the `message` stream, but I think that we want to ack it AFTER the join has happened.

Regarding seeing dup data, I believe you.  I am just trying to wrap my head around how it happens, but, as they say, the proof of the pudding is in the eating. :)

cestella:
It occurs to me that what may be happening is that an enrichment may be taking longer than the timeout that storm is using to wait on that ack.  If that is the case, I could see duplicated data.

Imagine the following situation, with the storm timeout `x` and an enrichment taking `x + 1`.  The enrichment would finish and send the enriched data from the enrichment adapter to the join bolt but storm would've already triggered a replay.  The enrichment completing would have triggered the join to happen and the joined message to be emitted and the replay would trigger another copy of the message.

In this case, I'd suggest ensuring that either your enrichment is capped at maximum http://storm.apache.org/releases/current/javadocs/org/apache/storm/Config.html#TOPOLOGY_MESSAGE_TIMEOUT_SECS or adjusting the message timeout in storm to be higher than this setting in storm.

DomenicPuzio:
Sorry for the delayed response, @cestella - I was taking some time off for the holiday.

While I don't believe custom enrichment is pushing us over the timeout (we've clocked it at 15ms at the worst), that's certainly a possibility and something that we should also guard against. With this in mind, we now have two tasks to look into:
1. Ensuring all messages get acked after a successful join by building out the `streamMessageMap`
2. Making sure that our cache refresh is set to some fraction of the Storm timeout interval

Are these both items that we would like to complete? Should we split these into separate JIRAs and PRs? Should I work on the acking changes within this PR?

cestella:
Yep, so my personal suggestion is to make sure that all messages get acked on a successful join by augmenting `streamMessageMap` and ensure that the cache refresh is *longer* than the storm timeout (if it is shorter, then it'll never join because the partial messages will timeout before the final message comes through).

Anyone else have thoughts?

cestella:
I think #603 should address this now.

merrimanr:
@DomenicPuzio, this has been resolved by https://github.com/apache/metron/pull/603.  Can we close this PR?

DomenicPuzio:
Yep, we can wrap it up! Thanks, all!

PR #360, opened 2016-11-16
METRON-567 enrichment types assume String values for CacheKeys

ottobackwards:
The CacheKeys store values as objects, and support gets as object, straight casting as passed .class or coercion to a class.

The enrichment adapters assume that the values in the CacheKeys are Strings, and thus pass String.class for lookups and retrievals.

It is possible to have non-string values for these fields, and when that happens enrichment fails with exceptions.

To address this issue, I am replacing calls to getValue(Class) with coerceValue(Class)

Tested - added failing unit tests for changed classes, addressed issue and retest
Integration Tests


cestella:
+1 by inspection; I like it!

cestella:
Just for posterity and to provide a tactical workaround in this case.  You can use stellar field transformations to coerce the whatever field into a string at the parser.  If your field is named `username`, then the parser config would look like:
```
{
  ...
"fieldTransformations" : [
    {
      "transformation" : "STELLAR"
    ,"output" : [ "username" ]
    ,"config" : {
      "username" : "TO_STRING(username)"
                     }
    }
                                       ]
}
```

PR #349, opened 2016-11-09
METRON-558 use posix for gnu tar with maven assembly plugin

ottobackwards:
This allows it to handle large group id's and resolves the following error

[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 09:36 min
[INFO] Finished at: 2016-11-09T10:13:17-05:00
[INFO] Final Memory: 177M/1916M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.6:single (build-tarball) on project metron_mpack: Execution build-tarball of goal org.apache.maven.plugins:maven-assembly-plugin:2.6:single failed: group id '397160923' is too big ( > 2097151 ). Use STAR or POSIX extensions to overcome this limit -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.

The typical answer is to configure the plugin to use posix for tarLongFileMode.

PR #361, opened 2016-11-17
METRON-630 Add tarLongFileMode posix to maven-assembly-plugin

atopian:
This fixes up some errant long UID messages that one gets when doing a full build from source on OSX (10.12)

### DIFF
--- a/metron-analytics/metron-maas-service/pom.xml
+++ b/metron-analytics/metron-maas-service/pom.xml
@@ -318,6 +318,7 @@
         <artifactId>maven-assembly-plugin</artifactId>
         <configuration>
           <descriptor>src/main/assembly/assembly.xml</descriptor>
+          <tarLongFileMode>posix</tarLongFileMode>
         </configuration>
         <executions>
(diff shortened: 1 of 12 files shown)


ottobackwards:
https://github.com/apache/incubator-metron/pull/349

This is a dupe

atopian:
#349 was a partial fix, but there were many, many other places that would error out when doing a `mvn clean package`.  This catches those other locations.

ottobackwards:
OK - so the previous PR resolved the issue completely on my managed macbook so I can't test if this 'fixes' the issue - but I have built this on that machine without issue, and we should be consistent through the projects.

+1

ottobackwards:
Can you create a JIRA for this issue?  Then change the title of the PR to begin with METRON-XXXX ( corresponding to issue )?

ottobackwards:
I have created METRON-630 for this.  Please change the title of the pull request.

ottobackwards:
Thanks for the contribution

PR #362, opened 2016-11-18
METRON-575 State from different profiles can be co-mingled incorrectly

nickwallen:
Added a keyed cache to the ProfileBuilderBolt so that it can maintain state for multiple profiles.  Previously, the same StellarExecutor was being used across multiple different Profile/Entity pairs, which was incorrect.

The cache is also used to define a finite lifetime for a profile.  A new property was added to allow the user to control the lifetime of a profile.

> `profiler.profile.ttl`: If a message has not been applied to a Profile in this period of time, the Profile will be forgotten and its resources will be cleaned up. This value should be defined along with profiler.ttl.units.


mattf-horton:
I would recommend using ordinary time units (seconds, minutes, hours, days) rather than a number of periods, for the TTL.  This puts the lifetime info in one place instead of two, and prevents questions like "I changed the period, does the TTL change too?"

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -20,7 +20,18 @@
 
 package org.apache.metron.profiler.bolt;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.sun.tools.javac.jvm.Profile;
```
cestella: Did you mean to include this?

nickwallen:
There are a couple of constraints on this parameter. (1) The profile TTL must be greater than the period duration.  (2) It doesn't make sense for a profile to be invalidated part way through a period.  I could have made the profile TTL specified by time units, but then I would have to manually validate these constraints.  My thought was that if I allow the user to specify the number of periods, then the constraints are mostly self-validating.

I also think, to some degree, it makes sense for the profile TTL to scale with the period duration.  But I am not sure I've fully convinced myself of this yet.

Still think time units are better?

mattf-horton:
@nickwallen  Seems to me using an actual time rather than a number of periods will be more intuitive for the user.  Does the user really want to say "If it's been more than 12 periods..." or "If it's been more than 3 hours..."?

Regarding validation, define the expiration to occur at the end of the period during which the expiration constraint becomes true, with the clarification that it will always run for at least one full period.  So all you have to validate is that the TTL is positive.

Anyone else have an opinion?

nickwallen:
It would be good to get other's opinion, but I could use time units, since you think that is more intuitive.  I'm too close to the code to really know what's more intuitive.  I trust you more than me at this point. ;)

cestella:
So, I agree that time units are more intuitive.  I would say that you should ensure that the TTL is strictly larger than the tick duration.  In the event that it does, it should just reset to 1.5 periods IMO.  I'd also be ok with the topology failing and the user being notified by a stacktrace, an error message, a carrier pigeon or a candygram.

nickwallen:
Thanks for the review guys.  Changed profile TTL to use time units.  Also fixed that errant import.

nickwallen:
Bump...

cestella:
I'm in.  Just one question, will this work ok with the modifications that I made in the outlier functions PR?  I can't think of a reason why it *wouldn't* but I'd like to just confirm.

nickwallen:
No problem with `tickUpdate`.  I already merged your changes in.

cestella:
Ok, +1 on inspection.  You smoke tested this on vagrant, right?  Great catch, @nickwallen

nickwallen:
Yes, I ran up on Quick Dev.  I'll do that once more just to make sure there are no problems with merging in the `tickUpdate` stuff.  

We should probably add a profiler integration test that includes a use of`tickUpdate` at some point.

cestella:
Agreed, better testing around that change could've happened.

nickwallen:
Everything is working in Quick Dev with this change.

cestella:
+1 stands

PR #363, opened 2016-11-21
METRON-578: Missing error handling bolts for enrichment and threat intel

justinleet:
In the flux files, added outputs for the various `GenericEnrichmentBolt` used throughout, to tie their error output streams to Kafka topics.  Specifically, enrichment is tied to enrichments_error and threat intel is tied to threatintel_error.

Quick-dev and the Ambari mpack are updated to ensure the topics are created, so that no errors occur on usage.  Both were run up to ensure that the topics were created and the flux files don't give errors.

To test, an adjustment was made to the `EnrichmentIntegrationTest` to have an enrichment bolt that always throws errors.  This bolt is derived from `GenericEnrichmentBolt` (which has been slightly refactored) to ensure that it uses the same error logic, but ignores the actual logic that leads to errors.  This was done because I couldn't find input that resulted in the error stream code path in actual practice.  Other errors can result, but don't go to the error stream (e.g. JSON decoding errors).

As a note, the `ErrorEnrichmentBolt` invariably ends up in the stack traces being printed (which is not handled by `ErrorEnrichmentBolt` or `GenericEnrichmentBolt`).

Given that the test explicitly calls `UnitTestHelper.verboseLogging()`, I'm not sure how to suppress that without killing actual errors at the same time is.  Open to suggestions if someone knows a more appropriate way to handle it (if we care).

### DIFF
--- a/metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
+++ b/metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
@@ -166,20 +166,14 @@ public class EnrichmentIntegrationTest extends BaseIntegrationTest {
 
       kafkaComponent.writeMessages(Constants.ENRICHMENT_TOPIC, inputMessages);
       ProcessorResult<List<Map<String, Object>>> result = runner.process(getProcessor(inputMessages));
+      // We expect failures, so we don't care if result returned failure or not
       List<Map<String, Object>> docs = result.getResult();
-      if (result.failed()){
-        StringBuffer buffer = new StringBuffer();
-        result.getBadResults(buffer);
-        buffer.append(String.format("%d Valid Messages Processed", docs.size())).append("\n");
-        dumpParsedMessages(docs,buffer);
-        Assert.fail(buffer.toString());
-      } else {
-        Assert.assertEquals(inputMessages.size(), docs.size());
-        List<Map<String, Object>> cleanedDocs = docs;
-        validateAll(cleanedDocs);
+      Assert.assertEquals(inputMessages.size(), docs.size());
+      validateAll(docs);
 
-        validateErrors(inputMessages, kafkaComponent);
-      }
+      List<byte[]> errors = result.getProcessErrors();
+      Assert.assertEquals(inputMessages.size(), errors.size());
+      validateErrors(result.getProcessErrors());
     } finally {
       runner.stop();
     }
@@ -204,12 +198,10 @@ public class EnrichmentIntegrationTest extends BaseIntegrationTest {
     }
   }
 
-  protected void validateErrors(List<byte[]> inputMessages, KafkaWithZKComponent kafkaComponent) {
-    List<byte[]> enrichmentErrors = kafkaComponent.readMessages(Constants.ENRICHMENT_ERROR_TOPIC);
-    Assert.assertEquals("Should have one error per message passed to enrichment", inputMessages.size(), enrichmentErrors.size());
-    for(byte[] enrichmentError : enrichmentErrors) {
+  protected void validateErrors(List<byte[]> errors) {
+    for(byte[] error : errors) {
       // Don't reconstruct the entire message, just ensure it contains the known error message inside.
-      Assert.assertTrue(new String(enrichmentError).contains(ErrorEnrichmentBolt.TEST_ERROR_MESSAGE));
+      Assert.assertTrue(new String(error).contains(ErrorEnrichmentBolt.TEST_ERROR_MESSAGE));
     }
   }
 
@@ -462,6 +454,7 @@ public class EnrichmentIntegrationTest extends BaseIntegrationTest {
       public ReadinessState process(ComponentRunner runner) {
         KafkaComponent kafkaComponent = runner.getComponent("kafka", KafkaComponent.class);
         List<byte[]> messages = kafkaComponent.readMessages(Constants.INDEXING_TOPIC);
+        errors = kafkaComponent.readMessages(Constants.ENRICHMENT_ERROR_TOPIC);
         if (messages.size() == inputMessages.size()) {
           docs = new ArrayList<>();
           for(byte[] message : messages) {
@@ -473,12 +466,6 @@ public class EnrichmentIntegrationTest extends BaseIntegrationTest {
           }
           return ReadinessState.READY;
         } else {
-          errors = kafkaComponent.readMessages(Constants.ERROR_STREAM);
-          invalids = kafkaComponent.readMessages(Constants.INVALID_STREAM);
-          if(errors.size() > 0 || invalids.size() > 0) {
-            messages = messages;
-            return ReadinessState.READY;
-          }
           return ReadinessState.NOT_READY;
         }
       }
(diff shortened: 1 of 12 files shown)


metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
```diff
@@ -196,6 +201,15 @@ public static void validateAll(List<Map<String, Object>> docs) {
     }
   }
 
+  protected void validateErrors(List<byte[]> inputMessages, KafkaWithZKComponent kafkaComponent) {
```
ottobackwards: I changed getProcessor() to attempt to read the ERROR and INVALID streams from Kafka already in a past PR to improve integration test error handling.  If they need to read from the bolts that you have, maybe you can change it there, and then pass the  results into here and not have to read them

justinleet: Thanks, I'll take a look at that.

ottobackwards: This should be true for all integration tests

justinleet:
Updated to use the processor more directly.  Fixed an issue where the processor previously attempted to read topics from Kafka that had the same name as the bolt's stream, rather than the corresponding topic name.  Changed the logic, so that the result having errors does not fail the test (because we expect a given failure set along with the successful results).

justinleet:
Making Travis recognize the new commit

justinleet:
Updated to fix conflicts brought on by https://github.com/apache/incubator-metron/pull/344

justinleet:
@ottobackwards That refactoring look like what you were looking for?

ottobackwards:
Sorry @justinleet I'll check that now

ottobackwards:
yes, that looks consistent, if you can think of an improvement however, I'm all ears.
I do have a question - but I'll put it inline

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
```diff
@@ -163,18 +166,14 @@ public void test() throws Exception {
...
-        result.getBadResults(buffer);
-        buffer.append(String.format("%d Valid Messages Processed", docs.size())).append("\n");
-        dumpParsedMessages(docs,buffer);
-        Assert.fail(buffer.toString());
-      } else {
-        Assert.assertEquals(inputMessages.size(), docs.size());
-        List<Map<String, Object>> cleanedDocs = docs;
-        validateAll(cleanedDocs);
-      }
+      Assert.assertEquals(inputMessages.size(), docs.size());
```
ottobackwards: I don't understand this part...  if you expect errors, why are you asserting that the input and doc size match?  would they match if there were errors?

ottobackwards:
another point with all the process runner stuff.... there is a LOT of repeated code through out the tests.  I might give a thought to how to get some re-usable boilerplate in there

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
justinleet: In the GenericEnrichmentBolt, in the event of an error that leads to the error stream, it still emits in order to not kill the entire processing in the event of one enrichment being broken (code snippet from that class down below). 

Both the regular output and the error stream get emitted, so the end result still has the enriched docs as before I hooked the error enrichment up to the topic (so inputMessages.size() should still be the same as docs.size(), except now there are also (nonfatal) errors we can check in addition to the regular results.

See:
```
    JSONObject enrichedMessage = new JSONObject();
    enrichedMessage.put("adapter." + adapter.getClass().getSimpleName().toLowerCase() + ".begin.ts", "" + System.currentTimeMillis());
    try {
      // Actual logic
      if (enrichedMessage != null && !enrichedMessage.isEmpty()) {
        collector.emit(enrichmentType, new Values(key, enrichedMessage, subGroup));
      }
    } catch (Exception e) {
      LOG.error("[Metron] Unable to enrich message: " + rawMessage, e);
      JSONObject error = ErrorUtils.generateErrorMessage("Enrichment problem: " + rawMessage, e);
      if (key != null) {
        collector.emit(enrichmentType, new Values(key, enrichedMessage, subGroup));
      }
      collector.emit("error", new Values(error));
    }
```

justinleet:
I completely agree about the code duplication (and a lot of dead code like unused fields and params)  throughout the tests, and was thinking about throwing a ticket in for cleaning up some of that code.  I can go ahead and create it if you want, otherwise if you're going to take a look and have some ideas I can let you throw it up and fill it with your thoughts.  Preference?

ottobackwards:
What I'm looking at is the Processor interface implementations and seeing if that can be reduced by providing some same base and derivation ( like the ES and Solr processor code is the same just about ).  I'm not looking as much at cleanup as refactoring to make it easier to do standard testing things.  If I understand what you are looking for, there would be two jiras.

ottobackwards:
I will throw up a jira for that refactoring, if you want to add one for the type of cleanup you are looking at you do that one

justinleet:
Sounds like a plan. Anything else on this PR?

ottobackwards:
METRON-587 if you want to put $.02 in.

ottobackwards:
No - how can I test this?  Just run the tests?

justinleet:
Yep, it's all tied into the integration test.  Travis output of the test from the check is actually at:
https://travis-ci.org/apache/incubator-metron/builds/178028223#L492, and you can see the errors getting kicked out, but the test still passing.

ottobackwards:
ok .  But just to mention it - you are going to have to refactor from master based on the changes I made to the components etc.  can you re-sync master etc?  I'll test that can give you the sign off

justinleet:
You mean from https://github.com/apache/incubator-metron/pull/344? I already fixed the conflicts from it.

I can merge master back in again if you want, but the only commits in master and not my branch appear to be:
METRON-584 STELLAR should have a TO_LONG Function (ottobackwards) closes apache/incubator-metron#368
METRON-575 State from different profiles can be co-mingled incorrectly (nickwallen) closes apache/incubator-metron#362

ottobackwards:
sorry, my mistake
i'll run this right away

ottobackwards:
Justin - if you look at the WriteBoltIntegrationTest in the processor - I had attempted to read from error and invalid queues - I don't think what is in there is correct base on this pr, can you take a peek?

justinleet:
Looks like you could probably do it an easier way, by constructing
```
ProcessorResult.Builder<Map<String,List<JSONObject>>> builder = new ProcessorResult.Builder();
return builder.withResult(messages).build();
```
more similar to 
```
ProcessorResult.Builder<List<Map<String,Object>>> builder = new ProcessorResult.Builder();
return builder.withResult(docs).withProcessErrors(errors).withProcessInvalids(invalids).build();
```

That object already has fields for errors/invalids, and at that point you can grab them separately from the output docs. Should avoid having to have the messages the Map keyed by the topic. 

What's there looks like it works (from a quick look), but using the results builtin error/invalids is going to be a little more explicit about what's happening and probably cleaner.

ottobackwards:
But the topics are correct?

ottobackwards:
+1 reviewed and tests run

justinleet:
Thanks a lot for reviewing this, Otto.  

Regarding the WriterBoltIntegrationTest, the topics looks good.  I had to do a bit of digging to confirm them, because I didn't realize they're kinda hidden in ParserTopologyBuilder, but it looks like ENRICHMENT_TOPIC, DEFAULT_PARSER_INVALID_TOPIC, and DEFAULT_PARSER_ERROR_TOPIC are what you want and what you have.

The reason this PR differs is because enrichment is specifically managed by the test.yaml, so the way components get hooked up is different and required me to define bolts to actually write the errors to Kafka topics.

PR #364, opened 2016-11-21
Metron-580: Remove hard-coded Metron version from Ambari MPack code

mmiklavc:
This PR addresses https://issues.apache.org/jira/browse/METRON-580

The purpose here is to keep us from having to grep for version numbers in the MPack code the next time we bump the Metron version number. The change sets up a "CURRENT" version directory for the mpack code. The Maven Assembly plugin is then used to perform two operations: token replacement in source files, and movement of source files from CURRENT -> METRON VERSION in the final packaging layout.

**Testing**

Follow install instructions here - https://www.evernote.com/shard/s530/sh/c5551fbd-0ac1-4861-89ce-9c5e37065c52/b13e05f39eaac1a6

EDIT: 
[
Please do not do this step -> "git clone https://github.com/justinleet/metron-rpm localrepo". Instead, build the project and MPacks from scratch doing the following:

From Metron home: `mvn clean package -DskipTests`
From metron-deployment: `mvn clean package -Pbuild-rpms && mvn clean package`

Then you'll want to mount the generated rpm's you built by setting the DOCKER_OPTS var
export DOCKER_OPTS="-v ${metron-project-home}/metron-deployment/packaging/docker/rpm-docker/RPMS/noarch:/localrepo"
]

**Note** - if you are using Z shell, there is some wonkiness with the ambari docker stuff. Run this using Bash and you should be ok.
Make sure you start any sensors *last* to ensure the install goes smoothly in a local dev machine. For this change and test, I only started up the bro sensor.

I also modified the sensor-test-mode yaf task to include "ignore_errors: yes" - there appears to be an issue with Ansible wrt OS'es with systemd - https://github.com/ansible/ansible-modules-core/issues/915


mmiklavc:
Some notes for completely removing Metron in the event of a botched install:

```
Uninstall metron
---------------
# stop sensors
service yaf stop && service pcap-replay stop && /usr/local/bro/bin/broctl stop && service snortd stop
# kill topologies
for i in snort yaf bro enrichment indexing; do storm kill $i; done
# enable kafka topic delete - in the Kafka config for your kafka-brokers, set delete.topic.enable=true (otherwise the topics will just sit around)
# delete topics
for i in snort bro yaf enrichments indexing; do ./kafka-topics.sh --zookeeper amb1.service.consul:2181 --delete --topic $i; done
# Remove hbase tables
$ hbase shell
    disable 'enrichment'
    disable 'threatintel'
    drop 'enrichment'
    drop 'threatintel'
# cleanup install:
yum -y erase mysql metron* && rm -rf /usr/metron && rm -fr /var/cache/yum/* && yum clean all
```

metron-deployment/packaging/ambari/metron-mpack/pom.xml
```diff
@@ -60,6 +64,19 @@
...
                     </execution>
+                    <execution>
+                        <id>metron-short-version</id>
+                        <goals>
+                            <goal>regex-property</goal>
+                        </goals>
+                        <configuration>
+                            <name>metron.short.version</name>
+                            <value>${project.parent.version}</value>
+                            <regex>\.</regex>
```
justinleet: Does this still work if we have something like 0.3.0-SNAPSHOT? Or does the regex need to handle things like the '-'?

mmiklavc: Hm, we haven't done a SNAPSHOT release version before. Agreed we need to handle hyphens, for instance. Any idea how Ambari handles it in the service advisor class name? Can we have "SNAPSHOT" in there?

justinleet: Yeah, SNAPSHOT can be in there, barring some craziness.  It used to have 020BETA in the name.  Just needs to line up with the version, and I'm entirely unsure how Ambari handles it unfortunately.

justinleet: @mmiklavc Did you ever have a chance to figure this out?  Given that we haven't agreed on version yet anyway, it may not matter too much, but I would be delighted if "-SNAPSHOT" worked.

mmiklavc: Yes, I will be making that change

mmiklavc:
@justinleet I set this to remove the -SNAPSHOT from the version info in both the short and long form version. 
In going through a testing effort to see what is affected by having "-SNAPSHOT" in jar versions, but not in the mpack stuff, I uncovered some issue with building the RPMs. SNAPSHOT will break RPM builds. I'd like to split off the work of end-to-end testing with a SNAPSHOT version to a separate Jira and PR and allow this one (as a precondition of doing the next Metron version rev) move forward.

justinleet:
@mmiklavc Thanks for taking care of that.  This is good, +1.

PR #365, opened 2016-11-21
METRON-565: apps/metron/enrichment/indexed directory path does not get created for metron cluster deployed via Ambari

mmiklavc:
This PR addresses https://issues.apache.org/jira/browse/METRON-565

Modified the parameters file to point to the correct path for indexing.

**Testing**
https://github.com/apache/incubator-metron/pull/364#issue-190758723

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.3.0/configuration/metron-env.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.3.0/configuration/metron-env.xml
@@ -31,6 +31,12 @@
         <display-name>Metron apps HDFS dir</display-name>
     </property>
     <property>
+        <name>metron_apps_indexed_hdfs_dir</name>
+        <value>{{metron_apps_hdfs_dir}}/indexing/indexed</value>
+        <description>Indexing bolts will write to this HDFS directory</description>
+        <display-name>Metron apps indexed HDFS dir</display-name>
+    </property>
+    <property>
         <name>metron_zookeeper_config_dir</name>
         <value>config/zookeeper</value>
         <description>Metron Zookeeper config dir. Relative path to Metron home.</description>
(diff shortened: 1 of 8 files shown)


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.3.0/configuration/metron-env.xml
```diff
@@ -246,7 +246,7 @@ bolt.hdfs.file.system.url={{ default_fs }}
 bolt.hdfs.wip.file.path=/paloalto/wip
 bolt.hdfs.finished.file.path=/paloalto/rotated
 bolt.hdfs.compression.codec.class=org.apache.hadoop.io.compress.SnappyCodec
-index.hdfs.output={{ metron_apps_enrichment_dir }}
```
justinleet: I agree that this works, but could we do
`index.hdfs.output={{ metron_apps_enrichment_dir }}/indexing`
or even defined in the config, e.g.
`index.hdfs.output={{ metron_indexing_hdfs_dir }}/indexing`

I don't have a particularly strong opinion on constructing it in the params vs metron-env, but we should be using metron_apps_enrichment_dir since we already calculate it in the params anyway.

mmiklavc: But that param/value is hidden from the user right?

justinleet: That is a fair point, since we do expose the metron-env in Ambari. I'm willing to leave it separately, although it does make me wonder if there's a better way to handle configs like that in general.  Having said that, I don't see a good way around it based on how Ambari needs things defined.

justinleet:
The Travis failure appears to be an unrelated profiler integration test error, although I'd like @mmiklavc to double-check.  Assuming it is unrelated, I'm +1, by inspection.  Thanks for picking this up!

james-sirota:
@nickwallen can you take a look and verify?

justinleet:
@mmiklavc Looks like https://github.com/apache/incubator-metron/pull/362 for METRON-575 got pulled in yesterday. Could you either kick this PR to get it to rerun and hopefully run correctly, or just merge in master so that Travis runs as expected?

mmiklavc:
Ended up making a few additional changes here. First, got rid of the /apps/metron/enrichment directory entirely. With indexing being split off into its own topology, this no longer made sense to live there. Furthermore, "indexed" was the only thing that lived in the enrichment directory in hdfs. The indexed doc path will now live in /apps/metron/indexing/indexed to keep consistent with the dir structure we used in the past. In addition, I've now exposed the directory as a property the user can both see and modify in the advanced properties section in Ambari.
image

image

Lastly, I updated the quickdev environment to reflect this change as well.

dlyle65535:
I'm +1 on this, good stuff. @justinleet - you still gtg with the additional changes?

justinleet:
Yep, I'm +1, too.

PR #366, opened 2016-11-21
METRON-576 Stellar function resolution takes too long on running cluster

nickwallen:
### [METRON-576](https://issues.apache.org/jira/browse/METRON-576)

When running the Stellar REPL in a cluster on AWS, function resolution takes 50-60 seconds. The user is not able to execute any functions in the REPL until this process completes.  

The default function resolver searches the entire classpath for Stellar functions.  This includes all library dependencies.  The delay is because there are just too many classes in the classpath to search on a running cluster. As more libraries are added as dependencies under `/usr/metron/0.3.0/lib` this problem just gets worse.

#### Testing

Tested on Quick Dev and showed definite performance improvement.  Before limiting the search path, the search takes roughly 18-20 seconds for me.  After adding a sensible "include" like `org.apache.metron.*`, the search takes roughly 2-3 seconds.

Have not yet deployed on AWS to compare the load time.

#### Changes

* A global property allows the user to specify packages that should be searched for Stellar functions.  `stellar.function.resolver.includes = org.apache.metron.*`  This significantly reduces the delay when waiting for the Stellar REPL to be ready to execute a function. 

* Another global property allows the user to exclude specific packages from being searched for Stellar functions.  `stellar.function.resolver.excludes = org.apache.metron.management.*`

* By default, the entire classpath is searched, which matches the previous behavior.  I'd be open to changing default behavior if there is interest.  

* All of the logic that was previously contained in SingletonFunctionResolver is now broken out across multiple classes.

     * The `BaseFunctionResolver` leverages Guava for lazy initialization and thread-safety, instead of the read/write locking scheme as before.  This is much simpler and I believe is effectively the same.  
     * The `ClasspathFunctionResolver` uses the Reflections API to search the classpath for Stellar functions as before.  This functionality was enhanced to explicitly include or exclude specific packages from function resolution.  
    * The `SingletonFunctionResolver` remains, but is really only a singleton that is used as the default function resolver.  Keeping this separate made testing simpler.
     * The `SimpleFunctionResolver` is a function resolver that allows single classes to be searched for Stellar functions.  This is useful when writing unit tests for Stellar functions.
    ```
    FunctionResolver functionResolver = new SimpleFunctionResolver()
            .withClass(KafkaFunctions.KafkaGet.class)
            .withClass(KafkaFunctions.KafkaPut.class)
            .withClass(KafkaFunctions.KafkaProps.class)
            .withClass(KafkaFunctions.KafkaTail.class);
    ```

* Added unit tests for the function resolvers.
* Removed the `clear()` function from the `FunctionResolver` API.  This was only needed for testing. This was originally only a work-around since it was not easy to swap in another `FunctionResolver` at the time.


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/resolver/ClasspathFunctionResolver.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  @Override
+  public void initialize(Context context) {
+    super.initialize(context);
+
+    // do we have access to the global configuration?
+    if(context != null) {
+
+      Optional<Object> optional = context.getCapability(GLOBAL_CONFIG, false);
```
cestella: While I think it's a sensible thing to pull these from the global config on a running cluster, on the situation where the REPL is used, it is likely that you may want to inject or ignore the includes and excludes from the global config, rather than inherit them.  Can we perhaps not couple function resolution with the global config?  I'd recommend it be a separate capability name called `STELLAR_CONFIG`.  Where we pull that capability from should depend on the context (e.g. in the REPL or MR jobs it should be probably passed in via some external parameters, in the running cluster it'll probably be in the global config or maybe a subsection of the global config like `globalConfig.get("stellarConfig")`.

metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/shell/StellarExecutor.java
```diff
@@ -143,11 +153,13 @@ public StellarExecutor(Console console) throws Exception {
 
   public StellarExecutor(String zookeeperUrl, Console console) throws Exception {
     this.variables = new HashMap<>();
-    this.functionResolver = new StellarFunctions().FUNCTION_RESOLVER();
     this.client = createClient(zookeeperUrl);
     this.context = createContext();
+    StellarFunctions.initialize(this.context);
+    this.functionResolver = StellarFunctions.FUNCTION_RESOLVER();
```
cestella: I'd prefer everyone's REPL to not include the same includes and excludes as the global config, which may be used on the running cluster.  Can we, instead, have the REPL take a JSON config that can set the stellar configs?  I'm ok if we set the default config to the one in global config, but the user should be able to override.

nickwallen: Yes, agreed.  I think that makes sense.  I'll hack on that a bit

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/resolver/BaseFunctionResolver.java
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * The Stellar execution context that can be used to inform the function resolution process.
+   */
+  protected Context context;
+
+  public BaseFunctionResolver() {
+    // memoize provides lazy initialization and thread-safety (the ugly cast is necessary for serialization)
+    functions = Suppliers.memoize((Supplier<Map<String, StellarFunctionInfo>> & Serializable) this::resolveFunctions);
+  }
```
nickwallen: If a function scan is currently running, then competing threads will block when calling `function.get()...`

nickwallen:
Instead of using the global properties, there is now a capability called `STELLAR_CONFIG`.  The includes/excludes are pulled from this.

In the context of the REPL, it attempts to load a file called `stellar.properties` from the classpath.  The values in this file become the `STELLAR_CONFIG`.

If none exists, it uses a default set of properties.  The default properties only include`org.apache.metron.*`.  This hopefully provides a better experience for casual users, but provides advanced users the ability to easily include additional packages in the function resolution process.

cestella:
Great!  One minor thing; could we have the shell pull the properties file from a CLI argument first, then from the classpath and then with the default?

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/resolver/ClasspathFunctionResolver.java
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  @Override
+  public void initialize(Context context) {
+    super.initialize(context);
+    if(context != null) {
+
+      Optional<Object> optional = context.getCapability(STELLAR_CONFIG, false);
+      if (optional.isPresent()) {
+        Map<String, String> stellarConfig = (Map<String, String>) optional.get();
```
cestella: Can we not assume the StellarConfig is a `Map<String, String>`, rather a `Map<String, Object>` (we might want more structure to the map as we go forward or numerical values).

nickwallen: Sure.  Done.

At some point, it might be nice to define the type as part of the capability rather than assuming what the type is when you use it.

nickwallen:
BTW, I also checked the Pcap Query tool to make sure there were no unintended side effects, since that also uses Stellar to some degree.  TL;DR;  I don't see any problems.

On Quick Dev, I started up Pcap Replay and Pycapa.  I then started the pcap topology to land data in HDFS.  I ran a basic query that leveraged Stellar to extract pcap.  I received the expected data and it was correctly filtered.  I was also able to open it up in Wireshark.

```
service pcap-replay start
service pycapa start
bin/start_pcap_topology.sh
bin/pcap_query.sh query -q "ip_src_addr == '192.168.138.158'" -st 500
```

nickwallen:
Kicking

ottobackwards:
Is that failure legit?

nickwallen:
I am trying to figure out why it is failing in Travis.  When I run it locally, it works just fine.

ottobackwards:
I will give it a try

ottobackwards:
were you planning on sync'ing with master?

nickwallen:
Oops.  I thought I had pushed that out.  Maybe that is the problem.  Thanks

ottobackwards:
worked locally for me as well.

nickwallen:
Needed sync'd with master.  Working now.  Yay.

nickwallen:
Any other feedback guys?  Just waiting on +1s if you guys are happy with the change.

cestella:
+1 by inspection

PR #367, opened 2016-11-22
METRON-522: Need to mandate Client installation on Metron Host

justinleet:
Added metainfo.xml dependency on HBASE_CLIENT.  Modified service_advisor to force colocation of metron_parsers with ZOOKEEPER_CLIENT and HDFS_CLIENT, along with metron_enrichment colocating with HBASE_CLIENT.

Given that the enrichment master has to be colocated with parsers master in the current setup, I didn't duplicate the constraints there for brevity, although if anybody has a preference towards explicitly adding them in code (beyond the comments mentioning it) I could be pretty easily persuaded.

Tested by spinning a cluster as described in https://www.evernote.com/shard/s530/sh/c5551fbd-0ac1-4861-89ce-9c5e37065c52/b13e05f39eaac1a6

After setup, the cluster had to be modified to ensure each message could be hit in the appropriate location (e.g. ensure a given node has Kafka Broker + HDFS client, but not ZK client, etc.)

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.3.0/metainfo.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/0.3.0/metainfo.xml
@@ -84,6 +84,13 @@
               </auto-deploy>
             </dependency>
             <dependency>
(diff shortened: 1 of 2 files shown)


dlyle65535:
+1, looks good!

PR #369, opened 2016-11-22
METRON-585 Pcap Replay Fails to Stop

nickwallen:
The service script for Pcap Replay often fails to actually stop the `tcpreplay` process.

The `tcpreplay` process is a huge consumer of resources when running on Quick Dev and actually having it stop when asked, would be super nice.

Needed to kill it until it was good and dead.  Tested the change on Quick Dev.


nickwallen:
This is a rather annoying bug when working on Quick Dev.  The `tcpreplay` process that is started by Pcap Replay is very resource intensive.  If you stop Pcap Replay, but `tcpreplay` keeps running, it annoyingly continues to consume a lot of resources.  I would love to get this merged, but need some reviewers from the community. :)

cestella:
+1 by inspection

PR #370, opened 2016-11-23
METRON-586 add FILL_LEFT and FILL_RIGHT functions to STELLAR

ottobackwards:
Add functions to stellar so that you an pull or pad a string out to a required length


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -222,4 +222,71 @@ public Object apply(List<Object> args) {
...
+      return input;
+    }
+
+    String fill = fillObject.toString().substring(0,1);
+    int requiredLength = 0;
+    if(requiredLengthObject instanceof Integer){
+      requiredLength = (int)requiredLengthObject;
+    }else{
+      return input;
+    }
```
nickwallen: If I accidentally pass in a String for the requiredLength, then it will give me back the origin input, right?  Does that make sense?  Should it blow up instead and report the problem?

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -222,4 +222,71 @@ public Object apply(List<Object> args) {
...
+    String input = inputObject.toString();
+
+    if(requiredLengthObject == null || fillObject == null) {
+      return input;
+    }
+
+    String fill = fillObject.toString().substring(0,1);
+    int requiredLength = 0;
+    if(requiredLengthObject instanceof Integer){
+      requiredLength = (int)requiredLengthObject;
```
nickwallen: Would it make sense to use our `ConversionUtils.convert` utility here?  Just to be more forgiving of what might get passed in as the length?

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -222,4 +222,71 @@ public Object apply(List<Object> args) {
...
+  }
+
+  private static Object fill(FillDirection direction, Object inputObject, Object fillObject, Object requiredLengthObject){
+    if(inputObject == null) {
+      return null;
+    }
+    String input = inputObject.toString();
+
+    if(requiredLengthObject == null || fillObject == null) {
+      return input;
```
nickwallen: Should this blow-up and throw an exception if we don't have a length or fill?

Consider the case where I write an expression for the length or fill.  In many cases, if something unexpected happens in that expression, like a missing field, it returns null.  So the null would propagate here and the original string would be returned.   I would wonder why I am just getting the original string back when I expected it to be filled.  That seems hard to debug.

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
```diff
@@ -593,6 +595,39 @@ public void testStringFunctions_advanced() throws Exception {
...
+      put("foo", null);
+      put("bar", null);
+      put("notInt","oh my");
+    }};
+
+    Object left = run("FILL_LEFT('123','X', 10)",new HashedMap());
+    Object right = run("FILL_RIGHT('123','X', 10)", new HashedMap());
+    Object same = run("FILL_RIGHT('123','X', 3)", new HashedMap());
+    Object nullIn = run("FILL_RIGHT('123',foo,bar)", variableMap);
+    Object notInt= run("FILL_RIGHT('123',foo, notInt)", variableMap);
```
nickwallen: Would it make sense to add more tests for the failure behavior?  Like nulls for the length and fill arguments.  Also the case where the actual length is greater than the fill length. There's probably some other corner cases we could hit too.

nickwallen: I don't fully get the order of your tests.  The tests get executed above, but all validated below.  It might be more clear to run the test, then validate it on the next line.  At least to me that would be more understandable.

```
// test fill left
Object left = run("FILL_LEFT('123','X', 10)",new HashedMap());
Assert.assertNotNull(left);
Assert.assertEquals(10,((String)left).length());
Assert.assertEquals("XXXXXXX123",(String)left);

// test fill right
Object right = run("FILL_RIGHT('123','X', 10)", new HashedMap());
Assert.assertNotNull(right);
Assert.assertEquals(10,((String)right).length());
Assert.assertEquals("123XXXXXXX",(String)right);

...
```

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -222,4 +222,71 @@ public Object apply(List<Object> args) {
...
+    }
+    int howMany = requiredLength - actualLength;
+    String fillString = fill;
+    for(int i = 0; i < (howMany - 1); i++){
+      fillString += fill;
+    }
+    if(direction == FillDirection.LEFT) {
+      return fillString += input;
+    }
+    return input += fillString;
```
nickwallen: Could we just use `org.apache.commons.lang.StringUtils` `leftPad` and `rightPad`?  Might be easier than implementing this ourselves.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
ottobackwards: I agree, that is a malformed expression, I'll modify

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
ottobackwards: Sure, I am sorry it didn't come to mind immediately.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
ottobackwards: I think using convert utils will change this?  I'll change this over and take another look.  In general - the behavior ( return null or be forgiving ) is something I was hoping to fall out of the review.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
ottobackwards: That sounds good.  I don't want to lose my @apache.org but I have not used commons enough to have it be the first thing I think of.

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
ottobackwards: No problem, I'll refactor.

ottobackwards:
Refactored based on feedback.
Moved the stellar utility functions to a utility class

cestella:
+1 by inspection.  Looks good!

ottobackwards:
@nickwallen are you ok with the changes based on your review?

nickwallen:
Looks great.  Many thanks for the effort. +1

ottobackwards:
I am kicking this to travis again.  I'm seeing integration test run timeout errors locally trying to verify before commit and I want to sanity check

PR #373, opened 2016-11-28
METRON-591: Make the website in compliance with ASF standards

cestella:
It has come to my attention that the website is out of compliance in the following ways:
* Our "Download" link is to the github repository releases page, which may or may not be officially apache releases (every tag and RC results in a release there). It should go to the official Apache releases website for Metron.
* We are insufficiently prominent with our "Incubating" status. Really, when we refer to "Apache Metron" it should come with the parenthetical "Incubating."  It appears that this requirement is a bit fluid, so outside of the index page, just the first mention of Metron is corrected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 [![Build Status](https://travis-ci.org/apache/incubator-metron.svg?branch=master)](https://travis-ci.org/apache/incubator-metron)
 
-# Metron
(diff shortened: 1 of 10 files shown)


mmiklavc:
+1

james-sirota:
+1

PR #374, opened 2016-11-28
METRON-587 Integration tests should use common processor implementations where possible

ottobackwards:
We have 7 or so integration tests that implement the Processor interface. Much of the code in these implementations is the same. It would be better for the system if we had some common implementations of the Processor interface.

    This would make it easier to write correct tests
    This would reduce errors from copy/paste test generation
    This would clean up the codebase and make it easier to understand

This PR creates a KafkaProcessor class which implements the common processor code used for all the Parser Integration Test based classes, The WriterBolt Integration Test and the Enrichment Integration tests.

The pcap and simple hbase enrichments are 'different' enough to continue to use their own interface implementations.  The indexing tests are less likely to have new additions to them, and can be part of a subsequent effort if required.


ottobackwards:
Re-Run travis - build working locally

ottobackwards:
I am going to close this until I figure out what in the latest sync is throwing of the timing in the tests

ottobackwards:
Resolved the issue:   The enrichment test is different in that it expect both errors and messages of the same amount.  The code was not handling the circumstance where the errors arrived before the messages from kafka correctly.

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java
```diff
@@ -153,9 +168,17 @@ public void test() throws Exception {
       fluxComponent.submitTopology();
 
       kafkaComponent.writeMessages(Constants.ENRICHMENT_TOPIC, inputMessages);
-      ProcessorResult<List<Map<String, Object>>> result = runner.process(getProcessor(inputMessages));
+      ProcessorResult<List<Map<String, Object>>> result = runner.process(getProcessor());
       // We expect failures, so we don't care if result returned failure or not
       List<Map<String, Object>> docs = result.getResult();
+      // OTTO>>>>>>>>>>>>>>>>>>>
```
cestella: *might* not want the OTTO there, unless you want to be immortalized in Metron code @ottobackwards  ;)  Lord knows the `StellarTest` has enough mentions of my name, though, so I can't talk.

cestella: I also assume those printlns are debug code and should be removed.

ottobackwards:
Done.  Sorry about that

PR #376, opened 2016-11-29
METRON-593: Enable an automated static analysis tool in the build

justinleet:
Setup everything to use [Error Prone](http://errorprone.info/) with errors fixed and warnings on.  This increases build output, but it still seems fairly manageable.

We had a lot of the poms redefining the maven-compiler-plugin entirely, or just adding ```<compilerArgument>-Xlint:unchecked</compilerArgument>```, so I made them all just inherit from the root pom with ```-Xlint:unchecked``` on.  It seems reasonable to leave it on everywhere, and simplifies the POMs, but if anyone has a strong objection, it's easy to change it to whatever we want.

Assuming we're good with this approach to our static analysis, we'd create a follow on ticket (or tickets?) to address them.  After addressing them, we probably should swing back around and upgrade some or all of our warnings to errors, so we don't have to deal with them cropping up again.

Tests were run successfully and quick-dev was spun up to with some data pushed through.

### DIFF
--- a/metron-analytics/metron-maas-common/pom.xml
+++ b/metron-analytics/metron-maas-common/pom.xml
@@ -129,33 +129,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.3</version>
-        <configuration>
-          <compilerId>javac-with-errorprone</compilerId>
-          <forceJavacCompilerUse>true</forceJavacCompilerUse>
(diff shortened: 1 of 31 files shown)


justinleet:
Updated the root pom's maven-compiler-version to match what it was.  Also noticed that the overridden versions in child poms tend to use 3.1.  Does anyone know if that's intentional, or just an artifact of moving chunks of poms around?  If it is intentional, that's a reason to keep them separate.  I was able to package things up (admittedly skipping tests) with 3.5.1 without issues.

dlyle65535:
+1, thanks!

PR #377, opened 2016-11-29
METRON-595: ES Writer uses more than one IP

JonathanRider:
Simple fix. Before, the client was reinitialzed for each ip, so only one was actually used. By initialing the client outside the loop, it uses all of the ip's.

### DIFF
--- a/metron-deployment/packer-build/bin/bento
+++ b/metron-deployment/packer-build/bin/bento
@@ -1,20 +1,4 @@
 #!/usr/bin/env ruby
-#
-#  Licensed to the Apache Software Foundation (ASF) under one or more
-#  contributor license agreements.  See the NOTICE file distributed with
-#  this work for additional information regarding copyright ownership.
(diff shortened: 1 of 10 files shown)


nickwallen:
Good catch. Thanks!  +1 by inspection, pending CI build

JonathanRider:
Looks like the build failed for a completely unrelated test. The latest commit in master fails the build on the same test.

Should I wait until that is fixed and then merge master into my branch?

james-sirota:
@nickwallen somehow this fix failed a profiler test. can you take a look?

nickwallen:
Definitely seems to be some kind of sporadic failure with those integration tests.  For now, close and reopen the PR and it would likely fix it. 

I created a JIRA to track the sporadic failures.   I have yet to be able to replicate this locally. Sorry, I know these sporadic failures are annoying.  https://issues.apache.org/jira/browse/METRON-597

ottobackwards:
I think you should >fetch apache && merge apache/master && push origin METRON-FOO

PR #378, opened 2016-11-29
METRON-596 Eliminate Maven warnings from build

justinleet:
Moving the maven-jar-plugin to a global variable and updating it.

Removing a duplicate dependency that had both 'provided' and 'test' scope (dropped 'test', because it's already covered by 'provided').

Changed ```${parent.version}``` to ```${project.parent.version}```

Ran ```mvn clean install``` to test.

Does anybody know if there's a way to force Maven to break the build if new warnings are introduced?  I'm not sure we'd actually want to do it, but at the same time I do like having these sorts of things enforced.

### DIFF
--- a/metron-analytics/metron-maas-common/pom.xml
+++ b/metron-analytics/metron-maas-common/pom.xml
@@ -120,7 +120,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
-        <version>2.5</version>
+        <version>${global_jar_version}</version>
         <executions>
           <execution>
             <goals>
--- a/metron-analytics/metron-maas-service/pom.xml
+++ b/metron-analytics/metron-maas-service/pom.xml
@@ -249,7 +249,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
-        <version>2.5</version>
+        <version>${global_jar_version}</version>
         <executions>
           <execution>
             <goals>
--- a/metron-deployment/packaging/ambari/metron-mpack/pom.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/pom.xml
@@ -89,7 +89,7 @@
             </plugin>
             <plugin>
                 <artifactId>maven-jar-plugin</artifactId>
-                <version>2.3.1</version>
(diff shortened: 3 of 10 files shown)


dlyle65535:
+1 (travis pending), glad to see the end of those. 
One small nit- if you find yourself doing wholesale dependency versioning, would you consider using the dependencyManagement section in the top-level pom?

dlyle65535:
On the warning/failure deal, does [this](http://maven.apache.org/plugins/maven-dependency-plugin/examples/failing-the-build-on-dependency-analysis-warnings.html) do what you're looking for?

justinleet:
On the dependency versioning, it is a good idea to use the dependencyManagement section, but I'd probably consider that cleanup out of scope for this ticket. I didn't want to start a larger refactor here, just clear out the warnings.

On point two, it's related, but different.  We should fix our issues with that (to see them run ```mvn dependency:analyze```), then I'd be interested in in exactly that.  The difference is that your link covers dependency issues, but I'm interested in general pom issues (e.g. using the deprecated variable).

I'm also inclined to consider point two more important, partially because I don't want to generalize a bunch of dependencies that end up not actual being general anyway if they're unused. Any objections to that plan?

I'm inclined to create two follow up tickets, one for each of them.  I'm happy to pick up the second one, given that it's pretty related to this ticket.

justinleet:
Kicking Travis due to https://issues.apache.org/jira/browse/METRON-597

justinleet:
This still runs locally, and appears to just incredibly unluckily be breaking.  If this one fails, I'm going to wait to pull in https://github.com/apache/incubator-metron/pull/380.

justinleet:
@nickwallen This seems to pretty consistently failing the ProfilerIntegrationTest on Travis (not local, unfortunately), if you're looking for something seemingly repeatable.

Could you glance at this to see if I'm doing something that has an unexpected affect on the test? All I've really done is bumped the maven-jar-plugin version, removed an unrelated dependency, and migrated from a deprecated variable in an unrelated pom.

nickwallen:
Thanks @justinleet.  I do not see anything you've done here that would cause any problems.  

I am able to fairly reliably reproduce the bug from METRON-597 when run in Travis CI.  So maybe this isn't more likely to happen with the changes that you've made, than any others.

justinleet:
Travis errors unrelated and fixed by https://github.com/apache/incubator-metron/pull/383

PR #379, opened 2016-11-30
METRON-595: ES Writer uses more than one IP

JonathanRider:
Resubmitting pull request due to sporadic test failure as per comments on the last pull request.

Simple fix. Before, the client was reinitialized for each ip, so only one was actually used. By initialing the client outside the loop, it uses all of the ip's.

### DIFF
--- a/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
+++ b/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
@@ -73,11 +73,11 @@ public class ElasticsearchWriter implements BulkMessageWriter<JSONObject>, Seria
     Settings settings = settingsBuilder.build();
 
     try{
+      client = TransportClient.builder().settings(settings).build();
       for(HostnamePort hp : getIps(globalConfiguration)) {
-        client = TransportClient.builder().settings(settings).build()
-                .addTransportAddress(
-                        new InetSocketTransportAddress(InetAddress.getByName(hp.hostname), hp.port)
-                );
+        client.addTransportAddress(
+                new InetSocketTransportAddress(InetAddress.getByName(hp.hostname), hp.port)
+        );
       }
 
 
--- a/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
+++ b/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
@@ -106,6 +106,25 @@ public class ElasticsearchWriter implements BulkMessageWriter<JSONObject>, Seria
       return Collections.emptyList();
     }
     if(ipObj instanceof String
+            && ipObj.toString().contains(",") && ipObj.toString().contains(":")){
+      List<String> ips = Arrays.asList(((String)ipObj).split(","));
+      List<HostnamePort> ret = new ArrayList<>();
+      for(String ip : ips) {
+        Iterable<String> tokens = Splitter.on(":").split(ip);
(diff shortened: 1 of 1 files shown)


dlyle65535:
Thanks for this change, @JonathanRider - can you tell me how you tested it?

JonathanRider:
Just realized we had some custom code since we couldn't get the List object working for feeding in multiple IP's. So I added in the code so you can just have a comma separated list of IP's. 

We have a dev cluster that has 3 master nodes, but I only gave the writer 2 ip addresses. I sent data through, then stopped the elasticsearch service on that machine. Sent data through and it works (previously this would give a NoNodeAvailable exception). I then started ES back up on that server, and shut down the ES Service on the other server and sent data through again and it also works.

JonathanRider:
Is there anything else I need to do in order to get this merged in, @dlyle65535 ?

dlyle65535:
Could you give it a run up on Quick Dev using your patch rebased on master?

JonathanRider:
Is quickdev working currently? I just spun it up and tried to start up a parser topology but get an error. Looking at it, it seems like the error seen here: https://community.hortonworks.com/questions/68798/metron-enrichment-fail-to-start-with-ambari-instal.html

Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/storm/Config
	at org.apache.metron.parsers.topology.ParserTopologyCLI$ParserOptions.<init>(ParserTopologyCLI.java:199)
	at org.apache.metron.parsers.topology.ParserTopologyCLI$ParserOptions.<clinit>(ParserTopologyCLI.java:47)
	at org.apache.metron.parsers.topology.ParserTopologyCLI.main(ParserTopologyCLI.java:266)
Caused by: java.lang.ClassNotFoundException: org.apache.storm.Config
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 3 more

ottobackwards:
quick dev is working - i spun one up this morning from a new branch off of master

JonathanRider:
And you can start a topolgy in it? This is the command I'm running:
/usr/metron/0.3.0/bin/start_parser_topology.sh -z node1:2181 -k node1:6667 -s bro

And storm version gives 0.10.0.2.3.4.7-4
And the HDP stack in ambari says: HDP-2.3.4.7-4

Can you confirm that these match yours?

JonathanRider:
Just ran it on quickdev where I set the es.host in the global json file to "node2,node1" where node2 doesn't exist as it is a single node vagrant.

Using my patch, it writes to ES without a problem as it fails to use node2, so it switches to node1.

I then compiled master and did the same thing and received this error since node2 doesn't exist and it fails to use the second host.:
2016-12-01 21:41:18.942 o.a.m.w.BulkWriterComponent [ERROR] Failing 5 tuples
org.elasticsearch.client.transport.NoNodeAvailableException: None of the configured nodes are available: [{#transport#-1}{10.120.164.190}{node2/10.120.164.190:9300}]

JonathanRider:
@dlyle65535 is there anything specific that you wanted me to test on quickdev that I haven't already done?

dlyle65535:
Nope, looks like you did a test run with changes and @ottobackwards did the clean one I was going to do, so I'm +1.
Thanks!

ottobackwards:
@dlyle65535 I did not test, I was answering the question of if quick-dev works at all

dlyle65535:
Oh, thanks for that. Let me run it up really quickly then. Totally mis-read you, @ottobackwards.

dlyle65535:
This looks good to me, @JonathanRider - do you want me to use the email on your profile for the commit?

JonathanRider:
Yep. Thanks!

PR #380, opened 2016-11-30
METRON-597 Sporadic Failures of Profiler Integration Tests

nickwallen:
Disabling the Profiler integration tests which are failing sporadically.  Short-term work around to ease some pain.  Will re-enable once a true fix has been found.

### DIFF
--- a/metron-analytics/metron-profiler/pom.xml
+++ b/metron-analytics/metron-profiler/pom.xml
@@ -260,6 +260,10 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>2.12.4</version>
                 <configuration>
+                    <excludes>
+                      <!-- TODO need to re-enable these test per METRON-597 -->
+                      <exclude>**/ProfilerIntegrationTest.java</exclude>
+                    </excludes>
                     <!-- Skip the default running of this plug-in (or everything is run twice...see below) -->
                     <argLine>-Xmx2048m</argLine>
                     <skip>true</skip>
--- a/metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
+++ b/metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
@@ -126,7 +126,8 @@ public class ProfilerIntegrationTest extends BaseIntegrationTest {
   /**
    * Tests the first example contained within the README.
    */
-  @Test
+  //@Test
+  // TODO need to re-enable these test per METRON-597
   public void testExample1() throws Exception {
 
     setup(TEST_RESOURCES + "/config/zookeeper/readme-example-1");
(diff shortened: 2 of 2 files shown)


ottobackwards:
can this be done in the pom/surefire?

nickwallen:
Dunno.  What are the advantages to doing it that way?

ottobackwards:
Configuration only change.  So, let's say that there is another pr that changes that test, it can do so without worrying about the commented out code etc.  not a big deal,  really just curious

cestella:
Agreed, this should be changed in the pom.  http://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

JonZeolla:
Doesn't this conflict with the merge requirements on the draft [Coding Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) page?  Specifically:
* A patch that breaks tests, or **introduces regressions by changing or removing existing tests** should not be merged. Tests must always be passing on master. This implies that the tests have been run...

I understand the current scenario means this isn't straightforward - perhaps that's an indication that the dev guidelines need updated to account for exceptions?  Or maybe we take another tact to fix the issue.  I'll shoot an email to the current [DISCUSS] thread regarding this.

ottobackwards:
Would it be worth closing/reopening a PASSING PR to see if it too fails?

dlyle65535:
Alternatively, if you wanted a one-liner, I think you could @Ignore the test class.

I didn't see @JonZeolla - add to the [DISCUSS] thread, so I'll answer here- I don't agree that this introduces a regression, it corrects one. At some point, something was introduced that has a damaging interaction with this test. As a result, the build is broken. I'd agree that we cannot introduce any Profiler changes until we get this sorted, but I don't see a problem with disabling the broken test as it is unrelated to any code changes to the classes under test.

JonZeolla:
@dlyle65535 while we slightly disagree on semantics, I agree that this should move forward for the short term.  I'm primarily indicating that the dev guide needs updated to account for situations like this.  I sent an email out to the [DISCUSS] thread a couple of minutes back.

dlyle65535:
@JonZeolla - oh yeah, totally in agreement there.

dlyle65535:
Probably a dumb question, but doesn't integration-test run Failsafe?

nickwallen:
I believe that #381 is the real solution to this problem.  This PR being simply a work around.  I would prefer to push #381 and close this PR.  

Open to other's opinions though.  Still need some votes to do anything.

JonZeolla:
@nickwallen Sounds reasonable, once #381 gets fleshed out.

ottobackwards:
I think you closed the wrong one

nickwallen:
I was not happy with #381 . I will reopen it soon once I have a fix that I am happy with.

nickwallen:
Closing this PR.  About to open another one with the fix.

PR #381, opened 2016-12-01
METRON-601 MockHTable Put Log is Not Thread Safe

nickwallen:
[METRON-601](https://issues.apache.org/jira/browse/METRON-601)

The MockHTable uses an ArrayList to store a log of Puts that have been submitted against the MockHTable. The MockHTable, along with the put log, is accessed from multiple threads during the integration tests. Access to the Put log is not thread safe, which is likely at least one cause of [METRON-597](https://issues.apache.org/jira/browse/METRON-597).

The Put log is used by multiple tests, but more so by the ProfilerIntegrationTest. This tests polls the list to block the thread until the expected number of Puts have been submitted. This is likely why this test is more impacted by this issue than others.

The Put Log needs to made thread safe.  See `org.apache.metron.test.mock.MockHTable.putLog`.  This is the smallest change that I could think of that would address the issue.

I have triggered this code to run in Travis CI many times without seeing a failure.  Prior to this the Travis CI builds would more likely fail than not.  I have also continually run the impacted tests over many hours without replicating the problem.  In all fairness, on the current master branch without this fix, I was only able to replicate the problem once on my local machine (after running many hours.)

### DIFF
--- a/metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
+++ b/metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
@@ -18,6 +18,9 @@
 package org.apache.metron.test.mock;
 
 
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
+import com.google.common.collect.ImmutableList;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Message;
 import com.google.protobuf.Service;
@@ -28,7 +31,18 @@ import org.apache.hadoop.hbase.HColumnDescriptor;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.KeyValue;
 import org.apache.hadoop.hbase.TableName;
-import org.apache.hadoop.hbase.client.*;
+import org.apache.hadoop.hbase.client.Append;
+import org.apache.hadoop.hbase.client.Delete;
+import org.apache.hadoop.hbase.client.Durability;
+import org.apache.hadoop.hbase.client.Get;
+import org.apache.hadoop.hbase.client.HTableInterface;
+import org.apache.hadoop.hbase.client.Increment;
+import org.apache.hadoop.hbase.client.Put;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Row;
+import org.apache.hadoop.hbase.client.RowMutations;
+import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.client.coprocessor.Batch;
 import org.apache.hadoop.hbase.filter.CompareFilter;
 import org.apache.hadoop.hbase.filter.Filter;
@@ -37,7 +51,15 @@ import org.apache.hadoop.hbase.util.Bytes;
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.NavigableMap;
+import java.util.NavigableSet;
+import java.util.NoSuchElementException;
+import java.util.TreeMap;
 
 /**
  * MockHTable.
@@ -53,9 +75,11 @@ public class MockHTable implements HTableInterface {
       HTableInterface ret = _cache.get(tableName);
       return ret;
     }
+
     public static HTableInterface getFromCache(String tableName) {
       return _cache.get(tableName);
     }
+
     public static HTableInterface addToCache(String tableName, String... columnFamilies) {
       MockHTable ret =  new MockHTable(tableName, columnFamilies);
       _cache.put(tableName, ret);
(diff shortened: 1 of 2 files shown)


nickwallen:
Travis is sleepy.  Hasn't started the build.  Kick.

metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
```diff
@@ -437,15 +463,14 @@ public ResultScanner getScanner(byte[] family, byte[] qualifier)
     return getScanner(scan);
   }
 
-  List<Put> putLog = new ArrayList<>();
```
justinleet: Couldn't you just change this to
```List<Put> putLog = Collections. synchronizedList(new ArrayList<>());```?

I'd still keep the `ImmutableList.copyOf()` in `getPutLog()`, because a synchronized list still requires a synchronized block during iteration and callers could do whatever.  Notably, `ImmutableList.copyOf()` specifically states "This method is safe to use even when elements is a synchronized or concurrent collection that is currently being modified by another thread."

At that point, it should be thread-safe.  It doesn't expose the underlying list, the putLog doesn't get iterated over in this class, and it only exposes thread safe copies of the putLog.

I'm also not convinced that the Supplier solution is actually thread-safe, which may be my own ignorance. The `get()` itself is thread-safe (per `memoize()`), but just returns a regular ArrayList (```this.putLog = Suppliers.memoize(() -> new ArrayList<Put>());```).  The `add()` won't necessarily be thread-safe, because two threads could each call `get()` and still attempt to put onto the same list at the same time. Assuming that's correct (which I'd like a second opinion on), it's probably just more that shuffling things happens to ease the race condition.

Somebody feel free to hop in and correct any misconceptions I have here.  I'm not as familiar with Java's concurrency as I probably (definitely?) should be.

File: metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java [hunk 2]
```diff
@@ -437,15 +463,14 @@ public ResultScanner getScanner(byte[] family, byte[] qualifier)
...
-
   public List<Put> getPutLog() {
-    return putLog;
+    return ImmutableList.copyOf(putLog.get());
   }
 
   @Override
   public void put(Put put) throws IOException {
-    putLog.add(put);
+    putLog.get().add(put);
```
cestella: To @justinleet's point, while the `Supplier.get()` call is thread-safe, the resulting list is not.  I would recommend adjusting the construction of the `ArrayList<Put>` to `Collections.synchronizedList(new ArrayList<>())` to ensure that the resulting `add` call is happening in a thread-safe way.  At that point, it's unclear why we would use the supplier at all.  

On the whole, it's interesting that this may be at the root of the unit test weirdness.  Have you gotten the tests to fail in a somewhat repeatable manner and does this change correct it?

Standard disclaimer applies: all of this is based on my understanding of Java, which can be obviously incomplete, so let me know if I got something wrong here.

metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
nickwallen: Good that we're digging into this.  I'm not totally sure that this is thread-safe.  You may be right.  I have used `memoize` for a thread-safe, lazy-loading cache.  That's the use case I've seen others use it for.  But maybe with that hammer, everything is looking like a nail to me. :)

But in my opinion if `memoize` is not thread-safe, then `synchronizedList` is not either.  See the code snippets below.  What `synchronizedList` is doing seems to be a subset of what `memoize` does.  No?

```
    MemoizingSupplier(Supplier<T> delegate) {
      this.delegate = delegate;
    }

    public T get() {
      if(!this.initialized) {
        synchronized(this) {
          if(!this.initialized) {
            Object t = this.delegate.get();
            this.value = t;
            this.initialized = true;
            return t;
          }
        }
      }

      return this.value;
    }
  }
```

versus

```
    ThreadSafeSupplier(Supplier<T> delegate) {
      this.delegate = delegate;
    }

    public T get() {
      Supplier var1 = this.delegate;
      synchronized(this.delegate) {
        return this.delegate.get();
      }
    }
  }
```

metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
nickwallen: See my response to @justinleet per `synchronizedList`.

Per your testing question: "I have triggered this code to run in Travis CI many times without seeing a failure. Prior to this the Travis CI builds would more likely fail than not."

metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
nickwallen: Oh, sorry guys.  Ignore me.  You're talking Collections.synchronizedList, not Guava's synchronized supplier.  Yeah, I think you're right.

nickwallen:
After 3 commits we might finally have something sensible. :)

Iteration is protected by a defensive copy.  Adding Puts is protected by the synchronizedList which explicitly synchronizes on a mutex for an operation like `add`.

metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
```diff
@@ -437,15 +462,14 @@ public ResultScanner getScanner(byte[] family, byte[] qualifier)
     return getScanner(scan);
   }
 
-  List<Put> putLog = new ArrayList<>();
-
   public List<Put> getPutLog() {
-    return putLog;
+    return ImmutableList.copyOf(putLog);
```
nickwallen: In changing to a `synchronizedList`, I don't think the defensive copy is thread-safe any longer.  I'm going to close this PR and test the whole thing holistically, since it is different enough from what I originally submitted.

PR #383, opened 2016-12-01
METRON-597 Sporadic Failures of Profiler Integration Tests

nickwallen:
#### MockHTable Put Log

The MockHTable uses an ArrayList to store a log of Puts that have been submitted against the MockHTable. The MockHTable, along with the put log, is accessed from multiple threads during the integration tests, but it is not thread safe.  The Put log is used by multiple tests, but more so by the ProfilerIntegrationTest. This tests polls the list to block the thread until the expected number of Puts have been submitted. This is likely why this test is more impacted by this issue than others.

#### MockHTable Scanner

Previously the test was using `MockHTable.getScanner` to validate the results.  This was changed to grep the same information from the Put log itself, instead of relying on the scanner implementation in MockHTable.

#### Test Validation

Some of the test validation was very opinionated about the number of times the Profiler itself had flushed.  As the integration test is being validated, the Profiler is running in the background and will continue to flush measurements.  

The validation was changed to not care how many times the Profiler has flushed measurements to HBase.

### Testing

Prior to these changes, the Travis CI build was reliably failing.  After making these changes, the Travis CI build has passed at least 6 times in a row.

I have also run this repeatedly on my local machine and have not experienced any failures.

### DIFF
--- a/metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
+++ b/metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
@@ -18,6 +18,7 @@
 package org.apache.metron.test.mock;
 
 
+import com.google.common.collect.ImmutableList;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Message;
 import com.google.protobuf.Service;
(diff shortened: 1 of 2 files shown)


metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
```diff
@@ -70,7 +93,7 @@ public static void clear() {
   private final String tableName;
   private final List<String> columnFamilies = new ArrayList<>();
   private HColumnDescriptor[] descriptors;
-
+  private List<Put> putLog;
```
justinleet: Why not do ```private List<Put> putLog = Collections.synchronizedList(new ArrayList<>);``` here?  It would eliminate the manual synchronization in `getPutLog()` and `addToPutLog()` and I doubt there's significant perf impact either way.  Correct me if I'm wrong on that.

To elaborate, `add()` is directly taken care of with synchronized list.  The only other concern is the `getPutLog()`, and Guava's docs state `ImmutableList.copyOf()` is fine to use on synchronized collections.  Looking into their code, it uses of `toArray()` (which is synchronized) instead of iteration to copy the List.

ottobackwards:
So - if I test this by getting the pr and pushing to my own github that is tied to travis - that would be the only thing to try with this right

ottobackwards:
+1
Ran pr on travis

metron-platform/metron-test-utilities/src/main/java/org/apache/metron/test/mock/MockHTable.java
nickwallen: That's what I initially tried and it was not working.  Thinking back, I later found two other issues and had to fix those also (described above as MockHTable scanner and Test Validation.)  So the problem may not have been caused by ImmtableList.copy + synchronizedList, but by those other issues.  Dunno.

nickwallen: @justinleet Do you think it is worth it to try `synchronizedList` again?  IMO, that code is just in a test class, so I am not too concerned.  It also takes a lot of effort to test this.  But if you think it is worth it, I could be convinced.

justinleet: Yeah, given that it's a test class and the amount of testing effort, I'm not really worried about it too much.  I'm good, +1.

PR #382, opened 2016-12-01
METRON-598 Add Kyle Richardson to committers

kylerichardson:
Add Kyle Richardson to list of committers on website

Verified site by following instructions on wiki [1].

[1] https://cwiki.apache.org/confluence/display/METRON/Website+PR+Merge

PR #384, opened 2016-12-02
METRON-603 Update website to use Apache release repo

kylerichardson:
Update the DOWNLOAD link on the navigation menu of the website to point to the Apache release repository.

Verified site by following instructions on wiki [1].

NOTE: I'm holding off committing my recent PR (#382) to asf-site until this is fully merged to ensure consistency on the website.

[1] https://cwiki.apache.org/confluence/display/METRON/Website+PR+Merge

### DIFF
--- a/site/_includes/primary-nav-items.html
+++ b/site/_includes/primary-nav-items.html
@@ -10,6 +10,6 @@
     <li class="training-menu"><a href="/news/">News</a>
     </li>
     <li>
-        <button class="button-default button-green trigger-top-form"> <a href="https://github.com/apache/incubator-metron/releases" target="new">Download </a></button>
+        <button class="button-default button-green trigger-top-form"> <a href="https://dist.apache.org/repos/dist/release/incubator/metron/" target="new">Download </a></button>
     </li>
 </ul>


site/_includes/primary-nav-items.html
```diff
@@ -10,6 +10,6 @@
     <li class="training-menu"><a href="/news/">News</a>
     </li>
     <li>
-        <button class="button-default button-green trigger-top-form"> <a href="https://github.com/apache/incubator-metron/releases" target="new">Download </a></button>
+        <button class="button-default button-green trigger-top-form"> <a href="https://dist.apache.org/repos/dist/release/incubator/metron/" target="new">Download </a></button>
```
mattf-horton: Hi Kyle, this change is on the right path, since we definitely should be publishing downloads on "dist".  However, http://www.apache.org/dev/release-download-pages.html says:
- All links to the distribution artifacts **must not** reference the main Apache web site ["dist"]. They should use the standard mechanisms to distribute the load between the mirror sites [see http://www.apache.org/dyn/closer.cgi/ ].
- All links to checksums, detached signatures and public keys **must** reference the main Apache web site and should use https (SSL).

The web page then goes on in some detail about how to set up a script to automatically redirect the user to an appropriate mirror.  

If the above isn't clear, there is a better written explanation at http://incubator.apache.org/guides/releasemanagement.html#understanding-mirroring
But it points back to http://www.apache.org/dev/release-download-pages.html to talk about the script.

An example of a good download page that uses an automatic mirror-direction script is
https://maven.apache.org/download.cgi
An example of a download page that is "dumber" but therefore easier to set up is
https://hadoop.apache.org/releases.html

The quick-and-dirty fix is to simply point at http://www.apache.org/dyn/closer.cgi/incubator/metron/ instead of 
https://dist.apache.org/repos/dist/release/incubator/metron/ , BUT that doesn't give the user access to the secure signatures, and that's not acceptable.

ptgoetz: @kylerichardson A separate downloads page a good approach that a lot of projects take.

Basically you have links to the source/binary artifacts that go through closer.cgi, and the links to the checksums/signatures go to dist.apache.org. The reason behind that is that checksums/signatures are not synced with the mirrors for security reasons.

mattf-horton: @kylerichardson , I agree, a downloads page is more or less the only way to satisfy the simultaneous requirements of directing the code download through a mirror and the signature download through the dist site.

I would suggest keeping jira METRON-603 open (adding "via Mirror script" to the title) and using it for the correct enhancement later on.  Meanwhile, I have no objection to committing this PR as an improvement to the current situation of having the actual site out of sync with the source code.  We are still sufficiently low volume that we aren't likely to cause problems with Infra.

cestella: Just FYI, METRON-600 was created to capture the website changes that need to happen.  That commit was to get around an immediate and urgent issue with the download link (namely, it was pointing to github which hosted non-apache released artifacts).  METRON-600 is intended to fix the website with a proper download page (among other things).

kylerichardson: @cestella Thanks, I didn't see METRON-600. Is it worth pursuing this PR? My thought is yes, so that master is in sync with what is deployed to asf-site. Otherwise, committing #382 to asf-site will revert your changes to the download link.

kylerichardson: @cestella Wow, looks like I need to do my homework a little better :). Thanks! I'll close this out as well as the associated JIRA.

mattf-horton: It looks like METRON-600 is now the correct place to state the need for a Downloads page to support the mirroring requirement.  I will add it there, so suggest you go ahead and close METRON-603.

PR #387, opened 2016-12-03
METRON-606 Profiler Overwriting Previously Written Values

nickwallen:
[METRON-606](https://issues.apache.org/jira/browse/METRON-606)

The Profiler is repetitively overwriting previously written values. Where I would expect to see a series of multiple values in Hbase, one written per flush, there is only a single value. This same value is being repetitively overwritten with new data.

The following changes are included in the PR.

* Refactored the ProfileBuilderBolt so that the logic for building a profile no longer lives in the bolt itself, but in a separate ProfileBuilder.  The bolt simply maintains a cache of ProfileBuilder instances.  This has helped simplify testing of this scenario. 

* Ensured that ProfileMeasurement objects are not reused, which reduces the likelihood of another defect like this in the future.

* Added a Clock interface that allows me to advance time in the unit tests.  This was necessary for testing this specific defect.  This will also help with METRON-590 in dealing with processing versus event time.

* Added a specific unit test for this defect.

* Added a unit test for the recent 'tickUpdate' addition to the Profiler.

The change was tested on a live AWS cluster along with the 'Quick Dev' environment.


cestella:
Did this get tested in vagrant?  
Specifically, did we try the following things:
* A global profile
* A non-global profile
* 2 profiles that share the same variable names in their update logic

nickwallen:
I had hit the first two, but not the last.  I just setup an additional manual test that covers all of these scenarios.  More specifically, here are the profiles that I defined.  

```
{
  "profiles": [
    {
      "profile": "counter",
      "foreach": "'global'",
      "init":  { "s": 0 },
      "update":  { "s": "s + 1" },
      "result":  "s"
    },
    {
      "profile": "by-ip",
      "foreach": "ip_src_addr",
      "init":  { "s": 0 },
      "update":  { "s": "s + 1" },
      "result":  "s"
    },
    {
      "profile": "always-zero",
      "foreach": "ip_src_addr",
      "init":  { "s": 0 },
      "update":  { "s": 0 },
      "result":  "s"
    }
  ]
}
```

One of the profiles counts messages that it sees "globally", the next counts the number of messages by source IP, and the last likes to always zero-out the same variable named "s", just to see if it can muck with the other two profiles.  

The `always-zero` profile as expected remains 0.  I can see additional values being written as time progresses.
```
[Stellar]>>> PROFILE_GET("always-zero","72.34.49.86", 5, "MINUTES")
[0, 0]
[Stellar]>>> PROFILE_GET("always-zero","72.34.49.86", 5, "MINUTES")
[0, 0, 0]
```

All of these profiles share the same variable 's', so if there was some kind of bad intermingling, I would expect the `always-zero` profile to seriously muck things up for the other two profiles.  But I do not see any of that.  

I can see new values being added to the `by-ip` profile.
```
[Stellar]>>> PROFILE_GET("by-ip","72.34.49.86", 5, "MINUTES")
[10.0, 106.0]
[Stellar]>>> PROFILE_GET("by-ip","72.34.49.86", 5, "MINUTES")
[10.0, 106.0, 58.0]

[Stellar]>>> PROFILE_GET("by-ip","192.168.138.158", 5, "MINUTES")
[296.0, 303.0, 302.0]
[Stellar]>>> PROFILE_GET("by-ip","192.168.138.158", 7, "MINUTES")
[296.0, 303.0, 302.0, 303.0]
```

The "global" profile is also behaving properly.
```
[Stellar]>>> PROFILE_GET("counter","global", 5, "MINUTES")
[833.0, 319.0, 746.0]
[Stellar]>>> PROFILE_GET("counter","global", 6, "MINUTES")
[833.0, 319.0, 746.0, 1538.0]
```

cestella:
Ok, I'm in.  +1 by inspection.  Great job tracking it down, @nickwallen

PR #388, opened 2016-12-05
METRON-610: OnlineStatisticsProvider serialization is broken at random in the REPL

cestella:
We rely on the t-digest library version 3.1 and elasticsearch brings along 3.0. There is a small API incompatibility between the two versions (namely the static method `TDigest.createAvlTreeDigest()` is available in 3.1 but not 3.0).

If the classpath for the Stellar REPL chooses the 3.0 version of the library, then deserialization is broken.  Strictly speaking this is not a problem of the serialized form being incorrect (we use t-digest's custom serialization), but a problem in the custom kryo serialization code in the class. It relies on the default constructor being called and then the digest being deserialized using the code within the t-digest library. Because the default constructor initializes the digest via a call that does not exist in 3.0, it breaks. The serialization logic is safe to use in both versions, but the object can't be constructed in 3.0.  This fix directly instantiates the AvlTreeDigest, which exists in both versions.


cestella:
I wrote out serialized objects from 3.0 and 3.1 versions of the t-digest library and verified that the objects were able to mutually read from both versions of the library.

nickwallen:
Is there not a way to ensure that our code uses a single version of the t-digest library?  I'd be worried that we might run into another incompatibility in the future.  

If not, is there some ingenious way to automate the testing of this scenario?

cestella:
It's a good point, @nickwallen.  We could shade and relocate the jar to enforce our version is used rather than a random one being picked up on the classpath.

metron-analytics/metron-statistics/pom.xml
```diff
@@ -52,4 +52,62 @@
...
+                                        <resource>LICENSE.txt</resource>
+                                        <resource>ASL2.0</resource>
+                                        <resource>NOTICE.txt</resource>
+                                      </resources>
+                                </transformer>
+                                <!-- UNCOMMENT THIS IF YOU NEED TO REGENERATE THE BEST GUESS NOTICES FILE WHICH REQUIRES PRUNING EVERY RELEASE -->
+                                <!--transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
+                                    <addHeader>false</addHeader>
+                                    <projectName>${project.name}</projectName>
+                                </transformer-->
```
dlyle65535: Would this be better as a wiki entry for release managers?

cestella: Probably.  It's not needed here in any case since this isn't a jar that gets distributed.

cestella:
Ok, I tested this in vagrant.  The testing methodology was as follows:
* Created a simple global profile that adds a `value` field to a stats object and serializes it:
```
{
  "profiles": [
    {
      "profile": "stats",
      "foreach": "'global'",
      "onlyif": "true",
      "init" : {
        "s": "STATS_INIT()"
               },
      "update": {
        "s": "STATS_ADD(s, value)"
                },
      "result": "s"
    }
  ]
}
```
* Created a simple CSV parser called `mad` that takes in comma separated pairs and adds the mean from the last 10 minute window:
```
{
 ,"sensorTopic" : "mad"
 ,"parserConfig" : {
    "columns" : {
       "value_str" : 0
      ,"source" : 1
                }
                   }
 ,"fieldTransformations" : [
    {
    "transformation" : "STELLAR"
   ,"output" : [ "value" , "mean"]
   ,"config" : {
      "value" : "TO_DOUBLE(value_str)",
      "mean" : "STATS_MEAN(STATS_MERGE(PROFILE_GET('stats', 'global', 10, 'MINUTES')))"
               }
    }
                           ]
}
```
* Created an enrichment that adds a field `enrichment_mean` via a stellar enrichment:
```
{
  "index": "mad",
  "batchSize": 1,
  "enrichment": {
    "fieldMap": {
      "stellar" : {
        "config" : {
          "enrichment_mean" : "STATS_MEAN(STATS_MERGE(PROFILE_GET('stats', 'global', 10, 'MINUTES')))"
        }
      }
    }
  ,"fieldToTypeMap": { }
  },
  "threatIntel": {
    "fieldMap": { },
    "fieldToTypeMap": { },
    "triageConfig" : {
      "riskLevelRules" : {
      },
      "aggregator" : "MAX"
    }
  }
}
```
* Generated a stream of data of the form `value,type` (e.g. 2.45,synthetic)
* Open up the stellar shell after 10 minutes and run `STATS_MEAN(STATS_MERGE(PROFILE_GET('stats', 'global', 10, 'MINUTES')))` and ensure I get something that is not `NaN`

This ensures that statistics objects are able to be serialized and deserialized correctly from the:
* Enrichment topology
* Parser topology
* Stellar REPL

nickwallen:
+1 Looks solid

PR #389, opened 2016-12-06
METRON-612: Clean up Error Prone generated warnings

justinleet:


justinleet:
The number of files looks a lot more intimidating than the PR actually is (Most is adding `@Override` rather than actual changes), but it can be split out if we want it to be.

Besides actual fixes, http://errorprone.info/bugpattern/DefaultCharset warnings are very common and I don't touch them at all.  My guess is that we'd be good with UTF-8, but it's large enough to be it's own separate discussion. Assuming people agree, I'll create a separate ticket.

Highlights.  In addition, a couple things that were in the neighborhood got touched.

- `@Override` added throughout. A couple can't be, because they're in generated code. http://errorprone.info/bugpattern/MissingOverride
- `QueueHandler.java` get a type on the internal Queue.  Just happened to catch it going be.  Nothing changes externally, just carries over the ZKQueue type.
- A couple classes (e.g. `ContainerTracker` and `PersistentAccessTracker`) have fields changed to final for synchronization. http://errorprone.info/bugpattern/SynchronizeOnNonFinalField
- `SimpleFieldTransformation` fixed logic.  Don't remember which Error Prone warning this was, because the answer was to remove the chunk of offending code entirely.
- Lot of things are unboxed (e.g. Boolean to boolean). http://errorprone.info/bugpattern/BoxedPrimitiveConstructor
- Couple of inner classes are made static because they don't depend on the enclosing class at all. http://errorprone.info/bugpattern/ClassCanBeStatic
- `newInstance()` generally changed to `getConstructor().newInstance()`.  Added catches where appropriate. http://errorprone.info/bugpattern/ClassNewInstance
- `FluxTopologyComponent` has try-finally rewritten to use try with resources. Integration tests still work. http://errorprone.info/bugpattern/Finally

There's a couple miscellaneous leftover things, most notably:
```
/Users/jleet/Documents/workspace/incubator-metron/metron-analytics/metron-maas-service/src/main/java/org/apache/metron/maas/service/runner/Runner.java:265: warning: [Finally] If you return or throw from a finally, then values returned or thrown from the try-catch block will be ignored. Consider using try-with-resources instead.
           throw new IllegalStateException("Unable to execute " + script + ".  Stderr is: " + stderr + "\nStdout is: " + stdout);
           ^ 
    (see http://errorprone.info/bugpattern/Finally) Note: Some input files use unchecked or unsafe operations. Note: Recompile with -Xlint:unchecked for details.
```
I wasn't sure enough of what was going on, so I punted, given how large the PR already is.

Along with this, there are a couple instances of http://errorprone.info/bugpattern/TypeParameterUnusedInFormals in helper code that I'm not terribly worried about, but we might want to take a look at (and possibly explicitly suppress)

Testing for this was making sure all the unit and integration tests still work (given that most changes are things like `@Override` that don't have actual execution effect) + ensuring things can still flow through topologies correctly.  The largest change is is the `FluxTopologyComponent`, whose usage is in integration tests and should be covered by that.

In addition to this, we could probably upgrade several of these to errors if we wanted to, specifically SynchronizeOnNonFinalField, BoxedPrimitiveConstructor, ClassCanBeStatic, and ClassNewInstance I believe are all gone.  Unfortunately, because of the generated code MissingOverride is here to stay.  I could do that in either this ticket if we're interested or in a follow-on.

dlyle65535:
That looked like a lot of effort, thanks for taking it on. Did you happen to spin up quick dev?

justinleet:
@dlyle65535 Yeah, I spun up quick dev and made sure things passed through on topologies through to ES. I didn't extensively test other components in quick-dev, given that most of these changes are actually just built-in IDE refactorings (even the catches and combining Exceptions from `getConstructor()` get handled by IntelliJ) and non-behavior changes (`@Overrides`).

Given that maas and the profiler have very minor changes and are covered by integration tests, I didn't spin those up.  If we do want those spun up (or anything else), I can definitely do so.

dlyle65535:
I'm +1 by inspection.

Thanks again!

justinleet:
@dlyle65535 Any preference on making any of the warnings into compile-time errors? Specifically:
SynchronizeOnNonFinalField, BoxedPrimitiveConstructor, ClassCanBeStatic, and ClassNewInstance

Notably, BoxedPrimitiveConstructor and ClassNewInstance both address issues with using code that'll be deprecated in Java 9 anyway.

They can be done with a pretty simple POM edit, e.g.

<compilerArgs>
  <arg>-Xlint:unchecked</arg>
  <arg>-Xep:SynchronizeOnNonFinalField:ERROR</arg>
  <arg>-Xep:BoxedPrimitiveConstructor:ERROR</arg>
  <...>
</compilerArgs>

dlyle65535:
Totally For Erroring on:
SynchronizeOnNonFinalField

Not against it for:
BoxedPrimitiveConstructor
ClassCanBeStatic

Okay with it for ClassNewInstance with additional input. I get why it's harmful, but is it more harmful than reflection in general? Or do you just want to not insert known future deprecated code?

ottobackwards:
+1 by inspection

WRT Errors: do we have any other projects to reference here?

justinleet:
@dlyle65535 @ottobackwards Created https://issues.apache.org/jira/browse/METRON-615 for discussion on errors.  I made it a more general ticket, brought up what Hadoop as a project does (and feel free to add more examples) along with the error prone discussion.  I copied over David's comment here (so please correct me if I made a mistake anywhere involving your opinion).

ottobackwards:
nope, it is fine.   I was just thinking that since we are looking outward for reference and inspiration, we should continue to do that on some of these issues as they come up

PR #390, opened 2016-12-07
METRON-364: Preserve Type for Arithmetic Expressions in Stellar

jjmeyer0:
updated stellar arithmetic expressions to be evaluated in a manner that keeps type. For example, if two integers are added then an integer will be returned.

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/evaluators/ArithmeticEvaluator.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/evaluators/ArithmeticEvaluator.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.metron.common.stellar.evaluators;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.metron.common.dsl.ParseException;
+import org.apache.metron.common.dsl.Token;
+
+import java.util.function.BiFunction;
+
+public class ArithmeticEvaluator {
+
+  public Token<? extends Number> evaluate(BiFunction<Number, Number, Token<? extends Number>> function,
+                                          Pair<Token<? extends Number>, Token<? extends Number>> p) {
+    if (p == null || p.getKey() == null || p.getValue() == null) {
+      throw new IllegalArgumentException();
+    }
+
+    final Number l = p.getKey().getValue();
+    final Number r = p.getValue().getValue();
+
+    return function.apply(l == null ? 0 : l, r == null ? 0 : r);
+  }
+
+  /**
+   * This is a helper class that defines how to handle arithmetic operations. The conversion between number
+   * types is taken for the Java spec: http://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2
+   */
+  public static final class ArithmeticEvaluatorFunctions {
+    public static BiFunction<Number, Number, Token<? extends Number>> addition() {
+      return (Number l, Number r) -> {
+        if (l instanceof Double || r instanceof Double) {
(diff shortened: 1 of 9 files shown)


metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/evaluators/ArithmeticEvaluator.java
```diff
@@ -0,0 +1,81 @@
...
+    final Number l = p.getKey().getValue();
+    final Number r = p.getValue().getValue();
+
+    return function.apply(l == null ? 0 : l, r == null ? 0 : r);
+  }
+
+  public interface ArithmeticEvaluatorFunctions {
+    static BiFunction<Number, Number, Token<? extends Number>> addition() {
+      return (Number l, Number r) -> {
+        if (l instanceof Double || r instanceof Double) {
```
cestella: What about if l or r is Float?

cestella:
Oops also we will want to preserve Longs.  Down converting to an int will lose precision and may end up being inaccurate.  I'd suggest we operate like so:
* If either are a double then operate as doubles
* else if either are float then operate as floats
* else if either are longs then operate as longs
* else operate as ints

justinleet:
I agree on operating Casey's rules, given that they match Java's rules for handling this.

http://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2

jjmeyer0:
The reason I neglected longs and floats is because I didn't see the Stellar dsl having the ability to represent them. For example, the test below fails with a parse exception:

`
  @Test
  public void testLongAddition() throws Exception {
    String query = "2147483650 + 2";
    Assert.assertEquals(2147483652L, run(query, new HashMap<>()));
  }
`

That being said it makes sense to still handle longs and floats, if Stellar will at some point. If not that, at the very least if this evaluator sees a long or float it should error out.

Should I add support for longs/floats and then create a Jira for Stellar to add them? Or am I mistaken and Stellar already has support?


Thanks for the feedback.

cestella:
You can return them from as results from stellar functions. Longs are
output from time stamp functions for instance.

metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/evaluators/ArithmeticEvaluator.java
ottobackwards: My only question here, and I may be out of my depth, is does this functionality match the behavior of java itself?  Or other languages?  Would the result of operations on mixed types be consistent (rule of least surprise)?  For example ( and again I'm just imagining here ) Float + Long = Float.    Long + Float = Long.

Hopefully this makes sense

ottobackwards:
[moving this to unbury it]
My only question here, and I may be out of my depth, is does this functionality match the behavior of java itself? Or other languages? Would the result of operations on mixed types be consistent (rule of least surprise)? For example:

 Float + Long = Float
Long + Float = Long

Hopefully this makes sense

cestella:
@ottobackwards Yeah, that was my concern too.  The suggestion I made was based around the Java spec (http://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2 as @justinleet linked earlier).  This is intended to make it follow the rule of least surprise (I think ;).  In your example, Long + Float  should be Float by my suggestion.

ottobackwards:
@cestella I see it now, but not when I made the comment.  Sorry for adding confusion.

+1 by inspection

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/StellarTest.java
```diff
@@ -56,6 +56,37 @@
...
+  public void addingDoubleAndIntegerWhereSubjectIsDoubleShouldYieldADouble() throws Exception {
+    String query = "2.1 + 1";
+    Assert.assertEquals(3.1, run(query, new HashMap<>()));
+  }
+
+  @Test
+  public void addingDoubleAndIntegerWhereSubjectIsIntegerShouldYieldADouble() throws Exception {
+    String query = "1 + 2.1";
+    Assert.assertEquals(3.1, run(query, new HashMap<>()));
+  }
```
cestella: Could we ensure that we have the following test cases and they return the proper types:
* float + long returns a float
* long + float returns a float
* double + int returns a double
* int + double returns a double
* float + double returns  a double
* double + float returns a double
* long + double returns a double
* double + long returns a double
* long + int returns a long
* int + long returns a long
* int + int returns int
* float + float returns float
* double + double returns double
* long + long returns long

cestella: Obviously, the same set for multiplication and division.  I'd parameterize that case ;)

ottobackwards: Casey - just like with the stringfunctionstest - I'm not sure that all of this should be in stellartest.  How do you feel about breaking it out?

cestella: @ottobackwards Definitely; agreed.  We should have a `StellarArithmeticTest` where we have this.

jjmeyer0:
I believe I have updated all tests and addressed each of the concerns. Thanks again for all the help!

cestella:
Cool!  Dig the tests.  +1 by inspection

PR #391, opened 2016-12-08
METRON-618 Eliminate Javac Warnings in metron-analytics

justinleet:


justinleet:
Cleaned up most of the warnings that come from the compiler in metron-analytics.  What's left is a couple Error Prone issues (primarily related to Charset, which has a separate ticket), and multiline annotations is tied to Java 6 (and not something we can much do about, nor am I worried about given that it still works).

- For the `toArgs()` methods, swapped out the `Utils` class for `Arrays.asList()`.  This itself is `@SafeVarargs`, and since we just iterate over it, out methods themselves should be fine as `@SafeVarargs`.
- Said `Utils` class is deleted entirely now.
- Variety of `@SuppressWarnings("unchecked")` added.  These are typically things derived from JSONObject or similar, because the JSON library we use only has raw types.  I tried to keep these at the minimum scope I could.
- Couple of instances of code get generics applied to them appropriately, including one that was previously flagged to suppress raw types (once a couple other things that were warnings got fixed with types, that could carry through)

Tested this by spinning up quickdev and running up MaaS and the profiler and making sure output came through both of them.

ottobackwards:
+1 on inspection

PR #392, opened 2016-12-09
METRON-616: Added support for float and long literals in Stellar.

jjmeyer0:
Added support for double and long literals in Stellar. The double and long literals are based on java's longs/doubles. However, octal, binary, and hexadecimal formats were not implemented. Int literals were also updated to not allow numbers such as 0001, 009, etc.


ottobackwards:
Just a question to start - does this only work if the input value has the marking ( D,L etc )?

jjmeyer0:
@ottobackwards Depends. It should work very similar to java. For example, doubles will not require a D/d (eg. .4, 0.5, 1., etc). All these examples are doubles. Floats require an f/F, and to be a long, it requires an l/L. It should be very similar to https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.10.1

ottobackwards:
So - the case I am thinking of is a stellar transform on an incoming string field that contains a number, but that number will not be marked - so only doubles would work unless I'm mistaken.  I am not sure that requiring the fields to be marked up works so well when you don't have control of the sources.

By specifying TO_XXXX are you explicitly saying you want it treated at a string of number marked as type?   Why impose the restriction of having to have the string marked?

cestella:
I think it should work as follows for literals:
* `1234` should be an int
* `1234.0` should be a double
* `1234L` should be a long
* `1234f` should be a float

My only question is whether `12345678910` should be a long.  My instinct is that it should probably *should*, but I'd like feedback on that assertion.

Numbers coming in via the parser or as a result of a Stellar function are the types that the function or parser creates them as.

ottobackwards:
@cestella  - should I be able to do TO_FLOAT(field) when the field is a string and could be 1231 or 12.31?

cestella:
@ottobackwards  Yep, the `TO_$TYPE` stellar functions should work with inputs that are strings or numbers of other types.  To wit,
* `TO_FLOAT('1231')` should yield a float (`1231.0f`) from the input string
* `TO_FLOAT('12.31')` should yield a float(`12.31f`) from the input string
* `TO_FLOAT(1231)` should yield a float (`1231.0f`) from the input int
* `TO_FLOAT(12.31)` should yield a float(`12.31f`) from the input double

ottobackwards:
OK - thanks for explaining

jjmeyer0:
@cestella I thought about the case where a number such as '12345678910' should be considered a long or not. The biggest reason why I chose not to go this way, was because we were basing many things on the way Java does things, and Java would throw a compile time error if this example was not followed by any l/L. Also, I had some concerns in a scenario where two integers were added, a long was automatically created from them, and then passed to a function that only accepts integers. In this case a very interesting error may occur, and confuse end users. Not sure if any of those are valid, but that was my thought process. What do you think?

jjmeyer0:
and I'll add some tests that specifically test expected type of these number literals.

cestella:
@jjmeyer0 I think that's fine; mimicking java is probably the right move here.

cestella:
I think we need the following test cases around numerical literals:
* In conditionals:
  * `1.0f < 2.0f`
  * `1.0l < 3.0f`
  * `1 < 3.0f`
  * `1.0 < 3.0f`
  * `if 1.0l < 3.0f then 'true' else 'false'`
* In function args:
  * `TO_FLOAT(1231)`
  * `TO_FLOAT(12.31)`
  * `TO_FLOAT(12.31f)`
  * `TO_FLOAT(12L)`
  * `TO_FLOAT(12.31) < 10.2f`
* In arithmetical expressions:
  * `1.2f + 3.7`
  * `12L*(1.2f + 7)`
  * `TO_FLOAT(12.2) * (1.2f + 7L)`

Anyone have any others?

jjmeyer0:
@cestella @ottobackwards is there anything else you all would like to see on this one?

cestella:
Nope, I like it.  +1 from me by inspection

ottobackwards:
+1 by inspection

PR #396, opened 2016-12-14
METRON-625: Parser Filters cannot be specified from the sensor config

cestella:
Parser filters cannot be specified because of a logic error in the prepare for the ParserBolt. Specifically, we are loading the filter from the config only if the filter member variable is null. Unfortunately, this will never be null, because it is initialized to a default implementation.


metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -76,12 +77,15 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
     if(getSensorParserConfig() == null) {
       filter = new GenericMessageFilter();
     }
-    else if(filter == null) {
+    if(filter == null) {
       getSensorParserConfig().getParserConfig().putIfAbsent("stellarContext", stellarContext);
-      filter = Filters.get(getSensorParserConfig().getFilterClassName()
-              , getSensorParserConfig().getParserConfig()
-      );
+      if(!StringUtils.isEmpty(getSensorParserConfig().getFilterClassName())) {
```
ottobackwards: If it is null, don't we want to put in the GenericMessageFilter() as well?

cestella: `GenericMessageFilter` just returns true all the time.  Instead, we just check to see if filter is null in the loop and pass things by if it's null.  Removes an unnecessary function call.  If you think it's more clear to set filter to `GenericMessageFilter` if it's not otherwise specified, I can do that.

ottobackwards: Well, it was confusing to see.  Why does GenericMessageFilter exist at all then?  At the very least I would think to put a comment in to explain why you are doing it that way for maintenance purposes if you leave it like that.

cestella: Agreed.  I removed `GenericMessageFilter` for clarity.

cestella:
What do you think of renaming `QUERY` to `STELLAR` like we did with `fieldTransformation`?  I don't think anyone is using this since it doesn't work, so it would likely not be a regression. ;)

ottobackwards:
Where is QUERY documented?  I can't find it.  What I did find is this in the Parser readme:

Message Filter - a filter defining which messages can be dropped. This feature is only present in the Java paerer adapters

Maybe you can fix the spelling of Java paerer?

cestella:
@ottobackwards LOL yes, this is an entirely undocumented feature evidently.  I don't think it's worked since we moved the parsers from Flux.  I just happened to trip upon it trying to do something else.  I'll add some documentation around it and make sure it works.  I will also create a follow-on JIRA to make sure that the filter can be changed on the fly (it cannot now and that fix would be out of scope for this JIRA).

ottobackwards:
Are there upgrade issues with the QUERY change?

cestella:
@ottobackwards I don't think so because it never worked (at least not in any of our apache releases; I checked yesterday).  Nobody could be depending on this functionality unless they fixed it like I have here.

ottobackwards:
So it there is nothing with that value out there so there are no serialization issues

cestella:
@ottobackwards There shouldn't be, no.

kylerichardson:
+1 by inspection. Thanks for fixing!

PR #401, opened 2016-12-21
METRON-637: Add a STATS_BIN function to Stellar.

cestella:
When passing parameters to models, it's often useful to pass the binned representation of a variable based on an empirical statistical distribution, rather than the actual variable. This function should accept a set of percentile bins and a statistical sketch and a value. It should return the index where the percentile of the value falls.

For instance, consider the value 17 who is percentile 27. If we use 25, 75, 95 to define our bins, this function would return 1, because its percentile, 27, is between 25 and 75.


metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
```diff
@@ -425,4 +428,74 @@ public Object apply(List<Object> args) {
...
+   * Calculates the statistical bin that a value falls in.
+   */
+  @Stellar(namespace = "STATS", name = "BIN"
+          , description = "Computes the bin that the value is in based on the statistical distribution."
+          , params = {
+          "stats - The Stellar statistics object"
+          , "value - The value to bin"
+          , "range? - A list of percentile bin ranges (excluding min and max) or a string representing a known and common set of bins.  " +
+          "For convenience, we have provided QUARTILE, QUINTILE, and DECILE which you can pass in as a string arg." +
+          " If this argument is omitted, then we assume a Quartile bin split."
```
mattf-horton: This is a list of bounds rather than ranges.  More precisely, the N'th bound specifies the closed upper bound of the N'th bin, where N is zero indexed; and if there is a list of M bounds then there are M+1 bins, numbered 0 through M. (Bin M is the bin for values greater than last bound in the list.)

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
```diff
@@ -425,4 +428,74 @@ public Object apply(List<Object> args) {
...
+  }
+          , returns = "Which bin the value falls in such that bin < value < bin + 1"
```
mattf-horton: One of the "<" has to be "<=", please.  From code, it's the upper one.

Suggest phrasing as:
returns = "Which bin N the value falls in such that bound(N-1) < value <= bound(N). " +
"No min and max bounds are provided, so values smaller than the 0'th bound go in the 0'th bin, " +
"and values greater than the last bound go in the M'th bin."

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
```diff
@@ -425,4 +428,74 @@ public Object apply(List<Object> args) {
...
+      }
+
+      double prevPctile = stats.getPercentile(bins.get(0));
+
+      if(value <= prevPctile) {
+        return 0;
+      }
+      for(int bin = 1; bin < bins.size();++bin) {
+        double pctile = stats.getPercentile(bins.get(bin));
+        if(value > prevPctile && value <= pctile) {
```
mattf-horton: Really no need to check the lower bound here, is there? :-)
Thus, you can dump "prevPctile" and start at bin = 0.

metron-analytics/metron-statistics/README.md
```diff
@@ -112,6 +112,13 @@ functions can be used from everywhere where Stellar is used.
   * Input:
     * stats - The Stellar statistics object
   * Returns: The variance of the values in the window or NaN if the statistics object is null.
+* `STATS_BIN`
+  * Description: Computes the bin that the value is in based on the statistical distribution. 
+  * Input:
+    * stats - The Stellar statistics object
+    * value - The value to bin
+    * range? - A list of percentile bin ranges (excluding min and max) or a string representing a known and common set of bins.  For convenience, we have provided QUARTILE, QUINTILE, and DECILE which you can pass in as a string arg. If this argument is omitted, then we assume a Quartile bin split.
```
cestella: It would, for sure, but probably not called `STATS_BIN`.  I was thinking of adding a proper `BIN` function as a follow-on and refactoring this one to use it.  Or do you think we should just bite the bullet and create it all here?

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
cestella: Yep, I'll fix that.  Good clarification.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
cestella: yup, will do; great suggestion.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
cestella: haha yes, you can.  I'll correct.

metron-analytics/metron-statistics/README.md
mattf-horton: It seems like it would be barely more effort than writing another StellarFunction wrapper, but maybe I'm being optimistic again.  I would support including both here, but if you prefer to separate them I'm fine with that.

cestella: The signature will be `BIN(value, list_of_bounds)` and `STATS_BIN` will call it.  Does that make sense to you?

mattf-horton: Sure!

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
```diff
@@ -356,6 +357,44 @@ public void testSkewness() throws Exception {
...
+    statsBinRunner(splits, null);
+  }
+
+  public void statsBinRunner(List<Double> splits, String splitsName) throws Exception {
+    int bin = 0;
+    for(Double d : stats.getSortedValues()) {
+      StatisticsProvider provider = (StatisticsProvider)variables.get("stats");
+      if(bin < splits.size()) {
+        double percentileOfBin = provider.getPercentile(splits.get(bin));
+        if (d > percentileOfBin) {
```
mattf-horton: Sorry if I'm wrong here, but I couldn't find the definition of stats.getSortedValues().
Isn't this line 380 comparing a raw value `d` to a percentile value `percentileOfBin` ?

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
```diff
@@ -356,6 +357,44 @@ public void testSkewness() throws Exception {
...
+          //we aren't the right bin, so let's find the right one.
+          // Keep in mind that this value could be more than one bin away from the last good bin.
+          for(;bin < splits.size() && d > provider.getPercentile(splits.get(bin));bin++) {
+
```
mattf-horton: Right, I'm not getting this.  The bound is being fed into provider.getPercentile, which returns digest.quantile(p/100.0).  Isn't this what should be done to "value" or "d", not the bound?
And the bound should just be divided by 100?  Maybe I stayed up too late last night.

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
```diff
@@ -356,6 +357,44 @@ public void testSkewness() throws Exception {
...
+  }
+
+  public void statsBinRunner(List<Double> splits) throws Exception {
+    statsBinRunner(splits, null);
+  }
+
+  public void statsBinRunner(List<Double> splits, String splitsName) throws Exception {
+    int bin = 0;
+    for(Double d : stats.getSortedValues()) {
+      StatisticsProvider provider = (StatisticsProvider)variables.get("stats");
```
cestella: That can be done, for sure.

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
cestella: We are walking the set of samples in sorted order and keeping track of the bin that the value is in as we walk.  When we move from one value to the other, we need to check the percentile associated with the current bin.  If it is larger than the current percentile, then we find the proper bin.

Take for example the following situation.  We will consider bounds `10` and `20`.  The  percentile mapping is as follows:
* 10th percentile is at point `17`
* 20th percentile is at point `50`

Let's take the values (in sorted order): `10`, `20`, `75`

For `10` we start with a bin of `0`.  We compare it to our current bin index, namely the 10th percentile and it's still less than the `17`, so it fits in bin `0`.

For `20` we are still at bin `0`, but now we compare it to the 10th percentile (`17`) and find that it is larger, so we need to find the next bin that is larger than `20`, which is bin `1` because the 20th percentile is 50 (so it's between the 10th percentile and 20th percentile)

For `75`, we compare to the 20th percentile (because we're at bin `1`) and it's larger, so we need to find the next percentile.  Unfortunately there's no other bound to go to (we're at the end), so it sticks on `2` or the length of bounds.

Because we can walk the sorted set, we can just move across the bins in sequential order as we pass the boundary marks.  This gives us the expected values for this test.

Does that clear things up?

metron-analytics/metron-statistics/README.md
mattf-horton: Sure!

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
cestella: Hopefully the explanation above makes things clearer.

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
mattf-horton: I see, so provider.getPercentile() actually takes a percentile value (as a number between 0 and 99.99) and returns a corresponding raw value?  That makes sense if so.  Please confirm.

cestella: Yes, `provider.getPercentile(x)` returns the value at percentile `x`.  So, to get the median, you'd call `provider.getPercentile(50.0)`. 

If you're thinking of the inverse (give me the percentile that a value falls at), we don't have that dude yet, but it's called the cumulative distribution function or (`ecdf` in R parlance).

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
mattf-horton: Yup, thanks.  Then the block (lines 378-387) could be better stated as
```
while ( bin < splits.size()  &&  d > provider.getPercentile(splits.get(bin)) ) {
       //increment the bin number until it includes the target value, or we run out of bins
       bin++; 
}
```

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
mattf-horton: Thanks.

cestella:
Thanks for the feedback, @mattf-horton!  I went ahead and incorporated your changes.  For posterity and so this gets replicated to the JIRA, I added a `BIN` function that just takes a list of bounds, not a list of percentiles to compute.  `STATS_BIN` could be encoded with `BIN` if we had a `MAP` function (i.e. `STATS_BIN` == `BIN( value, MAP( &STATS_GET_PERCENTILE(stats, _ ), [ 25.0, 50.0, 75.0])` where `MAP` takes a function pointer and applies it to a collection).  

We do not have this capability yet in Stellar, but it may be worth considering to enable these kinds of use-cases.  Just a thought.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
```diff
@@ -425,4 +428,61 @@ public Object apply(List<Object> args) {
...
+      }
+
+      public static List<Double> getSplit(Object o) {
+        if(o instanceof String) {
+          return BinSplits.valueOf((String)o).split;
+        }
+        else if(o instanceof List) {
+          List<Double> ret = new ArrayList<>();
+          for(Object valO : (List<Object>)o) {
+            ret.add(ConversionUtils.convert(valO, Double.class));
```
mattf-horton: During this step we need to validate that the bounds list is strictly increasing.

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
```diff
@@ -373,15 +373,16 @@ public void statsBinRunner(List<Double> splits) throws Exception {
...
       if(bin < splits.size()) {
         double percentileOfBin = provider.getPercentile(splits.get(bin));
         if (d > percentileOfBin) {
           //we aren't the right bin, so let's find the right one.
           // Keep in mind that this value could be more than one bin away from the last good bin.
-          for(;bin < splits.size() && d > provider.getPercentile(splits.get(bin));bin++) {
-
+          while ( bin < splits.size()  &&  d > provider.getPercentile(splits.get(bin)) ) {
+            //increment the bin number until it includes the target value, or we run out of bins
+            bin++;
```
mattf-horton: This whole block:
```
      if(bin < splits.size()) {
        double percentileOfBin = provider.getPercentile(splits.get(bin));
        if (d > percentileOfBin) {
          //we aren't the right bin, so let's find the right one.
          // Keep in mind that this value could be more than one bin away from the last good bin.
          while ( bin < splits.size()  &&  d > provider.getPercentile(splits.get(bin)) ) {
            //increment the bin number until it includes the target value, or we run out of bins
            bin++;
          }
        }
      }
``` 
can be replaced by:
```
      while ( bin < splits.size()  &&  d > provider.getPercentile(splits.get(bin)) ) {
        //increment the bin number until it includes the target value, or we run out of bins
        bin++;
      }
```

cestella:
Hmm good point.  How about keeping a guava timed cache in the STATS_BIN function that maps the stats object and raw bounds list to the computed percentile bounds list?  That should do the trick, no?

cestella:
Ok, so a couple of things.  I ran a quick perf test on this function as it stands.  On my macbook from 4 years ago, I ran the `STATS_BIN` function a million times with random values and it took ~5.5s.  Even at a throughput of 1M messages per second, if we assume that the messages are spread across the cluster, I think this keeps up.

Now, *EVEN GIVEN THIS*, normally I would go through the effort here of adding a caching layer to save us the computation of the percentile, but it's actually quite difficult to figure out when two `StatisticsProvider` objects are equivalent without resorting to actually calling percentiles.  I could do things like use non-computed attributes (number of data points, sum of data points, average and sum of data points), but the former Math grad student in me was uncomfortable in that.  It's just very hard to not be absolutely sure that you couldn't have all of those attributes the same and be different distributions.

I think given these things together that I'm going to recommend to cross the caching bridge when we come to it.  Now, it won't take much to convince me to go the other direction, so if you (or anyone else, really ;) feels strongly @mattf-horton , I'll go ahead and tackle that dude as best I can.

I'm going to correct the rest of your comments and check in the performance test I ran, so you can see that I have nothing up my sleeves and so it can be run periodically.

cestella:
I made the corrections suggested and also, as a compromise for performance, I changed things up a bit so that the list munging is optimized a bit:
* A new list of bins is not created per call in either `BIN` or `STATS_BIN`
* We use the list as passed, converting the `Number` to `Double` lazily.
* I do the monotonic increasing check as needed rather than prior to the function.

All that being said, caching would increase the performance, but I think we're in a decent spot at the moment.  I rejiggered the performance driver to give us a distribution of performance characteristics.  Current run with this change is at:
`Min/25th/50th/75th/Max Milliseconds: 2687.0 / 2700.5 / 2716.0 / 2733.5 / 3730.0`

Thoughts?

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
cestella: Done

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
cestella: Done, but not exactly in that spot.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/MathFunctions.java
```diff
@@ -59,4 +60,49 @@ public boolean isInitialized() {
...
+           "value - The value to bin"
+          , "bounds - A list of value bounds (excluding min and max) in sorted order."
+                    }
+          ,returns = "Which bin N the value falls in such that bound(N-1) < value <= bound(N). " +
+          "No min and max bounds are provided, so values smaller than the 0'th bound go in the 0'th bin, " +
+          "and values greater than the last bound go in the M'th bin."
+  )
+  public static class Bin extends BaseStellarFunction {
+
+    public static int getBin(double value, List<Double> bins, Function<Integer, Double> boundFunc) {
```
cestella: Thanks! :)

ottobackwards:
Do these methods *have* to run on every tuple?  Can you set some kind of sample rate parameter to give a way to play with performance?

cestella:
Hmm, well, the idea is to get the statistical bin for a given value.  The sampling portion is happening in the profiler as you construct the statistical distribution.  This is just querying the already constructed distribution.  This is intended to be an input into a model and to give a sense of where a value falls in the distribution of values that have preceded it, so if you want a message to be used by your model, then it'll need to be computed and passed.

I'll give you an example that may help clarify.  Let's say I'm building a really, really naive model over DNS requests to determine if a DNS request is being made for a synthetic domain created by a botnet.  One of the features I might be interested in may be the length of the domain.  However, I may also want to get a sense of if this domain's length is outside of the norm or not.  To do that, I'd create a profile that captures the statistical distribution of the length of the domains that have been requested in the past and when I'm calling the model (which is deployed via Model as a Service), I can pass the statistical bin that the length falls into (e.g. between the min-25th percentile, 25 - 50th percentile, 50th - 75th percentile or 75th - 95th percentile, 95th - max) by using this function.  So, every DNS request really needs to be scored in this scenario.

I hope that makes sense.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/MathFunctions.java
```diff
@@ -59,4 +60,45 @@ public boolean isInitialized() {
...
+          , "bounds - A list of value bounds (excluding min and max) in sorted order."
+                    }
+          ,returns = "Which bin N the value falls in such that bound(N-1) < value <= bound(N). " +
+          "No min and max bounds are provided, so values smaller than the 0'th bound go in the 0'th bin, " +
+          "and values greater than the last bound go in the M'th bin."
+  )
+  public static class Bin extends BaseStellarFunction {
+
+    public static int getBin(double value, int numBins, Function<Integer, Double> boundFunc) {
+      double lastBound = Long.MIN_VALUE;
```
mattf-horton: Double.NEGATIVE_INFINITY or -Double.MAX_VALUE would be better than Long.MIN_VALUE

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/MathFunctions.java
```diff
@@ -59,4 +60,45 @@ public boolean isInitialized() {
...
+      for(int bin = 0; bin < numBins;++bin) {
+        double bound = boundFunc.apply(bin);
+        if(bound < lastBound) {
+          throw new IllegalStateException("Your bins must be monotonically increasing");
```
mattf-horton: If the stats model is sparse, can two bounds have the same value, and it's okay?  Otherwise,
although it's pedantic, we really should test for (bound <= lastBound) and say
"Your bins must be strictly increasing" rather than monotonically increasing.
But I don't fully understand how the stats models work, so maybe it's better this way.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/MathFunctions.java
```diff
@@ -59,4 +60,45 @@ public boolean isInitialized() {
...
+        }
+        if(value <= bound) {
+          return bin;
+        }
+        lastBound = bound;
+      }
+      return numBins;
+    }
```
mattf-horton: Beautiful.  Definitely pared down to the minimum.

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StatisticalBinningPerformanceDriver.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * This is a driver to drive evaluation of the performance characteristics of the STATS_BIN stellar function.
+ * It gets the distribution of the time it takes to calculate the bin of a million random numbers against the quintile bins
+ * of a statistical distribution of 10000 normally distributed reals between [-1000, 1000].
+ *
+ * On my 4 year old macbook pro, the values came out to be
+ *
+ * Min/25th/50th/75th/Max Milliseconds: 2687.0 / 2700.5 / 2716.0 / 2733.5 / 3730.0
+ */
```
mattf-horton: Great to have this.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/MathFunctions.java
cestella: You're right, strictly increasing is correct.

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/MathFunctions.java
cestella: done

cestella:
One last bug here.  I was actually right originally to insist on just monotonic increasing, rather than strict increasing for the bins.  Consider the situation where I create a distribution of a single value `d`.  Every percentile will be `d`.  In that case I would want the function to return `0` for `x` where `x <= d` and the number of bins for `x` where `x > d`.

I created a test to validate this behavior.

mattf-horton:
+1 lgtm

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/StellarStatisticsFunctions.java
```diff
@@ -425,4 +428,57 @@ public Object apply(List<Object> args) {
...
+    @Override
+    public Object apply(List<Object> args) {
+      StatisticsProvider stats = convert(args.get(0), StatisticsProvider.class);
+      Double value = convert(args.get(1), Double.class);
+      final List<Number> bins = args.size() > 2?BinSplits.getSplit(args.get(2)):BinSplits.QUARTILE.split;
+
+      if (stats == null || value == null || bins.size() == 0) {
+        return -1;
+      }
+      return MathFunctions.Bin.getBin(value, bins.size(), bin -> stats.getPercentile(bins.get(bin).doubleValue()));
```
mmiklavc: Nice suggestion by Matt. And I like the math bin code reuse and ability to plug in a stats function provider.

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
```diff
@@ -356,6 +357,49 @@ public void testSkewness() throws Exception {
     assertEquals(stats.getSkewness(), (Double) actual, 0.1);
   }
 
+  @Test
+  public void testStatsBin() throws Exception {
```
mmiklavc: It's not immediately obvious to me what's being tested here - is this checking that default enums' list of splits is syntactically sound? What would make this test fail?

mmiklavc:
This looks really good! Other than a minor question on one of the stats bin unit tests, it looks good to me. +1 pending comment/clarification on that item.

metron-analytics/metron-statistics/src/test/java/org/apache/metron/statistics/StellarStatisticsFunctionsTest.java
cestella: This test is testing that the `STATS_BIN` function operates correctly by taking a sorted list of numbers, walking down it and ensuring that the `STATS_BIN` for each number yields the correct bin.  This is a reasonable test because we are not actually computing the bin so much as recognizing since the numbers are sorted, the bin will increase at the percentile boundaries, thus we have the expected bin without recreating the computation in the `STATS_BIN` function.

cestella: Given that it's not entirely obvious what this is doing and why it is a valid test, do you think it merits a comment with the above in the test?

ottobackwards: That would make sense

cestella: You got it :)

mmiklavc:
+1

PR #402, opened 2016-12-21
METRON-639: The Network Stellar functions need to have better unit testing

cestella:
We have very little unit test coverage around the Networking functions in Stellar at the edge level. When diving a bit deeper on real data, I found a number of bugs around:
* Domains with TLDs that are not part of the proper set of TLDs
* URIs with schemes that Java doesn't know about.
* IN_SUBNET takes multiple CIDRs, but only evaluates the first one.
* Generally calling validate on these methods can be unsafe because they do not handle null arguments correctly.

This just shores up the functions and provides more explicit and detailed unit testing.  If I missed a test case or two, let me know and I'll add it in.


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
```diff
@@ -163,12 +163,16 @@ public Object apply(List<Object> objects) {
   public static class URLToPort extends BaseStellarFunction {
     @Override
     public Object apply(List<Object> objects) {
```
ottobackwards: Can we be consistent with the variables, such that a URI is uri and URL is url?

metron-platform/metron-common/src/test/java/org/apache/metron/common/stellar/network/NetworkFunctionsTest.java
```diff
@@ -0,0 +1,164 @@
...
+  @Test
+  public void inSubnetTest_multiple() {
+    assertTrue((Boolean) run("IN_SUBNET(ip, cidr1, cidr2)"
+                            , ImmutableMap.of("ip", "192.168.1.1"
+                                             , "cidr1", "192.168.0.0/24"
+                                             , "cidr2", "192.168.1.0/24")
+              ));
+  }
+
+  private static void runSimple(String function, String argument, Object expected) {
```
ottobackwards: I think this should be in the StellarProcessorUtils

cestella:
Thanks for the review @ottobackwards !  I extracted out the `runSimple` utility, generalized it a bit and put it in `StellarProcessorUtils`  Let me know if you like what you see there.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
cestella: I did make these changes to the variables, but did not change the docs since the functions are literally called `URL_*`.

ottobackwards:
+1 by inspection

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/NetworkFunctions.java
```diff
@@ -182,8 +186,8 @@ public Object apply(List<Object> objects) {
   public static class URLToPath extends BaseStellarFunction {
```
justinleet: Can you add testing around this?  In particular, make sure that fragments are handled correctly (`URI` has a separate `getFragment()`, so I assume that they are not returned by `getPath()`.  I could be wrong about this, though, I haven't run a test on it.).  If this is the case, we may also want/need a `TO_FRAGMENT` function that handles this appropriately, although I'm fine with that being a separate conversation/jira.

cestella:
@justinleet It's a good point you make and a very good catch.  I am using URIs specifically because if Java doesn't know about the protocol scheme, then it bails.  I'd hate for someone to ingest URLs of a custom scheme or a scheme that is more esoteric than is supported by Java and have it bail.

That being said, after further investigation, URLs are not a subset of URIs.  As per http://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn/1984225#1984225:

> Not "all URLs are URIs". It depends on the interpretation of the RFC. For example in Java the URI parser does not like [ or ] and that's because the spec says "should not" and not "shall not".

I'll have to consider whether or not I want to revert back to URL's and make a comment for further study/analysis or whether I want to try something else here.

cestella:
@justinleet I thought about it a bit more and decided that a sensible approach would be to validate the URL via commons validator and THEN parse using java.net.URI.  This would mean that URIs which are not URL's would not parse, but URLs would parse regardless of scheme.

What other test cases would you suggest (preferably with example URLs, if you don't mind)?

justinleet:
@cestella
Do you know of an example URL that isn't a URL in Java, e.g. with the square brackets issue?  I didn't see one in that thread, but I could have just missed it.  Are we sure commons validator wouldn't have the same issue?  I'm fine with validating them and then parsing if we're stuck dealing with Java implementation quirks.

Specifically on the square brackets issue, it sounds like (and this is way out of my wheelhouse, so forgive my potential ignorance) that these characters (and others) qualify as unwise (https://www.ietf.org/rfc/rfc2396.txt) and should be escaped, which is why Java is strict about them.

Could we appropriately escape the offending characters and end up with a correct URI? If commons has something to validate a URL, is there something to do that escaping for us?  If that works, do we potentially want to be doing operations on URIs instead of URLs, e.g. `URI_TO_PATH`, `URI_TO_HOST`?  Obviously, this raises the question of what we do the URL* functions and coexistence.

In terms of test cases, I'd want to see something with the unwise character issue (but I don't have an example of that myself).  I can help you try to dig one up if you want.  In addition, I'd like to see something with a fragment defined (e.g. `http://java.sun.com/index.html#chapter1`) and make sure that the functions work appropriately (mostly just tests for `TO_PATH`).

cestella:
Ok, I'm reverting this back to just using java's URL class.  The URL validator is a bit picky too.  If this becomes an issue (dealing with nonstandard protocols, we can cross that bridge in another PR).

cestella:
@justinleet You ok with me just using Java URL?

justinleet:
Yep, I'm good with it.

Thanks for contributing and running through this. +1

PR #403, opened 2016-12-22
METRON-640: Add a Stellar function to compute shannon entropy for strings

cestella:
A common feature used for models (especially DGA models) is shannon entropy of strings. We should have the ability to compute it in stellar.


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -284,4 +287,36 @@ private static Object fill(FillDirection direction, Object inputObject, Object f
     }
     return org.apache.commons.lang.StringUtils.rightPad(input,requiredLength,fill);
   }
+
+  @Stellar( namespace="STRING"
+          , name="ENTROPY"
+          , description = "Computes the base-2 shannon entropy of a string"
```
mmiklavc: Can we link to the paper?

cestella: Good catch; this is, at this point, pretty classical work, so I link to the wikipedia definition rather than claude shannon's original work.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -284,4 +287,36 @@ private static Object fill(FillDirection direction, Object inputObject, Object f
...
+  )
+  public static class Entropy extends BaseStellarFunction {
+    @Override
+    public Object apply(List<Object> strings) {
+      /*
+      Shannon entropy is defined as follows:
+      \Eta(X) = - \sum(p(x_i)*log_2(p(x_i)), i=0, n-1) where x_i are distinct characters in the string.
+       */
+      Map<Character, Integer> frequency = new HashMap<>();
+      String input = (String)strings.get(0);
```
mmiklavc: Super minor, but we should probably catch the no-arg provided NPE scenario and throw new IllegalStateException showing usage to the user when they fail to provide an arg.

cestella: Good catch; I did an `IllegalArgumentException`

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
```diff
@@ -32,124 +33,143 @@
...
+      thrown = true;
+      Assert.assertTrue(pe.getMessage().contains("expects three"));
     }
+    Assert.assertTrue(thrown);
+  }
+
+  @Test
+  public void shannonEntropyTest() throws Exception {
+    //test empty string
+    Assert.assertEquals(0.0, (Double)run("STRING_ENTROPY('')", new HashMap<>()), 1e-6);
```
mmiklavc: The specified equality delta is 1e-6, but is there a possibility that the expected return value for these cases is anything other than 0.0? I get it for imprecise double calculations, but I'm wondering if we would want to allow for that for these base cases.

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
```diff
@@ -32,124 +33,143 @@
...
+    Now consider the string aaaaaaaaaabbbbbccccc or 10 a's followed by 5 b's and 5 c's.
+    The probabilities of each character is as follows:
+    p(a) = 1/2
+    p(b) = 1/4
+    p(c) = 1/4
+    so the shannon entropy should be
+      -p(a)*log_2(p(a)) - p(b)*log_2(p(b)) - p(c)*log_2(p(c)) =
+      -0.5*-1 - 0.25*-2 - 0.25*-2 = 1.5
+     */
+    Assert.assertEquals(1.5, (Double)run("STRING_ENTROPY(foo)", ImmutableMap.of("foo", "aaaaaaaaaabbbbbccccc")), 1e-1);
```
mmiklavc: How'd you come up with the delta value?

mmiklavc:
Short and sweet, nice contribution. Just a couple questions and it looks ready to go.

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
cestella: Answering both questions around equality bounds here.  The short answer was that it was just a mistake.  I've changed both of them to `Double.MIN_VALUE`.  In those calculations, it should be as near to exact as a double precision number gets.  I would have dropped the bounds, but the non-epsilon'd `assertEquals` is deprecated.  I guess, which would you prefer, a deprecated call, an `assertTrue` or an assertEquals with the minimum possible epsilon?  I chose the final one, but if you think otherwise, I'm not married to that decision.

cestella: I also could use `0.0` as the epsilon in these cases if you think that's clearer.

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
```diff
@@ -32,124 +33,143 @@
...
+    Now consider the string aaaaaaaaaabbbbbccccc or 10 a's followed by 5 b's and 5 c's.
+    The probabilities of each character is as follows:
+    p(a) = 1/2
+    p(b) = 1/4
+    p(c) = 1/4
+    so the shannon entropy should be
+      -p(a)*log_2(p(a)) - p(b)*log_2(p(b)) - p(c)*log_2(p(c)) =
+      -0.5*-1 - 0.25*-2 - 0.25*-2 = 1.5
+     */
+    Assert.assertEquals(1.5, (Double)run("STRING_ENTROPY(foo)", ImmutableMap.of("foo", "aaaaaaaaaabbbbbccccc")), 0.0);
```
mattf-horton: It does seem that an empty string should have entropy exactly == 0.0 (line 160).

In line 173, I assume the argument is that a string with per-letter probabilities that are small exact powers of 2 will have its entropy calculated exactly.  I suspect that's true, but one would have to know the implementation to be sure.

Otherwise one would have to use a delta of 4*Math.ulp(Double(1.5)) , I think.
But it must pass as is, so it's good :-)

cestella: yeah, I suspect the reason why we can get away with such a small epsilon here is that the frequencies are of the form `(1/2)^n`.  I was a bit shocked it worked, frankly.

mattf-horton: Yes, the question is whether the underlying standard requires log base 2 of (1/2)^n to be exactly integral, or if that's an implementation dependency.  If it's not guaranteed by the standard, then implementation-independence would support use of the ULP.

PR #404, opened 2016-12-23
METRON-624: Updated Comparison/Equality Evaluations in Stellar

jjmeyer0:
1. General cleanup in Stellar.g4
- Refactored the grammar to consistently use fragments instead of inlining things.

2. Updated the way Stellar transformations were validated
- Previously Stellar parsed/evaluated the transformation to validate it. In certain circumstances this would cause failures. For instances, all variables would be resolved to null which would cause an error to be thrown when evaluating transformations such as: '1 < foo', '1 < null'. Both of these statements are syntactically valid.
- Now the validation just checks whether the Stellar transformation is syntactically valid. It will not send the transformation through the listener as it once did. It will not validate that variables are defined. It will not validate functions are defined.

3. Updated comparison/equality expression evaluation
- Number 2 was required in order to update the comparison/equality evaluation.
- Now statements such as 1 == '1' will return false.
- When comparing numerical values, the standard promotion rules apply (http://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2).
- When comparing non-numeric types, compareTo method is used.
- When either side is null then '==' and '!=' are used. For all other comparison operators null value would throw exception.


cestella:
Hey, thanks for the great addition @jjmeyer0!

A few things to note and some context.

Around validation, the initial stellar implementation purposefully ran validation the way that it's implemented, but we can reconsider those things if we want.
* The variables resolve to `null` to ensure that functions could handle `null` without error.  I understand that Java and most programming languages would NPE on `1 < null`, but our data flowing through the topologies will be sparse, so not all variables will be defined for every message.  If a stellar statement is applied to a message where the variable is not defined, I wanted to ensure that it did not fail, but rather did an appropriate thing depending on the operation.
* We do function resolution during validation to ensure people have not made a syntactic error in their stellar statements when pushing configs to zookeeper.  We would rather fail fast here.

Around `==` for floating point numbers.  The initial stellar implementation intended to do an epsilon check on equals involving doubles (e.g. Math.abs(lhs - rhs) < epsilon) as you almost never want strict equality in floating point numbers.

What this does show is that we need more of a rigorous spec and more time spent on documenting the design of Stellar.

cestella:
One further thought: regarding the validation, I just want to point out that I'd be ok for validation to not *execute* the statement, but only validate syntactic correctness and function existence.  It just means that we will need to augment `StellarProcessorUtils` to execute with a map that returns null for its variables to ensure our tests continue to cover the same way.

I do think we should retain the functionality that comparisons involving nulls operate in a sensible way.  I think any comparison involving `null` should return `false` sort of like how doubles function with `NaN`.  Also, the decisions made here should be included in the stellar section of the README.md as part of this PR.

I'm coming around about `==` operating strictly for doubles.  I could see how that would violate the principle of least surprise for people coming from other languages.  I guess I never had much use for `==` with `double`s and thought I could automatically save people some obvious errors, but I think the current approach is problematic and probably should be scrapped for what this PR does.

jjmeyer0:
@cestella I updated the comparison expressions to return false when either side is null. I'll work on the validation stuff soon.

jjmeyer0:
@cestella if comparing floats/doubles is going to be a common thing maybe we should provide a way to do so easily. Maybe we create functions FLOAT_EQUALS(f1, f2, epsilon) and DOUBLE_EQUALS(d1, d2, epsilon).

cestella:
@jjmeyer0 Yeah, I tend to agree.  I'll make a follow-on JIRA to add an `APPROX_EQUALS`.  I feel better about it now that we have an `ABS` function.

cestella:
Yeah, this is good work.  I looked at it again after my vacation and I like it.  I give it a +1.

jjmeyer0:
Thanks @cestella. I appreciate it. If no one has already, I'll work on those functions we discussed next.

PR #405, opened 2016-12-31
METRON-641: Fixed kibana_master.py Python3 miss

DimDroll:
Kibana installed as part of Metron mpack winthin Ambari will fail during start with following error:
ValueError: zero length field name in format

This fixed it.

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/KIBANA/4.5.1/package/scripts/kibana_master.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/KIBANA/4.5.1/package/scripts/kibana_master.py
@@ -81,7 +81,7 @@ class Kibana(Script):
                   group=params.kibana_user
                   )
 
-        File("{}/kibana.yml".format(params.conf_dir),
+        File("{0}/kibana.yml".format(params.conf_dir),
              owner=params.kibana_user,
              content=InlineTemplate(params.kibana_yml_template)
              )
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/KIBANA/4.5.1/package/scripts/kibana_master.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/KIBANA/4.5.1/package/scripts/kibana_master.py
@@ -53,9 +53,9 @@ class Kibana(Script):
         majorVersion = OSCheck.get_os_major_version()
         Logger.info("CentOS/RHEL major version reported by Ambari: " + majorVersion)
         if majorVersion == "6" or majorVersion == "7":
-            repoName = "name=CentOS/RHEL {} repository for Elasticsearch Curator 4.x packages\n".format(majorVersion)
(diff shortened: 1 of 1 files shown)


mattf-horton:
I believe this is only an issue with Python 2.6. Don't we require Python 2.7?

mattf-horton:
Let's do this, but need to also fix the other three places in this file where '{}' is used with format method.

dlyle65535:
Yup, please change the other 3 and we'll get this in, thanks!

mmiklavc:
Hi, thanks for the contribution! Per Matt's comments, we do require Python 2.7. I don't believe that's in our main README, but it is listed in our deployment READMEs, e.g. https://github.com/apache/incubator-metron/tree/master/metron-deployment/vagrant/quick-dev-platform

That being said, we have made this exact fix/change in other files in the Ambari installation code and it should be here as well. Also agreed with Matt's comments about the other 3 braces.

dlyle65535:
Here's a fun fact- quick-dev uses 2.6.6. Had to make this exact change while unifying Ansible with the MPack, so if nothing else, we'll get it there. Still I'd like @DimDroll to get the deserved credit, so I'd be happy for this one to get there first.

mattf-horton:
> I'd like @DimDroll to get the deserved credit

Exactly, @DimDroll please spin a quick update.

DimDroll:
Oh my, thank you team for such a warm words, I really appreciate it. I have made mentioned changes.

Also, as a side note, while I was looking for {} in rest of the code base using following command in root:
`grep -R '{}' * | grep -v '\.java:\|README.md:\|LICENSE:\|\.json:\|\.js:\|\.jj:' | less`
I have found this guy: metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/KIBANA/4.5.1/package/scripts/dashboard/dashboard.p
Unless it is Matlab file I feel like it is misnamed py or most likely pyc file. Should I do anything about it?
Please advise.

Thank you.

mattf-horton:
@DimDroll , here's what I did to find the answer to your question:  
* I went to github and found https://github.com/apache/incubator-metron/blob/master/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/KIBANA/4.5.1/package/scripts/dashboard/dashboard.p
* Then I clicked the "Blame" button, which showed me the patch it was created by.
* Clicking on that, I found that @dlyle65535 committed it as part of METRON-365: Load demo dashboard on Kibana Service install (PR#211).  
* Searching for 'dashboard.p', I find both the file itself, and the use of it in line 131 of kibana_master.py. 
* I couldn't recognize the language it's in (but it sure isn't python).  So I googled "file format .p" and found it is probably a "python pickle" file, and indeed searching for "pickle" in the same commit found in dashboardindex.py uses of pickle.load and pickle.dump.

So it's not misnamed, and you were close with the .pyc guess.  The only question left is, does it belong there as a constant template file for the kibana dashboard?  Or is it an auto-gen file that shouldn't actually be in git?  That's a question for @dlyle65535 .

mattf-horton:
+1 on the fix btw.

mattf-horton:
Would someone consider committing this PR?  It's been open a while.  Thanks.

ottobackwards:
I will later tonight or first thing tomorrow am

PR #406, opened 2016-12-31
METRON-642 Correct path to ES gc log file

DimDroll:


mattf-horton:
Since you're correctly fixing this bit of code, please also add backslashes at the line-ends of the first two lines of ES_JAVA_OPTS.  Currently,  the value of ES_JAVA_OPTS incorrectly spans 3 lines. The lines must be terminated with backslashes to effectively become a single line. The current inclusion of newlines in the long string value is acceptable (although unusual) in shellscript, but not in a systemd EnvironmentFile. /etc/sysconfig/elasticsearch must function as both.

DimDroll:
Thank you Matt,

I know it's been almost 2 months, but I have added backslashes you noted and updated commit.

mattf-horton:
+1 This is a correct and badly needed fix.
@DimDroll (Dima), thanks for progressing this. 
@dlyle65535 , note this is one of the fixes also suggested in PR#425, contributed for your PR#436

mattf-horton:
@DimDroll , please note the title of this PR is different from the title of the Jira.  That means comments here will *not* be correctly cc'ed to the Jira.  Please edit the PR title and/or the Jira title so they are exactly the same (the PR title being prefixed by "METRON-642" and no colon).  Thanks.

DimDroll:
@mattf-horton , thanks Matt,

I've updated Jira's tittle as you suggested, will make sure to follow this in future.

I have a question though, if one Jira case handles multiple PRs, they should be added as sub-tasks or there is different procedure for that?

DimDroll:
@mattf-horton , however even after correcting Title in Jira and here I still see same comments in both Jira (posted by ASF GitHub Bot ) and here. So what's the difference?

mattf-horton:
@DimDroll , yes, I just noticed in PR#405 that the comments were being correctly cc'ed to Jira even tho the title was different.  Don't know.  Maybe the received wisdom about the titles having to be exactly the same is wrong?  Perhaps it is sufficient to either (a) have the correctly hyphenated and capitalized Jira key (METRON-nnn) at the beginning of the PR title, or (b) have a link to the PR in the Jira?

Can anyone else out there give a more precise answer for the requirement?  Thanks.

DimDroll:
@mattf-horton , you're right according to this [link](https://confluence.atlassian.com/adminjiracloud/connect-jira-cloud-to-github-814188429.html):
> When a developer makes a commit, they should add a JIRA Software issue key to the commit message, like this:

And this [link](https://blogs.atlassian.com/2014/04/connecting-jira-6-2-github/):
> At this point JIRA will begin scanning your GitHub instance looking for references to JIRA. JIRA looks for at least two uppercase letters, a dash, and a number as an issue key (ala TIS-12). If it finds an issue key in a branch name or a commit, it will show it the development panel for that issue.

And this [link](https://confluence.atlassian.com/adminjiraserver070/integrating-with-development-tools-776637096.html):
> a user simply needs to supply an issue key for the issue to be automatically linked

mattf-horton:
@DimDroll  - Dima, sorry it took so long to get this in.  It's still needed, so I've merged it.  Thanks for the contribution!

PR #407, opened 2016-12-31
METRON-643: Stellar function documentation needs to be updated

JonZeolla:
The documentation for some the `STATS_*` Stellar functions was moved from [here](https://github.com/apache/incubator-metron/tree/master/metron-platform/metron-common) to [here](https://github.com/apache/incubator-metron/tree/master/metron-analytics/metron-statistics) as of [this commit](https://github.com/apache/incubator-metron/commit/b1c89df6f0503e82d5a9c4b7ac9c9e82ea810491) but the documentation was not correctly updated.

This PR updates the core function table in [metron-common](https://github.com/apache/incubator-metron/tree/master/metron-platform/metron-common#stellar-core-functions) to point to the new locations, and fixes some minor typos.  

The alternative would be to simply remove the `STATS_*` entries from the table in metron-common, and to add a pointer to the other locations where Stellar functions are documented.  Open to feedback.


JonZeolla:
Note:  `BIN` and `STATS_BIN` may be added via [METRON-637](https://github.com/apache/incubator-metron/pull/401) but the documentation in metron-common was not updated, potentially indicating a preference to remove the `STATS_*` documentation from the metron-common `README.md`.

I would suggest that at the end of the day we make sure there's a single place where people can go for Stellar function documentation, even if it's just a page of pointers to the various READMEs.

nickwallen:
+1 Looks good, Jon.  

I agree with you that we need one single, searchable "thing" documenting all of the Stellar functions.  The approach that you took here gives us that. I prefer your approach over simply having one link to the 'metron-statistics' README which would make a user click through to multiple READMEs and search each one.

Longer term, we really need something along the lines of auto-generated documentation that gets published with each release.  The approach we currently have is difficult to maintain and is destined to fall out-of-sync from the underlying code.

cestella:
@JonZeolla I agree; I like this approach. I will commit and update the `STATS_BIN` PR to conform to the new approach.

PR #317, opened 2016-10-20
METRON-410: mysql_server's MySQL install causes mutually assured dest…

dlyle65535:
…ruction when installed on the same machine as the Ambari Hive MySQL

Allowed the user to indicate their using an existing MySQL instance:

a) Refactored the required database and index settings to a separate property sheet using enhanced configs.
b) Made the MySQL start/stop logic from the add user, remove user and geo ip database load conditional.
c) Auto-populated the Storm UI URL and MySQL Host Address from the Add Service Wizard Master Install Selection page.
d) Cleaned up some lingering metainfo.xml items since I was in the file anyway to add the theme.

Tested on Docker cluster following procedure here: [Metron MPack](https://www.evernote.com/l/AhLFVR-9CsFIYYnOnF43BlxSsT4F856qwaY)

Test case 1: Default install (Ambari installed MySQL)l
Test case 2: Pre-existing MySQL install.

PR #408, opened 2017-01-03
METRON-608 Mpack to install a single-node test cluster

mattf-horton:
This "metron-mpack-singlenode" is very similar to "metron-mpack", and can be usefully compared with it for code review, altho it also includes many bug fixes proposed in METRON-634.  As it says in the Jira, this is a short-term fix by providing a completely separate Mpack just for the single-node scenario, until we can work the bugs out of an mpack that works on all sizes of cluster.

This was tested on a Centos7 VM with 16GB of RAM, and all prerequisites installed, including Python 2.7.11.  The usual install process was followed, with Ambari 2.4.2 and HDP stack 2.5.3.  After successful startup, a few thousand 'bro' packets were injected, and were seen to correctly propagate through parser, enricher, and indexer topologies, and finally into elasticsearch.

An outstanding issue is that the configuration validator in the mpack doesn't work.  It fails with "unknown error" and must be ignored.  The multi-node mpack already had this problem, and I have run out of time to try to fix it here.  Consequently, my enhancement of trying to automatically flag the user if less than five Storm Supervisor Slots are allocated (commit 9ad8706), is untested.  I would like to open a new Jira for this issue, since it was pre-existing before these additions.
Thanks,
--Matt


mattf-horton:
Please note the Travis failures are not relevant.  All nine are due to https://issues.apache.org/jira/browse/METRON-648

mattf-horton:
kicking travis

mattf-horton:
Oops, now the 2017 fix (METRON-647) is committed.  Kicking Travis again.

JonZeolla:
Is "MySQL should not be installed on a node already running a MySQL instance" still accurate wrt Ambari MPack?  I thought this was fixed.  I see it noted as the 2nd bullet point [here](https://github.com/mattf-horton/incubator-metron/blob/b5b3a348527c03bc778a5fc5dd0d94ec3a00470b/metron-deployment/README.md#current-limitations).  Would we be able to add that tweak in to this PR since it's so minor, and it's [semi-]related to the changes made here?  I believe the actual fix was a part of [METRON-410](https://issues.apache.org/jira/browse/METRON-410) (#317)

dlyle65535:
This is great stuff. Can you talk a bit about why you split it off instead of just updated the existing MPack to support 1+ nodes? I'd prefer not to maintain two if we don't have to.

mattf-horton:
@JonZeolla , I'm not sure I follow the doc statement, since the issue is not installing two MySQL instances on the same server, but rather sharing the same MySQL instance between two services (Hive and Metron).

At any rate, I think you are correct that this is an obsolete statement in the doc, since [METRON-410](https://issues.apache.org/jira/browse/METRON-410) \([#317](https://github.com/apache/incubator-metron/pull/317)\) was fixed in November.

Are you suggesting that since I'm in this doc anyway, I should fix this doc issue by removing that statement?  Okay by me, just clarifying.  Thanks.

mattf-horton:
Hi @dlyle65535 , from a design perspective, I totally agree that we should have a single MPack that works on any size cluster, and I hope to replace both the 4(+)-node and single-node versions in the future with a more general solution.  I even have a prototype of such.  My reasons for proceeding with the split-off version at this time were entirely pragmatic, due to time pressure:

- I really needed a single-node that works with Ambari, for testing other PRs.  
- I spent several days trying to make Elasticsearch work on a single node installation, while retaining the generality of the multi-node case, but couldn't make it work.  This chewed up a lot of time.
- I put together a unified Mpack that used two different template files for Elasticsearch, depending whether it was for a single-node or multi-node install.  That worked just fine, on the surface, but was clearly going to require in-depth testing of 1, 2, 3, and 4 node cases.  I do not currently have time to thoroughly test all the special cases.
- Doing it this way I only had to test the single-node case, which was fairly fast.
- Having made one that works, I'd like to make it available to others.
- Now, if it is not accepted into the code, the high rate of change in metron-deployment will soon obsolete it.

That's pretty much it.  It's a throw-away, but not until we do the work of actually producing the generalized solution and testing it thoroughly.  Which I evaluate as lower priority than a number of other more urgent needs.  Thanks.

mattf-horton:
@JonZeolla , I removed the obsolete comments about MySQL installation in Mpack.  Thanks for the suggestion.

dlyle65535:
@mattf-horton - I need an MPack that works for 1-N nodes in order to complete [METRON-671](https://issues.apache.org/jira/browse/METRON-671). I suspect I'm working through stuff you've already fixed. Would it make sense to push out a PR with the unified branch and let me test it? Then I could use finish METRON-671 on top of your changes.

mattf-horton:
@dlyle65535 , I'll be happy to share my wip.  It will take me a few hours to brush it off, then I'll post it as a PR referring to [METRON-609](https://issues.apache.org/jira/browse/METRON-609).

mattf-horton:
@dlyle65535 , please see https://github.com/apache/incubator-metron/pull/425

nickwallen:
@mattf-horton Still needed?

mattf-horton:
We should ask @dlyle65535 whether he completely obsoleted this one. In the meantime, METRON-634 PR#532 is definitely needed.

ottobackwards:
What is the status of this PR?

mattf-horton:
Withdrawn.  The current installers adequately install a single node system.

PR #409, opened 2017-01-03
METRON-644 RPM builds only work with Docker for Mac

kylerichardson:
Currently using the rpm-docker image to build the project RPMs will only succeed using Docker for Mac due to the way it handles permissions. When using docker-machine/boot2docker, "Bad owner/group" build errors occur due to the SPEC and SOURCE files having an owner uid which does not map to an existing user within the container.

Some background. When using Docker for Mac any volume mounted from the host into the container is munged into being owned by the running user within the container. Hence there are no permissions issues like we see with docker-machine/boot2docker.

Modified the build.sh script to ensure the owner uid of the metron.spec file maps to an existing user in the container and that the rpms post-build are owned by the same user.

Tested on Docker for Mac, docker-machine/boot2docker, and a local docker-engine on Linux. RPMs build as expected in each case.

### DIFF
--- a/metron-deployment/README.md
+++ b/metron-deployment/README.md
@@ -163,7 +163,6 @@ Components in the RPMs:
 
 ### Prerequisites
 - Docker.  The image detailed in: `metron-deployment/packaging/docker/rpm-docker/README.md` will automatically be built (or rebuilt if necessary).
-- On Mac, Docker Toolbox's handling of ownership causes permissions issues in the Docker container.  Please use Docker for Mac instead.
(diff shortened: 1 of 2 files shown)


justinleet:
Thanks for doing this, it'll be really nice to not be dependent on Docker for Mac for this.  Could you also update the prereqs (and any other necessary docs) in README.md in metron-deployment?

kylerichardson:
@justinleet Good catch. Updated the README to remove the note about Docker for Mac.

justinleet:
I'm +1 by inspection, given that you've run through testing it on a couple platforms.

PR #411, opened 2017-01-04
METRON-635: Vagrant provisioning fails on CentOS hosts

JonZeolla:
## Problem
Vagrant provisioning of full or quick-dev fails when using various hosts (specifically tested on CentOS 6.8) with the following error message. 
```
fatal: [node1]: FAILED! => {"failed": true, "msg": "ERROR! failed to transfer file to /home/vagrant/.ansible/tmp/ansible-tmp-1483495876.66-189750710960299/setup:\n\nsftp: illegal option -- i\nusage: sftp [-1Cv] [-B buffer_size] [-b batchfile] [-F ssh_config]\n            [-o ssh_option] [-P sftp_server_path] [-R num_requests]\n            [-S program] [-s subsystem | sftp_server] host\n            sftp [user@]host[:file ...]\n            sftp [user@]host[:dir[/]]\n            sftp -b batchfile [user@]host\n"}
```

## Solution
Add `scp_if_ssh = True` in the appropriate ansible.cfg.

## Testing

Tested by running up full-dev-platform, quick-dev-platform, codelab-platform, and fastcapa-test-platform (testing fastcapa required merging [the PR](https://github.com/apache/incubator-metron/pull/410) for [METRON-645](https://issues.apache.org/jira/browse/METRON-645)) with and without the modifications in this PR.  I found that codelab and fastcapa did not require the modification in my environment to be successfully provisioned.

### DIFF
--- a/metron-deployment/vagrant/full-dev-platform/ansible.cfg
+++ b/metron-deployment/vagrant/full-dev-platform/ansible.cfg
@@ -23,5 +23,7 @@ log_path = ./ansible.log
 
 
 # fix for "ssh throws 'unix domain socket too long' " problem
+# and provisioning issues with certain host OSs
 [ssh_connection]
 control_path = %(directory)s/%%h-%%p-%%r
+scp_if_ssh = True
--- a/metron-deployment/vagrant/quick-dev-platform/ansible.cfg
+++ b/metron-deployment/vagrant/quick-dev-platform/ansible.cfg
@@ -23,5 +23,7 @@ log_path = ./ansible.log
 
 
 # fix for "ssh throws 'unix domain socket too long' " problem
+# and provisioning issues with certain host OSs
 [ssh_connection]
 control_path = %(directory)s/%%h-%%p-%%r
+scp_if_ssh = True


nickwallen:
@JonZeolla I want to make sure I understand under what conditions this bug occurs.  If I try to provision Quick Dev when running on a CentOS 6.8 host, I will hit this error?

Any idea why this happens?  Is it a difference between the Ansible build on Mac versus CentOS?  Or something else entirely?

JonZeolla:
@nickwallen Right.  So more granular details about how I tested:
1. Installed [CentOS 6.8 minimal](http://centos.pymesolutionsweb.com/7/isos/x86_64/CentOS-7-x86_64-Minimal-1611.iso)
1. `yum install -f git`
1. `git clone https://github.com/jonzeolla/development`
1. `cd development/Bash && chmod 755 setupMetron.sh`
1. `./setupMetron.sh -fvs full`
    * This should not show an issue.
1. Destroy the vagrant instance (see `/usr/local/metron/*/metron-deployment/vagrant/full-dev-platform`)
1. `./setupMetron.sh -fvst full`
    * You should get the issue.
        * Adding `-t` prevents `setupMetron.sh` from adding `scp_if_ssh` to `ansible.cfg`, and you will see the error.  For the most pristine testing, I reinstalled the OS fresh for each run of `setupMetron.sh` (hence why testing this was a bit of a pain).  Finally:
1. Kill the vagrant instance again
1. `./setupMetron.sh -fvst -p 411 full`
    * This runs the setup process without adding scp_if_ssh, but it merges in this PR before building Metron and starting vagrant.  You should not see the issue.

Notes:
* `full` can be substituted with `quick`, `fastcapa`, or `codelab` for testing the different vagrant deploy methods.  As you're aware, for a good fastcapa build, this requires `-p 410`.
* Obviously it's probably worth reviewing `setupMetron.sh` to see exactly what's happening, but the `-v` provides some amount of feedback.
* There is a `-d` flag which adds `ansible.verbose = "vvvv"` to the Vagrantfile prior to doing `vagrant up` and then removes it after completion for additional troubleshooting.
* All the script really does is setup up all the dependencies.  From time to time, running `setupMetron.sh` will fail because something upstream is being sync'd or performing maintenance/having issues, and a rerun usually fixes the issue.
* The script is currently super inefficient and doesn't check to see if something's already installed before attempting to install it.

Happy to help provide any additional detail as would be helpful.  I wasn't able to pin this down exactly, but I have seen 3 others independently report this issue and the fix worked for them.

JonZeolla:
Any thoughts on this?

JonZeolla:
Vagrant provisioning [isn't supported on CentOS 6.8](https://github.com/apache/incubator-metron/pull/436/commits/37ddde3ae2f26e4efebe1c1c3d2e43a4c70f4b0b) any more, so removing this PR.

PR #412, opened 2017-01-04
METRON-647 Parser unit test failures due to assumed year values

justinleet:
Short term fix for the build, along with some cleanup noted in METRON-648 (which was closed as a dupe of 647).

Two classes are adjusted:

`BasicAsaParserTest:testIp6Addr`: The rawMessage string has been adjusted to include 2016 explicitly.  In addition `assertTrue` has been replaced with `assertEquals` and params put in correct order.

`GrokWebSphereParserTest`: Tests no longer use explicit timestamps, but instead generate a timestamp based on the current year (UTC timezone is used, because GrokParser defaults to UTC if no timezone is explicitly provided). Using current year is not desirable long term behavior, but the tests now match existing behavior and the build can succeed. In addition the order of params to `assertEquals` is corrected and values are no longer unnecessarily coerced.

A follow-on ticket will be created for handling dates in a more robust manner.


nickwallen:
+1 Looks good to me

PR #418, opened 2017-01-17
METRON-666 Fix javadoc doclint errors

mattf-horton:
Fixes for Java 8 javadocs automatic doclint problems:
- replaced many uses of `<p/>` and `</p>` with simple uses of `<p>`
- wrapped uses of `<` or `>`  in `{@literal}` or equivalent
- a couple actual errors like \@return with no description or \@throws when nothing thrown

### DIFF
--- a/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileSplitterBolt.java
+++ b/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileSplitterBolt.java
@@ -147,11 +147,13 @@ public class ProfileSplitterBolt extends ConfiguredProfilerBolt {
 
   /**
    * Each emitted tuple contains the following fields.
-   * <p><ol>
+   * <p>
+   * <ol>
    * <li> entity - The name of the entity.  The actual result of executing the Stellar expression.
    * <li> profile - The profile definition that the message needs applied to.
    * <li> message - The message containing JSON-formatted data that needs applied to a profile.
-   * </ol></p>
+   * </ol>
+   * <p>
    */
   @Override
   public void declareOutputFields(OutputFieldsDeclarer declarer) {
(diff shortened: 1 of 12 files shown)


cestella:
I'd kick travis by closing and reopening this PR.  I think this has to do with Travis being overloaded so the test is timing out.

JonZeolla:
Yeah, it looks like the latest master build failed as well.

jjmeyer0:
If time outs are going to continue being an issue, maybe we should start looking into using Maven's parallel builds. For example, when I build locally I always do: `mvn -T 2C clean package`. This means build using 2 threads per core. It usually cuts the build time in half for me.

cestella:
So, one comment here for posterity and to make sure we're all on the same page.  This error started happening yesterday to my PR (#417).  Here's what happened:
* @mmiklavc asked for a doc change to a `README.md`
* Upon that commit, the SolrIntegrationTest failed because it timed out.
* I ran it 5 times in succession locally and noted that the total time was around the same as historically (32 - 35 minutes)
* Meanwhile, I updated the test to be a bit more permissive timeout-wise for the IndexingIntegrationTests

I concluded that it was travis latency based on:
* It is sporadic on travis but not locally
* Performance regressions did not appear locally
* It appeared suddenly on #417 after a no-code-change commit.

Just pointing out when I noticed this and some context.

cestella:
@jjmeyer0 It's timeouts in the test itself, so parallel execution wouldn't really help (this particular test sends data through and waits til it gets written out).

cestella:
@nickwallen replied on the dev list.  Replicating here for posterity:
@nickwallen said:
This is a self-imposed limit, no? See ComponentRunner.java:131.  Do we just
need to give Travis some more time?

cestella:
@nickwallen That would be one way to fix it, for sure.  I did bump the timeout a bit in #415, but it might could stand more.

mattf-horton:
Kicking Travis. Bad Travis!

mattf-horton:
Travis is back to normal.

mattf-horton:
Do we need any changes, to commit these modest fixes to javadoc comments?  I'd like to experiment with adding javadocs to the doc site.  Thanks.

justinleet:
I'm +1 on this, and definitely glad to see it.

ottobackwards:
+1

PR #420, opened 2017-01-19
METRON-668: Remove the "tickUpdate" profile config and make the "init" phase not reset variables

cestella:
Please see description at [METRON-668](https://issues.apache.org/jira/browse/METRON-668) for a full description.


mmiklavc:
Excellent. And I need these fixes for some outlier analysis with HLLP and MAD. +1 by inspection.

nickwallen:
I like this as it simplifies the Profile configuration.

How did you test this?  Did you run any profiles on "Quick Dev"?

cestella:
@nickwallen running as we speak.  The test plan will be to generate some data like we do for the MAD outliers and construct a couple simple profiles.  I'll make it a bit more explicit after I do it. :)

cestella:
Ok, test is complete, I made sure the profiler worked.  I also adjusted the timeout on the indexing integration tests so that travis doesn't barf.

cestella:
The explicit test that was done on quickdev was following the Median Absolute Deviation scenario in `metron-statistics`

cestella:
By the way, please don't merge this.  I'm using this branch to figure out what's going on with the build failures.

PR #424, opened 2017-01-26
METRON-672: SolrIndexingIntegrationTest fails intermittently

cestella:
This failure is due to a change in default behavior when indexing was split off into a separate configuration file.  The default batch size was changed from `5` to `1` in particular.  This, by itself, is not a problem, but the `IndexingIntegrationTest` (base class for Solr and Elastic search integration tests):
* submits the configs
* starts the indexing topology
* writes the input data

The writing of the input data may happen before the topology fully loads or the configuration fully loads, especially if the machine running the unit tests is under load (like with travis).  As a result, the first record may end up with the default batch size (of 1) and write out immediately because the indexing configs haven't loaded into zookeeper just yet.  In that circumstance, eventually the configs load and the batch size is set to `5`.  Meanwhile we've written 10 records and are expecting 10 in return, but because you wrote the first out already and then the next 5, we have another 4 pending to be written by the `BulkMessageWriterBolt`.

So, the failure scenario is as follows:
* Message 1 is received and the indexing config hasn't loaded yet, so the batch size is 1 and it immediately gets written out
* Message 2 - 5 are each received and the indexing config has loaded, so the batch size is 5 and it queues up
* Message 6 is received and the batch writes out
* Messages 7 - 10 are received, but never make a full batch, so we time out waiting for them to write out

The fix is to ensure that we don't write out messages to kafka until the configs are loaded, which is what this PR does.

### DIFF
--- a/metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/components/ConfigUploadComponent.java
+++ b/metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/components/ConfigUploadComponent.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 import static org.apache.metron.common.configuration.ConfigurationsUtils.*;
@@ -39,7 +40,7 @@ public class ConfigUploadComponent implements InMemoryComponent {
   private String enrichmentConfigsPath;
   private String indexingConfigsPath;
   private String profilerConfigPath;
-  private Optional<Function<ConfigUploadComponent, Void>> postStartCallback = Optional.empty();
+  private Optional<Consumer<ConfigUploadComponent>> postStartCallback = Optional.empty();
   private Optional<String> globalConfig = Optional.empty();
   private Map<String, SensorParserConfig> parserSensorConfigs = new HashMap<>();
   public ConfigUploadComponent withTopologyProperties(Properties topologyProperties) {
@@ -80,7 +81,7 @@ public class ConfigUploadComponent implements InMemoryComponent {
     return this;
   }
 
-  public ConfigUploadComponent withPostStartCallback(Function<ConfigUploadComponent, Void> f) {
+  public ConfigUploadComponent withPostStartCallback(Consumer<ConfigUploadComponent> f) {
         this.postStartCallback = Optional.ofNullable(f);
         return this;
   }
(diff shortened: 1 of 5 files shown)


cestella:
To test this locally, because it's extremely sporadic that it happens locally (1 out of every 50 times I run the test), I did the following:
* Built and installed the project in Maven: `mvn -DskipTests clean install`
* Ran the `metron-solr` project integration tests for at least 2 hours in a row ensuring that they don't fail: `echo "" > /tmp/output;while [ $(cat /tmp/output | grep "vs 6" | wc -l) -lt 1 ];do mvn install >& /tmp/output;done`

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/integration/IndexingIntegrationTest.java
```diff
@@ -139,11 +146,22 @@ public void test() throws Exception {
...
+              try {
+                waitForIndex(component.getTopologyProperties().getProperty(ZKServerComponent.ZOOKEEPER_PROPERTY));
+              } catch (Exception e) {
+                e.printStackTrace();
+              }
+              isLoaded.set(true);
+              return null;
+              }
+            );
             ;
```
justinleet: Can you kill the extra semicolon?

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/integration/IndexingIntegrationTest.java
```diff
@@ -184,6 +203,26 @@ public void test() throws Exception {
...
+        try {
+          bytes = ConfigurationsUtils.readSensorIndexingConfigBytesFromZookeeper(testSensorType, client);
+          Thread.sleep(1000);
+        }
+        catch(KeeperException.NoNodeException nne) {
+          //kindly ignore because the path might not exist just yet.
+        }
+      }
+      while(bytes == null || bytes.length == 0);
+      return;
```
justinleet: Drop the return, since it's a void method.

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/integration/components/ConfigUploadComponent.java
```diff
@@ -38,6 +39,7 @@
   private String enrichmentConfigsPath;
   private String indexingConfigsPath;
   private String profilerConfigPath;
+  private Optional<Function<ConfigUploadComponent, Void>> postStartCallback = Optional.empty();
```
justinleet: Could this just use Consumer instead of Function? Since the second type parameter is Void, it seems like the Function is just being a Consumer anyway

justinleet:
Thanks for taking the effort to dig into this.  Great work.  Other than a couple minor comments, I'm very happy with this.

cestella:
@justinleet comments addressed, let me know if there's anything else.

PR #430, opened 2017-01-30
METRON-658:  Updated Grammar to Handle More Uses of in/not in Expressions

jjmeyer0:
This PR allows for the use of in/not in expressions in more places. For example, the following expressions now work:

```
('grok' not in 'foobar') == true
'grok' not in ('foobar' == true)
'grok' not in 'foobar' == true
('grok' not in 'foobar') AND 1 < 2
```

It is important to note that github's diff of `Stellar.g4` isn't very nice. A vast majority of the changes are just whitespace.


nickwallen:
Is there a way to split out the core changes that you made from the whitespace changes?  Were the whitespace changes required?  Or should those be a separate "clean-up" PR?

jjmeyer0:
@nickwallen personally I don't think it should be a separate PR. To me it is just one of those things that should be cleaned up as we go. My reasoning is that I think we should encourage clean up as we go with minimal overhead. However, I understand that it is difficult to read on github. I do agree that I could have been more conscious of the way I made the commits though. I changed this PR to have two commits,1 with the changes and 1 with the cleanup. If this is still hard to review, I can split them into two PRs.

nickwallen:
I'm good with this split as separate commits.

cestella:
+1 by inspection, good contribution.  I like the refactoring of the grammar too.  Stellar is getting cleaner and cleaner with every contribution like this.

PR #434, opened 2017-02-02
METRON-687 Create String Formatting Function for Stellar

nickwallen:
Created a string formatting function for Stellar. 
* Uses Java's String formatting conventions.  
* Any non-existent variables will be formatted as 'null'.  This failure scenario would arise from a data quality issue.
* Forgetting to pass an argument required by the format string will throw an exception.  
* Not passing a format string will throw an exception.

#### FORMAT
* Description: Returns a formatted string using the specified format string and arguments.  Uses Java's string formatting conventions.
* Input:
    * format - string
    * arguments... - object(s)
* Returns: A formatted string.


cestella:
nice, quick, to the point and necessary.  +1 by inspection.

PR #435, opened 2017-02-02
METRON-684: Decouple Timestamp calculation from PROFILE_GET

cestella:
Currently `PROFILE_GET` only supports a static lookback of a fixed duration. As we have more complicated, potentially sparse, lookbacks (e.g. the same time slice every tuesday for a month), it would be nice to decouple the construction of timestamps from `PROFILE_GET` into its own set of functions.

The decoupling plus the default timestamp function is provided here, which is a fixed window starting from now.  This is implemented as `PROFILE_FIXED`.

Test plan to follow.

### DIFF
--- a/metron-analytics/metron-profiler-client/README.md
+++ b/metron-analytics/metron-profiler-client/README.md
@@ -29,8 +29,7 @@ The Stellar client consists of the `PROFILE_GET` command, which takes the follow
 REQUIRED:
     profile - The name of the profile
     entity - The name of the entity
-    durationAgo - How long ago should values be retrieved from?
-    units - The units of 'durationAgo'
+    timestamps - The list of timestamps (epoch millis) to grab
 OPTIONAL:
 	groups_list - Optional, must correspond to the 'groupBy' list used in profile creation - List (in square brackets) of 
             groupBy values used to filter the profile. Default is the empty list, meaning groupBy was not used when 
@@ -39,6 +38,21 @@ OPTIONAL:
             of the same name. Default is the empty Map, meaning no overrides.
 ```
 There is an older calling format where `groups_list` is specified as a sequence of group names, "varargs" style, instead of a List object.  This format is still supported for backward compatibility, but it is deprecated, and it is disallowed if the optional `config_overrides` argument is used.
+
+The `timestamps` field is (likely) the output of another Stellar function which defines the times to include.
+
+`PROFILE_FIXED`:      The timestamps associated with a fixed lookback starting from now
+```
+REQUIRED:
+    durationAgo - How long ago should values be retrieved from?
+    units - The units of 'durationAgo'.
+OPTIONAL:
+    config_overrides - Optional - Map (in curly braces) of name:value pairs, each overriding the global config parameter
+            of the same name. Default is the empty Map, meaning no overrides.
+
+e.g. To retrieve all the profiles for the last 5 hours.  PROFILE_GET('profile', 'entity', PROFILE_FIXED(5, 'HOURS'))
+```
+
 
 ### Groups_list argument
 The `groups_list` argument in the client must exactly correspond to the [`groupBy`](../metron-profiler#groupby) configuration in the profile definition.  If `groupBy` was not used in the profile, `groups_list` must be empty in the client.  If `groupBy` was used in the profile, then the client `groups_list` is <b>not</b> optional; it must be the same length as the `groupBy` list, and specify exactly one selected group value for each `groupBy` criterion, in the same order.  For example:
--- a/metron-analytics/metron-profiler-client/README.md
+++ b/metron-analytics/metron-profiler-client/README.md
@@ -29,7 +29,7 @@ The Stellar client consists of the `PROFILE_GET` command, which takes the follow
 REQUIRED:
     profile - The name of the profile
     entity - The name of the entity
-    timestamps - The list of timestamps (epoch millis) to grab
+    periods - The list of profile periods to grab
 OPTIONAL:
 	groups_list - Optional, must correspond to the 'groupBy' list used in profile creation - List (in square brackets) of 
             groupBy values used to filter the profile. Default is the empty list, meaning groupBy was not used when 
@@ -39,9 +39,9 @@ OPTIONAL:
 ```
 There is an older calling format where `groups_list` is specified as a sequence of group names, "varargs" style, instead of a List object.  This format is still supported for backward compatibility, but it is deprecated, and it is disallowed if the optional `config_overrides` argument is used.
 
-The `timestamps` field is (likely) the output of another Stellar function which defines the times to include.
+The `periods` field is (likely) the output of another Stellar function which defines the times to include.
 
-`PROFILE_FIXED`:      The timestamps associated with a fixed lookback starting from now
+`PROFILE_FIXED`: The profiler periods associated with a fixed lookback starting from now
 ```
 REQUIRED:
     durationAgo - How long ago should values be retrieved from?
--- a/metron-analytics/metron-profiler-client/README.md
+++ b/metron-analytics/metron-profiler-client/README.md
@@ -29,7 +29,7 @@ The Stellar client consists of the `PROFILE_GET` command, which takes the follow
 REQUIRED:
     profile - The name of the profile
     entity - The name of the entity
-    periods - The list of profile periods to grab
+    periods - The list of profile periods to grab.  These are ProfilePeriod objects.
 OPTIONAL:
 	groups_list - Optional, must correspond to the 'groupBy' list used in profile creation - List (in square brackets) of 
             groupBy values used to filter the profile. Default is the empty list, meaning groupBy was not used when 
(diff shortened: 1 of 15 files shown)


cestella:

## Preliminaries
`export METRON_HOME=/usr/metron/0.3.0` 

```
#!/usr/bin/python
import random
import sys
import time
def main():
  mu = float(sys.argv[1])
  sigma = float(sys.argv[2])
  freq_s = int(sys.argv[3])
  while True:
    print out
    sys.stdout.flush()
    time.sleep(freq_s)

if __name__ == '__main__':
  main()
```

```
```

   * `storm kill snort`
   * `storm kill bro`
* We won't need the enrichment or indexing topologies for this test, so you can kill them via:
   * `storm kill enrichment`
   * `storm kill indexing`

## Start the profiler

## Test Case

```
{
  "profiles": [
    {
      "profile": "stat",
      "foreach": "'global'",
      "onlyif": "true",
      "init" : {
               },
      "update": {
        "s": "STATS_ADD(s, value)"
                },
      "result": "s"
    }
  ]
}
```

`python ~/rand_gen.py 0 1 1 | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic indexing`
* Wait for at least 10 minutes and execute the following via the Stellar REPL:
```
# Grab the last 10 minutes worth of timestamps
PROFILE_FIXED( 10, 'MINUTES')
# Looks like 10 were returned, great.  Now, validate that I get 10 profile measurements back
PROFILE_GET('stat', 'global', PROFILE_FIXED( 10, 'MINUTES' ) )
# Ok, now look at the mean across the distribution
# STATS_MEAN( STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_FIXED( 10, 'MINUTES' ) )))
```
```
Stellar, Go!
{es.clustername=metron, es.ip=node1, es.port=9300, es.date.format=yyyy.MM.dd.HH, profiler.client.period.duration=1, profiler.client.period.duration.units=MINUTES}
[Stellar]>>> # Grab the last 10 minutes worth of timestamps
[Stellar]>>> PROFILE_FIXED( 10, 'MINUTES')
[24767772, 24767773, 24767774, 24767775, 24767776, 24767777, 24767778, 24767779, 24767780, 24767781, 24767782]
[Stellar]>>> # Looks like 10 were returned, great.  Now, validate that I get 10 profile measurements back
[Stellar]>>> PROFILE_GET('stat', 'global', PROFILE_FIXED( 10, 'MINUTES' ) )
[org.apache.metron.statistics.OnlineStatisticsProvider@44749031, org.apache.metron.statistics.OnlineStatisticsProvider@d2a7fbb9, org.apache.metron.statistics.OnlineStatisticsProvider@a217cfd7, org.apache.metron.statistics.OnlineStatisticsProvider@c5e42aed, org.apache.metron.statistics.OnlineStatisticsProvider@c4f4753d, org.apache.metron.statistics.OnlineStatisticsProvider@87a1606a, org.apache.metron.statistics.OnlineStatisticsProvider@e1b4c8dc, org.apache.metron.statistics.OnlineStatisticsProvider@fdb7b8d8]
[Stellar]>>> # Ok, now look at the mean across the distribution
[Stellar]>>> STATS_MEAN( STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_FIXED( 10, 'MINUTES' ) )))
-0.0077433441069769265
[Stellar]>>>
```

cestella:
By the way, I reran the acceptance tests after the last few commits to make sure it works on vagrant.

metron-analytics/metron-profiler-client/README.md
```diff
@@ -29,8 +29,7 @@ The Stellar client consists of the `PROFILE_GET` command, which takes the follow
 REQUIRED:
     profile - The name of the profile
     entity - The name of the entity
-    durationAgo - How long ago should values be retrieved from?
-    units - The units of 'durationAgo'
+    periods - The list of profile periods to grab
```
nickwallen: Would be good to mention what type 'periods' is; aka epoch milliseconds.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/FixedLookback.java
```diff
@@ -0,0 +1,74 @@
...
+      namespace="PROFILE",
+      name="FIXED",
+      description="The profiler periods associated with a fixed lookback starting from now.",
+      params={
+        "durationAgo - How long ago should values be retrieved from?",
+        "units - The units of 'durationAgo'.",
+        "config_overrides - Optional - Map (in curly braces) of name:value pairs, each overriding the global config parameter " +
+                "of the same name. Default is the empty Map, meaning no overrides."
+      },
+      returns="The selected profile measurement periods."
```
nickwallen: Also, describing the type here would be good too.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/Util.java
```diff
@@ -0,0 +1,117 @@
...
+            .of(required)
+            .filter(c -> !context.getCapability(c).isPresent())
+            .map(c -> c.toString())
+            .collect(Collectors.joining(", "));
+
+    if(StringUtils.isNotBlank(missing) || context == null) {
+      throw new IllegalStateException("missing required context: " + missing);
+    }
+  }
+  /**
```
nickwallen: Minor spacing issues (1 empty line between methods) here and later in the file.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/ProfilerConfig.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+package org.apache.metron.profiler.client.stellar;
+
+import org.apache.metron.common.utils.ConversionUtils;
+import org.apache.metron.hbase.HTableProvider;
+
+import java.util.Map;
+
+public enum ProfilerConfig {
```
nickwallen: Much better way to handle defaults.  I like it.  Clean.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/ProfilePeriod.java
```diff
@@ -51,7 +61,7 @@ public ProfilePeriod(long epochMillis, long duration, TimeUnit units) {
       throw new IllegalArgumentException(format(
               "period duration must be greater than 0; got '%d %s'", duration, units));
     }
-
+    this.timestamp = epochMillis;
```
nickwallen: I can't see where you are using this value, but maybe I am missing it.

I'd be cautious of using this value because it could be a timestamp anywhere throughout the period.  If you really need a timestamp for the period, then it might be better to use `getStartTimeMillis()`. That will give you the same timestamp for the period; no matter what initial timestamp was used to find that period.

metron-analytics/metron-profiler-client/README.md
nickwallen: Oops.  And here is an example already of the potential confusion between profile period and timestamp.  My own fault here, but it is a distinction that we need to try and make really clear.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/ProfilePeriod.java
cestella: That's good feedback, agreed.

metron-analytics/metron-profiler-client/README.md
cestella: Agreed and I think it should be the ProfilePeriod

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/FixedLookback.java
cestella: agreed

cestella:
@nickwallen Yeah, I was under the mistaken impression that period was milliseconds since epoch and it's not.  I think I like your idea better and just pass ProfilePeriods around instead.  I'll make that adjustment.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/Util.java
cestella: agreed

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/ProfilerConfig.java
cestella: thanks! :)

cestella:
Ok, I think I captured your comments, @nickwallen

PR #437, opened 2017-02-02
METRON-692: Update Upgrading.md for 0.3.0 -> 0.3.1

cestella:
We need a document to describe the manual upgrade issues from one release to the other.

### DIFF
--- a/Upgrading.md
+++ b/Upgrading.md
@@ -96,3 +96,46 @@ to:
 
 For every sensor enrichment configuration, you will need to migrate the `riskLevelRules` section
 to move from a map to a list of risk level rule objects.
+
+### [METRON-283: Make Threat Triage rules able to be assigned names and comments](https://issues.apache.org/jira/browse/METRON-283)
+
+#### Description
+
+As of 0.3.0, a MySQL database was used for storage and retrieval of
+GeoIP information during enrichment.
+As of 0.3.1, the MySQL database is removed in favor of using MaxMind's
+binary GeoIP files and stored on HDFS
+
+After initial setup, this change is transparent and existing enrichment
+definitions will run as-is.
+
+#### Migration
+
+While new installs will not require any additional steps, in an existing
+install a script must be run to retrieve and load the initial data.
+
+The shell script `geo_enrichment_load.sh` will retrieve MaxMind GeoLite2
+data and load data into HDFS, and update the configuration to point to
+this data.
+In most cases the following usage will grab the data appropriately:
+
+```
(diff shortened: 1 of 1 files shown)


justinleet:
While the migration from MySQL is transparent once data is in place, it requires running geo_enrichment_load.sh the first time to populate the GeoIP data into HDFS (or alternatively manually uploading the binary database file into HDFS and updating the global config to have geo.hdfs.file point to it). By default, the script just needs to be run as
`geo_enrichment_load.sh -z <zk_server:zk_port>`

I would also mention that if there is a MySQL instance dedicated to Metron's Geo data, it can be removed (AFTER appropriate testing).

You want me to do a write-up for that portion, or would you rather grab it?

cestella:
@justinleet Could you write up a description that fits the template in the doc and I'll cut and paste into the doc.

nickwallen:
I looked through all the 0.3.1 tickets and don't see any others that we definitely need to call out.

justinleet:
@cestella Let me know if there's anythign to change or adjust that you see in here. I'm not sure if you can get the raw version in GitHub like in Jira, so ping me if this isn't grabbable in the current form and I'll get it to you differently.

### [Migrate Geo Enrichment outside of MySQL](https://issues.apache.org/jira/browse/METRON-283)

#### Description
As of 0.3.0, a MySQL database was used for storage and retrieval of GeoIP information during enrichment.
As of 0.3.1, the MySQL database is removed in favor of using MaxMind's binary GeoIP files and stored on HDFS

After initial setup, this change is transparent and existing enrichment definitions will run as-is.

#### Migration
While new installs will not require any additional steps, in an existing install a script must be run to retrieve and load the initial data.

The shell script `geo_enrichment_load.sh`  will retrieve MaxMind GeoLite2 data and load data into HDFS, and update the configuration to point to this data.
In most cases the following usage will grab the data appropriately:
```
$METRON_HOME/bin/geo_enrichment_load.sh -z <zk_server>:<zk_port>
```

Additional options, including changing the source file location (which can be a file:// location if the GeoIP data is already downloaded), are available with the
`-h` flag and are also detailed in the metron-data-management README.me file.

One caveat is that this script will NOT update on disk config files.  It is recommended to retrieve the configuration using
```
$METRON_HOME/bin/zk_load_configs.sh -z <zk_server>:<zk_port> -m DUMP
```
The new config will be `geo.hdfs.file` in the global section of the configuration.  Append this key-value into the global.json in the config directory. A PUSH is unnecessary

justinleet:
Everything I know of is in here.  +1.  Thanks for putting this together.

PR #442, opened 2017-02-06
METRON-322 Global Batching and Flushing

mattf-horton:
DO NOT INTEGRATE YET.  This is a preliminary review request.
This Jira Ticket is to add timeout flushing to all Writers that do batch writes.  Currently most of them will wait indefinitely for a batch to fill, resulting in message recycling due to `topology.message.timeout.secs`.  This changeset so far implements timeout flushing for the BulkMessageWriterBolt.

I am now starting to work on unit tests, so while this code compiles and passes existing unit tests, the new functionality has not yet been tested.  However, I would be grateful for a quick review of the basic approach, focusing on changes in:
* BulkMessageWriterBolt
* BulkWriterComponent
* BatchTimeoutHelper (new), and 
* IndexingConfigurations.  

All of the other changes are basically boilerplate, just laying "getBatchTimeout()" facade methods alongside existing "getBatchSize()" facade methods.  (There's a lot of layers!)

The flush-on-timeout logic is fairly straightforward.  It was implemented by a refactoring of BulkWriterComponent and basic Tick Tuple reception in BulkMessageWriterBolt.

The tricky part was figuring out the appropriate setting for 'topology.tick.tuple.freq.secs' if the administrator configures non-default batchTimeouts.  It is necessary to enumerate the batchTimeout settings for all configured sensorNames, which is implemented in IndexingConfigurations$getAllConfiguredTimeouts().  Then multiple other factors must be taken into account to determine the allowed and recommended settings, which is implemented in BatchTimeoutHelper.  If there are better ways to accomplish these things, please share your ideas.

After feedback is incorporated, I need to make similar changes to the ParserWriter, and possibly other places in the code where batching is used and timeouts are not yet implemented.  That is separable work, which I'm aware also needs to be done.  Thanks for taking time to look at this.

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/IndexingConfigurations.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/IndexingConfigurations.java
@@ -24,14 +24,15 @@ import org.apache.metron.common.utils.JSONUtils;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
+import java.util.function.Supplier;
 
 public class IndexingConfigurations extends Configurations {
   public static final String BATCH_SIZE_CONF = "batchSize";
   public static final String BATCH_TIMEOUT_CONF = "batchTimeout";
   public static final String ENABLED_CONF = "enabled";
   public static final String INDEX_CONF = "index";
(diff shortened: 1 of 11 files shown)


mattf-horton:
I found a gap in my logic.  Will re-open after fixing.

mattf-horton:
Gap fixed.  (Added fix to decrease flush latency for most queues.)

Please give preliminary review feedback, as requested in the PR Intro.

mattf-horton:
Rebased to current master -- which apparently scotches this PR.

PR #449, opened 2017-02-09
METRON-701 Triage Metrics Produced by the Profiler

nickwallen:
### Problem

The motivating example is that I would like to create an alert if the number of inbound flows to any host over a 15 minute interval is abnormal.

In its most basic form, to perform Threat Triage I need to know two things.
1. What is the current value? 
2. What is an abnormal value?

The current problem with this is example is that the Threat Triage process is unable to answer the first question (What is the current value?). 

The value being interrogated here, the number of inbound flows, is not a static value contained within any single telemetry message. This value is calculated across multiple messages by the Profiler. The current Threat Triage process cannot be used to interrogate values that do not exist in a single telemetry message.

### Solution

I am proposing that we treat the Profiler as a source of telemetry. The measurements captured by the Profiler would be enqueued into a Kafka topic. We would then treat those Profiler messages like any other telemetry. This would allow us to enrich, triage, and index those messages.

* The Profiler topology was enhanced so that Profile measurements can be sent to HBase, Kafka, or both destinations.

    * Sending the Profile data to Kafka helps answer the first question; What is the current value? Prior to this PR, there is no way for the Triage process to 'see' a summarized value.  

    * The Profile data in HBase helps answer the second question; What is an abnormal value?  The triage rules would still use the Profiler Client and retrieve profile data from HBase to help define what an abnormal value is.  

* Enhanced the Profile's `result` field to allow for a `profile` expression that defines a value that is written to the profile store for later retrieval.

* Enhanced the Profile's `result` field to add one or more `triage` expressions that define values that are made available to the Threat Triage process.

* This change is backwards compatible with the current profile definition syntax.  
The current profile result syntax, for example, `"result": "2 + 2"` is equivalent to `"result": { "profile": "2 + 2" }`.

### Testing

* Alter the Profiler topology settings at `config/profiler.properties` to use a 1 minute period duration.  This is not necessary, but is  useful for the impatient.
```
profiler.period.duration=1
```

* Start only the Bro, Enrichment, and Profiler topologies.  Stop other unnecessary services like Elasticsearch, Kibana, Monit, etc.
```
service sensor-stubs start bro
$METRON_HOME/bin/start_parser_topology.sh -k node1:6667 -z node1:2181 -s bro
$METRON_HOME/bin/start_enrichment_topology.sh
$METRON_HOME/bin/start_profiler_topology.sh
```

* Create a Profile that simply counts messages.  The value of the counter will be stored in the profile and will also be made available to Threat Triage.
```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'bro'",
      "init":    { "count": "0" },
      "result": {
          "profile": "count",
          "triage": { "count": "count" }
      }
    }
  ]
}
```

* Look for messages to hit the 'enrichments' topic in Kafka.
```
[root@node1 0.3.1]#  /usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh --zookeeper node1:2181 --topic enrichments | grep period
{"period.start":1488233760000,"period":24803896,"profile":"test","count":591,"period.end":1488233820000,"is_alert":"true","entity":"global","timestamp":1488233781638,"source.type":"profiler"}
{"period.start":1488233820000,"period":24803897,"profile":"test","count":161,"period.end":1488233880000,"is_alert":"true","entity":"global","timestamp":1488233841600,"source.type":"profiler"}
```

* Ensure that the Profiler continues to write to HBase. (Be sure to set your Profiler Client period duration to match the Profiler itself).
```
$ bin/stellar -z node1:2181
Stellar, Go!
{es.clustername=metron, es.ip=node1, es.port=9300, es.date.format=yyyy.MM.dd.HH, bootstrap.servers=node1:6667, profiler.client.period.duration=1, profiler.client.period.duration.units=MINUTES}
[Stellar]>>>
[Stellar]>>> PROFILE_GET("test", "global", 5, "MINUTES")
[146, 147, 142, 120, 151]
```

* Add a threat triage rule that applies to the measurements coming out of the Profiler.  Run the following in a Stellar REPL.
```
# fetch the current config
conf := CONFIG_GET("ENRICHMENT", "profiler")

# create a rule 
rule := profile == "test" and count > 100
triage := { "name":"Test Rule", "rule": SHELL_GET_EXPRESSION('rule'), "score":"10" }
conf := THREAT_TRIAGE_ADD(conf, [triage])

# save the configuration
CONFIG_PUT("ENRICHMENT", conf, "profiler")
```

* Ensure that the threat triage rule is being applied correctly.  Look for a key named `threat.triage.level` when `count > 100`.
```
$ /usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh --zookeeper node1:2181 --topic indexing | grep profile

{"period.start":1488235260000,"period":24803921,"enrichmentsplitterbolt.splitter.end.ts":"1488235281611","profile":"test","count":161,"enrichmentsplitterbolt.splitter.begin.ts":"1488235281611","is_alert":"true","source.type":"profiler","threatintelsplitterbolt.splitter.end.ts":"1488235281616","threat.triage.level":10.0,"threatinteljoinbolt.joiner.ts":"1488235281618","enrichmentjoinbolt.joiner.ts":"1488235281613","period.end":1488235320000,"threatintelsplitterbolt.splitter.begin.ts":"1488235281616","entity":"global","timestamp":1488235281603}
```

These steps have shown how metrics produced by the Profiler can be triaged using the existing Threat Triage mechanism in Metron.


metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/KafkaDestinationHandler.java
```diff
@@ -0,0 +1,78 @@
...
+      JSONObject message = new JSONObject();
+      message.put("profile", measurement.getDefinition().getProfile());
+      message.put("entity", measurement.getEntity());
+      message.put("period", measurement.getPeriod().getPeriod());
+      message.put("periodStartTime", measurement.getPeriod().getStartTimeMillis());
+
+      // TODO How to serialize an object (like a StatisticsProvider) in a form that can be used on the other side? (Threat Triage)
+      // TODO How to embed binary in JSON?
+      message.put("value", measurement.getValue());
+
```
nickwallen: Everything works just fine if your profile produces a nice, easy to serialize value like a number.  If all you want to do is triage profiles that produce numbers, then this code works.  But how should this work if your profile is producing a complex object, like is done with the STATS package?

The Profiler has to serialize everything that the profiles produce.  Serializing objects into HBase works just fine.  But how should we serialize a complex object when sending it to Kafka?   

Random options follow.  What else might work?

* Sticking with the current architecture, I would have to embed binary data in the message JSON that is produced, which I don't think is possible or at least not ideal.

* I could just write the binary data alone to a Kafka topic, but then how do we pass the meta-information like the profile and entity name?  Do I create a wrapper object that contains the name, profile and value, then serialize all of that and write to a topic?  That would be drastically different from what we do today.

james-sirota: One of the problems that I see with this approach is that generally you wouldn't ask a question "do you think that the number of inbound connections from ip X is abnormal for the past 15 mins.  the question you would ask is the number of inbound connections for ip X abnormal for a Tuesday at 11.15AM.  To figure out the answer to this question we would have to retrieve telemetry from historical Tuesday data at 11.15AM and check if our value at hand for right now is an outlier based on that data.  METRON-690 is designed to make these retrieval patterns possible.  Because you retrieving values that are potentially more than 24 hours long I am not sure keeping this data in Kafka is a good idea.  I think you have to retrieve the values via a multi-get in Hbase

nickwallen: Pushing the data to Kafka is *not* meant as a replacement for HBase.  You still need to persist the Profile data in HBase, just like before.  The data is not pushed to Kafka to be used for retrieval.  The data is sent to Kafka to give us a hook into the triage process.  

Let me try to explain better.  To triage something, I need to know two things...
1. What is the current value? 
2. What is an abnormal value?

#### What is the current value? 
The Profile data in Kafka helps answer the first question. What is the current value? Prior to this PR, there is no way for the Triage process to 'see' a summarized value.  I tried to explain this in the PR description (probably poorly) as follows.

> The value being interrogated here, the number of inbound flows, is not a static value contained within any single telemetry message. This value is calculated across multiple messages by the Profiler. The current Threat Triage process cannot be used to interrogate values calculated by the Profiler.

#### What is an abnormal value?
The Profile data in HBase helps answer this second question.  What is an abnormal value?  The triage rules would still use the Profiler Client and retrieve the data from HBase that defines what normal is.  I think you are referring to this in your response.

Let me know if this provides any clarity or if I am misunderstanding the question.

cestella:
So, I think this is an interesting approach.  My issues with it are:
* You seem to be sending every profile into kafka, not just the configured ones.
* You seem to be assuming that one value only is being sent into the telemetry and it's the value that you store in HBase

I'd recommend, rather, that you make the `result` field more complex by making it a map where the key is the source (e.g. "hbase" or "kafka").  This allows you to separate the storage structure by storage medium.  You may, for instance, want to *STORE* a stats object in Hbase, but only send along the mean and standard deviation.  Also, I'd recommend allowing `result`to be either a string (which would presume only hbase is supported) or a Map, which would explicitly specify the structure for just the sources you want to write to.

Here's a worked example config for maximum clarity (!):
```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'squid'",
      "init":    { "stats": "STATS_INIT()" },
      "update":  { "stats": "STATS_ADD(stats, LENGTH(url))" },
      "result":  {
         "hbase" : "stats",
         "kafka" : "{ 'mean' : STATS_MEAN(stats), 'stddev' : STATS_SD(stats) }"
       }
      
    }
  ]
}
```

cestella:
Also, I will point out, this sets us up architecturally in the future to pull writer configs from zookeeper and support a series of other writers for the output of the profiler.

cestella:
This probably motivates us to allow `OUTLIER_MAD_SCORE` to accept median and median absolute deviation as parameters rather than just the State object as well.

nickwallen:
> You seem to be sending every profile into kafka, not just the configured ones

Just for clarity, you can define the destination for each profile.  It defaults to a `"destination" : ["hbase", "kafka"]`.  For example, if I only wanted to send to HBase.
```
{
   "profile": "profile-one-destination",
   "foreach": "ip_src_addr",
    "init":   { "x": "0" },
    "update": { "x": "x + 1" },
     "result": "x",
    "destination": ["hbase"]
}
```

But that is just a side point.  Your idea is really interesting.  We've talked before about having multiple result values, which I think is super useful.  I'll think on this a bit.  Thanks for the feedback.

nickwallen:
Outside the scope of your "multiple result" idea that I need to think more on...

The one thing I did not like about both approaches is the terminology.  Kind of silly, but important for usability.  

What I mean is the use of the terms 'hbase' and 'kafka'.  I think it doesn't make clear to the user why they would want to choose one over the other.  I would really like to find non-implementation-specific terms that describe the function of each better.

Random thoughts...
* 'kafka' -> 'triage' or ??
* 'hbase' -> 'store' or 'profile' or ??

nickwallen:
Still thinking through the implications, but it looks pretty clean and intuitive this way (at least more intuitive).  
```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'squid'",
      "update":  { "stats": "STATS_ADD(stats, LENGTH(url))" },
      "result":  {
         "profile" : "stats",
         "triage" : "{ 'mean' : 'STATS_MEAN(stats)', 'stddev' : 'STATS_SD(stats)' }"
       }     
    }
  ]
}
```

Maybe even get rid of "result" altogether?
```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'squid'",
      "update":  { "stats": "STATS_ADD(stats, LENGTH(url))" },
      "profile" : "stats",
      "triage" : "{ 'mean' : 'STATS_MEAN(stats)', 'stddev' : 'STATS_SD(stats)' }"
    }
  ]
}
```

cestella:
@nickwallen Hmm, how about the following, with `profile` and `triage` here being entirely user specifiable to break up the various ways you write:

```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'squid'",
      "update":  { "stats": "STATS_ADD(stats, LENGTH(url))" },
      "result":  {
         "profile" : {
                         "output" : "stats",
                         "if" : "STATS_COUNT(stats) > 0)",
                         "writer" : "hbase"
                         },
         "triage" : {
                       "output" : "{ 'mean' : STATS_MEAN(stats), 'stddev' : STATS_SD(stats) }",
                       "if" : "STATS_COUNT(stats) > 0 && STATS_SD(stats) > 10",
                       "writer" : "kafka"
                        }
       }     
    }
  ]
}
```

nickwallen:
I can see the value of the additional flexibility here.  Of course, the flip side is that I am always worried about too much complexity, as you probably guessed. 

I don't know if your proposal gets us all the way there in regards to user-focused over implementation-focused terminology.  Personally, I'd like to see profile definitions that are portable and work no matter where the Profiler is configured to persist data.  But maybe that is a pipe dream.

Your point on backwards compatibility is a good one.

cestella:
Well, let me try to make the case that this is user-focused while being aware of the limitations of implementation. ;)

The main aim for adaptability is to allow multiple representations to be stored in multiple datastores.  The representation has a 1 to n relationship with the data store, (maybe "writer" can be a list of writers or a single writer?).  This puts the top-level citizen as the representation associated with a naming about what it is intended to be used for.  Put simply, it's not that kafka can only handle JSON blobs, but rather it's that we need the kurtosis for the `kurtosis_triage` rule (by the way, `kurtosis_triage` should be part of the message constructed along with the `source.type` of `profiler`...maybe `profile.type`).

nickwallen:
I think what I mean is a little different (but maybe I've missed your point.)  

For example, when @james-sirota first reviewed this PR he was confused why we would send data to Kafka.  He thought it was a replacement for HBase, rather than an addition to that.  His mistake was totally understandable.  Using terms like 'kafka' and 'hbase', forces the user to under why they would want to send profile data to HBase and why they would want to send profile data to Kafka.  It forces the user to know the implementation.

But I am saying that users should not need to know the implementation details of the Profiler.  They should just tell us if they want the profile data stored for later and whether they want to triage the data from the Profiler.

So I am suggesting that to be "user focused" we use terms that focus on the functionality from the user's perspective, not terms based on how we've implemented the Profiler.  A user would tell us to 'triage' or not; they would not tell us 'kafka' or not.

nickwallen:
@cestella The core of your idea that really solves the problem that I was hung up on was adding two "result" expression; one for each destination.
* The HBase destination is just like it is now.  I can write any serializable Java Object.  
* The Kafka destination is more limited.  Its expression *must* result in a String.  We force the user to serialize the data.  And we kind of have to, we have no good way to embed a serialized object in a JSON message.

cestella:
@nickwallen Yep, I see what you mean.  I think we had different interpretations of "user focused."

I think where I landed here on what I'd like can be broken down into a near-term, medium-term and a long-term vision for the profiler.

**Near Term**
In the near term, for this PR, we need the ability to:
* Write from the profiler to kafka so we can triage the output of the profiler
* Adjust the representations of the data the profiler writes based on the places it writes to
  * For HBase, any kryo serialized object
  * For Kafka, any fundamental structure (e.g. number, string) or Map of fundamental structures.
* Be backwards compatible with the current syntax.

Strictly speaking, I'll adopt your approach here of separating representation by its destination, where that destination is restricted to the possible destinations inside of Metron's current architecture.  So called *destination-focused*, rather than separating representation by its storage mechanism, where those storage mechanisms are restricted to the possible mechanisms that we support in Metron.  So called *writer-focused*

In the following examples, every tick, the following happens:
* 1 message is written to HBase with the stats function
* 1 message is written to Kafka with a message that looks like this:
```
{
  'profile' : 'test',
  'entity' : 'global'
  'mean' : ####,
  'stddev' : ####
}
```

This looks like: 
```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'squid'",
      "update":  { "stats": "STATS_ADD(stats, LENGTH(url))" },
      "result":  {
         "profile" : "stats",
         "triage" : "{ 'mean' : STATS_MEAN(stats), 'stddev' : STATS_SD(stats) }"
       }     
    }
  ]
}
```

**Medium Term**

This gets expanded to allow for multiple elements written per profile.

In the following examples, every tick, the following happens:
* 2 message is written to HBase for profile `test`
  * entity: `global:stats`
  * entity: `global:count`
* 2 message is written to Kafka with a message that looks like this:
```
{
  'profile' : 'test',
  'entity' : 'global'
  'result_type' : 'baseline_stats'
  'mean' : ####,
  'stddev' : ####
}
```
and
```
{
  'profile' : 'test',
  'entity' : 'global'
  'result_type' : 'kurtosis'
  'kurtosis' : ####
}
```

This looks like:
```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'squid'",
      "update":  { "stats": "STATS_ADD(stats, LENGTH(url))" },
      "result":  {
         "profile" : {
                      "stats" :  "stats",
                      "count" : "STATS_COUNT(stats)"
         "triage" : {
                      "baseline_stats" : "{ 'mean' : STATS_MEAN(stats), 'stddev' : STATS_SD(stats) }",
                      "kurtosis" : "STATS_KURTOSIS(stats)"
                       }
                   }   
    }
  ]
}
```


**Longer Term**

This is where, in my mind, the writer-focused morphs into 'writer configuration' focused, which is to say, not just the transport, but also the destination.  In this world, we can directly associate the representation of the things we're writing from the profiler with the destination.  Our point of configuration for new writers in Metron is the `MessageWriter` and `BulkMessageWriter` interfaces.  We recently pulled out the configs into their own indexing configs, keyed by writer (kafka, elasticsearch, etc).  Imagine that the writers are configured entirely there and that it's not writer-oriented, but use-case oriented.  Instead of what we have now in the indexing config, we can make it:
```
{
  "writers" : {
     "kafka" : {
      "batchSize" : 1,
      "enabled" : true
     },
   "hbase_profile" : {
      "batchSize" : 5,
      "enabled" : true
   }
   },
  "endpoints" : {
     "triage" : {
         "writer" : "kafka",
         "queue" : "enrichments"
                     },
      "profile" : {
         "writer" : "hbase_profile",
         "table" : "profile:P"
      }
  }
}
```

here, the two forms merge into one because we can represent using our core abstractions the capability-driven design that you are focused on, @nickwallen .  In this world, the profiler is simple, it just writes messages out to the indexing topology.  The structure looks of the tuple looks like:
* message
* endpoint

The indexing topology will use the source type to pull the config and, since the endpoint is specified in the tuple, it will use the endpoint to write the message to the appropriately configured destination.

In this world, the example in the medium term does not change:
```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'squid'",
      "update":  { "stats": "STATS_ADD(stats, LENGTH(url))" },
      "result":  {
         "profile" : {
                      "stats" :  "stats",
                      "count" : "STATS_COUNT(stats)"
         "triage" : {
                      "baseline_stats" : "{ 'mean' : STATS_MEAN(stats), 'stddev' : STATS_SD(stats) }",
                      "kurtosis" : "STATS_KURTOSIS(stats)"
                       }
                   }   
    }
  ]
}
```
`profile` and `triage` are interpreted to mean endpoint names, who can be looked up in the indexing configuration.

ottobackwards:
I think routing from stellar within routing in storm is confusing.

mmiklavc:
@nickwallen I agree with you about exposing implementation details via our API. I think it better to name things according to feature/function, not the underlying implementation. 

@cestella I like the summary you provide for short/medium/long term solutions to this problem. I especially like the idea of being able to customize the writers used for the various endpoints.

nickwallen:
**Near Term:**  I like it.   I think we've converged on "near term".  Yay!  I will tackle these items as part of this PR.

> **Longer Term:** ... In this world, the profiler is simple, it just writes messages out to the indexing topology. 

Great! Yep, this is what I was hoping to do when we first starting batting this idea around.

nickwallen:
@cestella Thanks for laying out your other ideas for Medium and Longer term.  We can open those up for community debate on separate JIRAs, but it was very worthwhile for you to begin laying those out here.  They provided some good context.

@ottobackwards Not sure which part your comment was in reference to.  Do you have any concerns specifically with the "Near Term" items that I will tackle as part of this PR?  I'd like to make sure we reach consensus on those first.  We can debate the other items later (if that works for you.)

ottobackwards:
I'm sorry, no Nick I don't

nickwallen:
> Be backwards compatible with the current syntax.

This proposed syntax isn't directly backwards compatible.  Were you assuming we would do a translation of sorts?  

Like translate this...
```
{ "profiles": [
    {
      ...
      "result":  "stats"
    }
  ]
}
```

To this...
```
{ "profiles": [
    {
      ...
      "result":  {
         "profile" : "stats",
       }     
    }
  ]
}
```

cestella:
@nickwallen yes, I was

nickwallen:
> Ninja Edit: I think the kafka topic written to should be pulled from zookeeper...

@cestella I remember now why I settled on making the topic name a static configuration from the topology properties.  Our `BulkMessageWriterBolt` & `KafkaMessageWriter` classes seem to have been designed to only work with statically defined topic names.  I would have to change those core classes, if I want to set the topic name dynamically from Zookeeper.  And I tend to get smacked around when coming close to core classes :)

Am I missing something?  Is there a way to define the topic dynamically while using the `BulkMessageWriterBolt` & `KafkaMessageWriter` classes unchanged?

If not, I'd prefer to keep the topic name as a static property defined in the topology properties, at least for this PR. My second choice would be to open a completely separate PR to update those core classes to accept dynamic topic names.  It does not seem like a complex change, but those are core classes and will likely cause some debate.

cestella:
Ok, I'm ok with that.  The writer should be more adaptable here and that shouldn't hold your PR up, agreed.  Can we make it the enrichment queue, though?

nickwallen:
Ok, good.  I don't see any problems there.  Will use the 'enrichments' queue.

nickwallen:
> Am I missing something? Is there a way to define the topic dynamically while using the BulkMessageWriterBolt & KafkaMessageWriter classes unchanged?

Created [METRON-738](https://issues.apache.org/jira/browse/METRON-738) to track this 'wish list' enhancement.

nickwallen:
I made the required changes and updated the PR description to reflect that.  Please take a look and review.

cestella:
Looks great, quick question.  If I submit a profile that looks like:
```
{
  "profiles": [
    {
      "profile": "test",
      "foreach": "'global'",
      "onlyif": "source.type == 'bro'",
      "init":    { "count": "0" },
      "result": {
          "profile": "count",
          "triage": "{ 'blah' : count, 'zork' : 'zork'}"
      }
    }
  ]
}
```
Will I get messages in kafka that look like:
```
{"period.start":1488233820000
,"period":24803897
,"profile":"test"
,"blah":161
,"zork":"zork"
,"period.end":1488233880000
,"is_alert":"true"
,"entity":"global"
,"timestamp":1488233841600
,"source.type":"profiler"
}
```

I think that's an important aspect as people will probably want to submit multiple things to further triage or give context since they cannot send along our summary objects.


Also, if someone tries to submit something that JSON can't handle (like a stats object), will it get dropped or will an exception occur?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/KafkaDestinationHandler.java
```diff
@@ -0,0 +1,106 @@
...
+    message.put("is_alert", "true");
+
+    // append each of the triage values to the message
+    measurement.getTriageValues().forEach((key, value) -> {
+
+      if(isValidType(value)) {
+        message.put(key, value);
+
+      } else {
+        throw new IllegalArgumentException(String.format("invalid type for triage value: profile=%s, entity=%s, key=%s, value=%s",
```
nickwallen: @cestella asked...

> Also, if someone tries to submit something that JSON can't handle (like a stats object), will it get dropped or will an exception occur?

Good question.  The type check for the `triage` block occurs here in `KafkaDestinationHandler`.  This is called by `ProfileBuilderBolt.handleTick` when a tick tuple is received.  The `KafkaDestinationHandler` throws an exception which is then caught and logged by the bolt.

I was originally thinking this was OK, but after looking over this again, an exception here would impact any of the Profile-Entity pairs being managed by the same `ProfileBuilderBolt` instance.  I was originally thinking that only the specific problematic profile would be impacted, but this is not the case.  

I think I need to change this to log the error and move on to the next triage expression.  Of course, as a user I'd like to know explicitly if I make this mistake.  Hmm... I'll have to noodle on this.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/KafkaDestinationHandler.java
```diff
@@ -0,0 +1,106 @@
...
+  @Override
+  public void emit(ProfileMeasurement measurement, OutputCollector collector) {
+
+    JSONObject message = new JSONObject();
+    message.put("profile", measurement.getDefinition().getProfile());
+    message.put("entity", measurement.getEntity());
+    message.put("period", measurement.getPeriod().getPeriod());
+    message.put("period.start", measurement.getPeriod().getStartTimeMillis());
+    message.put("period.end", measurement.getPeriod().getEndTimeMillis());
+    message.put("timestamp", System.currentTimeMillis());
```
nickwallen: The `timestamp` comes from `System.currentTimeMillis()`.  I don't like this very much, but I am not sure of a better alternative at the moment.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/KafkaDestinationHandler.java
cestella: I'd recommend reporting the error and not killing the topology so it shows in the storm UI.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/KafkaDestinationHandler.java
```diff
@@ -0,0 +1,106 @@
...
+    message.put("source.type", sourceType);
+    message.put("is_alert", "true");
+
+    // append each of the triage values to the message
+    measurement.getTriageValues().forEach((key, value) -> {
+
+      if(isValidType(value)) {
+        message.put(key, value);
```
nickwallen: @cestella said...

>     {
>       "profile": "test",
>       "foreach": "'global'",
>       "onlyif": "source.type == 'bro'",
>       "init":    { "count": "0" },
>       "update":  { "count": "count + 1" },
>       "result": {
>           "profile": "count",
>           "triage": "{ 'blah' : count, 'zork' : 'zork'}"
>       }
> 
> Will I get messages in kafka that look like:
> 
> ```
> {
> "period.start":1488233820000
> ,"period":24803897
> ,"profile":"test"
> ,"blah":161
> ,"zork":"zork"
> ,"period.end":1488233880000
> ,"is_alert":"true"
> ,"entity":"global"
> ,"timestamp":1488233841600
> ,"source.type":"profiler"
> }
> ```

Yes, that is exactly what you would get.  The keys 'blah' and 'zork' would be added to the message with the result of evaluating the associated triage expression.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/KafkaDestinationHandler.java
cestella: Hmm, maybe the period.start?

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/KafkaDestinationHandler.java
nickwallen: It currently doesn't kill the topology.  It would just impact the flush of any other profiles that happen to be handled by the same bolt instance.  But yes, needs fixed.

nickwallen:
I think this is ready for final review.  Come one, come all.  Would love to get this closed out.

cestella:
Yep, looks good, got my +1

PR #450, opened 2017-02-10
METRON-690: Create a DSL-based timestamp lookup for profiler to enable sparse windows

cestella:
Creating a small DSL to allow specifying profiles from windows of time that may:
* repeat non-contiguously (e.g. the same hour every day for a week)
* have inclusions and exclusions (e.g. the same hour every day for a week excluding holidays and weekends)

This also provides a `PROFILE_WINDOW` Stellar function which accepts this DSL as a string and can return the set of profiler periods based on the times selected and suitable for passing to `PROFILE_GET`.  To be clear, you can compose the two functions.

 i.e. `PROFILE_GET('profile', 'entity', PROFILE_WINDOW('1 hour window every 24 hours starting from 14 days ago including the current day of the week excluding weekends, holidays:us'))` would retrieve all the profile measurements written for the profile "profile" and entity "entity" for the last hour on the same weekday excluding weekends and US holidays across the last 14 days

For a complete description with examples, see [here](https://github.com/cestella/incubator-metron/tree/METRON-690-squashed/metron-analytics/metron-profiler-client#profile_window).

Acceptance testing plan will be submitted as a follow-on comment.


cestella:
For reviewers, please note that as this is an antlr DSL, you may skip the generated code (the java files in the `org.apache.metron.profiler.client.window.generated` package , which is the bulk of this PR.

cestella:
I wanted to point out one thing; while this is a DSL called from a DSL (stellar), the way the parser is written, we cache the output of the parser and can reuse them for evaluation, thus saving us much of the impact of the parsing of this DSL.

metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/stellar/IntervalPredicateTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class IntervalPredicateTest {
+  @Test
+  public void testBasicTest() {
+    List<Range<Long>> intervals = new ArrayList<Range<Long>>() {{
+      add(Range.between(0L, 10L));
+      add(Range.between(20L, 30L));
+      add(Range.between(40L, 50L));
+    }};
+    IntervalPredicate predicate = new IntervalPredicate.Identity(intervals);
```
justinleet: Can you make this (and the similar instances) IntervalPredicate<Long> so my compiler likes me again?

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/window/WindowProcessor.java
```diff
@@ -0,0 +1,321 @@
...
+
+  private void enterList() {
+    stack.push(LIST_MARKER);
+  }
+
+  private List<Function<Long, Predicate<Long>>> getPredicates() {
+    LinkedList<Function<Long, Predicate<Long>>> predicates = new LinkedList<>();
+    while (true) {
+      Token<?> token = stack.pop();
+      if (token == LIST_MARKER) {
```
justinleet: Should this be .equals()?

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/window/WindowProcessor.java
```diff
@@ -0,0 +1,321 @@
...
+    stack.push(DAY_SPECIFIER_MARKER);
+  }
+
+  @Override
+  public void exitSpecifier(WindowParser.SpecifierContext ctx) {
+    LinkedList<String> args = new LinkedList<>();
+
+    while (true) {
+      Token<?> token = stack.pop();
+      if (token == DAY_SPECIFIER_MARKER) {
```
justinleet: Same deal, .equals()

metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/stellar/IntervalPredicateTest.java
cestella: Sure, but your compiler will never like you.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/window/WindowProcessor.java
cestella: Nope, it's reference equals.  `LIST_MARKER` is a specific instance of Token.

justinleet: Sounds good, thanks.

cestella: I should say, it's a specific static instance used for the parser to denote the end of the list of values in the stack.

james-sirota:
I think the API looks great.  Can we provide a grammar in the comments for constructing the PROFILE_WINDOW function?  I think the API is so flexible that it may be hard to wrap your head around how to construct the profile retrieve string

cestella:
The Readme additions were intended to break the expressions down into the possible phrases.  Do you think those sections need to be structured differently?

james-sirota:
I think taking the string as an argument is really powerful, but it's also really flexible.  "1 hour window every 24 hours starting from 14 days ago including the current day of the week excluding weekends, holidays:us" is a really long string.  I was thinking maybe define a grammar to look something like:

<command> = <window_duration> <window_period> <window_limit> <window_inclusions> <window_exlcusions> <holidays_spec>
<window_duration> = number <unit_spec>
<unit_spec> = MINUTS|HOURS|DAYS

...or something like that...a grammar for constructing this string

cestella:
@james-sirota Yep, agreed.  I'll modify the docs.

cestella:

   * `storm kill snort`
   * `storm kill bro`

## Preliminaries
`export METRON_HOME=/usr/metron/0.3.1` 

```
#!/usr/bin/python
import random
import sys
import time
def main():
  mu = float(sys.argv[1])
  sigma = float(sys.argv[2])
  freq_s = int(sys.argv[3])
  while True:
    print out
    sys.stdout.flush()
    time.sleep(freq_s)

if __name__ == '__main__':
  main()
```

```
```

## Deploy the custom parser
* Edit the value parser config at `$METRON_HOME/config/zookeeper/parsers/value.json`:
```
{
  "sensorTopic":"value",
  "fieldTransformations" : [
    {
    "transformation" : "STELLAR"
   ,"output" : [ "num_profiles_parser", "mean_parser" ]
   ,"config" : {
      "num_profiles_parser" : "LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us')))",
      "mean_parser" : "STATS_MEAN(STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us'))))"
               }
    }
                           ]
}
```

* Edit the value enrichment config at `$METRON_HOME/config/zookeeper/enrichments/value.json`:
```
{
  "enrichment" : {
   "fieldMap": {
      "stellar" : {
        "config" : {
        "num_profiles_enrichment" : "LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us')))",
        "mean_enrichment" : "STATS_MEAN(STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us'))))"
                  }
      }
    }
  }
}
```
* Create the value kafka topic:
  `/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper node1:2181 --create --topic value --partitions 1 --replication-factor 1`
* Push the configs via `$METRON_HOME/bin/zk_load_configs.sh -m PUSH -i $METRON_HOME/config/zookeeper -z node1:2181`
* Start via `$METRON_HOME/bin/start_parser_topology.sh -k node1:6667 -z node1:2181 -s value`


## Start the profiler

```
{
  "profiles": [
    {
      "profile": "stat",
      "foreach": "'global'",
      "onlyif": "true",
      "init" : {
               },
      "update": {
        "s": "STATS_ADD(s, value)"
                },
      "result": "s"
    }
  ]
}
```

## Test Case

`python ~/rand_gen.py 0 1 1 | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic value`
* Wait for at least 32 minutes and execute the following via the Stellar REPL:
```
# Grab the profiles from 1 minute ago to 8 minutes ago
LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('from 1 minute ago to 8 minutes ago')))
# Looks like 7 were returned, great.  Now try something more complex
# Grab the profiles in 5 minute windows every 10 minutes from 2 minutes ago to 32 minutes ago:
#  32 minutes ago til 27 minutes ago should be 5 profiles
#  22 minutes ago til 17 minutes ago should be 5 profiles
#  12 minutes ago til 7 minutes ago should be 5 profiles
# for a total of 15 profiles
LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us')))
```
```
Stellar, Go!
{es.clustername=metron, es.ip=node1, es.port=9300, es.date.format=yyyy.MM.dd.HH, profiler.client.period.duration=1, profiler.client.period.duration.units=MINUTES}
[Stellar]>>> # Grab the profiles from 1 minute ago to 8 minutes ago
[Stellar]>>> LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('from 1 minute ago to 8 minutes ago')))
7
[Stellar]>>> # Looks like 7 were returned, great.
[Stellar]>>> # Grab the profiles from 2 minutes ago to 32 minutes ago
[Stellar]>>> LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('from 2 minutes ago to 32 minutes ago')))
30
[Stellar]>>> # Looks like 30 were returned, great.
[Stellar]>>> # Now try something more complex
[Stellar]>>> # Grab the profiles in 5 minute windows every 10 minutes from 2 minutes ago to 32 minutes ago:
[Stellar]>>> #  32 minutes ago til 27 minutes ago should be 5 profiles
[Stellar]>>> #  22 minutes ago til 17 minutes ago should be 5 profiles
[Stellar]>>> #  12 minutes ago til 7 minutes ago should be 5 profiles
[Stellar]>>> # for a total of 15 profiles
[Stellar]>>> LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us')))
15
[Stellar]>>> STATS_MEAN(STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us'))))
0.028019658637877063
[Stellar]>>>
```
* Delete any value index that currently exists (if any do) via `curl -XDELETE "http://localhost:9200/value*"`
* Wait for a couple of seconds and run 
  * `curl "http://localhost:9200/value*/_search?pretty=true&q=*:*" 2> /dev/null` 
  * You should see values in the index with non-zero fields:
     * `num_profiles_enrichment` should be 15
     * `num_profiles_parser` should be 15
     * `mean_enrichment` should be a non-zero double
     * `mean_parser` should be a non-zero double
For reference, a sample message for me is:
```
 "_index" : "value_index_2017.02.17.18",
      "_type" : "value_doc",
      "_id" : "AVpNPI8JQV00TRR_I4zn",
      "_score" : 1.0,
      "_source" : {
        "adapter:stellaradapter:end:ts" : "1487354498620",
        "threatinteljoinbolt:joiner:ts" : "1487354498628",
        "enrichmentsplitterbolt:splitter:end:ts" : "1487354498576",
        "num_profiles_parser" : 15,
        "enrichmentsplitterbolt:splitter:begin:ts" : "1487354498571",
        "enrichmentjoinbolt:joiner:ts" : "1487354498622",
        "mean_enrichment" : 0.025770908095283665,
        "adapter:stellaradapter:begin:ts" : "1487354498578",
        "source:type" : "value",
        "original_string" : "{ \"value\" : -0.274471660322 }",
        "threatintelsplitterbolt:splitter:begin:ts" : "1487354498625",
        "num_profiles_enrichment" : 15,
        "threatintelsplitterbolt:splitter:end:ts" : "1487354498625",
        "value" : -0.274471660322,
        "mean_parser" : 0.025770908095283665,
        "timestamp" : 1487354498529
      }
```

Here we've validated that the new window function can be called from the relevant topologies as well as the REPL and give consistent results that make sense.

ottobackwards:
Not for nothing @cestella but those steps should just be a script if we are going to need them all the time

metron-analytics/metron-profiler-client/src/main/antlr4/org/apache/metron/profiler/client/window/generated/Window.g4
```diff
@@ -0,0 +1,138 @@
...
+}
+COMMA : ',';
+COLON : ':';
+
+WINDOW : 'window' | 'windows';
+
+INCLUDE : 'include' | 'INCLUDE' | 'includes' | 'INCLUDES' | 'including' | 'INCLUDING';
+EXCLUDE : 'exclude' | 'EXCLUDE' | 'excludes' | 'EXCLUDES' | 'excluding' | 'EXCLUDING';
+
+NOW : 'NOW' | 'now';
```
justinleet: It doesn't look like "NOW" is used.  Can we drop this token?

justinleet:
Ran this up and tested per @cestella instructions, along with testing out the language a bit more.  Everything worked well for me.

My only gotcha, and this is more a general Stellar issue, rather than an issue here is that if you make a mistake with a syntax error, e.g. `STATS_MEAN(STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window starting ago'))))`, results in `org.apache.metron.common.dsl.ParseException: Unable to execute: expected at least 3 argument(s), found 2
`.  While technically true (because the PROFILE_WINDOW() threw an error, resulting in PROFILE_GET() only having two args), it's not a particularly helpful error.  I'm curious if you have any thought on bubbling up the most specific error @cestella 

Otherwise, I'm +1 on this.  It was very flexible, easy to use, and a great contribution.  Thanks a lot!

cestella:
@justinleet yeah, there's a bug right now with stellar not bubbling up the inner exception.  I created https://issues.apache.org/jira/browse/METRON-732 to track it.

cestella:
@ottobackwards Yes, your observation is a good one.  I wanted to toe the line between "obvious step-by-step" and "automated" and ended up on the side of "step-by-step obvious" because 
* I wanted people to be able to tinker and play with the steps to find things that I didn't
* I didn't want to have to deal with weird one-off timing errors around some of the things specified here

It's probably worth discussing acceptance testing scripts as a general category of testing that we might be able to do with some of these scripts forming the beginnings of them.

ottobackwards:
Honestly, I was thinking more about the 'kill everything' part, which seems to be repeated a lot

cestella:
Oh, well, I was honestly hoping that'd go away naturally if we can trim down vagrant to give us enough headroom to not have to kill things to have it run smoothly, but your point is well taken.

metron-analytics/metron-profiler-client/README.md
```diff
@@ -91,37 +60,268 @@ want to change the global Client configuration so as not to disrupt the work of
...
+* A duration starting 1 hour ago and ending now
+   * <span style="color:blue">`from 1 hour ago`</span>
+   * <span style="color:blue">`from 1 hour`</span>
+   * <span style="color:blue">`starting from 1 hour ago`</span>
+   * <span style="color:blue">`starting from 1 hour`</span>
+* A duration starting 1 hour ago and ending 30 minutes ago: 
+   * <span style="color:blue">`from 1 hour ago until 30 minutes ago`</span>
+   * <span style="color:blue">`from 30 minutes ago until 1 hour ago`</span>
+   * <span style="color:blue">`starting from 1 hour ago to 30 minutes ago`</span>
+   * <span style="color:blue">`starting from 1 hour to 30 minutes`</span>
```
nickwallen: Should I be able to do this?  It does not seem to like floats and throws an exception.
```
from 1.5 hours ago
```

metron-analytics/metron-profiler-client/README.md
```diff
@@ -91,37 +60,268 @@ want to change the global Client configuration so as not to disrupt the work of
...
+* `time_interval` - A time amount followed by a unit (e.g. 1 hour).  The unit may be "minute", "day", "hour" with any pluralization.
+* `TO` - Can be the words "until" or "to"
+* `AGO` - Optionally the word "ago"
+
+The `TO time_interval AGO` portion is optional.  If unspecified then it is expected that the time interval ends now.
+
+Due to the vagaries of the english language, the from and the to portions, if both specified, are interchangeable
+with regard to which one specifies the start and which specifies the end.  
+
+In other words <span style="color:blue">`starting from 1 hour ago to 30 minutes ago`</span> and
```
nickwallen: Are the time intervals exclusive or inclusive?    Might be good to call that out specifically, if you haven't.

metron-analytics/metron-profiler-client/README.md
```diff
@@ -91,37 +60,268 @@ want to change the global Client configuration so as not to disrupt the work of
...
+###### <span style="color:blue">Total Temporal Duration</span>
+
+Total temporal duration is specified by a phrase: `FROM time_interval AGO TO time_interval AGO`
+This indicates the beginning and ending of a time interval.
+* `FROM` - Can be the words "from" or "starting from"
+* `time_interval` - A time amount followed by a unit (e.g. 1 hour).  The unit may be "minute", "day", "hour" with any pluralization.
+* `TO` - Can be the words "until" or "to"
+* `AGO` - Optionally the word "ago"
+
+The `TO time_interval AGO` portion is optional.  If unspecified then it is expected that the time interval ends now.
```
nickwallen: One thing to consider is that we seem to be assuming "processing time" with the grammar.   For example, in the testing notes that you provided, we are enriching the message with an expression like this.

```
STATS_MEAN(STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us'))))
```

When we grab the profile from "2 minutes ago" this will be 2 minutes ago from processing time (aka system time) rather than the event time of the message.

Is there any way to support event time now?  I'm not sure that this is needed right now, but thought we should have a discussion around this at the very least.

metron-analytics/metron-profiler-client/README.md
```diff
@@ -91,37 +60,268 @@ want to change the global Client configuration so as not to disrupt the work of
...
+Skip distance is the amount of time between temporal window beginnings that the next window starts.
+It is, in effect, the window period.  
+
+It is specified by the phrase `EVERY time_interval`
+* `time_interval` - A time amount followed by a unit (e.g. 1 hour).  The unit may be "minute", "day", "hour" with any pluralization.
+* `EVERY` - The word/phrase "every" or "for every"
+
+**Examples**
+
+* A repeating 30 minute window starting 2 hours ago and repeating every hour until now.
```
nickwallen: How do I select the last 1 hour window for the past 8 Tuesdays (assuming today is Tuesday)?

How do I select the last 1 hour window for the past 8 'current day of the week'?

metron-analytics/metron-profiler-client/README.md
```diff
@@ -91,37 +60,268 @@ want to change the global Client configuration so as not to disrupt the work of
...
+
+Note that the `config_overrides` parameter operates exactly as the `config_overrides` argument in `PROFILE_GET`.
+The only available parameters for override are:
+* `profiler.client.period.duration`
+* `profiler.client.period.duration.units`
+
+##### The Profile Selector Language
+
+The domain specific language can be broken into a series of clauses, some optional
+* <span style="color:blue">Total Temporal Duration</span> - The total range of time in which windows may be specified
```
cestella: It's rendered in the site-book via doxia

metron-analytics/metron-profiler-client/README.md
cestella: All time intervals are inclusive, but it's worth mentioning in the docs.

metron-analytics/metron-profiler-client/README.md
cestella: Yes, you can pass the event time as the second argument to `PROFILE_WINDOW`.

metron-analytics/metron-profiler-client/src/main/java/Window.tokens
```diff
@@ -0,0 +1,17 @@
+COMMA=1
```
cestella: I haven't found a way.  I looked into it with Stellar and it didn't seem possible at the time.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/window/WindowProcessor.java
```diff
@@ -0,0 +1,321 @@
...
+
+  private static String normalizeTimeUnit(String s) {
+    String ret = s.toUpperCase().replaceAll("[^A-Z]", "");
+    if(ret.endsWith("S")) {
+      return ret.substring(0, ret.length() - 1);
+    }
+    return ret;
+  }
+
+  public static Window parse(String statement) throws ParseException {
```
cestella: What would you like it to be called?  Maybe `process`?

metron-analytics/metron-profiler-client/README.md
```diff
@@ -91,37 +60,268 @@ want to change the global Client configuration so as not to disrupt the work of
...
+
+One *must* specify either a total temporal duration or a temporal window width.
+The remaining clauses are optional.
+During the course of the following discussion, we will color code the clauses in the examples.
+
+From a high level, the language fits the following three forms:
+
+* <span style="color:red">`time_interval WINDOW?`</span><span style="color:purple">`(INCLUDING specifier_list)? (EXCLUDING specifier_list)?`</span>
+* <span style="color:red">`time_interval WINDOW?`</span><span style="color:green">`EVERY time_interval`</span><span style="color:blue">`FROM time_interval (TO time_interval)?`</span><span style="color:purple">`(INCLUDING specifier_list)? (EXCLUDING specifier_list)?`</span>
+* <span style="color:blue">`FROM time_interval (TO time_interval)?`</span>
```
cestella: Regarding 2, that is the intention of the color coding of the examples, which is lost in github but available in the site-book.

metron-analytics/metron-profiler-client/README.md
cestella: We do not support floats just yet as we utilize `TimeUnit` for our time conversion.  If you want more granular than hours, then it was intended that you go down to minutes: `90 minutes ago`

metron-analytics/metron-profiler-client/README.md
cestella: * `1 hour window every 24 hours starting from 56 days ago including tuesdays`
* `1 hour window every 24 hours starting from 56 days ago including this day of the week`

This was by far my favorite question so far.  A subtle overflow bug was caught during its answer.  There's now a unit test for this one.  ;)

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/window/WindowProcessor.java
```diff
@@ -0,0 +1,321 @@
...
+    parser.removeErrorListeners();
+    parser.addErrorListener(new ErrorListener());
+    parser.window();
+    if(treeBuilder.throwable != null) {
+      throw new ParseException(treeBuilder.throwable.getMessage(), treeBuilder.throwable);
+    }
+    return treeBuilder.getWindow();
+  }
+
+  public static String syntaxTree(String statement) {
```
cestella: Nobody is using syntaxTree, that's currently only for debugging issues with the grammar going forward for people adapting the language for new use-cases.  Also, @mattf-horton asked for it during the dev discussion.  It was marginally useful during the construction of the grammar to understand how things were being parsed.  I'll document it as such.

metron-analytics/metron-profiler-client/README.md
cestella: I'm going to think on these good suggestions and tackle them tomorrow am.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/WindowLookback.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+@Stellar(
+      namespace="PROFILE",
+      name="WINDOW",
+      description="The profiler periods associated with a fixed lookback starting from now.",
```
cestella: Good catch

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/IntervalPredicate.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.commons.lang3.Range;
+
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class IntervalPredicate<T> implements Predicate<T> {
```
cestella: That's a fair ask.  I'll rethink the docs and add javadocs tomorrow AM.

cestella:
Ok, good catch, I've made the functionality and some of the doc changes requested @nickwallen and I'll finish the rest in the morning.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/window/WindowProcessor.java
nickwallen: Sure `process` or `execute` sound good to me.

metron-analytics/metron-profiler-client/README.md
nickwallen: I'll have to checkout the site book.  Its too bad, Github doesn't render that.

metron-analytics/metron-profiler-client/README.md
nickwallen: That's great.

metron-analytics/metron-profiler-client/README.md
cestella: Yeah, I was bummed.

metron-analytics/metron-profiler-client/README.md
nickwallen: Very nice.  If there is not an example like this in the README (I probably missed it) it would be good to add.  This is closer to the type of use case that you were trying to solve initially with this PR.

cestella: good feedback, I'll add that.

metron-analytics/metron-profiler-client/README.md
cestella: Actually, the color coding should link the three major forms with the types of clauses used to construct the major forms, so for 1 it becomes more clear if you look at it from the site-book as well.

cestella:
Ok @nickwallen I think I have adjusted the docs and functionality sufficiently well that I have addressed your issues.  Would you please give it another look and let me know if I missed anything?

metron-analytics/metron-profiler-client/README.md
cestella: I made this more explicit in the docs.

metron-analytics/metron-profiler-client/README.md
cestella: I went ahead and made it color coded *and* made links.  I also verified that it renders properly in the site-book.

PR #451, opened 2017-02-11
METRON-157: Added CEF Parser

simonellistonball:
There is some discussion of using an external library on the jira ticket for this issue. The library in question is excellent, and covers the spec well, but applies types which don't get along with Metron (IP address, MAC address). This has a simplified parser focussing on speed and sticking to basic types.


kylerichardson:
@simonellistonball, thanks for picking this one up! I have unassigned the JIRA from myself as I've clearly not had the time to work on it recently.

simonellistonball:
@kylerichardson no problem at all, would really appreciate it if you could review, and add anything from any work you have on this.

trixpan:
@simonellistonball seems like the code will hit the same issue we hit in NIFI-3466? The CEF spec doesn mention but it seems like ArcSight behavior is to be able to MMM even when MMM is written with localised values for "Short Month Name"

metron-platform/metron-parsers/src/test/resources/org/apache/metron/parsers/cef/adallom.cef
```diff
@@ -0,0 +1 @@
+2016-04-01T09:29:11.356-0400 CEF:0|Adallom|Adallom|1.0|56fe779ee4b0459f4e9a484a|ALERT_CABINET_EVENT_MATCH_AUDIT|0|msg=Activity policy 'User download/view file' was triggered by 'scolbert@gmail.com' suser=wanderson@rock.com start=1459517280810 end=1459517280810 audits=["AVPR-4oIPeFmuZ3CKKrg","AVPR-wx80cd9PUpAu2aj","AVPR-6XGPeFmuZ3CKKvx","AVPSALn_qE4Kgs_8_yK9","AVPSASW3gw_f3aEvgEmi"] services=["APPID_SXC"] users=["lvader@hotmail.com"] cs6=https://abcd-remote.console.arc.com/#/alerts/56fe779ee4b0459f4e9a484a cs6Label=consoleUrl
```
trixpan: May be a good idea to use ASF owned domains on emails as other domains may be under trademark and/or copyright

metron-platform/metron-parsers/src/test/resources/org/apache/metron/parsers/cef/cyberark.json
```diff
@@ -0,0 +1,21 @@
+{
+    "\"Other info\"": "101.198.70.93",
+    "\"Safe Name\"": "Security Vulnerability Mgmt",
+    "\"Ticket Id\"": "Needed to verify config files being pulled ",
+    "deviceAction": "Retrieve password",
+    "deviceAddress": "120.99.70.3",
```
trixpan: I suspect metron will take care of the IP parsing but it may be a good idea to add ipv6 to the test units?

simonellistonball: There is no IP parsing. IPs are purely represented as strings in Metron, so this shouldn't be an issue. This would only be an issue at the validation stage, which is out of scope here.

simonellistonball:
Syslog timestamp capture looks to be locale sensitive here, though all other date parsing is SimpleDateFormat based, so should be robust to locale. Do you see this issue on syslog headers?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/cef/CEFParser.java
```diff
@@ -0,0 +1,272 @@
...
+			obj = mutate(obj, "spt", "ip_src_port");
+			obj = convertToInt(obj, "ip_src_port");
+
+			obj = mutate(obj, "act", "deviceAction");
+			// applicationProtocol
+			obj = mutate(obj, "app", "protocol");
+
+			obj.put("original_string", cefString);
+
+			// apply timestamp from message if present, using syslog timestamp,
```
simonellistonball: Good point, submitting a fix now.

metron-platform/metron-parsers/src/test/resources/org/apache/metron/parsers/cef/adallom.cef
simonellistonball: Another good point, cleaning up the test data a bit now.

trixpan:
No.  And under RFC 3164, Syslog's Mmm is English only but this certainty is not present in the CEF spec states MMM as SimpleDateFormat and makes no reference over locale. This in theory means it should be locale agnostic. 

 If they adopt the syslog approach, locale should not be an issue but being CEF God knows. :-) 

Regarding robustness, SDF should not be able to automatically recognise MMM in French on  metron cluster running under user.language=en. From the top of my head for this to occur, the code must invoke SDF specifying the locale used for parsing. 

This whole localised dates shouldn't be an issue for servers, as they frequently run without locale settings but its particularly complex within multinationals operating under multiple languages. Think about all those agents insisting in using local settings when crafting CEF messages...

Shouldn't happen but after witnessing vendors violating their own standards I lost faith :-) 

BTW, I am happy to forward you whatever I get back from HPE around this issue.

simonellistonball:
The joys of international date parsing, right? Seems like a the CEF standard is not the most well read among device vendors. A number of the 'from the wild' examples we've got in the tests already violate the rt standard set down by HPE, hence the DateUtils class separating the list of "according to the standard" and "found in the wild".

This feels like a much wider issue we should handle elsewhere, maybe in the new DateUtils class I introduced here. To do so properly we would have to have a way of feeding the source locale for the log feed into the parser. We should really open a discuss on the best way to do that in general, but maybe it's something bound to a general parser config (i.e. each parser can specify date locale) and this can be propagated on a general basis rather than parser to parser.

trixpan:
Yep. I would say unless HPE clarifies Mmm being English only, providing the parsers with the ability to set locale would be ideal. 

And I didn't even mentioned that they use Zzz aka the timezone representation from hell... :-) 

One my favourite posts on the subject is:

http://stackoverflow.com/a/28554443/888876

Which could be summarised by the question BST which of them? (also kindly illustrated here https://www.timeanddate.com/time/zones/)

simonellistonball:
Agreed, let's pull the date discussion into a wider forum. Apart from this, is there anything else you see in this parser specifically to block merging?

trixpan:
Seems ok to me.

The only last comment which certainly is not a blocker (and if I read the code correctly, is already addressed [here](https://github.com/apache/incubator-metron/pull/451/files#diff-5100b5d781b88b14f7335b604268905aR44) ) is to add escaped '='  to one of extension fields used in test as this is a quite common pattern and covered by specific escaping rules:

> If an equal sign (=) is used in the extensions, it has to be escaped with a backslash (\). Equal signs in
> the header need no escaping.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/cef/CEFParser.java
```diff
@@ -0,0 +1,274 @@
...
+		sb.append("\\|");
+
+		// extension capture:
+		sb.append("(?<extensions>.*)");
+		String pattern = sb.toString();
+
+		p = Pattern.compile(pattern);
+
+		// key finder for extensions
+		pext = Pattern.compile(EXTENSION_CAPTURE_PATTERN);
```
james-sirota: We can probably do this in Grok as well

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/cef/CEFParserTest.java
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			// test against an explicit json example
+			if (!targetJson.isEmpty()) {
+
+			}
+		}
+	}
+
+	protected boolean validateJsonData(final String jsonSchema, final String jsonData) throws Exception {
```
james-sirota: I like this part. Would be nice to eventually make this global

metron-platform/metron-parsers/src/test/resources/org/apache/metron/parsers/cef/adallom.cef
```diff
@@ -0,0 +1 @@
+2016-04-01T09:29:11.356-0400 CEF:0|Adallom|Adallom|1.0|56fe779ee4b0459f4e9a484a|ALERT_CABINET_EVENT_MATCH_AUDIT|0|msg=Activity policy 'User download/view file' was triggered by 'person@example.com' suser=auser@example.com start=1459517280810 end=1459517280810 audits=["AVPR-4oIPeFmuZ3CKKrg","AVPR-wx80cd9PUpAu2aj","AVPR-6XGPeFmuZ3CKKvx","AVPSALn_qE4Kgs_8_yK9","AVPSASW3gw_f3aEvgEmi"] services=["APPID_SXC"] users=["another@example.com"] cs6=https://abcd-remote.console.arc.com/#/alerts/56fe779ee4b0459f4e9a484a cs6Label=consoleUrl
```
james-sirota: Can anyone from the community provide more test data for this?

james-sirota:
+1 via inspection.  This is a great contribution. Thanks, Simon.  I think it would be nice to get more test data from different instances of Arcsight.  Does anyone in the community have access to this data set?  This is the official CEF spec. chrome-extension://ihgdgpjankaehldoaimdlekdidkjfghe/viewer.html#https://www.protect724.hpe.com/servlet/JiveServlet/downloadBody/1072-102-9-20354/CommonEventFormatv23.pdf Would be nice to find a test set that covers most of this.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/cef/CEFParser.java
trixpan: I believe grok is how the fireeye parser currently works?

trixpan:
@james-sirota these are synthetic but should cover all valid field types:

https://github.com/apache/nifi/blob/master/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestParseCEF.java#L37

https://github.com/apache/nifi/blob/master/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/test/java/org/apache/nifi/processors/standard/TestParseCEF.java#L51

And this may help you as well, although if I recall correctly, the fireeye CMS had some minor non compliances to the CEF spec.

https://github.com/apache/incubator-metron/blob/4a4cb8b117dbb66bbfb4915bca9d871a06682c28/metron-platform/metron-integration-test/src/main/sample/data/SampleInput/FireeyeExampleOutput#L6

Worth of notice:

1. Be mindfull that downstream ArcSight data will contain fields not present in the CEF specficiation. These are internal to ArcSight and to the best of my knowledge not openly documented. 

An example of such fields can be found [here](https://my.vertica.com/docs/7.1.x/HTML/Content/Authoring/FlexTables/LoadCEFData.htm):

```
CEF:0|ArcSight|ArcSight|6.0.3.6664.0|agent:030|Agent [test] type [testalertng] started|Low| 
eventId=1 mrt=1396328238973 categorySignificance=/Normal categoryBehavior=/Execute/Start 
categoryDeviceGroup=/Application catdt=Security Mangement categoryOutcome=/Success 
categoryObject=/Host/Application/Service art=1396328241038 cat=/Agent/Started 
deviceSeverity=Warning rt=1396328238937 fileType=Agent 
cs2=<Resource ID\="3DxKlG0UBABCAA0cXXAZIwA\=\="/> c6a4=fe80:0:0:0:495d:cc3c:db1a:de71 
cs2Label=Configuration Resource c6a4Label=Agent 
IPv6 Address ahost=SKEELES10 agt=888.99.100.1 agentZoneURI=/All Zones/ArcSight 
System/Private Address Space 
Zones/RFC1918: 888.99.0.0-888.200.255.255 av=6.0.3.6664.0 atz=Australia/Sydney 
aid=3DxKlG0UBABCAA0cXXAZIwA\=\= at=testalertng dvchost=SKEELES10 dvc=888.99.100.1 
deviceZoneURI=/All Zones/ArcSight System/Private Address Space Zones/RFC1918: 
888.99.0.0-888.200.255.255 dtz=Australia/Sydney _cefVer=0.1
```


2. Nearly every CEF extension field has both type and length requirements which we may want to address in due course.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/cef/CEFParser.java
```diff
@@ -0,0 +1,274 @@
...
+			obj = mutate(obj, "app", "protocol");
+
+			obj.put("original_string", cefString);
+
+			// apply timestamp from message if present, using rt, syslog
+			// timestamp,
+			// default to current system time
+
+			if (obj.containsKey("rt")) {
+				String rt = (String) obj.get("rt");
```
kylerichardson: The CEF spec provides for rt (and other datetime fields) to be in epoch milliseconds. Can we add support for that as well?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/utils/DateUtils.java
```diff
@@ -0,0 +1,78 @@
...
+ * Various utilities for parsing and extracting dates
+ * 
+ */
+public class DateUtils {
+
+	public static List<SimpleDateFormat> DATE_FORMATS_CEF = new ArrayList<SimpleDateFormat>() {
+		{
+			// as per CEF Spec
+			add(new SimpleDateFormat("MMM dd HH:mm:ss.SSS zzz"));
+			add(new SimpleDateFormat("MMM dd HH:mm:ss.SSS"));
```
kylerichardson: What are the defaults in the case that no year and/or no timezone are provided? For example, in SysLogUtils we assume any datetime more than 4 days in the future is actually in the past. Would a parser config setting make sense to hold these default assumptions?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/cef/CEFParser.java
simonellistonball: True. I was wondering if someone would care enough to notice that. I'll add.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/utils/DateUtils.java
simonellistonball: So, the default at present in the case of no year is the default for SDF, which is 1970. Probably not what we're after. I'll fix that to present year. In the case of no timezone, I would trend to prefer we assumed UTC (I would actually prefer it if all dates anywhere in Metron were assumed to be UTC targeted, but that's a bigger debate). At present it would assume TZ for the locale, which is not passed, ergo defaults to UTC. This feels like an easy option, though there is an argument for pulling TZ fields out of CEF. Of course the spec doesn't specify anything on TZ, or indeed any fields that would indicate the local TZ of the log source, which means we can't really do anything but assume UTC. 

Please let me know if you're aware of anything that might make this clearer, or if you have anything from HPE on this.

mattf-horton: Systems that don't specify time zone are often time zone naive, meaning they use local time settings.  That may be a better guess than UTC, altho it only works correctly when the source and the collector app are in the same zone.  I've seen some apps provide a per-source time zone setting to improve this.

simonellistonball:
Kicking travis.

cestella:
+1 by inspection, this is great stuff.  Are there outstanding asks for this PR that would prevent it from being committed?

kylerichardson:
Not from me. +1, great contribution.

Ravi0204:
Hello Guys,

I am new to Metron. I need to use CEF parser to parse the logs coming from a source in CEF format. Could anyone guide me how can I use CEF java parser?

Regards.

cestella:
Hey @Ravi0204 you might have better luck sending an email to the Metron user@ list with that question.  This PR is closed and the range of people paying attention to it is constrained.

PR #452, opened 2017-02-12
METRON-715: Removed MySQL from Enrichment Diagram

simonellistonball:


james-sirota:
+1 by inspection. thanks for catching this, simon

JonZeolla:
Please rename the PR to "Metron-715: Enrichment Diagram still includes MySQL, which is no longer used".  Also +1 (non-binding) by inspection

ottobackwards:
Please rename the PR to start with "METRON-715".  Then we can land this

cestella:
+1 by inspection, thanks for the contribution, Simon :)

PR #458, opened 2017-02-20
METRON-725: Javadoc is broken by the use of apiNote

justinleet:
As noted in the ticket, there are 3 options for fixing this

1. Just delete the offending inherited javadoc
2. Explicitly inherit the Javadoc
3. Enable the `@apiNote` tag

This PR just does the first one, but it's easy to change if anybody has a strong preference. I lean pretty heavily towards 1 or 2 (because 3 is apparently more intended towards the Java language, rather than outside usage)

To test, just run mvn javadoc:javadoc.

This also suggests that we should be running this as part of our normal build process (and I'll create a ticket if people agree).

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/utils/file/ReaderSpliterator.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/utils/file/ReaderSpliterator.java
@@ -65,18 +65,7 @@ public class ReaderSpliterator implements Spliterator<String> {
       throw new IllegalStateException(e);
     }
   }
-  /**
-   * If a remaining element exists, performs the given action on it,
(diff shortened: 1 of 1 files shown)


cestella:
Yeah, I'm ok with this fix.  I shouldn't have just green-lit the inherited javadoc suggested by the IDE.

Also, I think we probably *should* do the following during builds:
* generate javadoc
* generate the site-book

My only concern is that we already have hefty build times, but maybe that's not a good enough reason to avoid these kinds of errors.

Anyway, +1 by inspection.

justinleet:
I agree on the "should", and share the same concerns. It would be nice to get everything generated and run regularly, but improving our build is a prereq for that.

There's probably a discussion to be had on that, but I suspect it's not worth having until after the build times are a little more under control.

Thanks for reviewing this.

PR #466, opened 2017-02-26
METRON-742: Generated code for profile window selector DSL did not get committed as part of METRON-690

cestella:
Lest we be concerned this implies that the testing done on 690 is voided, this was due to the removal of the `NOW` lexical token, which was unused in the grammar, along with a really dumb refactoring mistake changing "parse" to "processing" that only hit generated code.  The unit tests, etc. ran the code in this PR since it was generated as part of the build.

This brings me to an interesting question that we might want to discuss.  Early on, we decided to submit generated code to ease potential issues around pulling code into IDEs (generated code means we don't have to either remind people to rerun `mvn generate-sources` or to instruct their IDEs to regenerate sources automatically).  Maybe we should revisit that point, but I don't know, it's just a thought.


ottobackwards:
So - the test is - build without changing anything after clone, and there should not be shown in git status right?

cestella:
Yep

ottobackwards:
Does building or running the tests have anything to do with it?

cestella:
It shouldn't because the tests are already run with the code submitted in this PR.  This is just about getting git in sync with what's generated.

PR #472, opened 2017-03-05
METRON-700: Add hadoop container to metron-docker

kylerichardson:
### For all changes:

### For code changes:
- [x] Have you ensured that the full suite of tests and checks have been executed in the root incubating-metron folder.

- [x] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via site-book/target/site/index.html.

EDIT: Removed N/A items from template.

### DIFF
--- a/metron-docker/.gitignore
+++ b/metron-docker/.gitignore
@@ -1,8 +1,6 @@
 /compose/.env
 /compose/elasticsearch/es_templates
-/compose/kafkazk/common
-/compose/kafkazk/parser
-/compose/kafkazk/enrichment
+/compose/kafkazk/packages/
 /compose/hbase/data-management
 /compose/storm/elasticsearch
 /compose/storm/enrichment
--- a/metron-docker/README.md
+++ b/metron-docker/README.md
@@ -13,6 +13,7 @@ Metron Docker includes these images that have been customized for Metron:
   - Storm (with all topologies deployed)
   - Elasticsearch
   - Kibana
+  - HDFS
 
 Setup
 -----
@@ -80,6 +81,7 @@ Then, assuming a host ip of `192.168.99.100`, the UIs and APIs are available at:
 * HBase - http://192.168.99.100:16010/
 * Elasticsearch - http://192.168.99.100:9200/_plugin/head/
 * Kibana - http://192.168.99.100:5601/
+* HDFS (Namenode) - http://192.168.99.100:50070/
 
 The Storm logs can be useful when troubleshooting topologies.  They can be found on the Storm container in `/usr/share/apache-storm/logs`.
 
--- a/metron-docker/compose/docker-compose.yml
+++ b/metron-docker/compose/docker-compose.yml
@@ -37,6 +37,12 @@ services:
       - "/opt/hbase-1.1.6/conf"
     depends_on:
       - kafkazk
+  hadoop:
+    build:
+      context: ./hadoop
+    ports:
+      - "50070:50070"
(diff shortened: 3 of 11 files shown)


kylerichardson:
### Description
Added a Hadoop/HDFS container to metron-docker. As part of the change, the geo enrichment database is now staged in HDFS to closer mimic a real cluster deployment.

### Testing
Successfully ran through testing with bro sensor data based on the test plan previously documented in the [README](https://github.com/apache/incubator-metron/blob/master/metron-docker/README.md#run-sensor-data-end-to-end).

merrimanr:
Looks good.  The "Run sensor data end to end" didn't work for me and I had to make a couple changes to get data flowing through the indexing topology.  When I started the indexing topology, I got insufficient indexing config errors and realized the index zookeeper configs are not deployed to the kafkazk image.  Once I added those and uploaded to zookeeper, everything worked and I could see data archived in the HDFS container.  Did you run into that?

I believe the missing index configs is a separate issue and not related to the PR.  I will package the changes I had to make up into a follow-on PR.  Nice work +1.

merrimanr:
While reviewing this I had an idea for another PR that would make this easier to use.  The slowest part of spinning this up is downloading the kafka and hbase distributions.  What would you think of splitting these Dockerfiles into 2 separate images:  one with just the service installed (kafka or hbase) and another that extends the service and installs the Metron bits on top?  This would allow us to quickly deploy Metron changes and rebuild.  To get that to work we would need to a way to get the base image installed locally.  Initially this could just be a build script that needs to be run once before spinning the application up and long term it could live in Docker Hub.  

What do you think of this idea?  Would it be helpful?  I'm a little hesitant to add a manual step that needs to be done before the application will work but I don't want to add that hook into the build process, even if we do add a check for a running Docker machine.  Maybe publishing to Docker Hub now is the answer.  Any feedback or ideas?

kylerichardson:
@merrimanr That's interesting. I did get warnings in the logs about the indexing configs but it seemed like the defaults applied and I was still able to see data indexed into HDFS. Did you have to make any other modifications? Do you want me to add in the indexing configs to zookeeper as part of this PR? It wouldn't be difficult.

I like the idea of creating base kafka/zookeeper and hbase container images that could be hosted on Docker Hub. This aligns with how we are handling the other containers and allows us to separate the Metron specific aspects into dedicated Dockerfiles/images. I would be in favor of going the Docker Hub route sooner than later to avoid the additional manual steps.

merrimanr:
If you don't mind adding the indexing configs in this PR that would be great.  I had to make 4 changes to get it to work:
- updated the pom.xml to copy the indexing archive into kafkazk
- updated the kafkazk Dockerfile to add the indexing archive and extract it
- updated the init-zk.sh script to create the /metron/topology/indexing path in zookeeper
- added /compose/kafkazk/indexing to .gitignore

Thanks for the feedback, I'll work on getting base kafka and hbase containers uploaded to Docker Hub.

merrimanr:
Bump.  I think this is ready to go @kylerichardson.

kylerichardson:
@merrimanr Just pushed the changes for adding the indexing configs. Should be ready to go now. Can you give it another glance?

merrimanr:
Looks good.  I like the packages approach instead of listing out individual modules.  +1

kylerichardson:
Thanks! Next time the storm Dockerfile needs a tweak, I'll rework it to use the packages approach too.

PR #476, opened 2017-03-15
METRON-764 DST bug in metron-profiler-client Unit Tests

mattf-horton:
This is a Unit Test bug.  Please see https://issues.apache.org/jira/browse/METRON-764 for detailed description and how to reproduce the problem.

I only changed a Unit Test file and a docs file, therefore I only tested Unit Tests, which pass as shown by Travis.  I also re-ran the reproduction manually and showed it no longer happens with the fix.

=================

### For all changes:

### For code changes:
- [NA] Have you ensured that the full suite of tests and checks have been executed in the root incubating-metron folder via:
  ```
  ```

- [NA] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?


cestella:
+1, great catch @mattf-horton !

PR #478, opened 2017-03-16
METRON-767: Clean up license

cestella:
There's some verbiage at the end that shouldn't exist and we're missing inclusion of the actual licenses. I corrected both.  This was pointed out during the 0.3.1 release review and is corrected here.


justinleet:
Actually, looking into this, it's my fault. I split the MIT license stuff with the geo share alike.  It should be moved up.

@cestella You want to quick fix that and put up a new PR?

cestella:
Nah, I slightly broadened the scope of this one, renamed the JIRA and will adjust the title accordingly

justinleet:
I'm +1 on this by inspection (pending Travis).  Thanks for taking care of the mentor feedback on this so quickly.

PR #481, opened 2017-03-16
METRON-322 Global Batching and Flushing

mattf-horton:
This patch starts by adding "batchTimeout" essentially every place "batchSize" is used.  Since the configuration-related classes are many-layered, this addition is trivial in many cases.

The flush-on-timeout logic is fairly straightforward. It was implemented by a refactoring of `BulkWriterComponent` to extract an explicit "flush" from the "write" logic, and then call flush from Tick Tuple processing in `BulkMessageWriterBolt`.

The tricky part was figuring out the appropriate setting for `topology.tick.tuple.freq.secs` if the administrator configures non-default batchTimeouts. It is necessary to enumerate the batchTimeout settings for all configured sensorNames, which is implemented in `IndexingConfigurations::getAllConfiguredTimeouts()`. Then multiple other factors must be taken into account to determine the allowed and recommended settings, which is implemented in `BatchTimeoutHelper`. If there are better ways to accomplish these things, please share your ideas.

For test and review, it may be useful to know that the default value of `batchTimeout`, and hence for `topology.tick.tuple.freq.secs`, is 14 seconds (which is the default value of `topology.message.timeout.secs` (30 sec) / 2 - 1).

To unit test this feature, we added `Clock` logic and implemented a writable `FakeClock`.  This may be generally useful for testing other time-based behavior as we add such functionality.

All configuration issues were done in such a way as to default to "good" behavior, with no changes needed in current configs of previously installed systems.  In fact, recommended configuration is to leave batchTimeout unconfigured, or set it to zero, thereby accepting the system calculated default.

After this patch is reviewed and accepted, similar work needs to be done for the `ParserWriter`, and possibly other sub-components.  That will be in a separate PR.

It is possible to split this patch into sections related to each of sub-tasks METRON-516, METRON-577, METRON-329 and METRON-330.  Let me know if you would like me to do so.  In the past, people seem to prefer to see a whole feature development in one PR, even though it gets a little long, so that's the way I'm submitting this one, for now.

Test Plan is appended below.

## Pull Request Checklist

### For all changes:

### For code changes:
- [NA] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
    Yes, see the two modified .md files, and the above information.
  ```
  ```

### Automated Unit Testing
There are so many permutations it only seemed reasonable to automate them in unit test, and so I did.  As part of code review, please provide your opinion on whether the provided unit tests are adequate, or what additional test cases should be added.

I enlarged the elasticsearch values for batchSize and batchTimeout in metron-integration-test/src/main/config/indexing/test.json, and mocked procedures in the unit tests that test the following cases:

1. Test that flush on batchSize continues to work correctly, without time-based triggers going off.
1. Test that flush occurs on less-than-full queue, when a message comes in and the queue is older than batchTimeout.  (Age of the queue is determined as currentTime() - timeOfEnqueingFirstMsg().)
1. Test that flush occurs on less-than-full queue, when a TickTuple comes in and the queue is older than batchTimeout.
1. Test that default batchTimeout is correctly calculated as (TOPOLOGY_MESSAGE_TIMEOUT_SECS / 2 / batchTimeoutDivisor - 1).
1. Test that when default batchTimeout is smaller than configured batchTimeout, that the default is used.
1. Test that when default batchTimeout is greater than configured batchTimeout, that the configured batchTimeout is used.

All of the above test cases are tested, with mocked writers and a FakeClock, in the BulkMessageWriterBoltTest unit test file, with the latest commits.

### Manual End-to-End Testing
Manual end-to-end testing, if you are so moved, consists of six scenarios for a given sensor queue:
1. Under **heavy continuous load** the batchSize still controls flushing behavior, because the queue size always exceeds batchSize before queue age exceeds batchTimeout.
2. Under **light continuous load**, where each queue continues to receive at least one message per second, and batchSize is large enough it is never exceeded, then the batchTimeout for each queue should control flushing behavior within +/- 1 sec, because each new message triggers a check of the queue age and potential timeout flush.
  - NOTE: If the configured batchTimeout is set to a large number, bigger than `1/2 topology.message.timeout.secs - 1` (which equals **14 sec** by default), then it will be replaced by an effective value equal to `1/2 topology.message.timeout.secs - 1`.  Flushing will occur within +/- 1 sec of each _effective_ batchTimeout interval, rather than the _configured_ batchTimeout interval.
3. Under **light intermittent load**, where less than batchSize messages queue up, and gaps between messages may exceed the timeout interval, then age checks and potential flush events may be triggered by _either_ incoming messages or TickTuple events, depending on the phase relationship between intermittent bursts of messages, and the TickTuple system tick.  The TickTuple interval is guaranteed to be < `1/2 topology.message.timeout.secs`, hence the default TickTuple interval is 14 seconds.  But if the smallest batchTimeout configured for any sensor queue on the Bolt is < the default TickTuple interval, then that smallest value becomes the actual TickTuple interval.  This produces three sub-cases, all of which guarantee a flush event before any message gets recycled due to aging past `topology.message.timeout.secs`:
  - If the queue's configured batchTimeout is the smallest (or only) such on this Bolt, and that number is smaller than the default TickTuple interval, then it _becomes_ the actual TickTuple interval.  The queue is guaranteed to flush between 1x and 2x this interval.
  - If the queue's configured batchTimeout is not the smallest such, but still is < the default TickTuple interval, then the queue is guaranteed to flush between its own `configured batchTimeout` and its `configured batchTimeout + actual TickTuple interval` (which is less than 2x its own `configured batchTimeout`).
  - If the queue's configured batchTimeout is > the default TickTuple interval (14 sec default), then its effective batchTimeout is set to the default TickTuple interval.  The queue is guaranteed to flush between this `effective batchTimeout` and its `effective batchTimeout + actual TickTuple interval`.

The upshot is that:
* "Configured batchTimeout" should be thought of as "minimum age before you'll allow a time-based flush" (capped by default TickTuple interval, aka 1/2 `topology.message.timeout.secs`)
* "Actual TickTuple interval" is the "maximum time between age checks".  It will be <= all the configured batchTimeouts for the various sensors on the Bolt.
* When a flush actually happens may be up to "effective batchTimeout" + "actual TickTuple interval", depending on exactly when intermittent message events and periodic Tick events happen.


cestella:
Should we rely on a tick tuple?  The reason I ask, is that we also use the `BulkWriterComponent` directly in the `ParserBolt`.

It seems to me that maybe a cleaner approach here would be to have a `Timer` in the `BulkWriterComponent`, this would have the benefit of being self-contained and working for both the enrichment and the parsers.  Thoughts?

mattf-horton:
@cestella , thanks for looking at this. The primary motivation for adding batchTimeout was to prevent tuple recycling due to "topology.message.timeout.secs".  Thus, correctly configuring the timeout period requires interacting with the Bolt.  You're correct that this means each Bolt that uses `BulkWriterComponent` must be modified to include the tick tuple processing, as noted in my opening comments:
```
After this patch is reviewed and accepted, similar work needs to be done for the ParserWriter, and possibly other sub-components. That will be in a separate PR.
```
I implemented the changes in `BulkWriterComponent` such that it would default to conservative behavior if the containing Bolt didn't configure it.

I considered using a timer thread instead of tick tuples, but:
1. This is precisely one of the use cases contemplated by the Storm team when they created Tick Tuples, as discussed in the [article here](https://hortonworks.com/blog/apache-storm-design-pattern-micro-batching/) cited in the jira for METRON-322.
1. It isn't sufficient to just create a timer thread.  One must also monitor that thread, be able to restart it if it dies, make sure it doesn't do anything non-thread-safe, etc.  These add significant complexity to the code, and uncertainty in the case of the thread-safeness, since any pattern we create here will surely be imitated by other developments down the road, and Bolt code is not typically thread-safe.
1. On the other hand, using the built-in Tick Tuples avoids both the complexity, since it handles the reliability issues internal to Storm, and uncertainty, since the Tick Tuple is processed in the single flow of control of normal Bolt processing.

So I think it's cleaner to use the feature provided by the Storm environment.  I'm open to arguments to the contrary.

cestella:
Ok, fair points all-the-way-around.  I'm convinced that we're thinking it through.  I'll give this a bit more of a careful look today.  Thanks for the contribution!

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
```diff
@@ -92,22 +177,51 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
-      WriterConfiguration writerConfiguration = configurationTransformation.apply(new IndexingWriterConfiguration(bulkMessageWriter.getName(), getConfigurations()));
+      if (isTick(tuple)) {
+        if (!(bulkMessageWriter instanceof WriterToBulkWriter)) {
+          //WriterToBulkWriter doesn't allow batching, so no need to flush on Tick.
+          LOG.debug("Flushing message queues older than their batchTimeouts");
+          writerComponent.flushTimeouts(bulkMessageWriter, configurationTransformation.apply(
+                  new IndexingWriterConfiguration(bulkMessageWriter.getName(), getConfigurations()))
+                  , messageGetStrategy);
+        }
+        collector.ack(tuple);
```
cestella: Should this be in a finally?

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/IndexingConfigurations.java
```diff
@@ -73,7 +74,30 @@ public boolean isDefault(String sensorName, String writerName) {
...
+    return getBatchTimeout(getSensorIndexingConfig(sensorName, writerName));
+  }
+
+  public List<Integer> getAllConfiguredTimeouts(String writerName) {
+    //The configuration infrastructure was not designed to enumerate sensors, so we synthesize.
+    //The use of SENSOR_MARKER_STRING assumes the config key for a sensor name consists of a
+    //prefix string (dependent on the ConfigurationType) followed by the sensor name.
+    //This then allows us to determine the keyPrefixString, and thence which of all the
+    //configurations.keySet are actually sensor keys.  Then we look under those for timeouts.
+    String markerKeyString = getKey(SENSOR_MARKER_STRING);
```
cestella: Could you perhaps go a bit into why you're using this `SENSOR_MARKER_STRING` again?  I read the comment and I am still not entirely sure.  I think, at a high level, what you're trying to do is, for a given writer (e.g. `hdfs`) you want a list of all of the configured timeouts across all sensors, right?

mattf-horton: Yup, this is unnecessarily complex.  I'm going to eliminate SENSOR_MARKER_STRING and replace:
```
    String markerKeyString = getKey(SENSOR_MARKER_STRING);
    int prefixStringLength = markerKeyString.indexOf(SENSOR_MARKER_STRING);
    String keyPrefixString = markerKeyString.substring(0, prefixStringLength);
```
with just:
```
String keyPrefixString = getKey("");
```

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
mattf-horton: Good idea, will change to separate "try" wrappers around the tick tuple vs message tuple processing.

ottobackwards:
What is the status of this PR?

mattf-horton:
This is simply waiting for me to have time to return to it.  Would like to leave it open, if agreeable to community.

dlyle65535:
@mattf-horton - Can I offer some help getting back to this? Is a pretty important fix.

mattf-horton:
Hi @dlyle65535 , I would welcome help.  I think this weekend I can update to current master, and integrate the (already done) fixes for @cestella 's suggestions, including stateful testing with mock FakeClock to observe positive and negative flush test cases.  At that point I think it is self-consistent and reviewable.  But if extensive additional testing is requested, I'd need help, and of course the Writers other than BulkMessageWriterBolt remain to be implemented in other Jiras (METRON-750 and METRON-751).

dlyle65535:
Awesome! I can run it up on a live system and see what's what. Thanks!

mattf-horton:
@cestella said:
```
Since this is a core change, it bears probably developing an acceptance test plan. 
Something like creating a dummy sensor, creating a batch size of 10 and sending 
in 5 messages and ensuring things get flushed.

Also, it might be worth adjusting the IndexingIntegrationTest to set the hdfs writer 
config in metron-integration-test/src/main/config/indexing/test.json to a large batch 
size (say 50) and a timeout of, say, 5 seconds to ensure that batches get flushed as 
part of the integration tests.
```
Unfortunately a lot of existing tests have fairly sensitive dependence on hdfs parameters in indexing/test.json, so tweaking it arbitrarily is not feasible.  However, I was able to enlarge the elasticsearch values, and I have mocked procedures in the unit test that test the following cases:

1. Test that flush on batchSize continues to work correctly, without time-based triggers going off.
1. Test that flush occurs on less-than-full queue, when a message comes in and the queue is older than batchTimeout.  (Age of the queue is determined as currentTime() - timeOfEnqueingFirstMsg().)
1. Test that flush occurs on less-than-full queue, when a TickTuple comes in and the queue is older than batchTimeout.
1. Test that default batchTimeout is correctly calculated as (TOPOLOGY_MESSAGE_TIMEOUT_SECS / 2 / batchTimeoutDivisor - 1).
1. Test that when default batchTimeout is smaller than configured batchTimeout, that the default is used.
1. Test that when default batchTimeout is greater than configured batchTimeout, that the configured batchTimeout is used.

All of the above test cases are tested, with mocked writers and a FakeClock, in the BulkMessageWriterBoltTest unit test file, with the latest commit above.

mattf-horton:
@dlyle65535 , integrated and fixed unit tests as intended, and updated to current master.  Thanks.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/writer/SingleBatchConfigurationFacade.java
```diff
@@ -18,6 +18,8 @@
 
 package org.apache.metron.common.configuration.writer;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
```
ottobackwards: It would be nice if there was some javadoc on this class as to it's purpose and usage

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/bolt/BulkMessageWriterBoltTest.java
```diff
@@ -164,4 +174,100 @@ public void test() throws Exception {
     verify(outputCollector, times(1)).emit(eq(Constants.ERROR_STREAM), any(Values.class));
     verify(outputCollector, times(1)).reportError(any(Throwable.class));
   }
+
+  @Test
```
ottobackwards: Maintainers will thank you for throwing some comments here

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
```diff
@@ -92,24 +177,58 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
+      writerComponent.setDefaultBatchTimeout(defaultBatchTimeout);
+      bulkMessageWriter.init(stormConf, context, writerconf);
     } catch (Exception e) {
       throw new RuntimeException(e);
     }
   }
 
+  /**
+   * Used only for unit testing.
+   */
```
ottobackwards: Maybe we can name this testPrepare? So it is explicit to caller, who may not 'peek' inside the source

ottobackwards:
This is a lot of work, and a great job @mattf-horton.  A couple of nit comments I would like to get response to before a +1

ottobackwards:
I am seeing errors like
```bash

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.apache.metron.solr.integration.SolrIndexingIntegrationTest
Waiting for zookeeper...
Found index config in zookeeper...
0 vs 10 vs 0
0 vs 10 vs 0
0 vs 10 vs 0
0 vs 10 vs 0
0 vs 10 vs 0
0 vs 10 vs 0
Processed target/indexingIntegrationTest/hdfs/test/enrichment-hdfsIndexingBolt-1-0-1499888936956.json
10 vs 10 vs 10
Processed target/indexingIntegrationTest/hdfs/test/enrichment-hdfsIndexingBolt-1-0-1499888936956.json
2017-07-12 15:48:57 ERROR ClientCnxn:524 - Error while calling watcher
java.util.concurrent.RejectedExecutionException: Task org.apache.solr.common.util.ExecutorUtil$MDCAwareThreadPoolExecutor$1@9651f04 rejected from org.apache.solr.common.util.ExecutorUtil$MDCAwareThreadPoolExecutor@d5a9af0[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 16]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2047)
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:823)
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1369)
	at org.apache.solr.common.util.ExecutorUtil$MDCAwareThreadPoolExecutor.execute(ExecutorUtil.java:135)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:112)
	at org.apache.solr.common.cloud.SolrZkClient$3.process(SolrZkClient.java:261)
	at org.apache.zookeeper.ClientCnxn$EventThread.processEvent(ClientCnxn.java:522)
	at org.apache.zookeeper.ClientCnxn$EventThread.run(ClientCnxn.java:498)
2017-07-12 15:49:01 ERROR ReadClusterState:181 - Failed to Sync Supervisor
java.lang.RuntimeException: java.lang.InterruptedException
	at org.apache.storm.utils.Utils.wrapInRuntime(Utils.java:1507)
	at org.apache.storm.zookeeper.Zookeeper.getChildren(Zookeeper.java:260)
	at org.apache.storm.cluster.ZKStateStorage.get_children(ZKStateStorage.java:174)
	at org.apache.storm.cluster.StormClusterStateImpl.assignments(StormClusterStateImpl.java:153)
	at org.apache.storm.daemon.supervisor.ReadClusterState.run(ReadClusterState.java:126)
	at org.apache.storm.event.EventManagerImp$1.run(EventManagerImp.java:54)
Caused by: java.lang.InterruptedException
	at java.lang.Object.wait(Native Method)
	at java.lang.Object.wait(Object.java:502)
	at org.apache.storm.shade.org.apache.zookeeper.ClientCnxn.submitRequest(ClientCnxn.java:1342)
	at org.apache.storm.shade.org.apache.zookeeper.ZooKeeper.getChildren(ZooKeeper.java:1588)
	at org.apache.storm.shade.org.apache.zookeeper.ZooKeeper.getChildren(ZooKeeper.java:1625)
	at org.apache.storm.shade.org.apache.curator.framework.imps.GetChildrenBuilderImpl$3.call(GetChildrenBuilderImpl.java:210)
	at org.apache.storm.shade.org.apache.curator.framework.imps.GetChildrenBuilderImpl$3.call(GetChildrenBuilderImpl.java:203)
	at org.apache.storm.shade.org.apache.curator.RetryLoop.callWithRetry(RetryLoop.java:108)
	at org.apache.storm.shade.org.apache.curator.framework.imps.GetChildrenBuilderImpl.pathInForeground(GetChildrenBuilderImpl.java:200)
	at org.apache.storm.shade.org.apache.curator.framework.imps.GetChildrenBuilderImpl.forPath(GetChildrenBuilderImpl.java:191)
	at org.apache.storm.shade.org.apache.curator.framework.imps.GetChildrenBuilderImpl.forPath(GetChildrenBuilderImpl.java:38)
	at org.apache.storm.zookeeper.Zookeeper.getChildren(Zookeeper.java:255)
	... 4 more
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 45.675 sec - in org.apache.metron.solr.integration.SolrIndexingIntegrationTest

Results :
```
Have you seen these?

mattf-horton:
Comment on testing:  There are so many permutations it only seemed reasonable to automate them in unit test, and so I did.  As part of code review, please provide your opinion on whether the provided unit tests are adequate, or what additional test cases should be added.

Manual end-to-end testing, if you are so moved, consists of six scenarios for a given sensor queue:
1. Under **heavy continuous load** the batchSize still controls flushing behavior, because the queue size always exceeds batchSize before queue age exceeds batchTimeout.
2. Under **light continuous load**, where each queue continues to receive at least one message per second, and batchSize is large enough it is never exceeded, then the batchTimeout for each queue should control flushing behavior within +/- 1 sec, because each new message triggers a check of the queue age and potential timeout flush.
  - NOTE: If the configured batchTimeout is set to a large number, bigger than `1/2 topology.message.timeout.secs - 1` (which equals **14 sec** by default), then it will be replaced by an effective value equal to `1/2 topology.message.timeout.secs - 1`.  Flushing will occur within +/- 1 sec of each _effective_ batchTimeout interval, rather than the _configured_ batchTimeout interval.
3. Under **light intermittent load**, where less than batchSize messages queue up, and gaps between messages may exceed the timeout interval, then age checks and potential flush events may be triggered by _either_ incoming messages or TickTuple events, depending on the phase relationship between intermittent bursts of messages, and the TickTuple system tick.  The TickTuple interval is guaranteed to be < `1/2 topology.message.timeout.secs`, hence the default TickTuple interval is 14 seconds.  But if the smallest batchTimeout configured for any sensor queue on the Bolt is < the default TickTuple interval, then that smallest value becomes the actual TickTuple interval.  This produces three sub-cases, all of which guarantee a flush event before any message gets recycled due to aging past `topology.message.timeout.secs`:
  - If the queue's configured batchTimeout is the smallest (or only) such on this Bolt, and that number is smaller than the default TickTuple interval, then it _becomes_ the actual TickTuple interval.  The queue is guaranteed to flush between 1x and 2x this interval.
  - If the queue's configured batchTimeout is not the smallest such, but still is < the default TickTuple interval, then the queue is guaranteed to flush between its own `configured batchTimeout` and its `configured batchTimeout + actual TickTuple interval` (which is less than 2x its own `configured batchTimeout`).
  - If the queue's configured batchTimeout is > the default TickTuple interval (14 sec default), then its effective batchTimeout is set to the default TickTuple interval.  The queue is guaranteed to flush between this `effective batchTimeout` and its `effective batchTimeout + actual TickTuple interval`.

The upshot is that:
* "Configured batchTimeout" should be thought of as "minimum age before you'll allow a time-based flush" (capped by default TickTuple interval, aka 1/2 `topology.message.timeout.secs`)
* "Actual TickTuple interval" is the "maximum time between age checks".  It will be <= all the configured batchTimeouts for the various sensors on the Bolt.
* When a flush actually happens may be up to "effective batchTimeout" + "actual TickTuple interval", depending on exactly when intermittent message events and periodic Tick events happen.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/writer/SingleBatchConfigurationFacade.java
mattf-horton: Well, there's some weirdness in this stuff, and I didn't try to get rid of it lest I break something I didn't understand.  In this case (and other places), it appears that non-batched writers are essentially faked by using batched writers with a batchSize of "1".  Doesn't seem real efficient to me, but obviously decreases the overall quantity of code and complexity of testing.  Anyway, since I only did a trivial change to this file, I didn't feel obligated to try to explain it.

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/bolt/BulkMessageWriterBoltTest.java
mattf-horton: @ottobackwards , how about "Mocking sucks"? :-)
Or at least "Mocking complex objects is excruciating."
But seriously, yes I'll add some comments.  Not planning on a line-by-line walk-thru, tho.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
mattf-horton: Hm.  With respect, this one I don't really agree with.  It's just a polymorphism of the regular prepare() method.  Since its signature differs from the "regular" prepare() method, system code will never call it.  It is documented, at javadoc level, as being only for use in unit testing, which is by definition code-level.  Those who aren't willing to look at code shouldn't be using it, and indeed will have a difficult time finding it at all.  I'd rather let the polymorphism be evident.

I will, however, add further comment that this is "used only for unit testing, with injection of a FakeClock for mocking passage of time in a controlled way, to test queue timeout behavior in the WriterComponent."  Would that be okay?

mattf-horton:
@ottobackwards , thanks for looking at this.  I did not see those error behaviors while running integration tests.  Please tell me the environment and the exact invocation you used.  Thanks.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
ottobackwards: let the intellisense fall where it may then

ottobackwards:
+1

mattf-horton:
Before committing, I'm going to retest to make sure I'm not missing the problems @ottobackwards saw.  @dlyle65535 , also interested in your feedback if you have time.

dlyle65535:
@mattf-horton - sorry about the hurry up and wait. I'll run this up tomorrow. Thanks!

dlyle65535:
My test environment is being funky, I'm +1 by inspection- thanks for the work, please don't let me hold you up any more.

dlyle65535:
@mattf-horton Oh noes! Looks like you got merge conflicted. Any chance this is on your radar still? Thanks!

mattf-horton:
@dlyle65535 yes, I'll clean this up.
I'll run integration tests one more time to make sure I don't see @ottobackwards 's problem, then commit on the strength of his and your +1.  Let me know if you disagree.

mattf-horton:
Updated and re-ran unit and integration tests.  @ottobackwards , metron-solr integration tests ran fine.  I did see a problem with metron-rest integration test org.apache.metron.rest.controller.UpdateControllerIntegrationTest , but it happens in current master branch without this PR too.  @dlyle65535 , proceeding with commit.

PR #486, opened 2017-03-22
METRON-793: Migrate to storm-kafka-client kafka spout from storm-kafka

cestella:
## Contributor Comments
In order to eventually support kerberos, the suggested path is to migrate to the new kafka spout (org.apache.storm:storm-kafka-client) which uses the new consumer API.

Testing plan to follow shortly.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  bin/generate-md.sh
  mvn site:site
  ```

#### Note:


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/resolver/ClasspathFunctionResolver.java
```diff
@@ -235,8 +235,16 @@ public void initialize(Context context) {
     }
 
     FilterBuilder filterBuilder = new FilterBuilder();
-    excludes.forEach(excl -> filterBuilder.exclude(excl));
-    includes.forEach(incl -> filterBuilder.include(incl));
+    excludes.forEach(excl -> {
```
cestella: This is an intermittent test bug that I noticed and fixed in order to get the tests to run consistently.

metron-platform/metron-integration-test/src/main/java/org/apache/metron/integration/components/FluxTopologyComponent.java
```diff
@@ -133,7 +138,25 @@ public void start() throws UnableToStartException {
   @Override
   public void stop() {
     if (stormCluster != null) {
-      stormCluster.shutdown();
+      try {
```
cestella: Note here, we moved to storm 1.0.3 in order to have a single API that would work between storm 1.0.1 in the HDP profile as well as the default profile.  Storm 1.0.3 writes out worker artifact logs and they need to be cleaned up.

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/spout/pcap/deserializer/Deserializers.java
```diff
@@ -16,31 +16,35 @@
...
+import org.apache.metron.common.utils.timestamp.TimestampConverters;
 import org.apache.metron.common.utils.timestamp.TimestampConverter;
-import org.apache.storm.kafka.KeyValueSchemeAsMultiScheme;
 
-public enum TimestampScheme {
-   FROM_KEY( converter -> new KeyValueSchemeAsMultiScheme(new FromKeyScheme().withTimestampConverter(converter)))
-  ,FROM_PACKET(converter -> new FromPacketScheme().withTimestampConverter(converter));
+import java.util.function.Function;
+
+public enum Deserializers {
```
cestella: It's worth noting here that the new spout does not use multischemes.  We are doing a fair bit of byte munging on data coming into the pcap topology in those multischemes, but the abstraction was clunky.  This makes things a bit clearer.

pom.xml
```diff
@@ -67,20 +67,44 @@
...
+             (e.g. the Admin functions have different parameters) and behaves
+             differently than 0.10.x in subtle ways (e.g. KAFKA_GET doesn't work as implemented).
+
+             Rather than do this, we chose to depend on the HDP version of storm-kafka because 
+             it is compiled against 0.10.x and therefore would allow us to not fork our support
+             for kafka.  I do not like this bleeding of the HDP profile dependency into the default,
+             but I justify it by noting that this should be able to be removed when we migrate to
+             Storm 1.1.x, which properly supports Kafka 0.10.x.
+          -->
+        <global_storm_kafka_version>1.0.1.2.5.0.0-1245</global_storm_kafka_version>
```
cestella: The comment above this line is worth pointing out.

cestella:
I owe:
* Better Javadocs around some of the pcap infrastructure
* An acceptance test plan

metron-analytics/metron-profiler/src/main/config/profiler.properties
```diff
@@ -39,4 +39,4 @@ profiler.hbase.flush.interval.seconds=30
 
 kafka.zk=node1:2181
 kafka.broker=node1:6667
-kafka.start=WHERE_I_LEFT_OFF
+kafka.start=UNCOMMITTED_EARLIEST
```
cestella: This needs documentation.  We had our own abstractions around the kafka offsets which we no longer need because the new spout has better ones.

cestella:
# Testing Plan
## Preliminaries

* Please perform the following tests on the `full-dev` vagrant environment.
`export METRON_HOME=/usr/metron/0.3.1` 


## Ensure Data Flows from the Indices
Ensure that with a basic full-dev we get data into the elasticsearch
indices and into HDFS.

## (Optional) Free Up Space on the virtual machine

* Stop and disable Metron in Ambari

## Test the PCAP topology

A new kafka spout necessitates testing pcap.
### Install and start pycapa 
```
# set env vars
export PYCAPA_HOME=/opt/pycapa
export PYTHON27_HOME=/opt/rh/python27/root

# Install these packages via yum (RHEL, CentOS)
yum -y install epel-release centos-release-scl 
yum -y install "@Development tools" python27 python27-scldevel python27-python-virtualenv libpcap-devel libselinux-python

# Setup directories
mkdir $PYCAPA_HOME && chmod 755 $PYCAPA_HOME

# Create virtualenv
export LD_LIBRARY_PATH="/opt/rh/python27/root/usr/lib64"
${PYTHON27_HOME}/usr/bin/virtualenv pycapa-venv

# Copy pycapa
# copy incubator-metron/metron-sensors/pycapa from the Metron source tree into $PYCAPA_HOME on the node you would like to install pycapa on.

# Build it
cd ${PYCAPA_HOME}/pycapa
# activate the virtualenv
source ${PYCAPA_HOME}/pycapa-venv/bin/activate
pip install -r requirements.txt
python setup.py install

# Run it
cd ${PYCAPA_HOME}/pycapa-venv/bin
pycapa --producer --topic pcap -i eth1 -k node1:6667
```
### Ensure pycapa can write to HDFS
* Ensure that `/apps/metron/pcap` exists and can be written to by the
  storm user.  If not, then:
```
sudo su - hdfs
hadoop fs -mkdir -p /apps/metron/pcap
hadoop fs -chown metron:hadoop /apps/metron/pcap
hadoop fs -chmod 775 /apps/metron/pcap
``` 
* Start the pcap topology via `$METRON_HOME/bin/start_pcap_topology.sh`
* Start the pycapa packet capture producer on eth1 via `/usr/bin/pycapa --producer --topic pcap -i eth1 -k node1:6667`
* Watch the topology in the Storm UI and kill the packet capture utility from before, when the number of packets ingested is over 3k.  Ensure that at at least 3 files exist on HDFS by running `hadoop fs -ls /apps/metron/pcap`
* Choose a file (denoted by $FILE) and dump a few of the contents using the pcap_inspector utility via `$METRON_HOME/bin/pcap_inspector.sh -i $FILE -n 5`
* Choose one of the lines and note the protocol.
  * Note that when you run the commands below, the resulting file will be placed in the execution directory where you kicked off the job from.
* Run a Stellar query filter query by executing a command similar to the following, with the values noted above (match your start_time format to the date format provided - default is to use millis since epoch):
```
$METRON_HOME/bin/pcap_query.sh query -st "20160617" -df "yyyyMMdd" -query "protocol == 6" -rpf 500
```
* Verify the MR job finishes successfully. Upon completion, you should see multiple files named with relatively current datestamps in your current directory, e.g. pcap-data-20160617160549737+0000.pcap
* Copy the files to your local machine and verify you can them it in Wireshark. I chose a middle file and the last file. The middle file should have 500 records (per the records_per_file option), and the last one will likely have a number of records <= 500.

## Test the Profiler

### Setup
* Ensure that Metron is stopped and put in maintenance mode in Ambari

```
#!/usr/bin/python
import random
import sys
import time
def main():
  mu = float(sys.argv[1])
  sigma = float(sys.argv[2])
  freq_s = int(sys.argv[3])
  while True:
    print out
    sys.stdout.flush()
    time.sleep(freq_s)

if __name__ == '__main__':
  main()
```

```
```
### Deploy the custom parser

* Edit the value parser config at `$METRON_HOME/config/zookeeper/parsers/value.json`:
```
{
  "sensorTopic":"value",
  "fieldTransformations" : [
    {
    "transformation" : "STELLAR"
   ,"output" : [ "num_profiles_parser", "mean_parser" ]
   ,"config" : {
      "num_profiles_parser" : "LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us')))",
      "mean_parser" : "STATS_MEAN(STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us'))))"
               }
    }
                           ]
}
```

* Edit the value enrichment config at `$METRON_HOME/config/zookeeper/enrichments/value.json`:
```
{
  "enrichment" : {
   "fieldMap": {
      "stellar" : {
        "config" : {
        "num_profiles_enrichment" : "LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us')))",
        "mean_enrichment" : "STATS_MEAN(STATS_MERGE(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us'))))"
                  }
      }
    }
  }
}
```
* Create the value kafka topic:
  `/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper node1:2181 --create --topic value --partitions 1 --replication-factor 1`
* Push the configs via `$METRON_HOME/bin/zk_load_configs.sh -m PUSH -i $METRON_HOME/config/zookeeper -z node1:2181`
* Start via `$METRON_HOME/bin/start_parser_topology.sh -k node1:6667 -z node1:2181 -s value`


### Start the profiler

```
{
  "profiles": [
    {
      "profile": "stat",
      "foreach": "'global'",
      "onlyif": "true",
      "init" : {
               },
      "update": {
        "s": "STATS_ADD(s, value)"
                },
      "result": "s"
    }
  ]
}
```

### Test Case

`python ~/rand_gen.py 0 1 1 | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic value`
* Wait for at least 32 minutes and execute the following via the Stellar REPL:
```
# Grab the profiles from 1 minute ago to 8 minutes ago
LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('from 1 minute ago to 8 minutes ago')))
# Looks like 7 were returned, great.  Now try something more complex
# Grab the profiles in 5 minute windows every 10 minutes from 2 minutes ago to 32 minutes ago:
#  32 minutes ago til 27 minutes ago should be 5 profiles
#  22 minutes ago til 17 minutes ago should be 5 profiles
#  12 minutes ago til 7 minutes ago should be 5 profiles
# for a total of 15 profiles
LENGTH(PROFILE_GET('stat', 'global', PROFILE_WINDOW('5 minute window every 10 minutes starting from 2 minutes ago until 32 minutes ago excluding holidays:us')))
```
```
```
* Delete any value index that currently exists (if any do) via `curl -XDELETE "http://localhost:9200/value*"`
* Wait for a couple of seconds and run 
  * `curl "http://localhost:9200/value*/_search?pretty=true&q=*:*" 2> /dev/null` 
  * You should see values in the index with non-zero fields:
     * `num_profiles_enrichment` should be 15
     * `num_profiles_parser` should be 15
     * `mean_enrichment` should be a non-zero double
     * `mean_parser` should be a non-zero double
For reference, a sample message for me is:
```
```

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
```diff
@@ -0,0 +1,234 @@
...
+    Map<String, Object> spoutConfig = SpoutConfiguration.separate(kafkaProps);
+
+    SimpleStormKafkaBuilder builder = new SimpleStormKafkaBuilder(kafkaProps, topic, zkQuorum, fieldsConfiguration);
+    SpoutConfiguration.configure(builder, spoutConfig);
+    return new StormKafkaSpout(builder);
+  }
+
+  private static Map<String, Object> modifyKafkaProps(Map<String, Object> props, String zkQuorum) {
+    try {
+      if(!props.containsKey(KafkaSpoutConfig.Consumer.BOOTSTRAP_SERVERS)) {
```
justinleet: Same deal with the putIfAbsent

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * Create a StormKafkaSpout from a given topic, zookeeper quorum and fields.  Also, configure the spout
+   * using a Map that configures both kafka as well as the spout (see the properties in SpoutConfiguration).
+   * @param topic
+   * @param zkQuorum
+   * @param fieldsConfiguration
+   * @param kafkaProps  The aforementioned map.
+   * @return
+   */
+  public static StormKafkaSpout create( String topic
```
justinleet: Shouldn't this and the stuff calling it all be generic? Why bother typing it, if we're just killing the types during the create()?

I didn't dig too much into it, but it seems like at least a few things can be typed.

pom.xml
justinleet: I'm sure you've already thought of this, but assuming we do go with this, please make sure this gets a JIRA associated with it.

cestella:
@justinleet good points overall.  Yeah, I'm not thrilled with the HDP dependency either, as I stated already.

Regarding "If flux gets support for the Builder items, would we just axe those classes entirely?"
I believe that the extension will live on.  It has a couple of benefits beyond the builder pattern in flux:
* The parsers aren't using flux and do need a way to expose the configuration of the spout via properties.  This is part of the extension here.
* A component of extension in the storm-kafka-client Builder is via polymorphism (e.g. the TupleBuilder, etc).  This will just never work well in flux, I think.
* There is an assumption in the storm-kafka-client builder that we're passing in brokers directly, rather than reading them from zookeeper.
* This API is shifting dramatically.  It's totally different in storm 1.0.3 vs 1.0.1.  Creating a layer here insulates us from API changes and localizes their impact.

I made the TODOs because, while what we have here matched the capabilities of what we had before, we could do better in supporting multiple topics.  I wanted to bring out the places that would need to shift to support that.

cestella:
I was planning to make a JIRA around the multiple topic bit.

justinleet:
@cestella I'm good with keeping the extension points especially after the points you made. I think the TODOs are valuable, I just wanted to know the thought behind potentially building it out.

Given the API instability, unfortunately it seems like our dependencies aren't going to provide that insulation layer.  I'd rather have that be provided in a stable manner upstream from us, but that's not something we have any control over.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/topology/ParserTopologyBuilder.java
```diff
@@ -106,19 +105,26 @@ public static TopologyBuilder build(String zookeeperUrl,
...
    */
-  private static KafkaSpout createKafkaSpout(String zookeeperUrl, String sensorType, SpoutConfig.Offset offset, EnumMap<SpoutConfigOptions, Object> kafkaSpoutConfigOptions, SensorParserConfig parserConfig) {
-
+  private static StormKafkaSpout createKafkaSpout(String zkQuorum, String sensorType, Optional<Map<String, Object>> kafkaConfigOptional, SensorParserConfig parserConfig) {
+    Map<String, Object> kafkaSpoutConfigOptions = kafkaConfigOptional.orElse(new HashMap<>());
     String inputTopic = parserConfig.getSensorTopic() != null ? parserConfig.getSensorTopic() : sensorType;
-    SpoutConfig spoutConfig = new SpoutConfig(new ZkHosts(zookeeperUrl), inputTopic, "", inputTopic).from(offset);
-    SpoutConfigOptions.configure(spoutConfig, kafkaSpoutConfigOptions);
-    return new KafkaSpout(spoutConfig);
+    if(!kafkaSpoutConfigOptions.containsKey(SpoutConfiguration.FIRST_POLL_OFFSET_STRATEGY.key)) {
```
cestella: Yep

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/spout/pcap/HDFSWriterCallback.java
```diff
@@ -96,16 +109,18 @@ public HDFSWriterCallback withConfig(HDFSWriterConfig config) {
...
     @Override
     public List<Object> apply(List<Object> tuple, EmitContext context) {
-
-        List<Object> keyValue = (List<Object>) tuple.get(0);
-        LongWritable ts = (LongWritable) keyValue.get(0);
-        BytesWritable rawPacket = (BytesWritable)keyValue.get(1);
+        byte[] key = (byte[]) tuple.get(0);
+        byte[] value = (byte[]) tuple.get(1);
+        if(!config.getDeserializer().deserializeKeyValue(key, value, KeyValue.key.get(), KeyValue.value.get())) {
+            LOG.debug("Dropping malformed packet...");
```
cestella: I could, but, of course, they're raw bytes and the value could be rather big.  How about I compromise and make this a bit more helpful by including:
* Their lengths
* The key itself in hex

justinleet: I'm good with that. I had mostly discarded the worry about size because this is in debugging statements anyway and typically with storm you're setting reasonable timeouts on logging levels anyway.

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
cestella: yep, agreed

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
cestella: yep

cestella:
Follow-on JIRA to move back to apache storm for storm-kafka-client is at https://issues.apache.org/jira/browse/METRON-794

metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/spout/pcap/HDFSWriterCallback.java
cestella: yeah, you're right.

mmiklavc:
I just finished running some parser tests in a Kerberized environment to confirm that this will address both kerberized and non-kerberized configurations. Here is what I had to do to get this running. Note, I'm using Ambari to configure everything for Hadoop, but not the Metron components. There is separate work that will need to be done for Metron in the MPack for Ambari.

1. Spin up full dev as normal (ensure that the profile is correctly set to -P HDP-2.5.0.0 in metron-deployment/roles/metron-builder/tasks/main.yml). As of this writing, the current full-dev ansible install will do a complete build as part of the deployment.
2. service stop monit and kill all the topologies. You can go ahead and kill the other sensors per the commands from @cestella  as well (bro, pycapa, etc.). We'll spin up our own simple test topology and populate the kafka topic manually to test this out.
3. Setup symlinks
```
ln -s /usr/metron/0.3.1/lib/metron-*-uber.jar /usr/hdp/2.5.3.0-37/storm/extlib-daemon/
ln -s /usr/hdp/current/hadoop-client/conf/core-site.xml /etc/storm/conf
ln -s /usr/hdp/current/hbase-client/conf/hbase-site.xml /etc/storm/conf
ln -s /usr/hdp/2.5.3.0-37/storm/contrib/storm-hbase/storm-hbase-1.0.1.2.5.3.0-37.jar /usr/hdp/2.5.3.0-37/storm/extlib
ln -s /usr/hdp/2.5.3.0-37/storm/contrib/storm-hdfs/storm-hdfs-1.0.1.2.5.3.0-37.jar /usr/hdp/2.5.3.0-37/storm/extlib
```
4. Check that the jce security is setup in /usr/jdk64/jdk1.8.0_77/jre/lib/security/. If not, then you'll want to run through the following steps:
4.1 Download the jce policy from: http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html and copy it to your vm.
4.2 unzip -o -j -q /path-to-the-copied-zip/jce_policy-8.zip -d /usr/jdk64/jdk1.8.0_77/jre/lib/security/
5. Configure MIT Kerberos
```
yum -y install krb5-server krb5-libs krb5-workstation
sed -i 's/kerberos.example.com/node1/g' /etc/krb5.conf
cp /etc/krb5.conf /var/lib/ambari-server/resources/scripts
kdb5_util create -s
/etc/rc.d/init.d/krb5kdc start
/etc/rc.d/init.d/kadmin start
chkconfig krb5kdc on
chkconfig kadmin on 
```
6. Setup admin and personal user principals. You'll kinit as the "{your user name}" user. For my tests, I did "addprinc mike". I just made the passwords "password" for the sake of easy testing.
```
kadmin.local -q "addprinc admin/admin"
kadmin.local -q "addprinc {your user name}"
```
7. In Ambari, setup AutoHDFS by adding the following properties to custom-storm-site
```
nimbus.autocredential.plugins.classes=['org.apache.storm.hdfs.common.security.AutoHDFS','org.apache.storm.hbase.security.AutoHBase']
nimbus.credential.renewers.classes=['org.apache.storm.hdfs.common.security.AutoHDFS','org.apache.storm.hbase.security.AutoHBase']
hdfs.keytab.file=/etc/security/keytabs/hdfs.headless.keytab
hdfs.kerberos.principal=hdfs-metron_cluster@EXAMPLE.COM
hbase.keytab.file=/etc/security/keytabs/hbase.headless.keytab
hbase.kerberos.principal=hbase-metron_cluster@EXAMPLE.COM
nimbus.credential.renewers.freq.secs=82800
```
8. Kerberize the cluster via Ambari. The wizard is fairly straightforward, but further documentation can be found [here](http://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.5.3/bk_security/content/_enabling_kerberos_security_in_ambari.html). For this exercise, choose existing MIT KDC (this is what we setup and installed in previous steps.) Realm is EXAMPLE.COM. The admin principal will end up as admin/admin@EXAMPLE.COM when testing the KDC.
9. Let the cluster spin up, but don't worry about starting up Metron via Ambari - we're going to run the parsers manually against the rest of the Hadoop cluster Kerberized.
10. Kinit and provide the password you chose from earlier.
```
kinit <your username>
```
11. Create a kafka topic for the jsonMap parser.
```
/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper node1:2181 --create --topic jsonMap --replication-factor 1 --partitions 1
```
12. Setup the ACLs for the topic AND the consumer group. The consumer group config is new in this API as far as I can tell because we didn't have to do this in the "storm-kafka" version of the api. Note that the docs around kafka 0.10 will suggest using a "--consumer-group" option, but in HDP 2.5.x the option needs to be "--group". Also make sure you use your username in "--allow-principal User:<your username>" instead of "mike". 
```
/usr/hdp/current/kafka-broker/bin/kafka-acls.sh --authorizer kafka.security.auth.SimpleAclAuthorizer --authorizer-properties zookeeper.connect=node1:2181 --add --allow-principal User:storm-metron_cluster --allow-principal User:mike --topic jsonMap
/usr/hdp/current/kafka-broker/bin/kafka-acls.sh --authorizer-properties zookeeper.connect=node1:2181 --group jsonMap_parser --list --topic jsonMap
```
13. Setup an additional config for the Kafka consumer security protocol options. Create a file wherever is convenient and note the path. e.g. Here is my config output from root home.
```
# cat ~/extra_config.json
{
    "security.protocol":"PLAINTEXTSASL"
}
```
14. Configure the jsonMap parser to handle Kerberos. The parserConfig key is the new piece required for Kerberos.
```
# cat /usr/metron/0.3.1/config/zookeeper/parsers/jsonMap.json
{
  "sensorTopic":"jsonMap",
  "parserConfig" : {
      "kafka.producerConfigs" : {
          "security.protocol" : "PLAINTEXTSASL"
      },
      "error.kafka.producerConfigs" : {
          "security.protocol" : "PLAINTEXTSASL"
      }
  }
}
```
15. Push the updated configs to Zookeeper
```
/usr/metron/0.3.1/bin/zk_load_configs.sh -z node1:2181 -m PUSH -i /usr/metron/0.3.1/config/zookeeper
```
16. Start the topology
```
./start_parser_topology.sh -k node1:6667 -z node1:2181 -s jsonMap -esc ~/extra_config.json
```
17. Create a file with some test data that we'll push into the parser Kafka topic
```
# cat ~/sampledata.txt
{ "string" : "bar", "number" : 2, "ignored" : [ "blah" ] }
{ "number" : 7 }
```
18. Push data into Kafa
```
cat ~/sampledata.txt | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --security-protocol PLAINTEXTSASL --topic jsonMap
```
19. Verify our data comes out on the other end by looking at the enrichments topic.
```
/usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh --zookeeper node1:2181 --from-beginning --security-protocol PLAINTEXTSASL --topic enrichments
```

The above process worked for me with this PR. That should be it.

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
```diff
@@ -0,0 +1,234 @@
...
+   * @param zkQuorum
+   * @param fieldsConfiguration
+   * @param kafkaProps  The aforementioned map.
+   * @return
+   */
+  public static <K, V> StormKafkaSpout<K, V> create( String topic
+                                                   , String zkQuorum
+                                                   , List<String> fieldsConfiguration
+                                                   , Map<String, Object> kafkaProps
+                                                   , Class<K> keyClazz
```
justinleet: The clazz object can be dropped, because this is all typed now, correct?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/topology/ParserTopologyBuilder.java
```diff
@@ -106,19 +105,22 @@ public static TopologyBuilder build(String zookeeperUrl,
...
    * @param sensorType              Type of sensor
-   * @param offset                  Kafka topic offset where the topology will start; BEGINNING, END, WHERE_I_LEFT_OFF
-   * @param kafkaSpoutConfigOptions Configuration options for the kafka spout
+   * @param kafkaConfigOptional     Configuration options for the kafka spout
    * @param parserConfig            Configuration for the parser
    * @return
    */
-  private static KafkaSpout createKafkaSpout(String zookeeperUrl, String sensorType, SpoutConfig.Offset offset, EnumMap<SpoutConfigOptions, Object> kafkaSpoutConfigOptions, SensorParserConfig parserConfig) {
-
+  private static StormKafkaSpout createKafkaSpout(String zkQuorum, String sensorType, Optional<Map<String, Object>> kafkaConfigOptional, SensorParserConfig parserConfig) {
```
justinleet: StormKafkaSpout's return type here will actually be StormKafkaSpout<Object, Object>, right?  Can we make that explicit, rather than untyped (and also drop the Object.class from the creates assuming the other typing change)?

cestella: yep, we can.

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
cestella: yes we can

justinleet:
+1, was able to follow Mike's instructions, with a couple caveats.

- Group authorization command was missing
```
/usr/hdp/current/kafka-broker/bin/kafka-acls.sh --authorizer kafka.security.auth.SimpleAclAuthorizer --authorizer-properties zookeeper.connect=node1:2181 --add --allow-principal User:storm-metron_cluster --allow-principal User:justin --group jsonMap_parser
```
- Topic authorization command on the enrichments topic side was missing.
```
 /usr/hdp/current/kafka-broker/bin/kafka-acls.sh --authorizer kafka.security.auth.SimpleAclAuthorizer --authorizer-properties zookeeper.connect=node1:2181 --add --allow-principal User:storm-metron_cluster --allow-principal User:justin --topic enrichments
```

justinleet:
@cestella Just noticed Travis after I commented.  I'm moderately surprised that the most recent PR would break it, do you know what the issue is?

mmiklavc:
@justinleet Thanks for checking this out - modified the instructions with the ACL command corrections. Must have copy-pasted the wrong commands from history, so thanks for that!

cestella:
@justinleet It is unclear and that test has had some intermittent issues, especially under load, I've noticed.  I added a more descriptive message to help diagnose.

cestella:
I found a couple of spots where we might not be cleaning things up and exceptions that are thrown that kill the spout that shouldn't.  I've run the test 10 times locally and will continue to run it in travis for the rest of the day to suss out any other lingering issues.  It almost assuredly is correlated to shutting down things during load, I think.

cestella:
Ok, update on the integration tests.  It appears that the issue is that storm 1.0.3 does not consistently kill the slots when shutting down.  It times out after a minute and dies.  The fix that I am testing is to directly close the slot in that case.  I've run the travis build about 8 times with no failures.  I'm running locally 20 times before I claim it's fixed.

james-sirota:
I was able to validate this on a single-node Suse-based Hadoop cluster once with Kerberos disabled and second time with Kerberos enabled.  That's a good first step.  I will try to run this up in AWS tomorrow on a larger cluster to see if it still works.  But so far so good.

cestella:
Ok, I ran it 20 times independently over the weekend and again through travis whenever I had a moment and the tests appear stable after I kill the slots manually as a last resort in the FluxComponent.

justinleet:
@cestella My +1 stands with the testing issues ironed out.  Thanks for looking into it.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
```diff
@@ -177,7 +177,7 @@ indexing.executors=0
 ##### Kafka #####
 kafka.zk={{ zookeeper_quorum }}
 kafka.broker={{ kafka_brokers }}
-kafka.start=WHERE_I_LEFT_OFF
+kafka.start=UNCOMMITTED_EARLIEST
 ##### Indexing #####
```
dlyle65535: Little late to the party, but I think this is causing an error in Quick Dev. The basis assumption is that the MPack is unchanged for a given quick dev image, so kafka.start remains WHERE_I_LEFT_OFF right now but that's not part of the enum, so it throws. I think the easy thing is to create a new image. What do you think?

PR #436, opened 2017-02-02
METRON-671: Refactor existing Ansible deployment to use Ambari MPack

dlyle65535:
Updated Update - I think this is ready for review.

Update - Documentation pending, this is pretty much ready to go and could use some eyes on. The major change is that you'll need Docker (Docker for Mac on Macs) running for the build to complete. This is because the MPack requires the RPMs be built. 

To test, run Quick Dev or Full Dev or both.

I'll be working on the documentation in the next day or so.

This is the first set of changes to enable Ansible installation to use the Ambari MPack. It currently works (in my environment) with full-dev using sensor-stubs. ~~It will not work with Quick-Dev or EC2 at this time~~. Update: All environments worked in my testing. 

It's (well past - sorry) starting to get large, so I wanted to push it out for feedback while I'm working issues with the distributed install.

Some points of interest:
- As discussed, components (ES, Kibana, Metron Topologies) that are managed by Ambari are removed from Monit
- I had added a notion of 'required configurations' to the blueprint and ambari_cluster_state script so blueprint verification would pass.
- I've refactored our Ansible host groupings (see inventory/full-dev-platform/hosts). The Metron group is now the host that we use as the Metron services master.
- I cleaned up the very minimum number of issues that @mattf-horton found in [METRON-609](https://issues.apache.org/jira/browse/METRON-609) required to make single-node deployment work with the MPack. It's my intention to work to get his changes incorporated on top of the final version of this changeset.
- I removed the Solr role from Ansible. My current intention is to open a Jira to add a Solr Indexing component to the MPack.

Immediate next steps:

- [X] Get Full-Dev working.
- [X] Get the EC2 distributed install working.
- [X] Get Quick-Dev with enrichment replacement working.
- [X] Adjust readmes and docs where required.
- [X] Document test plan

I also hope to receive and respond to feedback.

PR #488, opened 2017-03-24
METRON-796: Mpack uses wrong group for owning HDFS directories

justinleet:
## Contributor Comments
Reverts the group owner of a couple HDFS directories to be the hadoop group, rather than the metron group (which is just metron).  Right now, the topologies run as the storm user (which belongs to the hadoop group), and therefore didn't have permission to write to HDFS (including in quick and full dev).  This sets HDFS ownership to metron:hadoop, which lets it be handled appropriately.

Other items, such as configs and installation files, were just left as the metron group.

To test, just run up a dev environment and ensure files are being written and ownership makes sense (/apps/metron/indexing/indexed is metron:hadoop with 755 perms).  The individual sensors will be owned by storm:hadoop (proving that writes work).

For example:
```
[vagrant@node1 ~]$ hdfs dfs -ls /apps/metron/indexing
Found 1 items
drwxrwxr-x   - metron hadoop          0 2017-03-24 12:57 /apps/metron/indexing/indexed
[vagrant@node1 ~]$ hdfs dfs -ls /apps/metron/indexing/indexed
Found 3 items
drwxrwxr-x   - storm hadoop          0 2017-03-24 13:01 /apps/metron/indexing/indexed/bro
drwxrwxr-x   - storm hadoop          0 2017-03-24 13:01 /apps/metron/indexing/indexed/error
drwxrwxr-x   - storm hadoop          0 2017-03-24 13:01 /apps/metron/indexing/indexed/snort
[vagrant@node1 ~]$ hdfs dfs -ls /apps/metron/indexing/indexed/bro
Found 1 items
-rw-r--r--   1 storm hadoop     211393 2017-03-24 13:01 /apps/metron/indexing/indexed/bro/enrichment-null-0-0-1490360489968.json
```

As a note, metron_group existed twice in params_linux.py, so only the first instance is changed to hadoop_group and pulled appropriately.  The second is left as-is.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

#### Note:

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
@@ -75,6 +75,8 @@ else:
 
 # hadoop params
 stack_root = Script.get_stack_root()
+# This is the cluster group named 'hadoop'. Its membership is the stack process user ids not individual users.
+# The config name 'user_group' out of our control and a bit misleading, so it is renamed to 'hadoop_group'.
 hadoop_group = config['configurations']['cluster-env']['user_group']
 hadoop_home_dir = stack_select.get_hadoop_dir("home")
 hadoop_bin_dir = stack_select.get_hadoop_dir("bin")
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
@@ -39,7 +39,7 @@ config = Script.get_config()
 tmp_dir = Script.get_tmp_dir()
 
 hostname = config['hostname']
-metron_group = config['configurations']['cluster-env']['metron_group']
+hadoop_group = config['configurations']['cluster-env']['user_group']
 metron_home = status_params.metron_home
 parsers = status_params.parsers
 geoip_url = config['configurations']['metron-env']['geoip_url']
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
@@ -76,7 +76,7 @@ else:
 # hadoop params
 stack_root = Script.get_stack_root()
 # This is the cluster group named 'hadoop'. Its membership is the stack process user ids not individual users.
-# The config name 'user_group' out of our control and a bit misleading, so it is renamed to 'hadoop_group'.
+# The config name 'user_group' is out of our control and a bit misleading, so it is renamed to 'hadoop_group'.
 hadoop_group = config['configurations']['cluster-env']['user_group']
 hadoop_home_dir = stack_select.get_hadoop_dir("home")
 hadoop_bin_dir = stack_select.get_hadoop_dir("bin")
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
@@ -39,7 +39,6 @@ config = Script.get_config()
 tmp_dir = Script.get_tmp_dir()
 
 hostname = config['hostname']
-hadoop_group = config['configurations']['cluster-env']['user_group']
 metron_home = status_params.metron_home
 parsers = status_params.parsers
 geoip_url = config['configurations']['metron-env']['geoip_url']
@@ -76,6 +75,7 @@ else:
 
 # hadoop params
 stack_root = Script.get_stack_root()
+hadoop_group = config['configurations']['cluster-env']['user_group']
 hadoop_home_dir = stack_select.get_hadoop_dir("home")
 hadoop_bin_dir = stack_select.get_hadoop_dir("bin")
 hadoop_conf_dir = conf_select.get_hadoop_conf_dir()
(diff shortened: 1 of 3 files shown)


simonellistonball:
Opening this to the hadoop group feels wrong from a security perspective. That gives all the other hadoop users too much write access to the data store. A better solution would be to have the topologies run as a metron user rather than as storm, or to add the storm user to the metron group to maintain a least access required principal.

justinleet:
I agree the topologies should run as the metron user, but this is just to get things back in a working state again (and it already used to be this way, so this isn't opening things up more than it was a couple weeks ago).

I actually thought there was a separate Jira for running as the Metron user, but the one I was thinking of is https://issues.apache.org/jira/browse/METRON-349.  The ticket should really be to consolidate everything under the metron user with appropriate ownership.  I don't have a preference for updating that ticket or closing it as a new one (and I'm not sure which the community would prefer).

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
```diff
@@ -39,7 +39,7 @@
 tmp_dir = Script.get_tmp_dir()
 
 hostname = config['hostname']
-metron_group = config['configurations']['cluster-env']['metron_group']
+hadoop_group = config['configurations']['cluster-env']['user_group']
```
mattf-horton: please put this line in the `hadoop params` section (circa line 80).

Also, why equate hadoop_group to 'user_group'?  In typical hadoop deployments there is already a group named 'hadoop'.  It is managed by Ambari.  In a recent test install of HDP+Metron, it was defined as follows (from /etc/group), which I think would meet your intended use since both storm and metron are included:
* hadoop: x :1004:hive,storm,zookeeper,infra-solr,zeppelin,metron,livy,elasticsearch,spark,flume,kafka,hdfs,yarn,kibana,mapred,hbase,hcat

justinleet: I'll go ahead and move the config.

On the group issue, that is the group named 'hadoop'.  The cluster level config is named 'user_group', I have absolutely no idea why. I only called it 'hadoop_group' here, so it was more obvious it shouldn't be killed in the future.  If there are objections to calling it 'hadoop_group', I could also carry it through as user_group and add a comment about the meaning in the params file.

For example, in HDP stack https://github.com/apache/ambari/blob/trunk/ambari-server/src/main/resources/stacks/HDP/2.0.6/configuration/cluster-env.xml#L158
```
    <name>user_group</name>
    <display-name>Hadoop Group</display-name>
    <value>hadoop</value>
    <property-type>GROUP</property-type>
    <description>Hadoop user group.</description>
```

This declaration carried through a couple other stack definitions that I looked at.

The use of this group also seems fairly common, e.g. in https://github.com/apache/ambari/blob/trunk/ambari-server/src/main/resources/common-services/HDFS/2.1.0.2.0/package/scripts/hdfs.py#L60
```
  if "hadoop-policy" in params.config['configurations']:
    XmlConfig("hadoop-policy.xml",
              conf_dir=params.hadoop_conf_dir,
              configurations=params.config['configurations']['hadoop-policy'],
              configuration_attributes=params.config['configuration_attributes']['hadoop-policy'],
              owner=params.hdfs_user,
              group=params.user_group
    )
```

justinleet:
For anybody looking, Matt's review comment is still relevant to discussion, but unfortunately hidden by GitHub thinking it's outdated after the last commit.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
mattf-horton: Yuck (but thanks for the data).  That's kinda perverse, but since it's in Ambari it's out of our scope to fix.  So I agree with your current usage.  A comment would be nice, since 
a. it's the `hadoop` group not the `users` group; and furthermore
b. its members are the *stack process user ids* not individual users.

justinleet: Updated with a comment to clarify things a bit.  Let me know if you think there's anything else we want to add or clarify.

mattf-horton:
Thanks.  Looks okay to me now, for the intended purpose.  +0 due to Simon's concerns; I'd prefer someone with more experience in Metron give you the plus one.

Regarding @simonellistonball 's comments:
The prior assignment of group ownership as `metron_group` in enrichment_commands.py, line 114, and indexing_commands.py, line 104, dates from the commit for 
* [METRON-671 Refactor existing Ansible deployment to use Ambari MPack (dlyle via justinleet)#436](https://github.com/justinleet/incubator-metron/commit/68a334a8717d2a6b82f7f9651c57bc75e256ba97)
Prior to PR#436 they were both set to `user_group`.

@dlyle65535 are you okay with setting these back to `user_group`, as they were before?
ie do you agree with @justinleet 's reasoning?  Thanks.

@justinleet , you say: "Other items, such as configs and installation files, were just left as the metron group."  Why is this okay?  Don't the topologies have to read config files too?  Thanks.

justinleet:
@mattf-horton Permissions on the other items are generally 775, so they can be read as needed (and should be scaled back once we have everything lined up with the user as @simonellistonball mentions). I just wanted to touch things as little as possible to get them back in a working state.

We should either expand out or replace METRON-349 (including closing off permissions) as the actual solution to the problem.

simonellistonball:
I agree, let's get this sorted, and then follow up with a general review of the permissions once we've got the rest of the security pieces in place.

mattf-horton:
+1 then, thanks.  Please expand METRON-349 to include these issues too.

dlyle65535:
I don't have an objection. I made that change because there was some issue with user_group. Unfortunately, I can't remember exactly what that was. In it's current state, what's the failure mode? I didn't see any errors in any of the final runs I performed in full dev or ec2.

justinleet:
@dlyle65535 Failure mode is that HDFS writes from Storm fail.  The directories are owned by metron:metron with 775.  Storm isn't in the metron group, so it fails to write.  The perms exception is thrown in the bolt and no output file is created.  Writes to ES work as expected.  Validating this is pretty easy, just run up full dev and see if the perms error shows up and if HDFS gets any files.

justinleet:
https://issues.apache.org/jira/browse/METRON-349 is updated to be more complete and reflect the current state of things.

cestella:
Piling on, +1  Thanks for the quick turnaround, @justinleet

PR #500, opened 2017-03-30
METRON-795: Install Metron REST with Ambari MPack

merrimanr:
## Contributor Comments
This PR adds the metron-rest module to the Ambari MPack.  This can be tested by building the rpms and mpack and spinning up full-dev-platform.  Once the ansible script has finished, navigate to "Ambari > Services > Metron" and there should be an additional "Metron REST" component in the Summary.  The "Quick Links" tab will have a link to the REST Swagger UI that should be functional if the service is running.  All the Ambari lifecycle commands should work.  I also tested starting/stopping the service with the "service metron-rest start/stop" commands and verified Ambari correctly reports status.

I ran into an issue with rpmlint complaining about missing chkconfig calls:

```
The package contains an init script but doesn't contain a %post with a call to
chkconfig
```

I believe this is benign because Ambari managements the REST application lifecycle but I can't figure out how to suppress them yet.  Also considering adding a couple lines to the metron-rest README that references the MPack install path.  Credentials for the REST app are "user/password" (will also add that to the metron-rest README).

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  bin/generate-md.sh
  mvn site:site
  ```

#### Note:


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/templates/application.yml.j2
```diff
@@ -0,0 +1,51 @@
...
+  broker:
+    url: {{kafka_brokers}}
+
+hdfs:
+  namenode:
+    url: {{default_fs}}
+
+grok:
+  path:
+    temp: ./patterns/temp
```
simonellistonball: should this be linked to the /apps/metron location, or use the /tmp in hdfs, or is it implicitly relative to /apps/metron already?

merrimanr: That is not an HDFS path.  Think of it as a staging area when testing grok statements before they are persisted to HDFS.

That being said, I would consider our strategy for testing grok parsers to be out of scope for this PR.  Am happy to discuss is further but I think it should be in a different thread.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_commands.py
```diff
@@ -0,0 +1,71 @@
...
+        yum_repo_types = {
+            'local': local_repo,
+            'remote': remote_repo
+        }
+        repo_type = self.__params.yum_repo_type
+        if repo_type in yum_repo_types:
+            yum_repo_types[repo_type]()
+        else:
+            raise ValueError("Unsupported repo type '{0}'".format(repo_type))
+
```
dlyle65535: Could you merge with current master and remove the repo setup stuff? It's been deprecated in favor of repoinfo.xml.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_master.py
```diff
@@ -0,0 +1,83 @@
...
+from resource_management.libraries.script import Script
+from resource_management.core.resources.system import Execute
+
+from rest_commands import RestCommands
+
+
+class RestMaster(Script):
+    def get_component_name(self):
+        # TODO add this at some point - currently will cause problems with hdp-select
+        # return "parser-master"
```
dlyle65535: Please remove. (I know they are in parser-master, but they probably shouldn't be.) :)

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_master.py
```diff
@@ -0,0 +1,83 @@
...
+        env.set_params(params)
+        commands = RestCommands(params)
+        commands.setup_repo()
+        self.install_packages(env)
+
+    def configure(self, env, upgrade_type=None, config_dir=None):
+        from params import params
+        env.set_params(params)
+
+        File(format("{metron_config_path}/application.yml"),
```
simonellistonball: Is this a good name for the config file? There should be some reference to it being the application.yml for the rest server.

metron-interface/metron-rest/src/main/scripts/metron-rest
```diff
@@ -0,0 +1,128 @@
...
+fi
+
+NAME=metron-rest
+DESC="Metron REST Application"
+SCRIPTNAME=/etc/init.d/$NAME
+LOGFILE="/var/log/metron-rest.log"
+METRON_VERSION=0.3.1
+METRON_HOME=/usr/metron/$METRON_VERSION
+METRON_REST_PORT=${METRON_REST_PORT:-8082}
+DAEMON="java -jar $METRON_HOME/lib/metron-rest-$METRON_VERSION.jar --spring.config.location=$METRON_HOME/config/application.yml --spring.profiles.active=dev --server.port=$METRON_REST_PORT"
```
simonellistonball: The dev profile shouldn't be activated by default in a production setup.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_commands.py
merrimanr: Done

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_master.py
merrimanr: Done

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_master.py
merrimanr: Good idea.  I changed the name to rest_application.yml.

metron-interface/metron-rest/src/main/scripts/metron-rest
merrimanr: The only thing that profile does is select JDBC authentication, set the database to H2 and load a test user.  This could potentially warrant it's own discussion:  how should we authenticate Metron services installed through Ambari MPack?  Should we tackle that here or create a follow-on Jira and possibly start a discussion?

merrimanr:
Just merged in master and pushed changes out to address the feedback so far.  

One question I have:  which account should this service run as?  My assumption would be the "metron" user but then we would likely need to update the permissions in HDFS which may be out of scope for this PR.  As it stands now it runs as root but I don't think this is correct.  Thoughts?

justinleet:
@merrimanr I'm pretty confident that is going to collide unpleasantly with https://github.com/apache/incubator-metron/pull/518 (The MPack should function in a kerberized cluster).

I haven't thought through the details, but do you think it's potentially sufficient to make this component optional (cardinality 0+, rather than 1)?  It's an extra step for the user, but it's only selecting it on the component list.  At that point, we say the REST API + UI is only supported on non-kerberized clusters (which is true anyway), and we (hopefully) don't break Kerberizing the core components.

justinleet:
"confident this PR", not "confident that".  In retrospect, it sounds like I'm referring to your last comment, not just in general.

merrimanr:
Just pushed out a commit that runs the REST service as the metron user and added configurable log and pid directories.

@justinleet, I tried setting cardinality 0+ and changed the component to a Slave.  While this did make it optional, it feels a little awkward for it to be listed as a 1/1 Slave component and the quick link no longer works.  Still not sure how we handle this.  I see 3 options:  

- we hold this PR up until the REST kerberos work is done
- we include it but document that it's not supported in a kerberized environment
- we continue investigating ways to make it optional

mattf-horton:
@merrimanr , do I understand correctly that the problem is:
* metron-rest should be a service, not a slave
* but as such, if it isn't installed the Metron as a whole says it isn't completely installed, because metron-rest is specified as a <componnt> of Metron.
Then it seems metron-rest should be a service dependent on Metron, but not vice versa.
Have you tried putting metron-rest at the <service> level as a peer to Metron, in that same metainfo.xml file?

merrimanr:
That's a really good idea @mattf-horton!  I think that would work but It's probably a moot point now. 
 The PR that kerberizes the REST application will be going up soon so making it optional won't be necessary anymore.

mattf-horton:
@merrimanr , since METRON-859 / PR#535 just went in, will this be going in too?  Thanks.

merrimanr:
Matt,

METRON-795 is ready for review but it's substantial and I imagine it will
take some time to make it through the process.

Ryan

justinleet:
I was able to run this up in a nonkerberized cluster, and then set up Kerberos.  Mostly things worked, but I spun out two tickets from this, [METRON-909](https://issues.apache.org/jira/browse/METRON-909) and [METRON-910](https://issues.apache.org/jira/browse/METRON-910).

909 is an entirely separate bug (one of the Storm calls references METRON_HOME and fails).

910 is more relevant, and could potentially be solved here if we want, but the REST API can't hit the Kafka endpoints in a Kerberized cluster, because of the user not having Kafka ACLs setup.

I'm not inclined to consider either of these blockers for getting this in (since they're preexisting), however, I would like to see if anyone has any other opinions.

merrimanr:
@justinleet, METRON-909 and METRON-910 have both been addressed in the latest commit.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
```diff
@@ -65,6 +65,102 @@
...
+        <description>JDBC Connection URL used by Metron</description>
+        <display-name>Metron JDBC URL</display-name>
+    </property>
+    <property require-input="true">
+        <name>metron_jdbc_username</name>
+        <value></value>
+        <description>Metron JDBC Username</description>
+        <display-name>Metron JDBC username</display-name>
+    </property>
+    <property require-input="true">
```
justinleet: I'm not particularly familiar with it, but Ambari has some setup for handling passwords.  We should look into what their recommended way of handling them is, so it's setup right.

Super brief digging has led to a couple things in hive.
```
<property require-input="true">
    <name>javax.jdo.option.ConnectionPassword</name>
    <value> </value>
    <property-type>PASSWORD</property-type>
    <display-name>Database Password</display-name>
    <description>password to use against metastore database</description>
    <value-attributes>
      <type>password</type>
      <overridable>false</overridable>
      <hidden>HIVE_CLIENT,WEBHCAT_SERVER,HCAT,CONFIG_DOWNLOAD</hidden>
    </value-attributes>
    <on-ambari-upgrade add="true"/>
```
eventually this becomes 
```
      create_schema_cmd = format("export HIVE_CONF_DIR={hive_server_conf_dir} ; "
                                 "{hive_schematool_bin}/schematool -initSchema "
                                 "-dbType {hive_metastore_db_type} "
                                 "-userName {hive_metastore_user_name} "
                                 "-passWord {hive_metastore_user_passwd!p} -verbose")
```
Notably with the `!p` suffix on the password, which I'm sure does some magic.

merrimanr:
The Ambari example you gave worked perfectly.  Should be good for a review now.

justinleet:
+1 pending Travis, thanks for taking care of the various comments and fixes here.

ottobackwards:
This is the best.  I just tested my new rest functions through it, and it was awesome.  Great job

merrimanr:
Thanks @ottobackwards!

PR #505, opened 2017-04-03
METRON-817: Customise output file path patterns for HDFS indexing

justinleet:
## Contributor Comments
Primarily this affects HdfsWriter by changing the output path from a set path (`/apps/metron/.../<sensor>`), and allow it to be defined via a Stellar Function.  Specifically, the base path is still defined the same (The `/apps/metron/.../` portion), but the `<sensor>` portion is dropped and can now be defined by a Stellar function.  By default, the original behavior of `<sensor>` is used.  This is defined in the `<sensor>.json` file as indicated in the new README.md for metron-writer.

### Notes
- This requires adding tracking things a bit more carefully (and if you're reviewing, please validate that it happens correctly).  When the outputFile is closed, we remove the sourceHandler from HdfsWriter's map.
  - I'm slightly concerned about the correctness of the implementation, but it seems necessary to ensure that we don't leave a bunch of SourceHandlers lying around as data changes (and we don't want an enormous number of output files being written to).
  - If there's a cleaner way to manage this, I'd love to hear it and can refactor pretty easily. It throws off the rotation count (because we kill the SourceHandler from the map itself), but I doubt we care about that since it really only shows up in the output filename anyway.
- This also adds an argument for max open files.  This is a flux level config. I defaulted this to 500.  500 was chosen because it was an arbitrary round number that wasn't enormous.
  - If someone has a default with any real reasoning behind it, I'll go ahead and change it.
- In HdfsWriter, we iterate through the messages, apply the Stellar function and then call the relevant handler. The entire group of message is treated as one single pass/fail (which is the same as the old behavior), rather than individually. The try/catch could potentially be moved into the for loop, but I don't think there's an explicit link between the message and the tuples that we can exploit to fail per message.  I don't think it needs to be addressed here, but I'm curious if there's thought on this.

### Testing
Unit tests are added to pretty much cover HdfsWriter, and this can be spun up in a dev environment.

To test in dev

- Spin up a dev environment
- Validate that the output matches the old format in HDFS (Nothing has an output function defined)
  ```
  [hdfs@node1 vagrant]$ hdfs dfs -ls /apps/metron/indexing/indexed/
  Found 3 items
  drwxrwxr-x   - storm hadoop          0 2017-04-03 13:11 /apps/metron/indexing/indexed/bro
  drwxrwxr-x   - storm hadoop          0 2017-04-03 13:11 /apps/metron/indexing/indexed/error
  drwxrwxr-x   - storm hadoop          0 2017-04-03 13:11 /apps/metron/indexing/indexed/snort
  ```
- Edit the indexing config for Bro to include an outputPathFunction in the hdfs section, e.g. in `/usr/metron/0.3.1/config/zookeeper/indexing/bro.json`
  ```
  {
    "hdfs" : {
      "index": "bro",
      "batchSize": 5,
      "enabled" : true,
      "outputPathFunction": "FORMAT('ipsrc-%s', ip_src_addr)"
    },
    "elasticsearch" : {
      "index": "bro",
      "batchSize": 5,
      "enabled" : true
    },
    "solr" : {
      "index": "bro",
      "batchSize": 5,
      "enabled" : true
    }
  }
  ```
- Push the config configs to ZooKeeper: `/usr/metron/0.3.1/bin/zk_load_configs.sh -z node1:2181 -m PUSH -i /usr/metron/0.3.1/config/zookeeper/`
- Let some more data run through and check the output folders, e.g.
  ```
  [hdfs@node1 vagrant]$ hdfs dfs -ls /apps/metron/indexing/indexed/
  Found 5 items
  drwxrwxr-x   - storm hadoop          0 2017-04-03 13:11 /apps/metron/indexing/indexed/bro
  drwxrwxr-x   - storm hadoop          0 2017-04-03 13:11 /apps/metron/indexing/indexed/error
  drwxrwxr-x   - storm hadoop          0 2017-04-03 13:14 /apps/metron/indexing/indexed/ipsrc-192.168.138.158
  drwxrwxr-x   - storm hadoop          0 2017-04-03 13:14 /apps/metron/indexing/indexed/ipsrc-192.168.66.1
  drwxrwxr-x   - storm hadoop          0 2017-04-03 13:11 /apps/metron/indexing/indexed/snort
  [hdfs@node1 vagrant]$ hdfs dfs -ls /apps/metron/indexing/indexed/ipsrc-192.168.138.158
  Found 1 items
  -rw-r--r--   1 storm hadoop     223182 2017-04-03 13:14 /apps/metron/indexing/indexed/ipsrc-  192.168.138.158/enrichment-null-0-0-1491225291377.json
  ```

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

- ~If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?~

  ```
  cd site-book
  bin/generate-md.sh
  mvn site:site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/HdfsWriter.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/HdfsWriter.java
@@ -17,8 +17,14 @@
  */
 package org.apache.metron.writer.hdfs;
 
+import org.apache.metron.common.configuration.IndexingConfigurations;
+import org.apache.metron.common.dsl.Context;
+import org.apache.metron.common.dsl.MapVariableResolver;
+import org.apache.metron.common.dsl.StellarFunctions;
+import org.apache.metron.common.dsl.VariableResolver;
+import org.apache.metron.common.stellar.StellarCompiler;
+import org.apache.metron.common.stellar.StellarProcessor;
 import org.apache.storm.tuple.Tuple;
-import org.apache.metron.common.configuration.EnrichmentConfigurations;
 import org.apache.metron.common.configuration.writer.WriterConfiguration;
 import org.apache.metron.common.writer.BulkMessageWriter;
 import org.apache.metron.common.writer.BulkWriterResponse;
@@ -39,8 +45,13 @@ public class HdfsWriter implements BulkMessageWriter<JSONObject>, Serializable {
   FileRotationPolicy rotationPolicy = new NoRotationPolicy();
   SyncPolicy syncPolicy = new CountSyncPolicy(1); //sync every time, duh.
   FileNameFormat fileNameFormat;
-  Map<String, SourceHandler> sourceHandlerMap = new HashMap<>();
+  Map<SourceHandlerKey, SourceHandler> sourceHandlerMap = new HashMap<>();
+  HashMap<String, StellarCompiler.Expression> sourceTypeExpressionMap = new HashMap<>();
+  int maxOpenFiles = 500;
+  transient StellarProcessor stellarProcessor;
   transient Map stormConfig;
+
+
   public HdfsWriter withFileNameFormat(FileNameFormat fileNameFormat){
     this.fileNameFormat = fileNameFormat;
     return this;
@@ -60,9 +71,15 @@ public class HdfsWriter implements BulkMessageWriter<JSONObject>, Serializable {
     return this;
   }
 
+  public HdfsWriter withMaxOpenFiles(int maxOpenFiles) {
+    this.maxOpenFiles = maxOpenFiles;
+    return this;
+  }
+
   @Override
   public void init(Map stormConfig, WriterConfiguration configurations) {
     this.stormConfig = stormConfig;
+    this.stellarProcessor = new StellarProcessor();
   }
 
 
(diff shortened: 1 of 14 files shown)


metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/HdfsWriter.java
```diff
@@ -74,17 +91,43 @@ public BulkWriterResponse write(String sourceType
...
   }
 
+  public String getHdfsPathExtension(String sourceType, String stellarFunction, JSONObject message) {
+    // If no function is provided, just use the sourceType directly
+    if(stellarFunction == null || stellarFunction.trim().isEmpty()) {
+      return sourceType;
+    }
+
+    StellarCompiler.Expression expression = sourceTypeExpressionMap.computeIfAbsent(stellarFunction, s -> stellarProcessor.compile(stellarFunction));
+    VariableResolver resolver = new MapVariableResolver(message);
```
ottobackwards: We should be able to find out from the function metadata/annotation the return type, without doing all this work shouldn't we?

ottobackwards: This makes me think of the UI case.  We configure the index configuration but have no way of validation before they save and deploy.

justinleet: Unfortunately, I don't think we can, unless we want to do more work to actually look up the function and validate. On top of it, things like MAP_GET essentially return Object anyway, so we'd still want to check if it's a String afterwards.

cestella: So, is there a reason why this isn't just:
```
//processor is a StellarProcessor();
VariableResolver resolver = new MapVariableResolver(message);
Object objResult = processor.parse(stellarFunction, resolver, StellarFunctions.FUNCTION_RESOLVER(), Context.EMPTY_CONTEXT());
if(!objResult instanceof String) {
  throw new IllegalArgumentException("Stellar Function <" + stellarFunction + "> did not return a String value. Returned: " + objResult);
}
return objResult == null?"":(String)objResult;
```

justinleet: @cestella I'm mostly concerned about the performance of function compile on every single message that comes through indexing.

If we keep the current approach, I would be interested in if there's a way to make things a little cleaner.

In retrospect, I think this should be an LRU cache, so that we don't keep around a given parse forever. Any thoughts on that, assuming performance would be enough of a concern to not just use your proposal?

cestella: Yeah, it's a good concern.  We *do* actually have a cache in the `StellarProcessor` so that compilations happen once and are cached afterwards.  As long as `StellarProcessor` is a transient member variable, I think you're good to do what I suggested.

justinleet: @cestella Made that change.  I did make the check `if(objResult != null && !(objResult instanceof String)`, to avoid having falling into the IAE when objResult is null.

cestella: After looking at this a bit further, while reusing the StellarProcessor *is* the right answer, it is apparent that we don't practice that everywhere...in fact, we practice it almost literally nowhere.  I have created a follow-on PR ( #508 ) to address that problem, which is a substantial performance issue, in fact.

cestella:
+1 by inspection

PR #516, opened 2017-04-06
METRON-830 Adding StringFunctions to Stellar - chop, prependifmissing, appendifmissing and countmatches

anandsubbu:
## Contributor Comments
Added stellar functions and unit tests for the following:
chop -> Remove the last character from a String.
prependifmissing -> Prepends the prefix to the start of the string if the string does not already start with any of the prefixes.
appendifmissing -> Appends the suffix to the end of the string if the string does not already end with any of the suffixes.
countmatches -> Counts how many times the substring appears in the larger string.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  bin/generate-md.sh
  mvn site:site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-common/README.md
+++ b/metron-platform/metron-common/README.md
@@ -177,7 +177,7 @@ The `!=` operator is the negation of the above.
   * Input:
     * string - The string to be appended.
     * suffix - The string suffix to append to the end of the string.
-    * suffixes - Optional - Additional string suffixes that are valid terminators.
+    * additionalsuffix - Optional - Additional string suffix that is a valid terminator.
   * Returns: A new String if prefix was prepended, the same string otherwise.
 
 ### `BLOOM_ADD`
@@ -471,7 +471,7 @@ The `!=` operator is the negation of the above.
   * Input:
     * string - The string to be prepended.
     * prefix - The string prefix to prepend to the start of the string.
-    * prefixes - Optional - Additional string prefixes that are valid.
+    * additionalprefix - Optional - Additional string prefix that is valid.
   * Returns: A new String if prefix was prepended, the same string otherwise.
 
 ### `PROFILE_GET`
--- a/metron-platform/metron-common/README.md
+++ b/metron-platform/metron-common/README.md
@@ -78,11 +78,15 @@ The `!=` operator is the negation of the above.
 |                                                                                                    |
 | ----------                                                                                         |
 | [ `ABS`](../../metron-analytics/metron-statistics#abs)                                             |
+| [ `APPEND_IF_MISSING`](#append_if_missing)                                                         |
 | [ `BIN`](../../metron-analytics/metron-statistics#bin)                                             |
 | [ `BLOOM_ADD`](#bloom_add)                                                                         |
 | [ `BLOOM_EXISTS`](#bloom_exists)                                                                   |
 | [ `BLOOM_INIT`](#bloom_init)                                                                       |
 | [ `BLOOM_MERGE`](#bloom_merge)                                                                     |
+| [ `CHOP`](#chop)                                                                                   |
+| [ `CHOMP`](#chomp)                                                                                 |
+| [ `COUNT_MATCHES`](#count_matches)                                                                 |
 | [ `DAY_OF_MONTH`](#day_of_month)                                                                   |
 | [ `DAY_OF_WEEK`](#day_of_week)                                                                     |
 | [ `DAY_OF_YEAR`](#day_of_year)                                                                     |
@@ -121,9 +125,10 @@ The `!=` operator is the negation of the above.
 | [ `MAAS_MODEL_APPLY`](#maas_model_apply)                                                           |
 | [ `MAP_EXISTS`](#map_exists)                                                                       |
 | [ `MONTH`](#month)                                                                                 |
+| [ `PREPEND_IF_MISSING`](#prepend_if_missing)                                                       |
 | [ `PROFILE_GET`](#profile_get)                                                                     |
-| [ `PROFILE_FIXED`](#profile_fixed)                                                                     |
-| [ `PROFILE_WINDOW`](#profile_window)                                                                     |
+| [ `PROFILE_FIXED`](#profile_fixed)                                                                 |
+| [ `PROFILE_WINDOW`](#profile_window)                                                               |
 | [ `PROTOCOL_TO_NAME`](#protocol_to_name)                                                           |
 | [ `REGEXP_MATCH`](#regexp_match)                                                                   |
 | [ `SPLIT`](#split)                                                                                 |
@@ -167,6 +172,14 @@ The `!=` operator is the negation of the above.
 | [ `WEEK_OF_YEAR`](#week_of_year)                                                                   |
 | [ `YEAR`](#year)                                                                                   |
 
+### `APPEND_IF_MISSING`
+  * Description: Appends the suffix to the end of the string if the string does not already end with any of the suffixes.
+  * Input:
+    * string - The string to be appended.
+    * suffix - The string suffix to append to the end of the string.
+    * suffixes - Optional - Additional string suffixes that are valid terminators.
+  * Returns: A new String if prefix was prepended, the same string otherwise.
+
 ### `BLOOM_ADD`
   * Description: Adds an element to the bloom filter passed in
   * Input:
@@ -193,6 +206,25 @@ The `!=` operator is the negation of the above.
   * Input:
     * bloomfilters - A list of bloom filters to merge
   * Returns: Bloom Filter or null if the list is empty
+
+### `CHOP`
+  * Description: Remove the last character from a String
+  * Input:
+    * string - the String to chop last character from, may be null
+  * Returns: String without last character, null if null String input
+
+### `CHOMP`
+  * Description: Removes one newline from end of a String if it's there, otherwise leave it alone. A newline is "\n", "\r", or "\r\n"
+  * Input:
+    * string - the String to chomp a newline from, may be null
+  * Returns: String without newline, null if null String input
+
+### `COUNT_MATCHES`
+  * Description: Counts how many times the substring appears in the larger string.
+  * Input:
+    * string - the CharSequence to check, may be null.
+    * substring/character - the substring or character to count, may be null.
+  * Returns: the number of non-overlapping occurrences, 0 if either CharSequence is null.
 
 ### `DAY_OF_MONTH`
   * Description: The numbered day within the month.  The first day within the month has a value of 1.
@@ -433,6 +465,14 @@ The `!=` operator is the negation of the above.
   * Input:
     * dateTime - The datetime as a long representing the milliseconds since unix epoch
   * Returns: The current month (0-based).
+
+### `PREPEND_IF_MISSING`
+  * Description: Prepends the prefix to the start of the string if the string does not already start with any of the prefixes.
+  * Input:
+    * string - The string to be prepended.
+    * prefix - The string prefix to prepend to the start of the string.
+    * prefixes - Optional - Additional string prefixes that are valid.
+  * Returns: A new String if prefix was prepended, the same string otherwise.
 
 ### `PROFILE_GET`
   * Description: Retrieves a series of values from a stored profile.
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
@@ -354,7 +354,7 @@ public class StringFunctions {
     @Override
     public Object apply(List<Object> strings) {
 
-      if(strings.size() == 0) {
+      if(strings == null || strings.size() == 0 || strings.get(0) == null || strings.get(0).toString().length() == 0) {
         throw new IllegalArgumentException("[CHOMP] missing argument: string to chomp a newline from");
       }
 
@@ -372,7 +372,7 @@ public class StringFunctions {
     @Override
     public Object apply(List<Object> strings) {
 
-      if(strings.size() == 0) {
+      if(strings == null || strings.size() == 0 || strings.get(0) == null || strings.get(0).toString().length() == 0) {
         throw new IllegalArgumentException("[CHOP] missing argument: string to be chopped");
       }
 
@@ -386,7 +386,7 @@ public class StringFunctions {
           , params = {
           "str - The string."
           , "prefix - The string prefix to prepend to the start of the string"
-          , "prefixes - Optional - Additional string prefixes that are valid"
+          , "additionalprefix - Optional - Additional string prefix that is valid"
   }
           , returns = "A new String if prefix was prepended, the same string otherwise."
   )
@@ -412,7 +412,7 @@ public class StringFunctions {
           , params = {
           "str - The string."
           , "suffix - The string suffix to append to the end of the string"
-          , "suffixes - Optional - Additional string suffixes that are valid terminators"
+          , "additionalsuffix - Optional - Additional string suffix that is a valid terminator"
   }
           , returns = "A new String if suffix was appended, the same string otherwise."
   )
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
@@ -21,7 +21,7 @@ package org.apache.metron.common.dsl.functions;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.collect.Iterables;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.metron.common.dsl.BaseStellarFunction;
 import org.apache.metron.common.dsl.ParseException;
 import org.apache.metron.common.dsl.Stellar;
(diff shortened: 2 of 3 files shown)


metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,89 @@ public Object apply(List<Object> args) {
...
+  @Stellar( name = "PREPENDIFMISSING"
+          , description = "Prepends the prefix to the start of the string if the string does not already start with any of the prefixes"
+          , params = { "str - The string.", "prefix - The prefix to prepend to the start of the string", "prefixes - Additional prefixes that are valid" }
+          , returns = "A new String if prefix was prepended, the same string otherwise."
+  )
+  public static class prependifmissing extends BaseStellarFunction {
+
+    @Override
+    public Object apply(List<Object> strings) {
+
```
ottobackwards: Invalid Cast exception?

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
```diff
@@ -207,4 +207,78 @@ public void testFormatWithNoArguments() throws Exception {
...
+
+  /**
+  * CHOP StringFunction
+  * @throws Exception
+  */
+  @Test
+  public void testChop() throws Exception {
+    Assert.assertEquals("ab",   run("CHOP('abc')", new HashedMap()));
+    Assert.assertEquals(null,   run("CHOP(null)", new HashedMap()));
+    Assert.assertEquals("abc",  run("CHOP(msg)", ImmutableMap.of("msg", "abc\r\n")));
```
mattf-horton: Why does this work, when the string "abc\r\n" is never processed by any of the shell or java string input processors that might convert it?

ottobackwards:
Thanks for the contribution!

Quick review, first thing please follow the style guide for naming classes.  You can see from the other stellar function classes.

It would be good to have some tests that aren't with string or null, to verify the behavior.

What happens if I pass in a number literal?

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
```diff
@@ -207,4 +207,78 @@ public void testFormatWithNoArguments() throws Exception {
...
+  }
+
+  /**
+   * PREPENDIFMISSING StringFunction
+   * @throws Exception
+   */
+  @Test
+  public void testPrependIfMissing() throws Exception {
+    Assert.assertEquals("xyzabc",     run("PREPENDIFMISSING('abc', 'xyz')", new HashedMap()));
+    Assert.assertEquals("xyzXYZabc",  run("PREPENDIFMISSING('XYZabc', 'xyz', 'mno')", new HashedMap()));
```
mattf-horton: Please add a case that exercises non-null alt prefix, eg:
```
Assert.assertEquals("mnoXYZabc",  run("PREPENDIFMISSING('mnoXYZabc', 'xyz', 'mno')", new HashedMap()));
```

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
```diff
@@ -207,4 +207,78 @@ public void testFormatWithNoArguments() throws Exception {
...
+    Assert.assertEquals(null,         run("PREPENDIFMISSING(null, null, null)", new HashedMap()));
+    Assert.assertEquals("xyz",        run("PREPENDIFMISSING('', 'xyz', null)", new HashedMap()));
+  }
+
+  @Test(expected = ParseException.class)
+  public void testPrependIfMissingWithIncorrectArgs() throws Exception {
```
mattf-horton: I may be wrong, but I think JUnit expected errors do not auto-continue inside an errored test case.  So, you're only testing the first sub-case.  

To fix, you either need 3 test cases, or you need to do try/catch/finally contexts around each `run()`, where the catch eats the exception but sets a marker flag, and the finally asserts the marker flag.

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
```diff
@@ -207,4 +207,78 @@ public void testFormatWithNoArguments() throws Exception {
...
+    run("PREPENDIFMISSING()", Collections.emptyMap());
+    run("PREPENDIFMISSING('abc')", Collections.emptyMap());
+    run("PREPENDIFMISSING('abc', 'def', 'ghi', 'jkl')", Collections.emptyMap());
+  }
+
+  /**
+   * APPENDIFMISSING StringFunction
```
mattf-horton: Same issues here of course, and in COUNTMATCHES.

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
cestella: Because it's coming in through a variable `msg`.  The idea being that it could be existent in a message flowing through the topologies.  Stellar can't represent `\n` and `\r`, but they can get in there honestly.

anandsubbu:
Hi @mattf-horton  and @ottobackwards , thank you for the review. Please see review comments addressed as a part of my latest commit, and also responses inline.

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
anandsubbu: Yup, you're right. Added try/catch blocks now.

metron-platform/metron-common/src/test/java/org/apache/metron/common/dsl/functions/StringFunctionsTest.java
anandsubbu: Done, added.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,89 @@ public Object apply(List<Object> args) {
...
+          throw new IllegalArgumentException("[APPENDIFMISSING] incorrect arguments. Usage: APPENDIFMISSING <String> <prefix> [<prefix>...]");
+      }
+      return suffixed;
+    }
+  }
+
+  @Stellar( name = "COUNTMATCHES"
+          , description = "Counts how many times the substring appears in the larger string"
+          , params = { "str - the CharSequence to check, may be null", "sub - the substring to count, may be null"}
+          , returns = "the number of occurrences, 0 if either CharSequence is null"
```
anandsubbu: Corrected.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,89 @@ public Object apply(List<Object> args) {
...
+    @Override
+    public Object apply(List<Object> strings) {
+
+      String prefixed;
+      switch (strings.size()) {
+        case 2: prefixed = org.apache.commons.lang3.StringUtils.prependIfMissing((String) strings.get(0), (String) strings.get(1));
+          break;
+        case 3: prefixed = org.apache.commons.lang3.StringUtils.prependIfMissing((String) strings.get(0), (String) strings.get(1), (String) strings.get(2));
+          break;
+        default: throw new IllegalArgumentException("[PREPENDIFMISSING] incorrect arguments. Usage: PREPENDIFMISSING <String> <prefix> [<prefix>...]");
```
anandsubbu: Now included the arguments provided by user as well, apart from Usage info.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,89 @@ public Object apply(List<Object> args) {
...
+          , returns = "A new String if prefix was prepended, the same string otherwise."
+  )
+  public static class prependifmissing extends BaseStellarFunction {
+
+    @Override
+    public Object apply(List<Object> strings) {
+
+      String prefixed;
+      switch (strings.size()) {
+        case 2: prefixed = org.apache.commons.lang3.StringUtils.prependIfMissing((String) strings.get(0), (String) strings.get(1));
```
anandsubbu: Changed the import statement to use lang3.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,89 @@ public Object apply(List<Object> args) {
...
+      }
+
+      String chop = StringUtils.chop((String) strings.get(0));
+      return chop;
+    }
+  }
+
+  @Stellar( name = "PREPENDIFMISSING"
+          , description = "Prepends the prefix to the start of the string if the string does not already start with any of the prefixes"
+          , params = { "str - The string.", "prefix - The prefix to prepend to the start of the string", "prefixes - Additional prefixes that are valid" }
```
anandsubbu: Sure, done.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,89 @@ public Object apply(List<Object> args) {
...
+      if(strings.size() == 0) {
+        throw new IllegalArgumentException("[CHOP] missing argument: string to be chopped");
+      }
+
+      String chop = StringUtils.chop((String) strings.get(0));
+      return chop;
+    }
+  }
+
+  @Stellar( name = "PREPENDIFMISSING"
```
anandsubbu: Sure, changed.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,89 @@ public Object apply(List<Object> args) {
       return String.format(format, formatArgs);
     }
   }
+
+  @Stellar( name="CHOP"
```
anandsubbu: IMHO the `chop` function would be helpful when fine-tuning string extraction operations through stellar (E.g. when extracting specific variations of hostnames or URLs). 

Aside of that, I went ahead and added the `chomp` string function as well since I agree that it would be of use.

ottobackwards: If we are going to do URL/URI work, we should create functions off of the URI / URL classes and use their specific parsing for uri named component parts

ottobackwards: That is not to say the CHOP cannot be useful however

anandsubbu: Okay, thanks. I stand corrected about the URI/URL.

cestella:
This looks good to me, anand.  Thanks so much for the contribution.  I'll spin it up and try the functions out in a REPL.  @mattf-horton @ottobackwards anything other outstanding concerns?

ottobackwards:
+1, thanks for the contribution

metron-platform/metron-common/README.md
```diff
@@ -167,6 +172,14 @@ The `!=` operator is the negation of the above.
 | [ `WEEK_OF_YEAR`](#week_of_year)                                                                   |
 | [ `YEAR`](#year)                                                                                   |
 
+### `APPEND_IF_MISSING`
+  * Description: Appends the suffix to the end of the string if the string does not already end with any of the suffixes.
+  * Input:
+    * string - The string to be appended.
+    * suffix - The string suffix to append to the end of the string.
+    * suffixes - Optional - Additional string suffixes that are valid terminators.
```
mattf-horton: Please state the format of "suffixes".  Is it a string, a regex, an array of strings, or what?  If it is a single string, how does it represent plural suffixes?

metron-platform/metron-common/README.md
```diff
@@ -434,6 +466,14 @@ The `!=` operator is the negation of the above.
     * dateTime - The datetime as a long representing the milliseconds since unix epoch
   * Returns: The current month (0-based).
 
+### `PREPEND_IF_MISSING`
+  * Description: Prepends the prefix to the start of the string if the string does not already start with any of the prefixes.
+  * Input:
+    * string - The string to be prepended.
+    * prefix - The string prefix to prepend to the start of the string.
+    * prefixes - Optional - Additional string prefixes that are valid.
```
mattf-horton: ditto

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,89 @@ public Object apply(List<Object> args) {
...
+          , params = { "the String to chop last character from, may be null"}
+          , returns = "String without last character, null if null String input"
+  )
+  public static class chop extends BaseStellarFunction {
+
+    @Override
+    public Object apply(List<Object> strings) {
+
+      if(strings.size() == 0) {
+        throw new IllegalArgumentException("[CHOP] missing argument: string to be chopped");
```
mattf-horton: Any thoughts about why treat nulls and empty strings differently?

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
```diff
@@ -343,4 +343,120 @@ public Object apply(List<Object> args) {
...
+      return prefixed;
+    }
+  }
+
+  @Stellar( name = "APPEND_IF_MISSING"
+          , description = "Appends the suffix to the end of the string if the string does not already end with any of the suffixes"
+          , params = {
+          "str - The string."
+          , "suffix - The string suffix to append to the end of the string"
+          , "suffixes - Optional - Additional string suffixes that are valid terminators"
```
mattf-horton: Same doc concern as in the .md file: Is "suffixes" a string, a regex, an array of strings, or what? If it is a single string, how does it represent plural suffixes?

metron-platform/metron-common/README.md
anandsubbu: Hi @mattf-horton , I was trying to following the same wording as the [apache commons documentation](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html) for PrependIfMissing, which uses the plural form. 

Let me know if it would be better if I changed the text as follows:

`* suffix - The string suffix to append to the end of the string.`
`* additionalsuffix - Optional - An additional string suffix that is a valid terminator.`

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
anandsubbu: Do you think adding a check for null as follows, would be the right way?

```
 if(strings == null || strings.size() == 0) {
    throw new IllegalArgumentException("[CHOP] missing argument: string to be chopped");
}
```

mattf-horton: I think you need 
```
if (strings == null || strings.size() == 0 || strings.get(0) == null || strings.get(0).length == 0) { ...
```

metron-platform/metron-common/README.md
mattf-horton: Ah, got it.  The documentation you point at for apache commons PrependIfMission method actually uses signature:
```
CharSequence... suffixes
```
so it is a varargs usage.  You can do the same with Stellar, or we have found it better to allow a list of strings instead (which you would explode into the calling varargs in the java code).  Using a list instead of varargs makes it much easier to extend the function later if we choose.

It's up to you whether to do this now.  We are trying to close a RC, so there is some time pressure.  Maybe you should just open a jira to enhance this later with a list arg.  

The current implementation of your call to PrependIfMissing is such that the "additionalsuffix" description you suggest would be correct.

anandsubbu: Okay, I will retain additionalsuffix for now. Created METRON-847 to track the enhancement.

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
anandsubbu: Thanks, made this change.

mattf-horton:
+1, thanks Anand!

metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java
anandsubbu: Hi @mattf-horton , with the above change where we check for `strings.get(0) == null || strings.get(0).toString().length() `, the following test results in a failure:

`Assert.assertEquals("abc",  run("CHOP(msg)", ImmutableMap.of("msg", "abc\r\n")));`

Pardon my limited understanding, but this is what I inferred:
- The `if` condition for null and 0 length check occurs twice.
- In the case of the `msg` test case where we check for `\n\r`, the initial contents of the `strings` is null and the variable substitution for `msg` occurs subsequently. So, this would end up failing for the 1st check. 

Due to this, I will modify the checks to only verify for `strings == null` and `string.size() == 0`. Please let me know if there is a better way this can be done.

anandsubbu:
Thanks a lot @cestella for your inputs and help. The latest commit addresses feedback from @mattf-horton and @cestella.

anandsubbu:
@mattf-horton - if you feel the latest commit addresses your feedback, can you please help with merging this pull request? I already have your +1 and @ottobackwards 's as well. Thank you!

PR #518, opened 2017-04-07
METRON-799: The MPack should function in a kerberized cluster

justinleet:
## Contributor Comments
Allows the Ambari Kerberos wizard to handle Metron setup.

Changes include:
- Creation of Keytabs
- Running everything as the Metron user, including Storm topology workers (on a Kerberized cluster).
- Setup for Metron user to actually be able to run (client_jaas setup, home Storm dir setup, etc.)
- Adjusting perms to 0755.  The exception is the HDFS output folder on a non-kerb cluster is left as 0775 because we don't have Storm running workers as metron user on.  When Kerberizing, the permissions will be restricted down to 0755.
- Kafka ACLs
- HBase ACLs
- Refactored Topic creation to use a common function so I didn't have to edit the same thing 3 times.
- Automated updating of Storm configs (the AutoTGT and running workers as user)

There's still more testing I want to do, but this is definitely far enough along to submit a PR.

I've spun this up on full dev with the now modified Kerberos setup instructions, with the caveat that Ambari's Storm service check fails (it's harmless, as far as I can tell).  See below for more details.  As this does not touch the sensors, data will need to be pushed manually (same as the old instructions).  I've been able to push data from Kafka to Elasticsearch/HDFS.

### The Bad News
I would love insight on a problem, if anybody has some.  I haven't edited the docs to reflect this yet, in the hopes it'll be resolved.

Storm's service check will fail during (and after) Kerberization.  Metron can immediately be started perfectly fine.  Nothing is legit wrong, but this setup means that the storm user is unable to submit to the cluster (it doesn't have it's home directory setup with some configs).  Unfortunately, Ambari runs the service check as the storm user.

This can be worked around by creating ~storm/.storm/storm.yaml
```
nimbus.seeds : ['node1']
java.security.auth.login.config : '/usr/hdp/current/storm-supervisor/conf/storm_jaas.conf'
storm.thrift.transport : 'org.apache.storm.security.auth.kerberos.KerberosSaslTransportPlugin
```
`java.security.auth.login.conf` can also be `/etc/storm/conf/storm_jaas.conf`, but the value above leads me to my next point.  All of these values already exist in storm.yaml.  The fact that they need to be specified again in the user's home is really strange. And it'll give an error that the TGT found is not renewable, not something you'd expected.

I'm unsure if there are restrictions on where Ambari chooses to run service check, so it's possible this would have to be setup on every node Storm lives on the cluster. I'm also unsure if we can actually have Ambari automate this if it turns out to be necessary, since we aren't the Storm service.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

- ~Have you written or updated unit tests and or integration tests to verify your changes?~
- ~If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?~ 

  ```
  cd site-book
  bin/generate-md.sh
  mvn site:site
  ```

#### Note:


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_security.py
```diff
@@ -0,0 +1,74 @@
...
+             mode=0755
+             )
+
+        File(metron_storm_dir + '/storm.yaml',
+             content=Template('storm.yaml.j2'),
+             owner=params.metron_user,
+             group=params.metron_group,
+             mode=0755
+             )
+
```
dlyle65535: Does this override the storm.yaml that is at /etc/conf and managed by Ambari?

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_security.py
```diff
@@ -0,0 +1,74 @@
...
+                  group=params.metron_group
+                  )
+
+        File(ambari_format('{client_jaas_path}'),
+             content=Template('client_jaas.conf.j2'),
+             owner=params.metron_user,
+             group=params.metron_group,
+             mode=0755
+             )
+
```
dlyle65535: Does this file end up on all the supervisor nodes too? Does it need to?

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
```diff
@@ -72,55 +93,46 @@ def remote_repo():
...
+
+        # Non Kerberized Metron runs under 'storm', requiring write under the 'hadoop' group.
+        # Kerberized Metron runs under it's own user.
+        ownership = 0755 if self.__params.security_enabled else 0775
+        Logger.info('HDFS indexing directory ownership is: ' + str(ownership))
         self.__params.HdfsResource(self.__params.metron_apps_indexed_hdfs_dir,
                                    type="directory",
                                    action="create_on_execute",
                                    owner=self.__params.metron_user,
                                    group=self.__params.hadoop_group,
```
dlyle65535: Did you intend to leave this as hadoop group ownership or change it to the metron group?

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/enrichment_commands.py
```diff
@@ -131,47 +153,35 @@ def init_geo(self):
...
-                                            retention_bytes))
-
-        Logger.info("Done creating Kafka topics")
+        Logger.info('Creating Kafka topics for enrichment')
+        # All errors go to indexing topics, so create it here if it's not already
+        metron_service.init_kafka_topics(self.__params, [self.__enrichment_topic, self.__params.metron_error_topic])
         self.set_kafka_configured()
 
+    def init_kafka_acls(self):
+        Logger.info('Creating Kafka topics')
```
dlyle65535: Did you mean something like "Setting ACL" or even to omit this line?

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_security.py
justinleet: Just the properties in here.  Couple thoughts as you bring this up.

We probably want to have a ticket to make sure turning off Kerberos works correctly in the future.  The properties in the file (except nimbus.seeds) are essentially set properties.  We need our own client_jaas, and the storm.thrift.transport has to be there for some reason and that's pretty much constant on a secure cluster.  That should also be made a property that flows down from Storm.

nimbus.seeds is wrong and I need to carry that over from the actual Storm property. And I even made sure it was in params_linux and forgot to use it.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_security.py
justinleet: All the stuff in (or referenced in the case of client_jaas) ~metron/.storm should just need to be on the Metron node, because everything gets kicked off there.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
justinleet: I decided not to mess with it.  If we have a preference on it not being owned by metron:hadoop, we can go ahead and do that, but I think we probably need a more thorough discussion of how we want all that owned and permissioned anyway.  Leaving it only readable seemed like a reasonable compromise for now.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_security.py
dlyle65535: Won't the supervisors need to read/write to HBase/HDFS? How are they authenticated?

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/enrichment_commands.py
justinleet: Yes, yes I did.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
dlyle65535: Seems reasonable to me too, thanks.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_security.py
dlyle65535: So, if we switch on HA Storm after Metron kerberization, there's an issue?

I'm a bit uncomfortable with all the all the additional management required to use AutoTGT rather than Auto[HDFS|HBase]TGT. I didn't deliberately choose those two when I started (I just happened to land on those docs first), but that seems less complex than what we're currently doing. Are there benefits I'm missing? Because managing a jaas file and storm.yaml from outside of the Ambari Storm Service seems to add difficulty and risk.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_security.py
justinleet: My understanding, and correct me if I'm wrong, is that nimbus distributed and renews the TGT from Kerberos.  I'm honestly not sure how we'd get everything to line up otherwise if we have to distribute keytabs and jaas files and everything else out.

This does raise the question of what happens when the max renewal of a Kerberos ticket has passed. I don't know enough about the implementation to know what happens, and I'd be interested in thoughts.

dlyle65535:
I have a (likely wrong) idea about the service check. A couple of things to check- can you verify that Ambari runs the service check as the storm user, because that's just weird, isn't it? If it does, make sure the user identity it's using is in nimbus.admins. If that doesn't work, maybe try adding the metron user to nimbus.admins.

justinleet:
@dlyle65535 I'll take a look at nimbus.admins.  I'm pretty confident it's running as Storm user though,
Partial output from Ambari:
```
Traceback (most recent call last):
 ...
    user=params.storm_user
```

dlyle65535:
Yeah, reading is FUNdamental. I didn't see that it was the Storm service check that was failing. I kind of made up in my head that it was the Metron service check, because that would make sense.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_security.py
justinleet: nimbus.seeds is updated to grab the actual value.  Oddly, storm.thrift.transport doesn't end up flowing correctly to the template.  I double checked all the spelling and so on, and it always ended up just being the variable, not the value.  I'm inclined to let that one go, since it's only even created in a Kerberized environment anyway and it's a constant.

dlyle65535: Just to confirm, the shadowed nimbus.seeds is changed when I change the value for Storm via Ambari?

justinleet: It's not possible to change it via Ambari, but it will populate the variable appropriately.  There's some translation magic happening somewhere, because Ambari lists the value of nimbus.seeds as 'node1', but it actually passes it down as a list (which is why the new commit doesn't surround it with square brackets). I'm guessing this magic is also why the thrift config doesn't get passed.

dlyle65535: I think I can change it by setting up Storm HA. I think that's allowable via Ambari. But good deal that it populates it.

justinleet: Yeah, you're right, you should be able to change it via HA.  I just meant I can't go and change it to something in full dev.

Re: the AutoTGT stuff
I'm unsure if HA Storm would be an issue, but assuming Storm handles spinning up secure nimbuses correctly it shouldn't be.  New nimbuses are required to make sure they get all code of active topologies.  I would assume this includes getting any TGTs (although I haven't verified this) to ensure that active topologies can continue to run.  We'd need to test on an actual HA cluster.

I disagree with AutoHDFS/AutoHBase being less complex, because if I recall correctly (and I don't, please correct me), that solution required setting up symlinks on each Storm node.  I don't even know that we easily have that capability in Ambari as the Metron service. Even if we do work around that, I don't see AutoTGT as complicated, and the Storm docs mention "On a kerberos secure cluster they should be set by default to point to org.apache.storm.security.auth.kerberos.AutoTGT."  It seems reasonable to go with Storm's recommendation unless we have a compelling reason not to.

The custom storm.yaml only contains 3 configs that we need to run
1. Nimbus seeds (gathered from Storm configs themselves)
2. Jaas file
3. Thrift protocol (which admittedly doesn't grab correctly from Storm, but is essentially a constant for the purpose of Kerberos).

Of these, only the jaas file really matters for management, and all it does it define named stanzas and it's not anything particularly complicated.

Having said that, the Storm service check could make AutoTGT a lot less attractive if it's not easily workable.

justinleet:
So coming back to the AutoTGT discussion and making it easier to see.  There's basically two approaches we can takes that (should) work.

1. The current use of AutoTGT.  This requires setting up .storm/storm.yaml for the Storm nodes, and users will have to do the same.
2. The AutoHDFS / AutoHBase solution mentioned by @dlyle65535.  This requires symlinking some jars and configs to make them available to Storm.  Should only be necessary for Nimbuses, but it does mean that HDFS/HBase upgrades can make the symlinks stale.

Either solution requires Ambari acting on a different node (potentially) than the one running the scripts.  I don't know if Ambari has any resources for handling that sort of things.  It could potentially be a command over ssh to another node, but presumably that requires passwordless ssh setup or someone to manually create the symlinks (which may be acceptable for this pass).

I'm fairly strongly inclined towards the second one, primarily because it requires less effort on the users part.  Ambari work is fairly similar either way.

justinleet:
Still need to do some additional testing (I'm spiin, but this is updated to pull the topology.auto-credentials down to the topology level.

This should let Storm actually run service check (because AutoTGT isn't initializing).  At the same time, Metron should be able to set these appropriately and run them.   This involves passing the config down to the various files (the .properties files that feed Flux) and updating integration tests slightly.

Oddly enough, the use of the tickets in the client_jaas file caused issues, so it's now using the ticket cache and running a kinit before acting on topologies.  I strongly dislike having the kinits, but there's no obvious reason for the difference in behavior between ticket cache and keytab.

justinleet:
Full dev is able to successfully startup, be Kerberized via Ambari without errors (including Storm service check), and have new data flow through.

justinleet:
Updated to use builtin check for security in stack advisor.  Old versions caused problems because of typing in Python and my lack of catching what it actually does.  Params files should still be fine, because the config objects underneath the covers in Ambari explicitly convert strings to booleans if appropriate.

justinleet:
Update on this:

After a significant amount of pain involved in making this work, we need to have the client_jaas.conf and the metron keytab distributed to the various supervisor nodes so that they can actually be authenticated/authorized as the metron user.

The best discovered solution to this is to have a client that essentially just sets them up (so we can actually create things on the various supervisor servers).  It's not ideal, but @dlyle65535 and I are testing that it works.

To actually set this up, just install Metron clients on all Storm supervisor nodes prior to Kerberization.  Docs will be updated once we confirm that this approach works.

justinleet:
I've run this up on an EC2 cluster, and been able to get data from the core topologies from Kafka to ES/HDFS.  There are a couple caveats spelled out in the metron-deployment README now in a short Kerberos section mentioning that the mpack supports Ambari's Kerberization process.

The client process managed to ensure keytabs existed and that the client_jaas was created appropriately.

dlyle65535:
I've had 2 successful runs:
1) Full dev + kerberos.
2) EC2 will full stack installed prior to enabling kerberos.

Both ran without error.

merrimanr:
I validated this on full dev.  Ran a vagrant up, followed the instructions in Kerberos-setup.md, and verified Bro data being written to the index.

+1.  Nice job.

james-sirota:
+1 I tested it on AWS and was able to get it to work. Great job!

PR #525, opened 2017-04-12
METRON-845 Increment build number to 0.4.0

mattf-horton:
Simply replaced all effective Metron version strings, from "0.3.2" to "0.4.0".
Used `mvn versions:set -DnewVersion=0.4.0` first, then did
```
for f in `grep -r -l "0\.3\.1" * | egrep -v '^site/'` ; do  echo " "; echo $f; grep "0\.3\.1" $f; done
```
to catch the non-maven files, and hand-edited them.
The remaining handful of .md and log files containing "0.3.2" are not effective Metron versions, but recordings of previous uses, so they are not changed.

NOTE: The new branch has not propagated to github, presumably because only commits stimulate the propagation, and a new branch copied from existing master is not technically a commit.  So we are using this low-risk commit to cause that propagation, even though it needs a few more hours to run full tests.

### For all changes:

### For code changes:
- [NA] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [NA] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] (IN PROCESS) Have you ensured that the full suite of tests and checks have been executed in the root incubating-metron folder via:
  ```
  ```

- [x] (IN PROCESS) Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

- [NA] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book?

### DIFF
--- a/metron-analytics/metron-maas-common/pom.xml
+++ b/metron-analytics/metron-maas-common/pom.xml
@@ -18,7 +18,7 @@
   <parent>
     <groupId>org.apache.metron</groupId>
     <artifactId>metron-analytics</artifactId>
-    <version>0.3.1</version>
+    <version>0.4.0</version>
   </parent>
   <artifactId>metron-maas-common</artifactId>
   <properties>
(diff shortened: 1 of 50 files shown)


james-sirota:
+ 1

PR #542, opened 2017-04-21
METRON-873: Stellar string literals do not support quote escaping

cestella:
## Contributor Comments
Right now, in stellar, we cannot represent a string literal that contains `'foo'` if the string is quoted with `'` or `"foo"` if the string is quoted with `"`.  This is unfortunate and should be corrected.

To test this out, start up the stellar REPL in fulldev *or* run it locally by running `mvn exec:java -Dexec.mainClass="org.apache.metron.common.stellar.shell.StellarShell"` from `metron-platform/metron-common` and try the following strings:
* `'\'foo\''` should yield `'foo'`
* `"\"foo\""` should yield `"foo"`
* `TO_UPPER('\'foo\'')` should yield `'FOO'`
* `TO_UPPER("\"foo\"")` should yield `"FOO"`

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  bin/generate-md.sh
  mvn site:site
  ```

#### Note:


justinleet:
What if I want to put a backslash character in my string?  e.g. I want to produce 'some \ string'.

I don't think this worked before, but it's particularly relevant because now we're doing stuff with backslash.  In particular, I don't think this worked before because (and correct me if I'm wrong)

```
fragment SCHAR:  ~['"\\\r\n];
```
This says that strings are NOT:
* Single Quote
* Double Quote
* Backslash
* Carriage Return
* Newline

Even escaping won't help this, because the backslash is outright disallowed in strings.  Is there a reason it's not allowe?

cestella:
We're not using SCHAR anymore.  The current commit should address the backslash issue.  Give it a try and let me know what you think. :)

justinleet:
```
/code [Stellar]>>> " bar \\\\ foo "
 bar \\\ foo
```
Shouldn't that produce 'bar \\ foo' (One less backslash)?  It seems like it's not handling chained backslash escapes properly (or at least the way Java does, which I assume is the template).

cestella:
Yes, you're right.  Take a look now.  I also added support for \n, \t and \r while I was there.

ottobackwards:
We need a stellar guide, more suited to learning stellar, than just documenting the facts of it.  Something that gets you in the shell and gives you some exercises.

cestella:
@ottobackwards you're totally right.  We desperately are in need of better stellar documentation:
* A language reference
* A set of introductory lessons in Stellar

metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/StellarCompiler.java
```diff
@@ -263,7 +263,56 @@ public void exitVariable(StellarParser.VariableContext ctx) {
 
   @Override
   public void exitStringLiteral(StellarParser.StringLiteralContext ctx) {
-    expression.tokenDeque.push(new Token<>(ctx.getText().substring(1, ctx.getText().length() - 1), String.class));
+    String rawToken = ctx.getText();
+    String literal = unquoteLiteral(rawToken);
+
+    expression.tokenDeque.push(new Token<>(literal, String.class));
+  }
+
```
ottobackwards: While I don't see a problem with this code, it does make me wonder if we are re-inventing the wheel here.  Isn't there some standard lib for doing this?

ottobackwards: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringEscapeUtils.html ?

justinleet:
+1.  Thanks for working on this.

metron-platform/metron-common/src/main/java/org/apache/metron/common/stellar/StellarCompiler.java
cestella: hmm, let me try, good catch.

cestella: Dude!  Great catch!  We're actually closer to JSON escaping because our strings can start with either `'` or `"`, so I used that escaping strategy and it worked just fine.  Thanks!

PR #538, opened 2017-04-20
METRON-868 Fix documentation on building RPMs

justinleet:
## Contributor Comments
Update doc to run install instead of package as a prereq to the RPM build.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron (Incubating).  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via `site-book/target/site/index.html`:

  ```
  cd site-book
  bin/generate-md.sh
  mvn site:site
  ```

#### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.
It is also recommended that [travis-ci](https://travis-ci.org) is set up for your personal repository such that your branches are built there before submitting a pull request.

PR #544, opened 2017-04-24
METRON-866 Quickdev and Fulldev readme misses Docker as prerequisite

nickwallen:
Updates to the README instructions for both "Quick Dev" and "Full Dev".  This adds "Docker" as a prerequisite, provides simple linkage/instructions, and also prompts the user to ensure that the Docker Service is running before trying to deploy Metron.

## Pull Request Checklist

  ```
  cd site-book
  bin/generate-md.sh
  mvn site:site
  ```

### DIFF
--- a/metron-deployment/vagrant/full-dev-platform/README.md
+++ b/metron-deployment/vagrant/full-dev-platform/README.md
@@ -37,13 +37,6 @@ Any platform that supports these tools is suitable, but the following instructio
 
 ### Deploy Metron
 
-1. Build Metron
-
-    ```
-    cd incubator-metron
-    mvn clean install -DskipTests
-    ```
-
 1. Ensure that the Docker service is running.
 
 1. Deploy Metron
--- a/metron-deployment/vagrant/full-dev-platform/README.md
+++ b/metron-deployment/vagrant/full-dev-platform/README.md
@@ -13,7 +13,9 @@ Getting Started
 The computer used to deploy Apache Metron will need to have the following components installed.
 
  - [Ansible](https://github.com/ansible/ansible) (2.0.0.2 or 2.2.2.0)
+ - [Docker](https://www.docker.com/community-edition)
  - [Vagrant](https://www.vagrantup.com) 1.8.1
+ - [Vagrant Hostmanager Plugin](https://github.com/devopsgroup-io/vagrant-hostmanager) `vagrant plugin install vagrant-hostmanager`
  - [Virtualbox](https://virtualbox.org) 5.0.16
  - Python 2.7.11
  - Maven 3.3.9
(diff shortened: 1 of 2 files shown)


metron-deployment/vagrant/full-dev-platform/README.md
```diff
@@ -24,35 +26,38 @@ Any platform that supports these tools is suitable, but the following instructio
...
-1. Install Vagrant Hostmanager.
+1. Build Metron
 
-  ```
-  vagrant plugin install vagrant-hostmanager
-  ```
+    ```
+    cd incubator-metron
+    mvn clean install -DskipTests
+    ```
```
dlyle65535: Metron is built during the provisioning process. There is a defect that makes a single "mvn install" call per release number required. See discussion [here](https://github.com/apache/incubator-metron/pull/538). My hope is the defect will be removed shortly. In the meantime, what would you think of either omitting this or adding some language that makes it clear that the manual install only needs to be called once per release until 538 is resolved (i.e. if we bumped the rev from 0.4.0, install would need to be called again)? Otherwise, following these instructions will add around 13 minutes to each build for no gain.

nickwallen: My goal was to get something into master that works right now.  I've seen a number of users run into problems. If things change per #538, then I would suggest we update these READMEs as part of that PR to remove any unnecessary steps.  Do you think that would work @dlyle65535 ?

dlyle65535: Yeah, completely understand. I don't like the idea of people running the build 2 times when it's absolutely not required. This is the 3rd PR we've introduced to "fix" the documentation without addressing the actual issue, which should be trivial- simply roll back METRON-827 which (I hope) is actually in work and will likely land before this does. @justinleet - is that a correct assumption?

justinleet: Yeah, it's a correct assumption.  It should be set to go pending possibly some docs tweaks.

nickwallen: So for "Quick Dev" it already had the "Build Metron" step.  All I did was change from 'package' to 'install'.  For "Full Dev" I just copied the same Build Step to the README as Quick Dev had.  What do I need to do to disentangle this PR from #538 ?  I'd rather not wait days to get such a simple fix into the README that ends up confusing a lot of users.

dlyle65535: Leaving the build step in quick dev must have been my error. It doesn't belong there. I think good omitting them both. Looks like 538 is on the way to closure.

nickwallen:
Done. Please take a look.

dlyle65535:
+1 Thanks for the backstop, Nick. I can't believe I left that doggone thing there.

metron-deployment/vagrant/full-dev-platform/README.md
```diff
@@ -13,7 +13,9 @@ Getting Started
 The computer used to deploy Apache Metron will need to have the following components installed.
 
  - [Ansible](https://github.com/ansible/ansible) (2.0.0.2 or 2.2.2.0)
+ - [Docker](https://www.docker.com/community-edition)
  - [Vagrant](https://www.vagrantup.com) 1.8.1
+ - [Vagrant Hostmanager Plugin](https://github.com/devopsgroup-io/vagrant-hostmanager) `vagrant plugin install vagrant-hostmanager`
```
simonellistonball: Is this pre-req or optional? (admittedly very nice optional, but some people might not like something that tramples their hosts file)

nickwallen: Required, at this point.  References that we make like `http://node1:8080` will not work without the Hostmanager plugin.  It just updates the host OS's 'hosts' file.  

In this PR, I moved the Hostmanager Plugin reference from the "Deploy Metron" section to the "Preerequisite" section.  I think this makes more sense as it is not something that you need to do everytime that you deploy Metron, just once; like a prereq.

nickwallen: But do agree on your point about it trampling the hosts file.  Definitely something we could look at doing differently at some point.

ottobackwards: I don't think anyone has complained

mmiklavc: I don't have a strong opinion on this, but I think the impact is minimal. As Nick stated, it's kind of nice to be able to have the deployment automatically setup a host name that we can use to interact with the env. I believe in our usage it should only append to /etc/hosts. And when you do `vagrant destroy` it will update the file by removing only the line wrapped/tagged with the corresponding ID, e.g.
```
## vagrant-hostmanager-start id: eded1905-6263-4699-8d30-45e30f693db6
192.168.66.121  node1

## vagrant-hostmanager-end
```

PR #556, opened 2017-04-28
METRON-903: Create a connections report in Zeppelin

justinleet:
## Contributor Comments
The added Zeppelin dashboard is pretty straightforward.

For each of the sensors (Bro/Snort/Yaf) and for all three as a whole, groups everything by (ip_src_addr,ip_dst_addr) for a given range, counts it all up, and spits out tables.

The range is provided by two fields, start and end, which are dates that can be formatted by a user provided formatting string.  e.g.
```
2017-04-28 14:19:19 -> yyyy-MM-dd HH:mm:ss
2017-04-28 -> yyyy-MM-dd
```

## Test Plan
To test, spin up full-dev. Ensure data flows through for the sensors.  For Yaf, which is not enabled by default, it'll be necessary to start the sensor-stub
```
service sensor-stubs start yaf
```

It'll also be necessary to add Yaf to the list of sensors run in Ambari. To do so, stop Metron, edit "Metron Parsers" to include "yaf" (or if the other sensors already have data output, feel free to just make it "yaf" only).

Let data flow through for a little while (enough so that we can reasonable make range filters).

Once some data has gone through, we'll need to have an instance of Zeppelin.  Because of the size of the Vagrant instance, we'll want to shut down unneeded services.  Shutdown Metron, Kibana, Storm, Kafka, and HBase.

Install Zeppelin from "Actions - Add Service".  It'll prompt you to install Spark and Hive, do so.  Configuration is pretty trivial, all that's necessary is to set an arbitrary Hive database password.  Let this run.  The Hive service check likes to fail on our Vagrant, but it's benign (some impersonation configuration issue unrelated to actually running our queries).  Ignore it and accept the installation.

From Metron's "Service Actions", run the "Zeppelin Notebook Import", to load our notebooks into Zeppelin.  Use the quick links to navigate to the Zeppelin UI.

Go into the "Metron - Connection Report" notebook.

Each service (and the aggregate) will have something similar to the following:
screen shot 2017-04-28 at 11 53 17 am

In my case, note the second row:
```
192.168.66.1	192.168.66.121	439
```

We want to validate this range, so we can check ES and see what a count against the same range returns
```
curl -XPOST 'node1:9200/y*/_count?pretty' -H 'Content-Type: application/json' -d'
{
   "query": {
      "filtered": {
         "filter": {
            "and": [
               {
                  "bool": {
                     "must": [
                        {
                           "term": {
                              "ip_src_addr": "192.168.66.1"
                           }
                        },
                        {
                           "term": {
                              "ip_dst_addr": "192.168.66.121"
                           }
                        }
                     ]
                  }
               },
               {
                  "range": {
                     "timestamp": {
                        "gte": 1493389159000,
                        "lt": 1493389859000
                     }
                  }
               }
            ]
         }
      }
   }
}
'
```
The result is:
```
{
  "count" : 439,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "failed" : 0
  }
}
```
They're the same, so we're good.

Additionally, the dates and formatting can be changed and run as well.

This exercise can be repeated across the sensors.
## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

#### Note:


dlyle65535:
Looks good, one question- why the UNION ALL (vs UNION) in the Top Connections paragraph? Will that overcount?

justinleet:
Since we're counting number of connections, we don't want to remove any dupes between tables.  Each data source can independently have the same ip -> ip connection any number of times, so it should be preserved in the count.

dlyle65535:
Wouldn't a single connection (e.g. looking up dns entry) create an event on more than one sensor?

justinleet:
Hmm, possibly?  I don't really know enough about what happens overlap there is in the sensors.  That's a good point though.

The problem is if we just do a UNION, any dupes get removed (and correct me if I'm wrong there, I may not remember the semantics correctly) making our counts wrong regardless.  If that's true, there's not a great way to do the total connections (unless we know the exact circumstances that lead to dupes).  In that case, we should probably just remove it entirely and leave it by sensor.

merrimanr:
Worked as advertised.  +1 from me assuming we work out the UNION ALL vs UNION requirement.

dlyle65535:
tl;dr - I'm +1 with the combined report, just wanted to understand the trade space.

You have it right. I think the combined view is useful, but we probably should do some noodling around what constitutes a unique event. The way I think about is this (though I may be all wet).

If we have sufficient clock granularity, src/dst/time is unique.
If we don't then UNION ALL runs a risk of overcounting but UNION runs the risk of undercounting.

Either way, I'm good and my default position would be to overcount.

justinleet:
I feel like sorting out uniqueness might be a little much to bite off here, just because it's likely other components in the stack might care about this type of analysis and I don't really want to set a precedent for a dashboard.

Having said that, I'm inclined to add a note to the description indicating that totals can be subject to double counting for events that trigger multiple sensors at once.  Is this a reasonable compromise for this ticket?  If we're good with that, I'm happy to go ahead and make that change.

dlyle65535:
I'm good with or without it. I wasn't suggesting we do said noodling as part of this effort.

justinleet:
Sorry, I wasn't trying to imply you were; just explaining my thought process on it a bit and offer that possibility up for you and Ryan.  I should have said "I agree that" instead of "I feel that".

@merrimanr Do you have a preference on this?

james-sirota:
This dashboard is designed to look at flows and we absolutely do want to count every flow.  So UnionALL is the correct implementation here

james-sirota:
Just spun up the dashboard.  The YAF portion is correct.  The YAF sensor produces flow information from A to B and you produce a table that counts and orders them.  Bro is almost correct.  Bro produces HTTP and DNS metadata for us.  So you need to differentiate by protocol.  In case of Bro-DNS your dashboard should say top DNS requests.  In case of Bro-HTTP your dashboard should say top HTTP requests.  Snort is an IDS that produces alerts.  So your dashboard should say top number of generated alerts.  

One thing that would be useful here is to add a second dashboard for YAF that would give me a  histogram for a connection of my choice.  For example, your top connection is 192.168.66.1->192.168.66.121 | 867.  If I could have a second follow-up dashboard where i could have 4 boxes: sourceIP, destIP, TimeBinSize, HowFarToReachBack.  An example use would be that I would take the top connection and do the following: SourceIP= 192.168.66.1, DestIP=192.168.66.121, TimeBinSize=5 mins, HowFarToReachBack=3hours.  Then it draws a histogram of the top connection for me in 5 min increments reaching back 3 hours.

justinleet:
@james-sirota I'll go ahead and split out the Bro data by protocol and adjust the terminology for the Snort. Do we want to do the Yaf histogram as part of this ticket, or do we want to push that to another Jira and have it as a follow-on activity? I haven't done it before, but I'd expect it to be fairly easy to do.

james-sirota:
@justinleet the top-level cumulative dashboard would not be useful because you are combining flows with DPI with Alerts data.  With respect to the histogram I would do it as a part of this PR because it seems like a quick and easy taslk

justinleet:
Updated with the Bro split, removing aggregate paragraphs, and Snort verbiage.  Working on histogram.

james-sirota:
Excellent.  I see the HTML table and it works great. This is very well done.  +1

justinleet:
@james-sirota Looking at the way this is setup, this seems like the histogram dashboard should be able to be time windowed the same way the other dashboards are (i.e. start/end/date_format), rather than just a straight lookback.  If I'm looking at flows from 3 days ago to 1 day ago, I probably want to be able to select that same range and then bucket on it.

I assume we also want this functionality for the Bro and Snort portions?  It's pretty easy to take what's setup for Yaf, and just apply it to each source.

justinleet:
screen shot 2017-05-02 at 7 25 30 am
 With the bracketed time windows following the other panes.

justinleet:
I went ahead and added histograms for all sources, since I had a free couple minutes.

At this point, histogram input is pretty close to the top conns/requests/alerts setups.  It would be really nice if there were a way to set the timestamps at the top and just carry them through globally.  I'm not sure if that exists, but unless we have a high need to dig into it, I'd be inclined to consider it a follow-on activity.  Let me know if anyone disagrees with that.

james-sirota:
Thanks for fixing the cumulative report. The histogram looks great +1

PR #565, opened 2017-05-04
METRON-931: Stellar REDUCE incorrectly returns null for fewer than 3 items in list

mmiklavc:
https://issues.apache.org/jira/browse/METRON-931

Reduce was checking input.size() instead of args.size(). The input var is the first arg to the reduce function, so it was returning null if fewer than 3 items existed in the list. I've corrected the function to return null if the number of arguments to the function itself is fewer than the expected 3.

A simple set of unit tests has been added. This can also be manually tested by starting the Stellar REPL and testing that the following calls do not return null:
```
h1 := REDUCE(['foo', 'bar'], (s, x) -> HLLP_ADD(s, x), HLLP_INIT(5, 6))
s1 := REDUCE([1,2], (s, x) -> STATS_ADD(s, x), STATS_INIT())
```

### For all changes:

### For code changes:
  ```
  ```


nickwallen:
+1 Good catch

PR #578, opened 2017-05-10
METRON-946 Full/Quick dev should default network_host to [ _local_, _site_ ]

mattf-horton:
## Contributor Comments
@dlyle65535 observed that even single-node installs should have ES bind an external interface (as well as loopback) so that external clients can access ES reports. Based on that, FullDev assumes "node1" is bound to an external i/f, which didn't work with the latest. This patch will fix the problem.

Dave also recommends as part of this to do a cleanup step: removing metron-deployment/inventory/devimage-vagrant/ and all its contents. It is obsolete and unused, and might be confusing to developers looking for examples to follow.

To test, install FullDev and see that the ES config is good.  
Also install single node manually and check ES config still good.
There shouldn't be any other observable changes.

### For all changes:

### For code changes:
- [NA] Have you ensured that the full suite of tests and checks have been executed in the root incubating-metron folder via:
  ```
  ```


dlyle65535:
+1, ran it up in full dev. Everything worked as expected. Thanks for the quick turn, @mattf-horton.

mattf-horton:
and thanks for the quick check, @dlyle65535 .  In process to commit...

PR #584, opened 2017-05-12
METRON-950: Migrate storm-kafka-client to 1.1

cestella:
## Contributor Comments
There are MAJOR performance issues with the storm-kafka-client. Throughput is roughly an order of magnitude faster in the 1.1.0 version vs the 1.0.3.
This also removes the requirement for the non-HDP profile to rely on the HDP repo.

I will point out that the storm-kafka-client in 1.1 does rely on `Time.nanoTime()`, a method call that was introduced in 1.1.  To get around this, I've created the `metron-storm-kafka-override` with a backported implementation of the `Time` class.  This will be removed when we migrate to Storm 1.1.

Normally, I would hold off this change until we take the effort to migrate to 1.1 properly, but the combination of empirical evidence (I saw it go an order of magnitude slower in 1.0.1 vs 1.1) and wide ranging change for Storm 1.1 makes me err on the side of getting this in faster.  It should help make every topology faster, but is in particular absolutely required to get sensible throughput in the pcap topology.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-storm-kafka-override/src/main/java/org/apache/storm/kafka/spout/internal/Timer.java
+++ b/metron-platform/metron-storm-kafka-override/src/main/java/org/apache/storm/kafka/spout/internal/Timer.java
@@ -1,0 +1,58 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.storm.kafka.spout.internal;
+
+import java.util.concurrent.TimeUnit;
+import org.apache.storm.utils.Time;
+
+public class Timer {
+  private final long delay;
+  private final long period;
+  private final TimeUnit timeUnit;
+  private final long periodNanos;
+  private long start;
+
+  public Timer(long delay, long period, TimeUnit timeUnit) {
+    this.delay = delay;
+    this.period = period;
+    this.timeUnit = timeUnit;
+    this.periodNanos = timeUnit.toNanos(period);
+    this.start = System.nanoTime() + timeUnit.toNanos(delay);
+  }
+
+  public long period() {
+    return this.period;
+  }
+
+  public long delay() {
+    return this.delay;
+  }
+
+  public TimeUnit getTimeUnit() {
+    return this.timeUnit;
+  }
+
+  public boolean isExpiredResetOnTrue() {
+    boolean expired = System.nanoTime() - this.start >= this.periodNanos;
+    if(expired) {
+      this.start = System.nanoTime();
+    }
+
+    return expired;
+  }
+}
--- a/metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
+++ b/metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
@@ -20,8 +20,10 @@ package org.apache.metron.storm.kafka.flux;
 
 import com.google.common.base.Joiner;
 import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.apache.kafka.common.serialization.Deserializer;
 import org.apache.metron.common.utils.KafkaUtils;
 import org.apache.storm.kafka.spout.*;
 import org.apache.storm.spout.SpoutOutputCollector;
@@ -30,10 +32,7 @@ import org.apache.storm.topology.OutputFieldsGetter;
 import org.apache.storm.tuple.Fields;
 import org.apache.storm.tuple.Values;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.function.Function;
 
 /**
@@ -113,11 +112,12 @@ public class SimpleStormKafkaBuilder<K, V> extends KafkaSpoutConfig.Builder<K, V
    * @param <K> The key type in kafka
    * @param <V> The value type in kafka
    */
-  public static class TupleBuilder<K, V> extends KafkaSpoutTupleBuilder<K,V> {
+  public static class SpoutRecordTranslator<K, V> implements RecordTranslator<K,V> {
     private List<FieldsConfiguration> configurations;
-    private TupleBuilder(String topic, List<FieldsConfiguration> configurations) {
-      super(topic);
+    private Fields fields;
+    private SpoutRecordTranslator(List<FieldsConfiguration> configurations) {
       this.configurations = configurations;
+      this.fields = FieldsConfiguration.getFields(configurations);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


cestella:
The test plan for this should be mainly a smoke test..make sure data flows through the default topologies in full dev.

metron-platform/metron-storm-kafka-override/src/main/java/org/apache/storm/kafka/spout/internal/Timer.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public long delay() {
+    return this.delay;
+  }
+
+  public TimeUnit getTimeUnit() {
+    return this.timeUnit;
+  }
+
+  public boolean isExpiredResetOnTrue() {
```
justinleet: I'm not sure how complicated the original implementation is vs this one.  Could we grab any unit tests on the original Timer and make sure that this Timer meets the same expectations?

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
```diff
@@ -202,31 +240,4 @@ public String getTopic() {
...
-
-  private static <K,V> KafkaSpoutTuplesBuilder<K, V> createTuplesBuilder(List<String> config, String topic) {
-    TupleBuilder<K, V> tb =  new TupleBuilder<K, V>(topic, FieldsConfiguration.toList(config));
-    return new KafkaSpoutTuplesBuilderNamedTopics.Builder<>(tb).build();
-  }
-
-
-  private static KafkaSpoutStreams createStreams(List<String> config, String topic) {
-    final Fields fields = FieldsConfiguration.getFields(FieldsConfiguration.toList(config));
-    return new KafkaSpoutStreamsNamedTopics.Builder(fields, STREAM, new String[] { topic} ).build();
```
justinleet: STREAM isn't used anymore, right? Can we kill the field (and no, I can't comment on the field because it's not technically part of the diff, sigh).

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
```diff
@@ -165,13 +177,39 @@ public SimpleStormKafkaBuilder( Map<String, Object> kafkaProps
...
+         , createDeserializer(Optional.ofNullable((String)kafkaProps.get(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG)), DEFAULT_DESERIALIZER)
+         , createDeserializer(Optional.ofNullable((String)kafkaProps.get(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG)), DEFAULT_DESERIALIZER)
+         , topic
+    );
+    setProp(kafkaProps);
+    setRecordTranslator(new SpoutRecordTranslator<>(FieldsConfiguration.toList(fieldsConfiguration)));
     this.topic = topic;
   }
 
+  private static <T> Class<Deserializer<T>> createDeserializer( Optional<String> deserializerClass
```
justinleet: This is nitpicky, so don't feel obligated to fix it, but couldn't this method just take a (possibly null) String, and wrap it in the Optional itself?  It seems pointless to make the caller wrap it when it's just as easy for the function itself to handle null.

pom.xml
```diff
@@ -106,7 +122,7 @@
              but I justify it by noting that this should be able to be removed when we migrate to
              Storm 1.1.x, which properly supports Kafka 0.10.x.
           -->
-        <global_storm_kafka_version>1.0.1.2.5.0.0-1245</global_storm_kafka_version>
+        <global_storm_kafka_version>1.1.0</global_storm_kafka_version>
```
justinleet: Couldn't the whole block above this be dropped/edited?  It's pure HDP profile now, right?  Or better yet, replaced with an explanation of the new 1.1.0 + Timer explanation.

cestella: yeah, good idea

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
cestella: I'm not super keen on passing in possibly null values around.  I rather like the notion of passing in that something is optional in the type signature.

metron-platform/metron-storm-kafka-override/src/main/java/org/apache/storm/kafka/spout/internal/Timer.java
cestella: That's a good catch.  It'd be nice to have that tested.

mmiklavc:
Just to clarify for others looking at this, there is a `Timer` class in `storm-kafka-client-1.1.0` which depends on a `Time` class (no 'r') in storm-core. In 1.0.3, the main version of Storm we depend on, there is no `nanoTime()` method. This was introduced in 1.1. @cestella is handling this via the temporary module `metron-storm-kafka-override` via `System.nanoTime()`.

We're only recommending upgrading the client currently because, first and foremost, it provides the desired performance improvements, and second the impact is significantly less than upgrading the whole of Storm.

mmiklavc:
@cestella How are we guaranteeing our version of Timer is included over the 1.1 version? Is this purely dependency ordering with the Maven Shade plugin?

cestella:
@mmiklavc it is.  To give guarantees beyond that, would require us to shade and relocate the `metron-kafka-spout` project, which I did not want to do without some good justification.

metron-platform/metron-storm-kafka/src/main/java/org/apache/metron/storm/kafka/flux/SimpleStormKafkaBuilder.java
cestella: Yep

mmiklavc:
OK, works for me

metron-platform/metron-storm-kafka-override/src/main/java/org/apache/storm/kafka/spout/internal/Timer.java
cestella: It is precisely the 1.1.0 version of the code with the exception of calling `System.nanoTime` instead of `Time.nanoTime` (see [here](https://github.com/apache/storm/blob/a4afacd9617d620f50cf026fc599821f7ac25c79/external/storm-kafka-client/src/main/java/org/apache/storm/kafka/spout/internal/Timer.java) ).

It should be noted that `Time.nanoTime` calls `System.nanoTime` (see [here](https://github.com/apache/storm/blob/a4afacd9617d620f50cf026fc599821f7ac25c79/storm-client/src/jvm/org/apache/storm/utils/Time.java#L162) ) in a production scenario (i.e. non-simulation).

cestella: Ok, done, added test.

mmiklavc:
When I spin this up in full dev, half the records don't output with the pcap_inspector. I'm running:
`pycapa --producer --interface eth0 --kafka-broker node1:6667 --kafka-topic pcap --max-packets 10`

And I get 10 records, as expected. 

/usr/metron/0.4.0/bin/pcap_inspector.sh -i /apps/metron/pcap/pcap_pcap_1494886095377022000_0_pcap-7-1494887672
```
TS: May 15, 2017 10:08:15 PM UTC
TS: May 15, 2017 10:08:17 PM UTC
TS: May 15, 2017 10:08:17 PM UTC
TS: May 15, 2017 10:08:20 PM UTC
TS: May 15, 2017 10:08:20 PM UTC
TS: May 15, 2017 10:08:22 PM UTC,ip_src_addr: 10.0.2.2,ip_src_port: 58390,ip_dst_addr: 10.0.2.15,ip_dst_port: 22,protocol: 6
TS: May 15, 2017 10:08:22 PM UTC,ip_src_addr: 10.0.2.15,ip_src_port: 22,ip_dst_addr: 10.0.2.2,ip_dst_port: 58390,protocol: 6
TS: May 15, 2017 10:08:25 PM UTC,ip_src_addr: 10.0.2.2,ip_src_port: 58390,ip_dst_addr: 10.0.2.15,ip_dst_port: 22,protocol: 6
TS: May 15, 2017 10:08:25 PM UTC,ip_src_addr: 10.0.2.15,ip_src_port: 22,ip_dst_addr: 10.0.2.2,ip_dst_port: 58390,protocol: 6
TS: May 15, 2017 10:08:27 PM UTC,ip_src_addr: 10.0.2.2,ip_src_port: 58390,ip_dst_addr: 10.0.2.15,ip_dst_port: 22,protocol: 6
```
Wireshark seems to be OK.
image

I'm looking into this and will update as I have more info.

nickwallen:
Does `pcap_inspector` only print out for TCP packets?  The first 5 of those are all UDP based (NTP & DNS).

mmiklavc:
@nickwallen Yeah, that's where I landed. Looking in our pcap helper class, I see the following.
```
  public static MetronEthernetDecoder createDecoder() {
    MetronEthernetDecoder ethernetDecoder = new MetronEthernetDecoder();
    IpDecoder ipDecoder = new IpDecoder();
    ethernetDecoder.register(EthernetType.IPV4, ipDecoder);
    return ethernetDecoder;
  }
```

ctramnitz:
Does this change have any impact on using Storm 1.1 (i.e. from HDP 2.6)?

justinleet:
I ran up the main topologies last night and everything seemed to correlate correctly and ran without issue.  Hadn't run up pcap yet, but it look like @mmiklavc is working on validating it.  Given that it's the same spout throughout, I'm not convinced it's necessary to retread it.  Anybody disagree?

cestella:
@ctramnitz While we don't really test on 1.1 yet, this shouldn't have an impact on the topologies running there.

mmiklavc:
+1. Data flowed through to ES as expected, and I was able to spin up pcap. Along with @justinleet having tested the other topologies, I'm happy with the results. Great work @cestella!

justinleet:
Same here, +1.  Thanks a lot for the contribution, this makes a big difference.

PR #587, opened 2017-05-15
METRON-952 Travis CI Link in README Pointing to Old Repo

nickwallen:
The Travis CI status link in the README is pointing to the old apache/incubator-metron repository.  I updated this point to apache/metron.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/apache/incubator-metron.svg?branch=master)](https://travis-ci.org/apache/incubator-metron)
+[![Build Status](https://travis-ci.org/apache/metron.svg?branch=master)](https://travis-ci.org/apache/metron)
  
 # Apache Metron
 


cestella:
+1 by inspection

justinleet:
I actually also noticed this last night, and was about to put in a patch until I realized there's a good amount of documentation in general that needs to point to the newer repos, and our docs and so on shouldn't be referring to incubator-metron dir anymore and so on.

I'm fine with doing this separately (and am +1 by inspection, since I did this exercise last night), but it seems like we probably should do the rest of it also now that we're TLP and everything is setup.

cestella:
agreed; there is more to this than just this PR.

justinleet:
I'll create a ticket and take care of it.  I have a lot of this already done, while I was looking through last night.  I was hoping to get a comprehensive ticket out with this change, but I'm not really worried about having this separate.

PR #600, opened 2017-05-31
METRON-976 KafkaUtils doesn't handle SASL_PLAINTEXT

justinleet:
## Contributor Comments
Used @cestella's original implementation before I apparently foolishly assumed we'd get Java compliant URIs.  See: https://github.com/apache/metron/pull/486/commits/5d36c79effec75f6ac95fa587f80da0bd5420135

Unit testing is updated, and now checks that SASL_PLAINTEXT works (not just PLAINTEXTSASL).  ~I still need to spin up full dev.~

@cestella Let me know how you'd like to handle credit, given that you originally wrote it.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

#### Note:


cestella:
I will happily assign over credit to you.  Thanks for catching the bug @justinleet

justinleet:
Ran up in full dev, kerberized, stopped metron, changed configs in Kafka referencing PLAINTEXT/PLAINTEXT_SASL to SASL_PLAINTEXT (listeners and security.inter.broker.protocol), and restarted kafka and metron.  Ensured data still ran through to ES successfully.

metron-platform/metron-common/src/main/java/org/apache/metron/common/utils/KafkaUtils.java
```diff
@@ -68,12 +68,14 @@
...
+  public List<String> fromEndpoint(String url){
     List<String> ret = new ArrayList<>();
     if(url != null) {
-      URI uri = new URI(url);
-      int port = uri.getPort();
-      ret.add(uri.getHost() + ((port > 0)?(":" + port):""));
+      Iterable<String> splits = Splitter.on("//").split(url);
+      if(Iterables.size(splits) == 2) {
+        String hostPort = Iterables.getLast(splits);
+        ret.add(hostPort);
```
mattf-horton: Hi @justinleet , perhaps it is because I don't know how SASL_PLAINTEXT endpoints work, but I don't get the change above. I fully understand the deleted lines, and don't see how the substituted lines are equivalent, or what was wrong with the former.  Would you mind giving a bit of explanation?  Thanks.

Also, I just went back and re-read your introduction to this PR, and followed the link to (5d36c79)[https://github.com/apache/metron/commit/5d36c79effec75f6ac95fa587f80da0bd5420135].  It appears you've exactly reversed a chunk of @cestella 's patch of Mar 23.  Was this intended?  I'm even more confused :-)

justinleet: The problem is that the URI based solution that was implemented in that patch (I believe based on my suggestion, sigh) doesn't allow for protocols with underscore characters.  It's entirely independent of SASL_PLAINTEXT, other than the fact that underscore is in it.

The intention was to exactly reverse the chunk you mention, because the original didn't care about underscores.

On further inspection (which I'm glad you did!), these are not equivalent before and after.  Which begs the question of why things worked in testing.  I'll dig in a bit more to figure out what's going on under the hood here.

mattf-horton: Thanks, @justinleet , and sorry if I'm being a PITA.  By inspection, I have the following concerns:

1. The URI solution always delivers exactly "host[:port]" (or fails). The split solution delivers the entire url except the protocol, ie "host[:port]/path/extension/if/any...", unless there's something upstream cutting off the path extension (if any).  Is there?  I didn't see anything, it looks like endpoint urls are simply read raw from a config JSON structure.  So I would have expected trimming the proposed split at both ends.  Unless endpoint URLs are supposed to allow path extensions, in which case the URI approach is clearly wrong.

2. The split code only returns a result if the split size is == 2.  This prevents it from accepting:
  * Endpoint urls without the protocol prefix
  * Endpoint urls with path extensions that accidentally have double slashes in them (a common problem with machine-generated URLs, and accepted by most URL path processors)
  It also does not provide any log feedback on why it silently ignores "bad" endpoint urls.

None of these issues would show up in testing unless the unit tests include thorough negative testcases for invalid configured endpoint URLs.

justinleet: @mattf-horton You're absolutely not being a PITA.  Honestly, I tried to get a solution to a known problem out a little too quickly because it was actually experienced, so I'm actually pretty happy you're double checking me on this so something half baked didn't go in.

So looking back at this a bit more fresh, this only gets called from `getBrokersFromZookeeper`.  The fact that's it's public isn't reflective of it's use (I'd probably argue package private, so it can be unit tested, or just better unit tested in general). 

In practice, these aren't general URLs.  They're specifically from https://cwiki.apache.org/confluence/display/KAFKA/Kafka+data+structures+in+Zookeeper
e.g.
```
  "endpoints": ["PLAINTEXT://host1:9092", "SSL://host1:9093"]
```

So it's exactly protocol://host[:port] (to the best of my knowledge).  So while the URI solution is probably technically better (outside of the underscore issue), fixing the split should be perfectly fine, if a bit kludgier.

Looking back at it, I'm pretty sure I didn't hit the right code path to trigger it in testing.  I'm going to adjust the unit tests to actually do something more useful, so we avoid these issues in the future. I'm also going to add a comment, so it's very explicit what the actual assumptions made by that function are.

justinleet:
@mattf-horton Updated with a more explicit set of unit tests around it and a comment explaining the assumption made by the function.  Let me know if there's anything else you want or need to make sure everything's both correct and easily understandable.

mattf-horton:
+1 by inspection.  Thanks for expanding the unit test to be more complete and understandable.

PR #602, opened 2017-06-01
METRON-906: Rest service storm configuration does not allow for proper URLs

justinleet:
## Contributor Comments
Allows for protocol to be used in the Storm REST URL (e.g. http:// and https://).  Prior to this, http:// was automatically prepended.  In the event that no protocol is given, http:// will be prepended for backwards compatibility.

Unit tests are added to ensure that the protocol is handled properly. Full dev was spun up to ensure that things still functioned as expected, with the protocol provided.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root incubating-metron folder via:
  ```
  mvn -q clean integration-test install && build_utils/verify_licenses.sh 
  ```

- [x] Have you written or updated unit tests and or integration tests to verify your changes?
- [x] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

#### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.
It is also recommended that [travis-ci](https://travis-ci.org) is set up for your personal repository such that your branches are built there before submitting a pull request.

PR #577, opened 2017-05-09
METRON-746: Build Custom Checkstyle and IDE formatting settings

justinleet:
## Contributor Comments

Pretty much just editing the POM to actually set up Checkstyle with the Google Code Style, per the dev thread.

See [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)

Note that the Checkstyle version is set to ~7.7~ 8.0.  I've tested it with the plugin version and had no issues at all, but if anyone is concerned, we can lower it to either the default or another version.

### IntelliJ warnings
To pull into IntelliJ, please see https://github.com/jshiell/checkstyle-idea#configuration
Essentially the instructions are to:

-  Install the plugin (Available in IntelliJ's directly)
- Go into Settings -> Other Settings -> Checkstyle
- Change "Checkstyle version" to ~7.7~ 8.0.
- Apply.  Otherwise the new file won't match the version and an error will be thrown.
- Add a new Checkstyle file
  - ~This can either be a local file or a remote one (remote will attempt to update periodically, so if you don't want to repeatedly connect to the URL, download the file somewhere and load it in).  Either works. For 7.7, use [google_checks.xml](https://raw.githubusercontent.com/checkstyle/checkstyle/master/src/main/resources/google_checks.xml)~ Use the checkstyle.xml file included in the root directory of the project.  This file is identical to [google_checks.xml](https://raw.githubusercontent.com/checkstyle/checkstyle/master/src/main/resources/google_checks.xml), but we may choose to change it in the future.
  - If you have errors, you most likely need to make sure you're set to the right version and have applied it.
- Select the checkbox for the new style
- Apply

To ensure it's setup properly, open a file, e.g. `metron-platform/metron-common/src/main/java/org/apache/metron/common/dsl/functions/StringFunctions.java`

New warnings should show up that are based on Checkstyle, e.g.
```
Checkstyle: WhitespaceAround: 'if' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement
```

### Code formatting
There are two options for code formatting. One is to import the checkstyle.xml file from the project root.  The other is to import Google's IntelliJ editor settings for coverage.  In practice, Google's direct IntelliJ settings seem to handle autoformatting better. The instructions are essentially the same for each.  Note that if we choose to modify the checkstyle.xml we may also want to update the IntelliJ settings.

[...]

PR #607, opened 2017-06-03
METRON-982 add new rest api for storm supervisor status

ottobackwards:
This pr adds a new rest endpoint for retrieving storm supervisor summary information from the "/api/v1/supervisor/summary" storm ui rest endpoint.

This information contains the currently configured slots for each supervisor, and how many are used.

This is in support of METRON-981, which will be able to use this endpoint to enable errors and warnings based on the lack of available slots.

###TESTING

In full_dev environment execute the new endpoint from swagger, you should see the supervisor information from storm

/supervisors  off of the storm controller

### For all changes:
- [ x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [ x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [ x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [NA ] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [ x ] Have you included steps or a guide to how the change may be verified and tested manually?
- [ x ] Have you ensured that the full suite of tests and checks have been executed in the root incubating-metron folder via:
  ```
  ```

- [ x ] Have you written or updated unit tests and or integration tests to verify your changes?
- [ NA ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [ x ] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?


metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/StormStatusServiceImpl.java
```diff
@@ -49,6 +44,11 @@ public StormStatusServiceImpl(Environment environment, RestTemplate restTemplate
   }
 
   @Override
+  public SupervisorSummary getSupervisorSummary(){
+    return restTemplate.getForObject("http://" + environment.getProperty(STORM_UI_SPRING_PROPERTY) + SUPERVISOR_SUMMARY_URL, SupervisorSummary.class);
```
cestella: What if the storm UI is https?

simonellistonball: see https://issues.apache.org/jira/browse/METRON-906

ottobackwards: I was just following the convention of the service as I found it.  I think that is a separate issue as @simonellistonball points out

justinleet: https://github.com/apache/metron/pull/602 changes how that works slightly.  Whichever one goes in second will need to adjust for the other one (which shouldn't be much work either way).  I don't have a preference on ordering, personally.

justinleet: Should have pointed this out the other day, but that PR is in now, so this probably should be:
```
return restTemplate.getForObject(getStormUiProperty() + SUPERVISOR_SUMMARY_URL, SupervisorSummary.class);
```

Ping me if there's any issues and I'm happy to do more than just glance at it.

ottobackwards: Thanks!  I'll hook that up

ottobackwards:
Bump - anyone have any feedback?

merrimanr:
Hey @ottobackwards, list has been on my todo list for a while.  I'll spin it up today or tomorrow.  Sorry for the delay.

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/model/SupervisorStatus.java
```diff
@@ -0,0 +1,127 @@
...
+    upTime	String	Shows how long the supervisor is running
+    slotsTotal	Integer	Total number of available worker slots for this supervisor
+    slotsUsed	Integer	Number of worker slots used on this supervisor
+    Sample Response:
+    json { "supervisors": [ { "id": "0b879808-2a26-442b-8f7d-23101e0c3696", "host": "10.11.1.7", "upTime": "5m 58s", "slotsTotal": 4, "slotsUsed": 3 } ] }
+
+  */
+  private String id;
+  private String host;
+  private String upTime;
```
merrimanr: This needs to be "uptime" instead of "upTime" or jackson doesn't map it correctly.

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/model/SupervisorStatus.java
```diff
@@ -0,0 +1,127 @@
...
+    if (!getHost().equals(that.getHost())) {
+      return false;
+    }
+    return getUpTime().equals(that.getUpTime());
+  }
+
+  @Override
+  public int hashCode() {
+    int result = id != null? id.hashCode() : 0;
+    result = 31 * result + host != null? host.hashCode() : 0;
```
merrimanr: I get a warning in Intellij.  Should there be parenthesis around `host != null? host.hashCode() : 0`, similar to other classes in metron-rest-client?

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/StormStatusServiceImpl.java
```diff
@@ -17,10 +17,7 @@
  */
 package org.apache.metron.rest.service.impl;
 
-import org.apache.metron.rest.model.TopologyResponse;
-import org.apache.metron.rest.model.TopologyStatus;
-import org.apache.metron.rest.model.TopologyStatusCode;
-import org.apache.metron.rest.model.TopologySummary;
+import org.apache.metron.rest.model.*;
```
merrimanr: Most coding standards (including ours) don't like wildcard imports.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/StormControllerIntegrationTest.java
```diff
@@ -182,6 +185,16 @@ public void test() throws Exception {
             .andExpect(jsonPath("$.status").value("SUCCESS"))
             .andExpect(jsonPath("$.message").value(TopologyStatusCode.STARTED.name()));
 
+    this.mockMvc.perform(get(stormUrl + "/supervisors").with(httpBasic(user,password)))
+            .andExpect(status().isOk())
+            .andExpect(content().contentType(MediaType.parseMediaType("application/json;charset=UTF-8")))
+            .andExpect(jsonPath("$.supervisors[0]").exists())
+            .andExpect(jsonPath("$.supervisors[0].id").exists())
+            .andExpect(jsonPath("$.supervisors[0].host").exists())
+            .andExpect(jsonPath("$.supervisors[0].upTime").exists())
```
merrimanr: This also needs to be "uptime"

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/StormStatusServiceImpl.java
mattf-horton: @merrimanr , heh, you made me go look this up!

As far as I can see, our coding standards, both [here](https://cwiki.apache.org/confluence/display/METRON/Development+Guidelines#DevelopmentGuidelines-2.2CodeStyle) and [here](http://www.oracle.com/technetwork/java/codeconvtoc-136057.html) are silent on this topic.
There is a good discussion of the pros and cons [here](https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad). Despite the latter being a fairly old article, google did not show me any more recent definitive policy statements.

I must admit that I think the "wildcard imports are evil" side have slightly more weight in the argument, yet I personally find long lists of explicit imports annoying and appreciate IntelliJ's default settings (more than 5 imports of the same package collapses to wildcard).  Can you provide cites that support your assertion regarding "most coding standards"?

If you are so moved, you might bring this up on the mailing list, perhaps providing the citations above, and propose a vote to modify the Coding Style.

merrimanr: Thanks for checking me on this @mattf-horton.  I was under the impression we already voted to switch our coding standard to Google's based on the [Code Style](http://mail-archives.apache.org/mod_mbox/metron-dev/201705.mbox/%3CCADu=U3A23q-aehhim6R0AceZapAJ9QJvG8hnurhR2hkDtvXNLw@mail.gmail.com%3E) discussion on the dev list back in the beginning of May.  There is also a pending [PR](https://github.com/apache/metron/pull/577) that sets up Checkstyle with the Google Code Style.

Was this not an official vote or was the wiki just not updated?  If we did switch, [Google Coding Style](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements) does not allow wildcard imports.

This is probably one of those topics with no clear correct answer, similar to spaces vs tabs.  I personally am not a fan of wildcard statements because I prefer to see exactly what is being imported.

justinleet: @merrimanr @mattf-horton We did not have an official vote.  Given the widespread acceptance of it, since it's not a dev guildelines and not a bylaw thing, my impression was we didn't need it. Having said that, if it needs or should have an actual vote, I'm definitely okay with it and it should be done before https://github.com/apache/metron/pull/577 goes in.

However, 577 is still outstanding pending additional feedback I requested.  I was inclined to not update the wiki, because we hadn't built out and validated the tooling at the time.  I personally lean towards updating the wiki once we're good with that PR going in.

ottobackwards: Not to interrupt you guys....

I'm OK with changing this.... I have tried to setup intellij to not do this ( and I am already checking style ), but I think this PR came before I set things up. 

--- OK, GO!

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/model/SupervisorStatus.java
ottobackwards: Where would I see that?  Am I missing a test?

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/model/SupervisorStatus.java
ottobackwards: done

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/StormControllerIntegrationTest.java
ottobackwards: I don't understand, the API spec says it is upTime

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/model/SupervisorStatus.java
merrimanr: When I tested this in full dev with Swagger, the upTime field in the response was null.  Here's an example response when I hit http://node1:8744/api/v1/supervisor/summary:
```
{"supervisors":[{"totalMem":3072.0,"host":"node1","id":"70bd7170-160f-4215-8f18-d9088993fb47","uptime":"47m 20s","totalCpu":400.0,"usedCpu":0.0,"logLink":"http:\/\/node1:8000\/daemonlog?file=supervisor.log","usedMem":3328.0,"slotsUsed":4,"version":"1.0.1.2.5.6.0-40","slotsTotal":4,"uptimeSeconds":2840}],"schedulerDisplayResource":false,"logviewerPort":8000}
```
Notice the field is actually "uptime".  I changed the upTime variable name and getter/setting methods and that fixed it.  

The StormControllerIntegrationTest uses a mock Storm REST API so you're not going to catch it there.  You would only catch this by hitting a real Storm REST API.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/StormControllerIntegrationTest.java
merrimanr: Which API spec are you referring to?  This [page](http://storm.apache.org/releases/1.0.3/STORM-UI-REST-API.html) has it as "uptime".

ottobackwards: ok, i don't know how that got formatted to upTime, I have a comment in there that I copied out of that page to go by.....

The getter and setter names can stay though?  getUpTime() etc?

merrimanr: I changed the getter/setter methods in my test.  You can try leaving them the same but I know "getUptime()" works.

ottobackwards: done

ottobackwards:
@merrimanr we all set?

merrimanr:
Just tested this on full dev and it worked great.  Thanks @ottobackwards. +1

PR #608, opened 2017-06-05
METRON-986 Enhance Fastcapa to Support Intel X520

nickwallen:
### Changes

The following changes have been made to Fastcapa to improve interoperability with the Intel X520 NIC, address bugs uncovered during performance testing, and some basic enhancements to improve ease-of-use.

  * Created a separate command-line parameter for the transmit worker burst size; `-w TX_BURST_SIZE`.  This defines the maximum number of packets dequeued from the transmit ring and enqueued for delivery to Kafka at one time.  Previously the burst size for the receive and transmit worker were the same.  In practice the transmit worker burst size needs to be roughly 10x greater than the receive worker burst size.

  * Fixed a bug that caused some failures reported through the message delivery callback to be ignored.  Both successful and failed deliveries are reported through this callback and the logic has been corrected to match this.

  * Error messages for each failed message can be logged when the `DEBUG` flag is used during compilation.  By default errors for each message are not logged due to performance implications.

  * Now logging all configuration parameters by default.

  * The code was unnecessarily limiting `-x TX_RING_SIZE` by using a uint16_t; a 32-bit value is perfectly acceptable.  Also, no need to validate the -x TX_RING_SIZE as DPDK's EAL does this for us.  

  * Added separate default port configurations for the receive and transmit side.  Previously the same configuration values were used for both.

  * Added logic to count, summarize and report all ethernet device errors.

  * Refactored logic so that all receive and transmit worker logic is contained within `worker.[c|h]` and all NIC logic contained within `nic.[c|h]`.

  * Updated the fastcapa service script to allow the process to be stopped, killed, and the logs tailed.

### Testing 

This change has been tested on a live cluster with the Intel X520 NIC. The probe was able to capture and land roughly 1 gbps of raw packet data into Kafka.  The change was also tested in the Fastcapa test environment.

To test the change yourself, simply spin-up the virtualized test environment which will deploy and validate that Fastcapa can land packets in Kafka correctly.

```
cd metron-deployment/vagrant/fastcapa-test-environment
vagrant up
```


cestella:
This is great work, @nickwallen Nice refactoring and good functionality changes.

Just curious for the uninitiated, if I'm running a non-X520 but another DPDK supported card, is that supported via configuration changes only or do those require code changes at this point?

nickwallen:
What we have **should** work out-of-the-box for most NICs.  Of course, I have only been able to test on the x520 and Cisco VNICs, so that is all I can really be sure of.

If changes would be needed, it would likely be in the initialization of the `rte_eth_conf` and `rte_eth_txconf` structs. The defaults are currently provided in `nic.c`.  

I was going to make this configurable, but decided not to in this PR due to the amount of work needed.  I would have had to define configurable values that populate all of the possible fields in each of these structs (which are quite numerous).  I would have had to allow configuration of these values via the config file, rather than just the command-line, since they are so numerous.

There is also the challenge of how to allow a user to dynamically configure values that are effectively macros, if that is even possible. For example, how should a user define something like the following in a configuration file?  I'd have to dig for a good way to handle this.

```
static const struct rte_eth_conf rx_conf_default = {
     .rxmode = {
         .mq_mode = ETH_MQ_RX_RSS,
         ...
      }
}
```

nickwallen:
Did I answer your question @cestella ?  Short answer is no and unfortunately I don't quite know how to get there.

cestella:
That sounds good; can you create a follow-on JIRA to investigate how to make the properties specified at run-time rather than as macros in code?

cestella:
+1 by inspection

nickwallen:
> That sounds good; can you create a follow-on JIRA to investigate how to make the properties specified at run-time rather than as macros in code?

FYI - Created https://issues.apache.org/jira/browse/METRON-1024

PR #609, opened 2017-06-06
METRON-987: Allow stellar enrichments to be specified by a list as well as a map

cestella:
## Contributor Comments
Currently, stellar enrichments are specified by a map associating fields with the stellar expressions associated with the fields. There is a significant downside here in that you cannot update previously assigned fields. For instance, the following cannot be represented currently:
```
 "fieldMap": {
       ...
      "stellar" : {
        "config" : {
          "hostname" : "if ENDS_WITH(hostname, '.') then CHOP(hostname) else hostname",
          "hostname" : "TO_LOWER(hostname)"
        }
      }
    }
```
This would now be allowed thusly:
```
"fieldMap": {
       ...
      "stellar" : {
        "config" : [
          "hostname := if ENDS_WITH(hostname, '.') then CHOP(hostname) else hostname",
          "hostname := TO_LOWER(hostname)"
        ]
      }
    }
```
A consequent of this deficiency is that we also cannot use temporary variables and unset them after their use inside an enrichment group.

The proposed change is to allow users to use lists of strings representing stellar expression assignments with the same syntax as the Stellar REPL. This would be as an alternative to maps, but the map syntax would also be supported for legacy.

## Test plan
* Follow the instructions located [here](https://github.com/apache/metron/tree/master/metron-platform/metron-enrichment#example-enrichment-via-stellar) to ensure no regressions
* Adjust the configuration for `$METRON_HOME/config/zookeeper/enrichments/squid.json` to the following and run data through the topologies again:
```
{
 "enrichment" : {
   "fieldMap": {
     "stellar" : {
       "config" : {
         "numeric" : [
                     "foo := 1 + 1",
                     "grok := foo + 3"
                     ]
         ,"ALL_CAPS" : "TO_UPPER(source.type)"
       }
     }
    }
 },
 "threatIntel" : {
   "fieldMap":{
    "stellar" : {
       "config" : [
         "bar := TO_UPPER(source.type)"
       ]
     } 
   },
   "triageConfig" : {
   }
 }
}
```
  * Ensure that each message has `ALL_CAPS`, `foo`, `grok` and `bar` fields.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/message/JSONFromPosition.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/message/JSONFromPosition.java
@@ -40,10 +40,12 @@ public class JSONFromPosition implements MessageGetStrategy {
 
   @Override
   public JSONObject get(Tuple tuple) {
+    String s = null;
     try {
-      return (JSONObject) parser.get().parse(new String(tuple.getBinary(position), "UTF8"));
+      s =  new String(tuple.getBinary(position), "UTF8");
+      return (JSONObject) parser.get().parse(s);
     } catch (Exception e) {
-      throw new IllegalStateException(e.getMessage(), e);
+      throw new IllegalStateException("Unable to parse " + s + " due to " + e.getMessage(), e);
     }
   }
 }
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/message/JSONFromPosition.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/message/JSONFromPosition.java
@@ -17,6 +17,7 @@
  */
 package org.apache.metron.common.message;
 
+import org.apache.commons.io.Charsets;
 import org.apache.storm.tuple.Tuple;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
@@ -42,7 +43,7 @@ public class JSONFromPosition implements MessageGetStrategy {
   public JSONObject get(Tuple tuple) {
     String s = null;
     try {
-      s =  new String(tuple.getBinary(position), "UTF8");
+      s =  new String(tuple.getBinary(position), Charsets.UTF_8);
       return (JSONObject) parser.get().parse(s);
     } catch (Exception e) {
       throw new IllegalStateException("Unable to parse " + s + " due to " + e.getMessage(), e);
--- a/metron-platform/metron-enrichment/README.md
+++ b/metron-platform/metron-enrichment/README.md
@@ -54,7 +54,9 @@ The `config` map is intended to house enrichment specific configuration.
 For instance, for the `hbaseEnrichment`, the mappings between the
 enrichment types to the column families is specified.
 
-The `fieldMap`contents are of interest because they contain the routing and configuration information for the enrichments.  When we say 'routing', we mean how the messages get split up and sent to the enrichment adapter bolts.  The simplest, by far, is just providing a simple list as in
+The `fieldMap`contents are of interest because they contain the routing and configuration information for the enrichments.  
+When we say 'routing', we mean how the messages get split up and sent to the enrichment adapter bolts.  
+The simplest, by far, is just providing a simple list as in
 ```
     "fieldMap": {
       "geo": [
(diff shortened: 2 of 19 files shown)


metron-platform/metron-enrichment/README.md
```diff
@@ -71,40 +73,94 @@ The `fieldMap`contents are of interest because they contain the routing and conf
...
+expressions (i.e. `profile_window`, `global_login_profile`, and `stats`) that are useful only within the context
+of this group of stellar expressions.  In this case, we would need to ensure that we use the list construct
+when specifying the group and remember to set the temporary variables to `null` so they are not passed along.
+
+In general, things to note from this section are as follows:
+* The stellar enrichments for the `stellar` enrichment adapter are specified in the `config` for the `stellar` enrichment
+adapter in the `fieldMap`
+* Groups of independent (i.e. no expression in any group depend on the output of an expression from an other group) may be executed in parallel
+* If you have the need to use temporary variables, you may use the list construct.  Ensure that you assign the variables to `null` before the end of the group.
+* **Ensure that you do not assign a field to a stellar expression which returns an object which JSON cannot represent.**
```
justinleet: I don't think it's in scope for this, but it feels like this should be enforceable at an early time.  It would be nice to be able to do some high level checks like this before allowing things like this.

justinleet:
A lot of this brings up the discussions around deprecating / dropping functionality (both for Stellar and in general).  It seems like ideally, we'd be deprecating the map functionality in favor of the list functionality, probably letting it ride a few releases, and eventually dropping it after a migration period.

Do we need to start getting together a set of discussions and plans for handling this sort of deprecation as we build out enhanced versions of older features?

metron-platform/metron-common/src/main/java/org/apache/metron/common/message/JSONFromPosition.java
```diff
@@ -40,10 +40,12 @@ public JSONFromPosition(int position) {
 
   @Override
   public JSONObject get(Tuple tuple) {
+    String s = null;
     try {
-      return (JSONObject) parser.get().parse(new String(tuple.getBinary(position), "UTF8"));
+      s =  new String(tuple.getBinary(position), "UTF8");
```
justinleet: Isn't the identifying string "UTF-8", not "UTF8"?

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/adapters/stellar/StellarAdapterTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+public class StellarAdapterTest extends StellarEnrichmentTest {
+  StellarProcessor processor = new StellarProcessor();
+
+  private JSONObject enrich(JSONObject message, String field, ConfigHandler handler) {
+    VariableResolver resolver = new MapVariableResolver(message);
+    return StellarAdapter.process( message
+                                 , handler
+                                 , field
+                                 , 1000l
```
justinleet: Out of my personality preference, can you make this 1000L, so I don't look at it again and think it's 10001?

metron-platform/metron-common/src/main/java/org/apache/metron/common/message/JSONFromPosition.java
cestella: `UTF-8` and `UTF8` are aliases in the `StandardCharsets` from looking through the java API code.

metron-platform/metron-enrichment/src/test/java/org/apache/metron/enrichment/adapters/stellar/StellarAdapterTest.java
cestella: But the confusion is part of the charm!

metron-platform/metron-enrichment/README.md
cestella: I agree, this was just filling in documentation that didn't exist.  There should be a follow-on ticket here for a solution.

metron-platform/metron-common/src/main/java/org/apache/metron/common/message/JSONFromPosition.java
justinleet: Can we just use StandardCharsets.UTF_8 then, and not even worry about the string being right?  I'm not sure where that alias is, so I'm not convinced it's guaranteed (although a source proving me wrong would also be acceptable)

cestella:
@justinleet yes, agreed.  I wanted to get this functionality in and have it be backwards compatible for the short term, but ultimately, I think that it's a better approach and we should work toward phasing out the map-based approach.  The issue is that there are a few other places that also need to be normalized:
* the field transformations
* the importer config
* possibly others that I am not thinking about

Anyway, to answer your question, yes I'll probably start a discuss thread on that tomorrow and construct some JIRAs to see if 
* other people think deprecation is a good thing to work towards
* how in the heck we should accomplish it

justinleet:
@cestella Perfect.  Definitely didn't want to imply that belongs in this ticket at all.  I'm in full support of it being backwards compatible for now, and keeping the discussion of how we prune old implementations separate.

justinleet:
+1, I was able to run this up, both the smoke test and the adjusted variant

ottobackwards:
Do we want to add a version number to these configuration now?  Then we can say no version is 0, this is 1 etc.

cestella:
@ottobackwards Yeah, that's why I went with backwards compatibility here, so we could defer that decision until we had a discussion.  I think the enrichment topology should really change quite a bit and the config really should reflect that.  It may be broader changes than just a version can encapsulate, in my mind.

PR #610, opened 2017-06-07
METRON-877 Extract core implementation and UDF support, create metron-stellar module

mattf-horton:
## Contributor Comments
The design document has been available for some time at https://docs.google.com/document/d/1EP7Jt4ePHe2A-_oboLl2QbN1muh7uKeET_kbpIgjcJM/edit#
I have now added a new section, starting on p. 7, titled "Retrospective: Changes made for Parts 1 & 2" that details the specific changes actually done in this PR to extract Stellar.

## Pull Request Checklist
In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
The intent of these changes is to achieve an architecture change WITHOUT any functional change.  The continued working of all unit tests, integration tests and real-world applications should result.  In addition, it generates (in metron-stellar/stellar-common/target) a stellar-common uber-jar (including all dependencies), which among other things can run the Stellar shell as a stand-alone Java application (although missing those Stellar functions that remain as part of Metron):
```
java -cp target/stellar-common-0.4.0-uber.jar \
    org.apache.metron.stellar.common.shell.StellarShell
```
Note that this PR does not yet have RPM generation (Step 6).  Thus, the Stellar jar will need to be manually added to the lib/ directory of test deployments.

- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root incubating-metron folder via:
  ```
  mvn -q clean integration-test install && build_utils/verify_licenses.sh 
  ```

- [x] Have you written or updated unit tests and or integration tests to verify your changes?
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [IN PROCESS] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

### For documentation related changes:
[...]

PR #468, opened 2017-02-28
METRON-744: Allow Stellar functions to be loaded from HDFS

cestella:
## Description
The benefit of Stellar is that adding new functionality is as simple as providing a Jar. This enables people who want to integrate with Metron to easy add enrichments or other functionality. The snag currently with this is that we provide a single jar, so all stellar functions that we have available must be dependencies of the main jar that drives the topology plus what local directories we can configure via the storm configs. This makes the process of adding 3rd party jars not as easy as it could be.
Adjust the the following to additionally load classes from a location in HDFS /apps/metron/stellar using something like accumulo ( https://accumulo.apache.org/blog/2014/05/03/accumulo-classloader.html)
* Profiler topology
* Parser topology
* Enrichment topology
* Enrichment Flat file loader
* Enrichment MR loader
* The REPL

One thing to note about this PR is that we shift the way that we load classes from the classpath.  We now use [ClassIndex](https://github.com/atteo/classindex) instead of Reflections.  This means that stellar classes are found by analyzing a file dropped in META-INF.  This has a couple of advantages:
* Blazing fast function loading because we're iterating over every jar, not every class.  So fast, there is no noticeable delay for me in starting the REPL.
* Reflections had some serious issues with non-default classloaders that we ran into

For the Stellar function author, this has no impact to you.  You just use the annotation and make sure to depend on metron-common (provided dependency) and the file in META-INF is automatically added.

# Testing Plan
## Preliminaries

Also, set an environment variable to indicate `METRON_HOME`:
* `export METRON_HOME=/usr/metron/0.3.1` 

### Setup Classpath 
* As the hdfs user, create a directory in HDFS for the stellar function
  jars:
`hadoop fs -mkdir /apps/metron/stellar`
* Place [custom-1.0-SNAPSHOT.jar](https://github.com/cestella/incubator-metron/blob/5c50cf5e4b1d53383c1d5d724dd0d20983f16cb8/metron-platform/metron-common/src/test/classpath-resources/custom-1.0-SNAPSHOT.jar?raw=true) in /apps/metron/stellar
* Edit `$METRON_HOME/config/zookeeper/global.json` and add the following properties:
```
"stellar.function.paths" : "hdfs://node1:8020/apps/metron/stellar/.*.jar"
```

### Set-up the Profiler

* Create the profiler hbase table
`echo "create 'profiler', 'P'" | hbase shell`

* Open `~/rand_gen.py` and paste the following:
```
#!/usr/bin/python
import random
import sys
import time
def main():
[...]

PR #616, opened 2017-06-10
METRON-877 Extract core implementation and UDF support, create metron-stellar module

mattf-horton:
## Contributor Comments
It appears that github permanently closes any PR that gets rebased, hence for the history and discussion to date, I can only point you at the previous incarnation of this PR: https://github.com/apache/metron/pull/610

The design document has been available for some time at https://docs.google.com/document/d/1EP7Jt4ePHe2A-_oboLl2QbN1muh7uKeET_kbpIgjcJM/edit#
I have now added a new section, starting on p. 7, titled "Retrospective: Changes made for Parts 1 & 2" that details the specific changes actually done in this PR to extract Stellar.

## Pull Request Checklist

### For all changes:

### For code changes:
The intent of these changes is to achieve an architecture change WITHOUT any functional change.  The continued working of all unit tests, integration tests and real-world applications should result.  In addition, it generates (in metron-stellar/stellar-common/target) a stellar-common uber-jar (including all dependencies), which among other things can run the Stellar shell as a stand-alone Java application (although missing those Stellar functions that remain as part of Metron):
```
java -cp target/stellar-common-0.4.0-uber.jar \
    org.apache.metron.stellar.common.shell.StellarShell
```
For testing within Metron, the stellar classes are still incorporated into the metron-common jar and RPM, so test installation should work as before.

For a general test plan for all Stellar functionality within Metron, please see #468 's test plan. 

  ```
  ```

- [IN PROCESS] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?


ottobackwards:
So, the test plan for this is...?  The test work etc?

ottobackwards:
So we have a StatefulExecutor, is there an idea that there will be more types of executors?  Will there be a non-stateful executor, maybe for performance reasons or other use cases?  Maybe in the future some write up on this would help with the __intent__

ottobackwards:
As a more stand alone module, I think the readme should be broken down a little bit, instead of just being 'here is the language'.

A top level statement document, with what stellar is at a high level, why it exists and when it has value to be used.
Below that more technical guides - language, examples, faq, discovery and loading

I wouldn't hold this PR up for this, but maybe it should be added to the follow on list.

ottobackwards:
Each example could be the seed of future development of encapsulating that functionality.  For example - how to use stellar in spark could be evolved into a stellar-spark integration package.

ottobackwards:
+ 1 on inspection, built and ran the tests -  This is a lot of work, great job.  As with other large PR's ( cough METRON-777 ) I think that this could be a feature branch.  Although this did have more involvement from other contributors ( @cestella ), the feature branch would help the complete review pool

mattf-horton:
@ottobackwards , thanks very much for tackling this admittedly large review.

> So, the test plan for this is...?

"For a general test plan for all Stellar functionality within Metron, please see #468 's test plan."
It's long, hence not reproduced here.

> So we have a StatefulExecutor...

The base executor is org.apache.metron.stellar.common.BaseStellarProcessor, formerly org.apache.metron.common.stellar.BaseStellarProcessor.  It parses and evaluates Stellar expressions, with caching of the parse, and with given context (variables), variableResolver, and functionResolver.  It can return any Stellar data type.  It also provides a validate() method.

The "stateful" executor, org.apache.metron.stellar.common.StellarStatefulExecutor, assists with Aggregators, where state needs to persist across multiple invocations of the same entity, and in fact was formerly located at org.apache.metron.profiler.stellar.StellarExecutor.  It was renamed when I relocated it, to avoid conflict with the Shell executor, org.apache.metron.stellar.common.shell.StellarExecutor, which is significantly different.  I deemed the stateful executor to be Stellar functionality rather than Profiler, and moved it.  The Aggregator functions will also be moved into Stellar, but that will be in Step 3, which will be a different PR.

I agree in future it would be nice to have design documents, for Stellar and other parts of Metron.

> I think the readme should be \[expanded into a Language Reference Manual :-) \]
> ...maybe it should be added to the follow on list.

Yah, that would be good.  I've opened [METRON-998](https://issues.apache.org/jira/browse/METRON-998), as a follow-on task.

> I think that this could be a feature branch.

Since this started as a regular PR, I'd like to finish this one that way.  But I agree, before starting Part 3 I'll petition to create a feature branch for the next set(s) of work.

ottobackwards:
Thanks Matt, I think your write up above should be captured somewhere...  We don't want to lose it in the PR

mattf-horton:
Hi @ottobackwards , I've captured those two paragraphs in a new section of the [GoogleDoc](https://docs.google.com/document/d/1EP7Jt4ePHe2A-_oboLl2QbN1muh7uKeET_kbpIgjcJM/edit#), titled "Design Observations".

mattf-horton:
@ottobackwards , is the "1" above a +1?
@cestella , since this is in your bailiwick, would you like to review it?
Thanks!

ottobackwards:
Yes, sorry.
+1

cestella:
Ah, yes, I do want that.

cestella:
+1 by inspection, this is great work, @mattf-horton

ottobackwards:
so can this be merged?

mattf-horton:
Thanks, @ottobackwards and @cestella !
Merge to current master was trivial, as seen here: https://github.com/apache/metron/pull/616/commits/87a44f600678ffe246ac7da49128e0ac8c3cf8c1
Proceeding with commit.

ottobackwards:
super!  great work @mattf-horton

mattf-horton:
This has been committed to git-wip-us.apache.org/repos/asf/metron.  Propagation seems a little slow today.

PR #622, opened 2017-06-22
METRON-1005 Create Decodable Row Key for Profiler

nickwallen:
To be able to answer the types of questions that I outlined in [METRON-450](https://issues.apache.org/jira/browse/METRON-450), we need a row key that is decodable. Right now there is no logic to decode a row key, nor is the existing row key easily decodable.

Once the row keys can be decoded, you could scan all of the row keys in the Profiler's HBase table, decode each of them and extract things like, the names of all your profiles, the names of entities within a profile, the period duration of a given profile.

- [ ] Do not merge.  Opening this PR for review and feedback. 
- [x] This change is now backwards compatible.

## Testing
    ```
    cd metron-deployment/vagrant/full-dev-platform/
    vagrant up
    ```

1. Free up a slot by stopping the Snort topology.
    ```
    storm kill snort
    ```

1. Install the Profiler according to the [installation instructions](https://github.com/apache/metron/tree/master/metron-analytics/metron-profiler#installation), but don't start the Profiler topology just yet.

1. Change the window/period duration to 1 minute in the Profiler.
    ```
    cd /usr/metron/0.4.0
    sed -i 's/15/1/' config/profiler.properties
    ```

1. Change the window/period duration to 1 minute in the Profiler Client.  Add the `profiler.client.period.duration` property to your global settings.
    ```
    [root@node1 0.4.0]# cat config/zookeeper/global.json
    {
        "es.clustername": "metron",
        "es.ip": "node1:9300",
        "profiler.client.period.duration": "1"
    }
    ```

1. Define a profile in a file located at `$METRON_HOME/config/zookeeper/profiler.json`.
    ```
    {
      "profiles": [
        {
          "profile": "hello-world",
          "foreach": "ip_src_addr",
          "init":    { "count": "0" },
          "result":  "count"
        }
      ]
    }
    ```

1. Upload these configuration changes to Zookeeper.
    ```
    bin/zk_load_configs.sh -m PUSH -i config/zookeeper/ -z node1:2181
    ```

1. Start the Profiler.
    ```
    cd $METRON_HOME
    bin/start_profiler_topology.sh
    ```

1. By default, the Profiler will continue to use the original RowKeyBuilder.  Validate that it is doing so successfully, by retrieving some data using the client API.
    ```
    [root@node1 0.4.0]# bin/stellar -z node1:2181
    Stellar, Go!
    {es.clustername=metron, 
      es.ip=node1:9300, 
      es.date.format=yyyy.MM.dd.HH, 
      parser.error.topic=indexing, 
      profiler.client.period.duration=1}

    [Stellar]>>> PROFILE_GET('hello-world','192.168.66.1', PROFILE_FIXED(2, "HOURS"))
    [1, 16, 129, 125, 133, 137]
    ```

1. Stop the Profiler
    ```
    storm kill profiler
    ```

1. Configure the Profiler to use the new row keys.  This requires that you edit the Profiler's Flux file to specify the RowKeyBuilder.
    ```
    sed -i 's/org.apache.metron.profiler.hbase.SaltyRowKeyBuilder/org.apache.metron.profiler.hbase.DecodableRowKeyBuilder/' flux/profiler/remote.yaml
    ```

1. Configure the Profiler Client to use the new row keys by adding the `profiler.client.row.key.builder` to the global configuration.
    ```
    [root@node1 0.4.0]# cat config/zookeeper/global.json
    {
        "es.clustername": "metron",
        "es.ip": "node1:9300",
        "profiler.client.period.duration": "1",
        "profiler.client.row.key.builder": "org.apache.metron.profiler.hbase.DecodableRowKeyBuilder"
    }
    ```

1. Upload these configuration changes to Zookeeper.
    ```
    bin/zk_load_configs.sh -m PUSH -i config/zookeeper/ -z node1:2181
    ```

1. Start the Profiler.
    ```
    cd $METRON_HOME
    bin/start_profiler_topology.sh
    ```

1. Validate that it uses the new row keys.
    ```
    [root@node1 0.4.0]# bin/stellar -z node1:2181
    SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
    SLF4J: Defaulting to no-operation (NOP) logger implementation
    SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
    Stellar, Go!
    {es.clustername=metron, 
      es.ip=node1:9300, 
      es.date.format=yyyy.MM.dd.HH, 
      parser.error.topic=indexing, 
      profiler.client.period.duration=1, 
      profiler.client.row.key.builder=org.apache.metron.profiler.hbase.DecodableRowKeyBuilder}

    [Stellar]>>> PROFILE_GET('hello-world','192.168.66.1', PROFILE_FIXED(2, "HOURS"))
    [189, 119, 64, 137, 131]
    ```

## Pull Request Checklist


cestella:
I strongly vote that you move these changes to a new RowKeyBuilder and keep the old one default until a major release.  The sad thing is, I don't think there *is* a way to migrate the old data, so we're kinda screwed here in terms of backwards compatibility.  I definitely agree that we should move to a profiler row key that is decodable, though, so it's very good work @nickwallen .

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
```diff
@@ -0,0 +1,382 @@
...
+ * <li>magic number - Helps to validate the row key.</li>
+ * <li>version - The version number of the row key.</li>
+ * <li>salt - A salt that helps prevent hot-spotting.
+ * <li>profile - The name of the profile.
+ * <li>entity - The name of the entity being profiled.
+ * <li>group(s) - The group(s) used to sort the data in HBase. For example, a group may distinguish between weekends and weekdays.
+ * <li>period - The period in which the measurement was taken. The first period starts at the epoch and increases monotonically.
+ * </ul>
+ */
+public class DecodableRowKeyBuilder implements RowKeyBuilder {
```
nickwallen: The new `RowKeyBuilder` implementation that is decodable.  Everyone should just use this, but the older implementation is left for backwards compatibility.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/SaltyRowKeyBuilder.java
```diff
@@ -44,7 +46,17 @@
...
+ * easily decodable.  More specifically, the profile, entity, groups and period that make up the row key
+ * cannot be extracted from a previously generated row key.  This makes it difficult to answer questions
+ * like; What entities are included in this profile?  What is the period for this profile?  Use the
+ * DecodableRowKeyBuilder instead.
+ *
+ * @deprecated Replaced by DecodableRowKeyBuilder
+ * @see DecodableRowKeyBuilder
  */
+@Deprecated
 public class SaltyRowKeyBuilder implements RowKeyBuilder {
```
nickwallen: I marked the old `RowKeyBuilder` as deprecated.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
```diff
@@ -0,0 +1,382 @@
...
+
+  /**
+   * A magic number embedded in each row key to help validate the row key and byte ordering when decoding.
+   */
+  protected static final short MAGIC_NUMBER = 77;
+
+  /**
+   * The version number of the row keys supported by this builder.
+   */
+  protected static final byte VERSION = (byte) 1;
```
nickwallen: I added a `VERSION` field to the row key, hoping that this might help future changes to the `RowKeyBuilder`.  With this, I could potentially start to parse the row key and then choose the right `RowKeyBuilder` implementation; the one used to create the row key.  This would make row key changes seemless to users.

metron-analytics/metron-profiler-client/src/main/java/org/apache/metron/profiler/client/stellar/RowKeyBuilderFactory.java
```diff
@@ -0,0 +1,125 @@
...
+   * @return A RowKeyBuilder instantiated using the global property values.
+   */
+  public static RowKeyBuilder create(Map<String, Object> global) {
+    String rowKeyBuilderClass = PROFILER_ROW_KEY_BUILDER.get(global, String.class);
+    LOG.debug("profiler client: {}={}", PROFILER_ROW_KEY_BUILDER, rowKeyBuilderClass);
+
+    // instantiate the RowKeyBuilder
+    RowKeyBuilder builder = ReflectionUtils.createInstance(rowKeyBuilderClass);
+    setSaltDivisor(global, builder);
+    setPeriodDuration(global, builder);
```
nickwallen: If I had some IoC-like functionality like Flux or Spring here, then this wouldn't be a problem at all.

mattf-horton:
Your proposal has the advantage of making data in HBase self-identifying (if one has the key), which I always like.  However, it's a large change and induces yet more complexity.  There's an alternative I've been noodling occasionally, which I put forward here for consideration:

Create a Profile Audit Log table in HBase.  Every time a Profiler is configured, started, or stopped, make one entry in the audit log.  The idea is to be able to answer exactly the kinds of questions posed in METRON-450, so the records should include things like the configuration, the first and last timestamps, and perhaps the key builder parameters.  This would prevent historical profiles from being "lost" because the would-be querier doesn't have access to the exact config parameters used to write the profile.

For the sake of housekeeping, one might do a scan, daily and/or at system restart, to assure that (a) the set Profiles with a "start" but not an "end" recorded in the audit log, and (b) the set of currently running Profiles, are actually consistent, and record "inferred end" entries in the audit log for orphans found.

This solution is somewhat backward-applicable to existing Profile data; I think there are brute-force ways to scan the existing HBase tables and infer audit log entries, especially if historical configuration data is still available.  We could write such a scanner.

nickwallen:
> Your proposal has the advantage of making data in HBase self-identifying (if one has the key), which I always like. However, it's a large change and induces yet more complexity

What do you find unnecessarily complex here?  The code base was already designed to accept different row key implementations.  So this change involves the following.

1. The new decodable row key 
2. Profiler client logic to instantiate row key builders
3. Profiler client logic to pass parameters to the instantiated row key builders

I would agree that I think item 3 is unnecessarily complex.  That's where I wanted feedback.  I think just passing parameters through an interface method would simplify this a lot.

nickwallen:
I also think this row key implementation would be very helpful (if not necessary) if we want to implement Profiler Client APIs in languages that aren't JVM based.  For example, the magic number can help identify byte-order differences, the version number helps us version changes over time, and the size fields allow us to reliably parse each value from the row key.

nickwallen:
> Create a Profile Audit Log table in HBase

This is an interesting idea, Matt.  I like it.  Along these same lines I was thinking of a table of contents.  I think your audit log idea is one way to implement a table of contents. 

I have looked at a TSDB implementation backed by HBase, OpenTSDB, and they use a table of contents approach.  The ToC records metadata about the time series data that is stored. 

But I don't think these ideas are mutually exclusive with a decodable row key.  The decodable row key would allow us to rebuild the ToC should it become corrupted or lost.

Are you thinking that a decodable row key is not needed at all?

mattf-horton:
Hi @nickwallen , 
- I didn't say "unnecessarily complex", I just said "adds complexity".  If we are to have a decodable row key, then something like this would be necessary.  The point of my comments on this is to shift to a top-down perspective:
- The requirements you're trying to address are, I think, to be able to query metadata about existing Profiles, for multiple reasons including to not have old Profiles become inaccessible due to loss of metadata needed for Profile queries.
- I think a ToC is a better way to do that than a decodable rowkey.  It localizes the metadata instead of having to do a full scan and reason about all the rowkeys for every metadata query.  I proposed the Profile audit log as a very simple way to implement a ToC, and thanks for saying that better than I did.
- I'm not against doing a _more_ decodable row key, it just in my mind is not the simplest nor best solution to the evident requirement.  But it does constitute an architectural improvement.
- As I briefly mentioned, I don't think our current row key is totally opaque, it just needs a brute-force approach to figure out.  Not suitable for interactive queries, but would be acceptable for a one-time pass to build (or re-build) the ToC.

nickwallen:
>  I don't think our current row key is totally opaque, it just needs a brute-force approach to figure out. Not suitable for interactive queries, but would be acceptable for a one-time pass to build (or re-build) the ToC.

For reference, here is what the existing row key looks-like.

 salt (16B) + profile name (?) + entity name (?) + groups (?) + time (8B)

How would you decode it?  The salt and the time components have known lengths; 16B and 8B respectively.  Other than those two components, I don't know how to distinguish the profile name, entity or groups.  I can only decode the row key if I already know either the profile name or the entity, which defeats the advantages of being able to decode it.

mattf-horton:
Let me take a look at this more deeply.

cestella:
So, in my mind the feature here is the enablement of batch analytics on the profiles.  To that end, I'm in general in favor of a decodable row key. I think that the question really isn't a ToC *or* a decodable rowkey.  I think, rather, we will want both.  The two will follow different access patterns.  A decodable rowkey sans ToC will be suitable only for full table scan-style access.  A ToC would enable to slice or dice by profile/entity/etc.  

That being said, a ToC without a decodable rowkey is substantially less nice.  Without being able to decode the rowkey, we will not be able to regenerate the ToC to provide alternative indexing.  I see this as a first step to enable a broader discussion on just what kind of access semantics beyond Get/Put we want to place on the profiles.

All that to say, I'm in favor of the effort.  I worry at the impact going forward to existing profiles, though.  From the point where we do this, we will create a fork whereby new profiles and old profiles diverge.  I think we need to discuss the migration story more explicitly and see if it is plausible to create a migration tool that is fuzzy (i.e. will look at the existing profiles and try to pick them apart).

I'd be ok for that work to be a follow-on, but I would want the plan to be very explicit and I would be -1 for a release until it's in.

cestella:
Also, while we're in here, is there a strong reason why the prefixed hash is so large?  It's just there for uniformity of distribution, correct?  I'd propose a non-cryptographic hash for this purpose like Murmur.

cestella:
I want to point out that I am also in favor of an audit log for the profiler, but I don't think it's a complete solution for the batch analytics use-case.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/SaltyRowKeyBuilder.java
```diff
@@ -81,20 +99,19 @@ public SaltyRowKeyBuilder(int saltDivisor, long duration, TimeUnit units) {
    * @return All of the row keys necessary to retrieve the profile measurements.
    */
   @Override
-  public List<byte[]> rowKeys(String profile, String entity, List<Object> groups, long start, long end) {
+  public List<byte[]> encode(String profile, String entity, List<Object> groups, long start, long end) {
     // be forgiving of out-of-order start and end times; order is critical to this algorithm
     end = Math.max(start, end);
     start = Math.min(start, end);
```
mattf-horton: Heh, this has been in the code for a long time, but isn't this a bug?  If it starts out in the wrong order, say end is 1 and start is 5, won't this pair of statements result in both end and start being equal to the larger, ie 5 ?  We need an intermediate variable for a binary swap!

mattf-horton:
Here's what I've got on decoding old rowkeys:
https://gist.github.com/mattf-horton/8e685e373b1a3fa6aeec8ef8828be096

The format of the keys is
`salt (4B) + profile name (?) + entity name (?) + groupvalues (?) + period (8B)`
with most of it (all but the salt and period number) in the clear as human-readable strings.

Deducing periodDuration has a nice arithmetic answer, I think.
The NLP issues are of course harder.  Enjoy the read, it's only two pages.

mattf-horton:
And btw, since there is no easily expressed algorithm for the NLP part of the problem, I'm +1 on doing both a decodable rowkey and a ToC.  For the existing profiles that @cestella expressed concern about, I would point out that as long as one DOES have the Profile specs still lying around, it's actually easy to re-write the old Profiles into new format with decodable rowkeys.  That is a very modest-sized program, the main problem being noticing and dealing with duplicate titled Profiles with different periodDurations.  But the info I pointed out in the paper helps sufficiently, I think.

cestella:
I would agree that it's easy to rewrite the old profiles in the new decodable format IF we can read the old profile keys and pull out their info (even fuzzily).

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
```diff
@@ -0,0 +1,382 @@
...
+  private byte[] encodeGroups(List<Object> groups) {
+
+    // encode each of the groups and determine their size
+    int lengthOfGroups = 0;
+    List<byte[]> groupBytes = new ArrayList<>();
+    for(Object group : groups) {
+      byte[] groupB = Bytes.toBytes(String.valueOf(group));
+      groupBytes.add(groupB);
+      lengthOfGroups += groupB.length;
+    }
```
mattf-horton: This output format does not provide delimiters for the (possible) several group values.  Recommend we should use comma (",") or other such character to delimit groupvalues if more than one is present.  If comma can be present in the groupvalues, have the user provide a delimiter character.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
```diff
@@ -0,0 +1,402 @@
...
+   * @param period The period
+   * @param saltDivisor The salt divisor
+   */
+  public static byte[] encodeSalt(long period, int saltDivisor) {
+    try {
+      // an MD5 is 16 bytes aka 128 bits
+      MessageDigest digest = MessageDigest.getInstance("MD5");
+      byte[] hash = digest.digest(encodePeriod(period));
+      int salt = Bytes.toShort(hash) % saltDivisor;
+      return Bytes.toBytes(salt);
```
simonellistonball: doesn't the profile name in the key not achieve this? we should probably enforce those being unique in config (we may already). I suppose there is an argument that if you change a profile config it can no longer be the same thing, and so should essentially lose its memory, but that's probably a bigger discussion.

mattf-horton: \@ All, please note that my overall review comment (as opposed to these detailed comments) weren't published to email by github.  Please see it at https://github.com/apache/metron/pull/622#pullrequestreview-51932717 
It does address the issue of changed profile specs being different profiles.

@simonellistonball , your question sort of brings up the issue of why we use salts at all.  Apparently the leading bytes are important; if they weren't there would be no reason to use a salt, but we could just trust in the later bytes to sufficiently differentiate the partition/region for the row.  As I understand hbase table partitioning (imperfectly), the more times partitioning has happened, and the broader, more well-mixed set of rowkeys are in the table, then the more likely that rowkeys will end up in the same partition (region) unless they differ early in the byte array.  So, if the hbase instance mostly only has data with similar rowkeys, then the bytes later in the rowkey, such as the profile name, will serve sufficiently.  But as the table fills up with a broader range of rowkeys, as it will with our salting strategy, and as it grows larger over time, then a bunch of inserts with the same salt (leading few bytes) is more likely to land in the same partition.

I'm not an hbase expert, so if you feel I've got this wrong, please clarify.

mattf-horton: I should acknowledge that the current salting strategy is perhaps sufficient, since each active Profile typically records a single row every few _minutes_ , so if 20 or 30 active Profiles bunch up in a single partition, and then they do it again a few minutes later in a different partition, it probably doesn't actually qualify as a "hotspot" -- unlike the problem with unsalted timestamp based rowkeys with thousands of inserts per second.

cestella: Chiming in in favor of a prefixed hash. @mattf-horton  you're right about the hash preventing hotspotting.  Consider the case where we are doing batch updates of profiles ( you have a new profile, you want to backfill data).  In that case, you could be sending possibly quite a bit of puts through.

cestella: The other reason to prefix with a hash is in the situation where you have unbalanced keys.  So, the situation where one profile has a lot of data and other profiles don't end up with as much data.  In the case where we want to anticipate splits, having a uniform distribution of keys help because you can just presplit uniformly (as opposed to pre-splitting with a bias for a given key).  Having the uniformly distributed hash prefixing gives a uniformly distributed key distribution, so you can just presplit uniformly.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
```diff
@@ -0,0 +1,402 @@
...
+            .order(byteOrder)
+            .putShort(MAGIC_NUMBER)
+            .put(VERSION)
+            .putInt(salt.length)
+            .put(salt)
+            .putInt(profileB.length)
+            .put(profileB)
+            .putInt(entityB.length)
+            .put(entityB)
+            .put(groupB)
```
cestella: I'd argue that the lengths should be `short`s, not `int`s.  Remember, the key is replicated for each column, not just for each row, in the HBase internal structure.

cestella:
I do not like the idea of a using a profile `profileSpecSerialNumber`.   I don't think this buys us much except some space savings in hbase at the expense of complexity on the read (we need to keep a profile index resident in memory for lookup that is in sync with hbase).

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
nickwallen: The downside of salting unfortunately is that doing a scan over a time range is tough.  While we don't do scans now, I think expanding the Profiler for more use cases will require us to do scans.  

With a salt, you have to submit multiple scans, one for every possible salt value, and then merge what you get back.  

That is why I was originally against using a salt.  But, of course, a salt has the advantage that @cestella outlined.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/SaltyRowKeyBuilder.java
nickwallen: This does look fishy.  I will open a separate JIRA and track this separately.  Thanks, @mattf-horton!

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
nickwallen: Currently, it prepends each group with its length instead of using a delimiter. There are some unit tests that exercise this.   

Were you thinking this did not work at all or were you suggesting a better way to do it?

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
nickwallen: @mattf-horton A 'length' the number of groups along with the size of each group name is encoded.  See the function `encodeGroup`.

@cestella I would be fine with shorts instead of ints.

mattf-horton:
@cestella , we would not need to keep an index resident in memory.  Most of the time we would just have the active Profiles in memory, exactly as we do today.  You only need to retrieve the Profile by serial number on the rare occasions that you have to decode rowkeys.  That said, it's fine with me to just use the profileHash.  I agree it decreases complexity.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
mattf-horton: @nickwallen , right, now I see that each group value is itself L/V encoded.  I would suggest then that a count of group values should precede the set of L/V encoded group values.  Again it's not strictly necessary, but allows deterministic decoding without depending on groups being the last variable-length element in the rowkey.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
mattf-horton: @nickwallen , I retract this comment.  Sorry I didn't read carefully enough.  Of course L/V encoding is better.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
```diff
@@ -0,0 +1,402 @@
...
+          throw new IllegalArgumentException(String.format("Invalid group length; max allowed '%d', got '%d'", MAX_FIELD_LENGTH, groupLength));
+        }
+
+        // decode the group
+        byte[] groupBytes = new byte[groupLength];
+        buffer.get(groupBytes);
+
+        String group = new String(groupBytes);
+        groups.add(group);
+      }
```
mattf-horton: The "overall review comments" are at https://github.com/apache/metron/pull/622#pullrequestreview-51932717 and may not be visible "above" depending on your current view.

cestella:
@mattf-horton Wouldn't you have to use the serial number to retrieve profiles?

mattf-horton:
@cestella , tl;dr: The discussion of serial numbers is a distraction.  Let's just use the profileHash and forget the serial number.  It was a micro-optimization.

Answer to your question:  Two cases:
- If you have the profileHash, then you can look up the Profile using an hbase wildcard query for rowkey \<profileHash\>\* , and since the profileHash is unique, it will be essentially as efficient as using the full rowkey.
- If you are trying to decode a rowkey and only have the serial number then I stated some assumptions: "The expectation is that we will seldom (almost never) need to reference back to the Profile specification, and the total usage of Profile specs will be human-scale finite, **so it is okay to "scan" the ProfileSpecs table to find the full Profile spec referenced by a profileSN.** If this is not true, use the full hash as both the rowkey in the PeriodSpecs table, and as the reference element in the Profile rowkeys."

cestella:
@mattf-horton Would this approach require scans on read in the critical path?

mattf-horton:
@nickwallen brought up the issue of wildcard queries on our rowkeys.  It has always bothered me that we can't do wildcard queries on groups.  If you have, for example, a single groupBy based on day of week, that's just 7 possible values, and if you want them all you could just do 7 queries and combine them.  But if you have three groupBy's, and they have 7, 31, and 256 possible values, then to simulate a wildcard query you would have to do over 55,000 individual queries!  Of course you would just do an hbase scan, but it would require a full table scan to select the time range desired.

I propose that we re-order the rowkey elements to support prefix queries on Profile and time range, with wildcarding for primarily groups, and secondarily entities, ie:
\<salt\>\<magic\>\<profileHash\>\<period\>\<entity\>\<groups\>

So if I want the results for all rows in a time range regarding entity "192.168.222.123" regardless of group, I can query it, and if I want all rows in a time range regardless of entity value or group, I can query that too, as efficiently as an ordinary time range query.  What do you think?

mattf-horton:
@cestella , 
> Would this approach require scans on read in the critical path?

I don't perceive that decoding rowkeys is on any critical path.  You only need to look up Profile by serial number (or hash) in the case of decoding rowkeys.  No?

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/hbase/DecodableRowKeyBuilder.java
mattf-horton: An example in support: suppose you decode rowkeys, and see a bunch of entity values that look like "192.168.52.122", "10.5.123.52", "192.147.130.204", "172.217.5.110".  Of course those are probably IP addresses.  But are they source addresses or destination addresses?  There's no way to tell without access to the Profile spec.

Group values, of course, are potentially even worse.

justinleet:
@nickwallen I haven't been following this discussion, but it seems like a useful feature / enhancement that's been hanging out awhile after active discussion petered out. What are the next steps here?  Does this PR need changes?  Should the discussion be revived on the user lists?  It doesn't seem like there was any consensus on the approach, but again, I like this enhancement a lot.

nickwallen:
There was a lot of good discussion on this, but I find it hard to summarize completely the positions of everyone including @mattf-horton and @cestella.  Here is my attempt in trying to do that.  Please correct anything that I have misstated.

1. Everyone agreed that a ToC (table of contents) is a useful additional feature for the Profiler.  The decodable row key would be needed in addition to, not instead of, a ToC.

1. In implementing a decodable row key, we do need to plan for future changes in row key format.  This was handled in this PR, but can be improved.

1. The decodable row key feature should be completed **before** a ToC so that the row keys can be used to generate (or regenerate) a ToC on-demand.

1. There were various suggestions made on how to shorten up the row key format.  Some of those I completed on this PR (like using a murmur hash) and others (like using shorts instead of ints) I would need to incorporate in a future PR for a decodable row key.

1. There is a need for a migration tool.  A tool that can read the existing row key format and rewrite the same data using a new format.  This tool is necessary even if it cannot be implemented deterministically with the current row key format.  The tool may not hints from the user like the names of known profiles.


Once I compile a summary of these changes, I will close this PR.  All enhancements around this will be implemented on new PRs.

PR #629, opened 2017-07-04
METRON-1012 Update Metron public web site for 0.4.0 release

mattf-horton:
## Contributor Comments
Although 52 files were nominally involved, I really only changed two things:
1. replaced the 0.3.1 current-book with the new 0.4.0 site-book, via direct copy from the release image (that's 51 of the files!)
2. updated documentation/index.md to change 0.3.1 to 0.4.0 and get rid of vestigial uses of "incubator" and "incubating"

### For all changes:

- [x] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? (Actually, in this case, the site.) 

Yes, the web site was rendered via the instructions at https://cwiki.apache.org/confluence/display/METRON/Website+PR+Merge .  It seems okay, with the expected change in links to 0.4.0.

Requires Ruby version 2.2+
* sudo gem install bundler
* sudo gem install jekyll
From incubator-metron/site execute
* bundle install
* bundle exec jekyll serve
Navigate to http://localhost:4000 to review the site
This will create the static site in the target directory.


justinleet:
+1. I ran up the site and checked it over.  Good to see incubating finishing working its way out of everything.

Minor note, the instructions mention going to incubator-metron/site. Anybody who checked it out (or adjusted their dirs) post graduation should look at metron/site.

JonZeolla:
Are we still planning to keep the old versions of the book on the website, like under /0.3.1/ instead of /current-book/ ?  We previously discussed this [here](https://lists.apache.org/thread.html/9b5d34da2b6c6a6266b1d42c94d76e43f308bd533af66cfb36c04f68@%3Cdev.metron.apache.org%3E).

mattf-horton:
Hi @justinleet , thanks for looking at this.  I removed instances of "incubator" from Website+PR+Merge, and had already fixed other cwiki pages I came across.  Any other instances you noticed?

@JonZeolla , once published to dist/release/metron, all older versions remain available in http://archive.apache.org/dist/metron/ .  This includes the book-site for each version starting with 0.3.1.  So the old docs remain available.  (Interestingly, the incubator-metron archives didn't migrate, and are still at http://archive.apache.org/dist/incubator/metron/ .  But those older versions didn't have site-books included.)

I'd like to see a pull-down list of them (both old books and old releases) added to the public web site, but I'm probably not the right person to do that, since I have the web design skills of a cow.  Could you please open a new jira so we can address that outside of the release mechanics?

justinleet:
@mattf-horton To the best of my knowledge, this was the last pass that gets rid of the remaining incubator terminology.  Only things left are old news items and so on that are actually from or about incubation. Thanks for taking care of the wiki, too.  I hadn't looked at it enough for incubating stuff.

JonZeolla:
@mattf-horton Okay great, sounds good to me.  I did a quick search before opening a new JIRA and it looks like you've already addressed this in [METRON-718](https://issues.apache.org/jira/browse/METRON-718).

ottobackwards:
+1

mattf-horton:
Hi @JonZeolla , you're quite right, I forgot I already opened METRON-718.
Are we okay to push this patch then?

JonZeolla:
I didn't review it, but I have no reason to hold it back.

mattf-horton:
Thanks, I'll proceed on @justinleet and @ottobackwards 's +1s.

PR #635, opened 2017-07-06
METRON-1018: Integration tests should reference flux yaml and property files deployed by Ambari

merrimanr:
## Contributor Comments
This PR makes a couple of improvements to our integration tests.  

Enrichment and integration tests now set properties using the same templates as the Ambari MPack.   This will make our tests easier to maintain because there is now a single source of properties.  Any addition/subtraction/modification to those properties will be detected in the tests.

The EnrichmentIntegrationTest now uses a single flux file instead of having a dedicated test flux file.  This was problematic because a) two flux files needed to be maintained for one topology and b) the test flux file could easily become stale and no longer accurately represent the enrichment topology (this happened to me on METRON-990).  The introduction of Stellar (this test was written a while ago) makes it possible to simulate an error so a modified topology/flux file with an error generating bolt is no longer required.

As a result of adding an error-generating Stellar statement I encountered a bug in the GenericEnrichmentBolt.  The exception thrown by the Stellar adapter was being reported twice: once on line 235 and again on line 275.  I attempted to fix this by removing the "error" flag and just reporting the error right when it happens.  There are likely multiple ways to do this so I want to highlight this class for reviewers.  Suggestions and scrutiny needed here.

Testing this is simple:  all tests should pass without issue.  Any change to the templates or enrichment flux file should break the tests.  I also ensured full dev starts up without issue. 

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchIndexingIntegrationTest.java
```diff
@@ -102,11 +102,16 @@ public void setAdditionalProperties(Properties topologyProperties) {
...
     topologyProperties.setProperty("es.port", "9300");
     topologyProperties.setProperty("es.ip", "localhost");
-    topologyProperties.setProperty("indexing.writer.class.name", "org.apache.metron.elasticsearch.writer.ElasticsearchWriter");
+    topologyProperties.setProperty("indexing_writer_class_name", "org.apache.metron.elasticsearch.writer.ElasticsearchWriter");
   }
 
   @Override
   public String cleanField(String field) {
     return field;
   }
```
ottobackwards: I am not sure about having things live in ambari vs. things being referenced by ambari. That doesn't seem right, like a violation of a separation of concerns.  I know we do this other places too ( like the ES Templates, although I'm trying to move away from that ).   I am not sure I have a different way to suggest, but I want to bring up the point.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchIndexingIntegrationTest.java
```diff
@@ -102,11 +102,16 @@ public void setAdditionalProperties(Properties topologyProperties) {
...
+
```
ottobackwards: This could be in test utils?

merrimanr: This was there previously and not something I added.  It's an implementation of an abstract method in IndexingIntegrationTest and it looks like this implementation differs from the one in SolrIndexingIntegrationTest.  

I don't see a way to move this to a util class without significantly changing IndexingIntegrationTest.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchIndexingIntegrationTest.java
merrimanr: I think it's a valid concern.  I'm also not sure what the ideal alternative is.  At this point I feel like Metron has succumbed to being tightly coupled with Ambari.

ottobackwards: this will be the first time pulling things out of ambari ( or metron-deployment really ) into the project for test or other.  I think we should think about alternative approaches

ottobackwards: I would feel better keeping the template with the code and doing a resource copy into the ambari template area, hoping someone else will chime in

justinleet: I do agree it does seem like a violation of separation of concerns, but to be honest, I'm not terribly concerned about it in this case.

The copy-resources seems reasonable, if it's not too much work, but given this particular problem, I'm mostly content with just fixing it.

merrimanr: The latest commits move the templates to their respective modules (metron-enrichment and metron-elasticsearch).  They are copied to metron-deployment at build time with copy-resources.  Let me know if this satisfies the separation of concerns concern.

ottobackwards: yes thank you

justinleet:
I'm +1 by inspection. Thanks for the contribution, it's really nice to not have the potential source of error from these being different.

PR #636, opened 2017-07-06
METRON-1022: Elasticsearch REST endpoint

merrimanr:
## Contributor Comments
The PR adds a simple Elasticsearch REST endpoint that can execute simple queries.  This will serve as the foundation for building more complex Elasticsearch functions.

This can be tested with integration tests or in full dev.  To test is full dev, navigate to Swagger through Ambari and look for the "search-controller".  Here you can perform simple queries against data in Elasticsearch.  For example, posting this query should give back correct search results:
```
{
  "from": 0,
  "indices": [
    "bro", "snort"
  ],
  "query": "*",
  "size": 10,
  "sort": [
    {
      "field": "timestamp",
      "sortOrder": "desc"
    }
  ]
}
```
Note:  Swagger gives an example query with sortOrder = 'DESC'.  As of now this needs to be lowercase but I'm considering changing it to accept both.  

I still need to come up with a good description and add it to the README so that is still pending.  Still wanted to put this up for review as it is fairly large and the first time any Elasticsearch-related code has been added to REST.  Looking forward to the community's feedback.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
I feel like this and the alert UI are building the transcontinental railroad from each end of the county, and are going to miss each other by a state.

How does this relate to that effort?  Have we done any coordination?

From that PR has discussion about NOT having an api direct to the realtime store, but rather abstracting that.  Should this endpoint not be an abstraction?

ottobackwards:
@merrimanr, I'm obviously feeling a little confused about what seems like 3 related things going on.  I don't mean to say this is the wrong approach.  Would you feel comfortable opening a [DISCUSS] thread on REST and or Real Time Search abstraction?

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/ElasticsearchServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+            .size(searchRequest.getSize())
+            .from(searchRequest.getFrom())
+            .query(new QueryStringQueryBuilder(searchRequest.getQuery()))
+            .fetchSource(true)
+            .trackScores(true);
+    for(SortField sortField: searchRequest.getSort()) {
+      FieldSortBuilder fieldSortBuilder = new FieldSortBuilder(sortField.getField());
+      fieldSortBuilder.order(sortField.getSortOrder() == org.apache.metron.rest.model.SortOrder.DESC ? SortOrder.DESC : SortOrder.ASC);
+      searchSourceBuilder = searchSourceBuilder.sort(fieldSortBuilder);
+    }
```
ottobackwards: Is there a limit to how much can be passed back here?  Can the user run a request that returns too much data?
Do we need paging?  Max return count?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
```diff
@@ -60,96 +61,8 @@ public ElasticsearchWriter withOptionalSettings(Map<String, String> optionalSett
   @Override
   public void init(Map stormConf, TopologyContext topologyContext, WriterConfiguration configurations) {
     Map<String, Object> globalConfiguration = configurations.getGlobalConfig();
```
ottobackwards: This is a great idea, to standardize all this!

cestella:
This is good work!  Thanks, Ryan.  We desperately need this abstraction.  I think the confusion may be that this PR is the general abstraction as well as a concrete implementation for one index.  I think, as we found with the indexing topology, that solr and ES won't be able to exist within the same project due to classpath issues (note we have 2 separate projects for solr and ES to isolate the dependencies completely).  I would think that the ES specific stuff may should be separated into a separate project (`metron-rest-elasticsearch`) where the abstraction should exist in `metron-rest`.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/ElasticsearchServiceImpl.java
merrimanr: This is a great suggestion.  We already have paging, just need to enforce page size and provide a sensible default.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
merrimanr: Agreed!

merrimanr:
@ottobackwards just created a discuss thread on the general approach here.

@cestella you're absolutely correct.  We need a way for solr and ES to coexist.  A precedent has already been set in our indexing module but just want to throw another suggestion out there.  The downside of having separate modules that extend a common module is that automated deployment through Ambari MPack is more complex.  In fact, the current MPack doesn't support this right now.  

Ideally we could leverage Spring to easily swap out the implementation at runtime but we would need both Solr and ES classes available on the classpath.  Would it be possible to add lucene classes (and possibly others) to the list of shaded classes in the elasticsearch-shaded module and create a similar module for Solr?  Or are there challenges I'm not thinking of and/or would it make things more confusing having different strategies in different parts of Metron?

ottobackwards:
What kind of isolation we do need?  The 777 introduces bundle loading gives classloader isolation
If they were loaded as plugins then it may work.

merrimanr:
The core issue is that both Solr and Elasticsearch depend on Lucene.  Our version of Elasticsearch (2.3.3) is old so there are lots of Lucene class version conflicts when adding Solr as a dependency.  We have the same problem with Guava because some libraries we depend on use old versions of Guava.  The fact that Guava aggressively deprecates it's API makes it challenging.  We solved this by creating a "shaded" ES module that shades Guava so that it doesn't conflict with the version of Guava used by other libraries.

I'm not that familiar with bundle loading.  How would we use that here?

ottobackwards:
If the es or solr code was packaged as a bundle ( it with it's dependencies -> the es and solr libs ), the bundle system let you load each implementation as a plugin, and will create a new classloader for each.

merrimanr:
The latest commit enforces a maximum page size.  A value of 1000 is set by default.  As it currently stands, you would need to add this property to the rest_application.yml file manually if it needed to be changed.  I will expose it as a property in Ambari as soon as METRON-1017 makes it in.

cestella:
Looking at this implementation and working a bit on the PoC for index data mutation, I think the abstraction here isn't in the right place.  It's too bound-up in the REST layer whereas data access has traditionally existed further down in the abstraction layer.  It's a smell that metron-rest has to depend on ES at all.  That dependency should be provided at runtime entirely.  Also, we're going to need data access in batch as well and I'd like a sensible abstraction in place to enable that (e.g. merging modified records when reading as part of a Spark or MR job).

It seems to me that we need a DAO abstraction living in `metron-writer` or even `metron-indexing` for interacting with indices with the concrete implementations existing in `metron-elasticsearch` and `metron-solr` for ES and Solr respectively.  I think `metron-rest` should work entirely on abstract classes.  The REST layer should choose the DAO implementation via reflection and we should adjust the classpath in the shell script to include one of `metron-elasticsearch` or `metron-solr`.  In essence, I'm recommending something similar to how JDBC works where you choose your driver implementation via reflection (often) and work with base JDBC classes from there.

cestella:
I just want to follow-up with something a bit more concrete suggestions.  I think the beginnings of an abstraction are there.  You pulled out a bunch of utility methods from `ElasticsearchWriter` which is good.  I'm looking for a DAO-like abstraction in `metron-elasticsearch` and the `ServiceImpl` in the REST layer uses the generic DAO implementation (which would not live in the REST layer so it's usable for lower level access) and the index impl is specified by at runtime.  The impl for the DAO would live in one of `metron-elasticsearch` or `metron-solr`.

Again, these are my $0.02 and just suggestions.

ottobackwards:
@cestella 
https://issues.apache.org/jira/browse/METRON-956
Is the jira I had created on this topic.

cestella:
@ottobackwards agreed, we should work to that goal.  I think we might want to make baby steps, though, and create the abstractions first and move to making it pluggable later.

cestella:
Bear with me, @merrimanr, I am going to submit a PR with the DAO abstraction I was talking about so we can hash it out.  I started it in collaboration with @justinleet to ensure the ideas in the mutating indexed data and the meta alerts discuss threads made sense.  I'll just contribute the core abstraction here.  Unfortunately I won't be able to get around to it until Tuesday (vacation and all ;).

cestella:
Ok, this appears to me like good infrastructure to base further index-oriented REST calls atop of.  I'm +1; good job, @merrimanr !

merrimanr:
What do you think @ottobackwards?  Are you good with this?

merrimanr:
I added this endpoint to the README with the latest commit and also tested on full dev with Kerberos enabled.

ottobackwards:
+1 - great job

PR #641, opened 2017-07-10
METRON-539: added HASH function for stellar.

jjmeyer0:
## Contributor Comments
Added a hash function to stellar. Many unit tests were created. Along with these tests I opened the Stellar terminal and did the following:

### Start Stellar
```bash
mvn exec:java -Dexec.mainClass="org.apache.metron.stellar.common.shell.StellarShell" -pl metron-stellar/stellar-common/
```

### Test `HASH` Function

Below is the output of some of the tests I ran. To get all available inputs for the second argument you can call `Security.getAlgorithms("MessageDigest")` and print out the array. However, 'SHA-1', 'MD5', and 'SHA-256' are required by all Java platforms. Please see [MessageDigest](https://docs.oracle.com/javase/8/docs/api/java/security/MessageDigest.html) and [Security](https://docs.oracle.com/javase/8/docs/api/java/security/Security.html). 

```bash

[Stellar]>>> some_string := 'my string to hash'
[Stellar]>>> HASH(some_string, 'md5')
E08863A4BE232F166C350F45C9DAA586
[Stellar]>>> some_serializable_object := [ 1, 2, 3 ]
[Stellar]>>> HASH(some_serializable_object, 'MD5')
2383CD011F1DA06A173ADE2F3F99CF62
[Stellar]>>> 
```

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.metron.stellar.dsl.functions;
+
+import org.apache.commons.codec.EncoderException;
+import org.apache.commons.codec.binary.Hex;
+import org.apache.metron.stellar.common.utils.hashing.DefaultHasher;
+import org.apache.metron.stellar.dsl.BaseStellarFunction;
+import org.apache.metron.stellar.dsl.Stellar;
+
+import java.nio.charset.StandardCharsets;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+import java.util.ArrayList;
+import java.util.List;
+
+public class HashFunctions {
+
+  @Stellar(
+    name = "GET_HASHES_AVAILABLE",
+    description = "Will return all available hashing algorithms available to 'HASH'.",
+    returns = "A list containing all supported hashing algorithms."
+  )
+  public static class ListSupportedHashTypes extends BaseStellarFunction {
+
+    @Override
+    public List<String> apply(final List<Object> args) {
+      if (args == null || args.size() != 0) {
+        throw new IllegalArgumentException("Invalid call. This function does not expect any arguments.");
+      }
+
+      return new ArrayList<>(Security.getAlgorithms("MessageDigest"));
+    }
+  }
+
+  @Stellar(
+    name = "HASH",
+    description = "Hashes a given value using the given hashing algorithm and returns a hex encoded string.",
+    params = {
+      "toHash - value to hash.",
+      "hashType - A valid string representation of a hashing algorithm. See 'GET_HASHES_AVAILABLE'.",
+    },
+    returns = "A hex encoded string of a hashed value using the given algorithm. If 'hashType' is null " +
+      "then '00', padded to the necessary length, will be returned. If 'toHash' is not able to be hashed or " +
+      "'hashType' is null then null is returned."
+  )
+  public static class Hash extends BaseStellarFunction {
+
+    @Override
+    public Object apply(final List<Object> args) {
+      if (args == null || args.size() != 2) {
+        throw new IllegalArgumentException("Invalid number of arguments: " + (args == null ? 0 : args.size()));
+      }
+
+      final Object toHash = args.get(0);
+      final Object hashType = args.get(1);
+
+      if (hashType == null) {
+        return null;
+      }
+
+      try {
+        return new DefaultHasher(hashType.toString(), new Hex(StandardCharsets.UTF_8)).getHash(toHash);
+      } catch (final EncoderException e) {
+        return null;
+      } catch (final NoSuchAlgorithmException e) {
+        throw new IllegalArgumentException("Invalid hash type: " + hashType.toString());
+      }
+    }
+  }
+}
--- a/metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
+++ b/metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
@@ -1,0 +1,205 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.metron.stellar.dsl.functions;
+
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.lang.SerializationUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Test;
+
+import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.Security;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static org.apache.metron.stellar.common.utils.StellarProcessorUtils.run;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class HashFunctionsTest {
+  static final Hex HEX = new Hex(StandardCharsets.UTF_8);
+  final HashFunctions.ListSupportedHashTypes listSupportedHashTypes = new HashFunctions.ListSupportedHashTypes();
+  final HashFunctions.Hash hash = new HashFunctions.Hash();
+
+  @Test(expected = IllegalArgumentException.class)
+  public void nullArgumentsShouldFail() throws Exception {
+    listSupportedHashTypes.apply(null);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void getSupportedHashAlgorithmsCalledWithParametersShouldFail() throws Exception {
+    listSupportedHashTypes.apply(Collections.singletonList("bogus"));
+  }
+
+  @Test
+  public void listSupportedHashTypesReturnsAtMinimumTheHashingAlgorithmsThatMustBeSupported() throws Exception {
+    final List<String> requiredAlgorithmsByJava = Arrays.asList("MD5", "SHA", "SHA-256"); // These are required for all Java platforms (see java.security.MessageDigest). Note: SHA is SHA-1
+    final Collection<String> supportedHashes = listSupportedHashTypes.apply(Collections.emptyList());
+    requiredAlgorithmsByJava.forEach(a -> assertTrue(supportedHashes.contains(a)));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void nullArgumentListShouldThrowException() throws Exception {
+    hash.apply(null);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void emptyArgumentListShouldThrowException() throws Exception {
+    hash.apply(Collections.emptyList());
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void singleArgumentListShouldThrowException() throws Exception {
+    hash.apply(Collections.singletonList("some value."));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void argumentListWithMoreThanTwoValuesShouldThrowException3() throws Exception {
+    hash.apply(Arrays.asList("1", "2", "3"));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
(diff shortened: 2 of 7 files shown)


metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+        assertEquals(BaseEncoding.base16().encode(m.digest()), actual);
+      } catch (NoSuchAlgorithmException e) {
+        throw new RuntimeException(e);
+      }
+    });
+  }
+
+  @Test
+  public void nonStringValueThatIsSerializableShouldBeSerializedAndThenEncodedUsingHex() throws Exception {
```
mattf-horton: This is another nit, but I'll share the thought anyway:  A more typical (and shorter) name would be "testNonStringValueThatIsSerializable".  Being a "test", it is understood that the normative result is to properly do whatever it should do without error.  Similarly, the next two items could more briefly be named "testHashFunctionsWithVariablesAsInput" and "testStellarFunctionWhereOnlyHashTypeIsAVariable".

The last one, "aNonNullObjectThatDoesNotImplementSerializableShouldReturnAValueOfNull" is for a non-normative case, so it provides useful information as is.  It could still be shorter though.  How about "testNonSerializableObjectReturnsNull"?

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
```diff
@@ -0,0 +1,169 @@
...
+    final Set<String> algorithms = Security.getAlgorithms("MessageDigest");
+
+    algorithms.forEach(algorithm -> {
+      try {
+        final MessageDigest m = MessageDigest.getInstance(algorithm);
+        m.update(valueToHash.getBytes(StandardCharsets.UTF_8));
+
+        assertEquals(BaseEncoding.base16().encode(m.digest()), hash.apply(Arrays.asList(valueToHash, algorithm)));
+      } catch (NoSuchAlgorithmException e) {
+        throw new RuntimeException(e);
```
jjmeyer0: @mattf-horton since I'm using a lambda, I don't believe this is possible since `NoSuchAlgorithmException` is a checked exception.

mattf-horton: @jjmeyer0 , quite right!  Thanks, I learned something :-)
For those not already familiar with this, it isn't that it's a lambda, it's that it's a lambda being used as the Consumer argument to a .forEach call, and Consumer.accept() is contractually specified to not throw checked exceptions.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+      "strings and values that implement java.io.Serializable.",
+    params = {
+      "toHash - value to hash.",
+      "hashType - A valid string representation of an algorithm supported by java.security.MessageDigest.",
+    },
+    returns = "A hex representation of a hashed value using the given hashing algorithm. If either argument is " +
+      "null then null will be returned. If the type of 'toHash' is neither a string nor of type java.io.Serializable, " +
+      "then null is returned."
+  )
+  public static class Hash extends BaseStellarFunction {
```
ottobackwards: My pr for encoding has this, but it returns a comma delimited string

mattf-horton: Well, SPLIT and JOIN let you move back and forth between the two formats, but wouldn't the List format be the most naturally usable one?

ottobackwards: I have no preference.  I picked the string thinking of the return of the call being for 'information' and display only.  Like from the shell.  I think that we should do the 'get' no params calls the same though, i'd change to list if that is what makes sense

ottobackwards: I changed mine to GET_ instead of LIST_ and return a List<String>
So we shall be consistent.

jjmeyer0: @ottobackwards I'll make sure to stay consistent with what you have.

mattf-horton: Thanks @ottobackwards , I agree.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.List;
+
+public class HashFunctions {
+  @Stellar(
+    name = "HASH",
```
ottobackwards: I am not sure that java.io.Serializable should bleed through here.  While we may limit ourselves in some places to having stellar conform to java limitations and behaviors, and other things, we don't really limit ourselves explicitly to java types.  The idea that the user 'knows' the type of the entity in a stellar statement in java.... doesn't seem right.

While this limitation may be true and enforced, I am not sure it should be expressed this way explicitly.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+    description = "Hashes a given value using the given hashing algorithm and returns a hex encoded string. This function only hashes " +
+      "strings and values that implement java.io.Serializable.",
+    params = {
+      "toHash - value to hash.",
+      "hashType - A valid string representation of an algorithm supported by java.security.MessageDigest.",
+    },
```
ottobackwards: may want to call out that it is a string of the hex value

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private MessageDigest getMessageDigest(String hashType) {
+      try {
+        return MessageDigest.getInstance(hashType);
+      } catch (NoSuchAlgorithmException e) {
+        throw new IllegalArgumentException("Invalid hash type: " + hashType);
+      }
+    }
+
```
ottobackwards: Why is this Object and not String?  Can't we just be clear and use the real type?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+        return getHashAsHex(messageDigest, toHash.toString().getBytes(StandardCharsets.UTF_8));
+      } else if (toHash instanceof Serializable) {
+        final byte[] serialized = SerializationUtils.serialize((Serializable) toHash);
+        return getHashAsHex(messageDigest, serialized);
+      }
+
+      return null;
+    }
+
+    private MessageDigest getMessageDigest(String hashType) {
```
ottobackwards: Is it possible that the MessageDigest and hashing functions may be usable to other Stellar functions?  I think they may.  If so, then putting them into a stellar/common/utils/hashing/xxxx area would be better, then you are just making Stellar functions exposing those functions, and not 'burying' them in the class.

Disclaimer: this is what I did with the Encodings helper class, for base32,64,binary,hex encodings

ottobackwards: Then again, I don't usually think of using the Stellar static classes on their own, or from 'other' stellar functions, maybe I'm not thinking of that correctly

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
mattf-horton: Still, if it's not serializable, what should HASH do?  This may be moot, since all normal objects Stellar can get its hands on are, I think, Serializable.  The exceptions may indeed be appropriate to not be able to hash.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
mattf-horton: If we expose MessageDigest objects, that would be a good example of non-Serializable objects it is reasonable to be unable to hash :-)

ottobackwards: I am not sure I understand :( right over my head.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
ottobackwards: My point isn't that the behavior or restriction in practice should change, but that the parameter documentation should be different.

mattf-horton: Ah, sorry I didn't understand this was only a comment about documentation.  I guess you're right; since most objects are indeed serializable, maybe it's okay to let them discover the occasional object that isn't (if any).

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/HashFunctions.java
mattf-horton: Never mind, I think I misunderstood you again. You were probably just proposing the code be in a separate directory instead of in the generic stellar/dsl/functions, right?  If so then I'm in left field here.  Prob didn't get enough sleep last night...

ottobackwards: I have this [Encodings.java](https://github.com/apache/metron/pull/642/files#diff-21626fcded1721297dd93114ccfd5ba1)  and [EncodingFunctions.java](https://github.com/apache/metron/pull/642/files#diff-295147ddc21d558ddd7af7f31d9bb8ca) fresh in my mind.

ottobackwards: if you were on irc we wound't have these problems ;)

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
jjmeyer0: @mattf-horton I understand that many of these are long. This is somewhat intentional on my part. I personally don't like the `test*` pattern. IIRC this is the way junit used to pick up tests. Since this isn't necessarily needed anymore, I prefer to use these longer form names. My reason is, when I write tests, I actually write a lot these before I actually start coding. It helps me think through things. That being said, I do agree that some of them are a quite long, and I will re-word them. Also, I'm not completely tied to it. If you still prefer the shorter "test*" let me know and I will update it.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
jjmeyer0: @mattf-horton thanks for the clarification. I never took the time to read more into it. You're comment made me take the time. thanks again!

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+  @Test
+  public void invalidNullAlgorithmArgumentShouldThrowException() throws Exception {
+    assertNull(hash.apply(Arrays.asList("value to hash", null)));
+  }
+
+  @Test
+  public void nullInputForValueToHashShouldProperlyThrowException() throws Exception {
+    assertNull(hash.apply(Arrays.asList(null, "md5")));
+  }
```
jjmeyer0: @mattf-horton just to clarify. The tests `invalidNullAlgorithmArgumentShouldThrowException` and `nullInputForValueToHashShouldProperlyThrowException` do not through exceptions. They return null. I originally had them throw exceptions, but changed it to return null, but forgot to change the names. I'll address this.

That being said, are you saying you to return a hex encoded string of "\u0000" when `toHash` is null?

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
mattf-horton: As noted, it's a nit.  Do as you choose, it won't interfere with a +1 from me, after the substantive items are taken care of.  My inclination, though, is that verbiage (beyond what's needed for clarity) should go in comments.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/HashFunctionsTest.java
mattf-horton: Ah, I was trusting your very detailed testcase names :-)  If they don't throw exceptions then of course the annotation should be consistent.

I remain of the opinion that it would be good for HASH to be able to return a conforming result for null objects, ie "00" instead of null.  But this is mostly an aesthetic judgment on my part, so I could be talked out of it.

You gave me a headache by prefixing the "00" string with the unicode symbol :-) and no, "\u0000" would not be the desirable return value; that would be the unicode codepoint for a single null character.  Yuck!  But I shouldn't have used the hex prefix "0x" either.  Since we are returning unadorned string values representing hex values, it should be "00" (a pair of zero digit characters in a String of length 2).

mattf-horton: Um, but if the hash values are always supposed to be of a certain length, then pad that out to as many zero digits as necessary to be conforming.

jjmeyer0: oh boy, @mattf-horton that makes much more sense. forgive me i had a long day yesterday. I had a moment haha

cestella:
I just want to chime in and say thanks for the contribution @jjmeyer0 !. :)

JonZeolla:
Yeah, this is exactly what I had in mind when I put the ticket in, thanks @jjmeyer0 .  

I would love to get some LSH and CCTP hashes added, but I think that's more reasonable as a follow on.

cestella:
Ok, I'm +1 on this by inspection as soon as we have the conflict resolved.  Great job, @jjmeyer0

ottobackwards:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/apache/metron/641#-:-KrWliAsg4-eqkjCEHxu:bnfp4nl)*

PR #644, opened 2017-07-10
METRON-1027: Errant Log Message When No GeoIP Hit on Valid, External IPv4 Address

nickwallen:
When a valid public IPv4 address does not match a record in the Maxmind GeoIP database, a log message indicates that an error occurred with the GeoLite2 DB.  

```
2017-07-05 12:50:34.266 o.a.m.e.a.g.GeoLiteDatabase WARN Metron GeoLite2 DB encountered an error
com.maxmind.geoip2.exception.AddressNotFoundException: The address X.X.X.X is not in the database.
at com.maxmind.geoip2.DatabaseReader.get(DatabaseReader.java:157) ~[stormjar.jar:?]
at com.maxmind.geoip2.DatabaseReader.city(DatabaseReader.java:202) ~[stormjar.jar:?]
at org.apache.metron.enrichment.adapters.geo.GeoLiteDatabase.get(GeoLiteDatabase.java:136) [stormjar.jar:?]
at org.apache.metron.enrichment.adapters.geo.GeoAdapter.enrich(GeoAdapter.java:47) [stormjar.jar:?]
at org.apache.metron.enrichment.adapters.geo.GeoAdapter.enrich(GeoAdapter.java:31) [stormjar.jar:?]
```

Really no error has occurred with the GeoIP database.  Not having a hit in the database is an expected condition and processing should continue without the errant log message.

### Testing

To test this change, setup a Metron environment with GeoIP enrichment enabled.  Ingest some telemetry containing an valid, external IPv4 address that is not contained within the GeoIP database.   You can use an IP like `203.0.113.1` as the range 203.0.113.0/24 is reserved for "TEST-NET-3" and should never be locatable.  Before this PR, you will see the noted error message in the Enrichment topology logs.  With this change, you will not see the error message in your logs.

## Pull Request Checklist

### DIFF
--- a/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/geo/GeoAdapter.java
+++ b/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/geo/GeoAdapter.java
@@ -46,7 +46,6 @@ public class GeoAdapter implements EnrichmentAdapter<CacheKey>, Serializable {
     JSONObject enriched = new JSONObject();
     Optional<HashMap<String, String>> result = GeoLiteDatabase.INSTANCE.get(value.coerceValue(String.class));
     if(!result.isPresent()) {
-      _LOG.error("GEO Enrichment failure: {}", value.coerceValue(String.class));
       return new JSONObject();
     }
 
--- a/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/geo/GeoLiteDatabase.java
+++ b/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/geo/GeoLiteDatabase.java
@@ -19,6 +19,7 @@ package org.apache.metron.enrichment.adapters.geo;
 
 import com.maxmind.db.CHMCache;
 import com.maxmind.geoip2.DatabaseReader;
+import com.maxmind.geoip2.exception.AddressNotFoundException;
 import com.maxmind.geoip2.exception.GeoIp2Exception;
 import com.maxmind.geoip2.model.CityResponse;
 import com.maxmind.geoip2.record.City;
@@ -162,7 +163,7 @@ public enum GeoLiteDatabase {
       }
 
       return Optional.of(geoInfo);
-    } catch (UnknownHostException e) {
+    } catch (UnknownHostException | AddressNotFoundException e) {
       LOG.warn("[Metron] No result found for IP {}", ip);
     } catch (GeoIp2Exception | IOException e) {
       LOG.warn("[Metron] GeoLite2 DB encountered an error", e);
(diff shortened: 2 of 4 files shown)


nickwallen:
@justinleet Another thing I noticed while in the code is that a lookup of a private IP address is expected to return an empty map.  Since we're using an Optional, wouldn't it make more sense for it to return `Optional.empty()`?  

The javadocs say that `Optional.empty()` indicates a lookup miss, not an empty map.  I could either make it return `Optional.empty()` or I could update the docs so that they match current behavior.  Let me know what you think.

Thanks

metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/geo/GeoAdapter.java
```diff
@@ -46,7 +46,6 @@ public JSONObject enrich(CacheKey value) {
     JSONObject enriched = new JSONObject();
     Optional<HashMap<String, String>> result = GeoLiteDatabase.INSTANCE.get(value.coerceValue(String.class));
     if(!result.isPresent()) {
-      _LOG.error("GEO Enrichment failure: {}", value.coerceValue(String.class));
```
justinleet: Can you make it log.debug()?  I think it's still useful to be able to tell what happened, it just shouldn't be error.

justinleet:
@nickwallen Off the top of my head, I think it could be Optional.empty(), but I'd like to see a debug log added so we can easily tell the difference between "IP not found" and "It wasn't even something we should have looked up".  Double check the logic on the function, but it should essentially be the same thing.

justinleet:
Specifically, I'd say add it in 
```
    if (isIneligibleAddress(ip, addr)) {
      return Optional.of(new HashMap());
    }
```

ottobackwards:
+1 by inspection.  I agree that the log should be moved in level and not dropped

metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/adapters/geo/GeoAdapter.java
nickwallen: We already get logging in this case from GeoLiteDatabase:167.  I didn't think we needed another.

justinleet:
+1 by inspection.  @nickwallen is right that the log already occurs in the GeoLiteDatabase, so I'm okay with dropping it.

nickwallen:
Thanks for the review guys.  Just trying to understand why Travis is reporting a unit test failure.  Its like it is not running the latest code.  Bah.

nickwallen:
Oh, nevermind.  Got it.  Sorry, Travis.  Thank you for your service.

nickwallen:
Also, ran this up on Full Dev and did some quick validation on `GEO_GET` just to be sure.

```
[Stellar]>>> GEO_GET("192.168.0.1")
2017-07-11 15:11:13,401 INFO  [AeshProcess: 3] stellar.GeoEnrichmentFunctions: Initializing GeoEnrichmentFunctions
2017-07-11 15:11:13,790 WARN  [AeshProcess: 3] util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
2017-07-11 15:11:14,636 WARN  [AeshProcess: 3] shortcircuit.DomainSocketFactory: The short-circuit local reads feature cannot be used because libhadoop cannot be loaded.
{}
[Stellar]>>> GEO_GET("75.118.136.186")
{country=US, dmaCode=535, city=Columbus, postalCode=43230, latitude=40.0345, location_point=40.0345,-82.8695, locID=4509177, longitude=-82.8695}
```

PR #646, opened 2017-07-11
METRON-1030: Metron JDBC platform REST option should accept a blank value

merrimanr:
## Contributor Comments
Ambari should allow the metron_jdbc_platform option to be set to '' which it currently does not.  To verify, you should be able to spin up full dev, navigate to the REST tab and set that property to ''.  This is needed when configuring MySQL for Metron REST because this property must be '' and not 'mysql'.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
@@ -139,9 +139,12 @@
     </property>
     <property>
         <name>metron_jdbc_platform</name>
-        <value></value>
         <description>Database platform used by Metron.  One of: hsqldb, h2, oracle, mysql, postgresql</description>
         <display-name>Metron JDBC platform</display-name>
+        <value/>
+        <value-attributes>
+            <empty-value-valid>true</empty-value-valid>
+        </value-attributes>
     </property>
     <property>
         <name>metron_jdbc_client_path</name>


nickwallen:
When would a user want to set this property to an empty string?

ottobackwards:
^^^^^^^ 
" This is needed when configuring MySQL for Metron REST because this property must be '' and not 'mysql'."

merrimanr:
Spring provides a mechanism for running sql scripts for different databases on startup.  This is useful for initializing tables and such.  We provide 2 right now:  schema-h2.sql and schema-mysql.sql.  If you set this property to "h2" it will run the schema-h2.sql script automatically at startup.  Same goes for mysql.  It's possible that someone may want to use a different database and in that case they would want to set this to a blank string.

ottobackwards:
That seems like a configuration that is rather confusing... 
Is there a way we can clarify that?

lvets:
"This is needed when configuring MySQL for Metron REST because this property must be '' and not 'mysql'.". So this needs to be empty when you use mysql? I'm not following... :)

merrimanr:
The comment "This is needed when configuring MySQL for Metron REST because this property must be" is actually incorrect now that I go back and look at it.  The schema-mysql.sql was initially incorrect because it was dropping the tables (MySQL doesn't support "create table if not exists") and recreating them every time.  

Fortunately Spring just ignores any errors thrown when running these scripts so the drop statements were just removed and creating a table that already existed is safely ignored.  This is just a convenience feature that keeps people from having to manually create database assets.

merrimanr:
@ottobackwards it's in metron-rest README but I can expand it a little bit.

ottobackwards:
It needs to be documented in the ambari ui, if it needs clarification.  The user in ambari isn't going to be using the readme

ottobackwards:
this is a general problem I think

merrimanr:
Here's the challenge with all this.  Maybe we can collaborate on the right approach.

The REST app requires a relational database and setting that up is a prerequisite, unless you want to use h2.  We can provide a way to automatically run a script and make this setup easier (we do that for h2 and mysql).  But it's likely that different orgs will want to use their own database and we probably don't want to try and maintain scripts for every possible database out there.

So what's the best approach?  Do we:

- Remove the convenience scripts and document setup as best we can (again providing the exact SQL statements for every database is not feasible).  In this case we would just remove this metron_jdbc_platform property from Ambari because it's not needed anymore.
- Limit our database support to a handful of databases and provide setup scripts for those?  In this case the metron_jdbc_platform could be a dropdown with the databases we support but users would be forced to use one of those.
- Keep it like it is now, provide a convenient way to initialize MySQL with other databases requiring manual setup.  In this case metron_jdbc_platform would accept h2, mysql, or empty meaning the user is responsible for setup.

Either way, the READMEs need to be improved which probably includes moving this database setup prerequisite to a higher level so that it's more obvious and providing more detail around manual setup.

ottobackwards:
What do other HDP / ambari configured services do to support multiple databases? There are other services/projects we can look to for ideas

ottobackwards:
Your fix, plus some clarity in the documentation / ambari label etc + some idea that we need to think of something long term would be good right?

merrimanr:
Hive provides a list of supported databases:  https://cwiki.apache.org/confluence/display/Hive/AdminManual+MetastoreAdmin#AdminManualMetastoreAdmin-Local/EmbeddedMetastoreDatabase(Derby)

nickwallen:
I don't think we need any major changes here.  Just something more descriptive.

> Keep it like it is now, provide a convenient way to initialize MySQL with other databases requiring manual setup. In this case metron_jdbc_platform would accept h2, mysql, or empty meaning the user is responsible for setup.

Yes, something along these lines (^^^)  is what I think we need.  

More specifically, I would suggest...
* Change the name of `metron_jdbc_platform` to something more descriptive.  Maybe `metron_rest_db_setup`, `metron_rest_ddl`, or something better?  

* Update the brief description in Ambari so that it reflects what you described above.

* Instead of an '' empty option let's call it something along the lines of 'Manual Database Setup'.

*  Updates to the README as you already suggested.

PR #650, opened 2017-07-13
METRON-1038: Stellar should have a better collection of basic math operations

cestella:
## Contributor Comments
At the moment the math functions are woefully incomplete. 
We should add at least the ones difficult or impossible to reconstruct using existing stellar primitives/math functions:
* log10
* log2
* ln
* sqrt
* ceil
* floor
* sin
* cos
* tan

It should be sufficient to test these out in the Stellar REPL for manual evaluation.  I have also refactored and added a more robust suite of unit tests for existing basic math functions as well as these new ones.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  mvn -q clean integration-test install && build_utils/verify_licenses.sh 
  ```

- [x] Have you written or updated unit tests and or integration tests to verify your changes?
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [x] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via `site-book/target/site/index.html`:

  ```
  cd site-book
  mvn site
[...]

PR #651, opened 2017-07-13
METRON-1037 Added POWER function

simonellistonball:
## Contributor Comments

This is a quick addition to the Math functions. It may be worth revising following the work @cestella did this morning to add a more general implementation on top of #650 

I will probably add 2 argument capabilities and port this work to #650 when it's in once that is in.

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

### DIFF
--- a/metron-stellar/stellar-common/README.md
+++ b/metron-stellar/stellar-common/README.md
@@ -154,6 +154,7 @@ In the core language functions, we support basic functional programming primitiv
 | [ `MAP`](#map)                                                                       |
 | [ `MAP_EXISTS`](#map_exists)                                                                       |
 | [ `MONTH`](#month)                                                                                 |
+| [ `POWER`](#power)                                                                                 |
 | [ `PREPEND_IF_MISSING`](#prepend_if_missing)                                                       |
 | [ `PROFILE_GET`](#profile_get)                                                                     |
 | [ `PROFILE_FIXED`](#profile_fixed)                                                                 |
@@ -517,6 +518,13 @@ In the core language functions, we support basic functional programming primitiv
   * Input:
     * dateTime - The datetime as a long representing the milliseconds since unix epoch
   * Returns: The current month (0-based).
+
+### `POWER`
+  * Description: Raises the number to a given power.
+  * Input:
+    * number - the number
+    * power - the exponent to raise the number to
+  * Returns: The power-th power of the number passed in.
 
 ### `PREPEND_IF_MISSING`
   * Description: Prepends the prefix to the start of the string if the string does not already start with any of the prefixes.
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
@@ -60,4 +60,38 @@ public class MathFunctions {
     }
   }
 
+  @Stellar(name="POWER"
+           ,description="Raises the number to a given power"
+           ,params = {
+                  "number - the number",
+                  "power - the exponent to raise the number to"
+                     }
+           , returns="The power-th power of the number passed in."
+  )
+
+  public static class Power implements StellarFunction {
+    @Override
+    public Object apply(List<Object> args, Context context) throws ParseException {
+      if(args.size() != 2) {
+        return Double.NaN;
+      }
+      Number b = (Number)args.get(0);
+      Number e = (Number)args.get(1);
+      if (b == null || e == null) {
+        return Double.NaN;
+      }
+      return Math.pow(b.doubleValue(), e.doubleValue());
+    }
+
+    @Override
+    public void initialize(Context context) {
+
+    }
+
+    @Override
+    public boolean isInitialized() {
+      return true;
+    }
+  }
+
 }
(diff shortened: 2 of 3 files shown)


mattf-horton:
I would rather see this implemented as a `^` math operator rather than a function, even though Java doesn't.  Opinions?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
```diff
@@ -60,4 +60,38 @@ public boolean isInitialized() {
     }
   }
 
+  @Stellar(name="POWER"
```
cestella: can we call it POW?

simonellistonball:
I get the point @mattf-horton and thought about it, but went this way for familiarity's sake, a lot of the likely authors of stellar statements are security analysts who will be more familiar with languages like SQL or Excel functions, or even python than languages which use the ^ operator (off the top of my head:  urgh... just tried python, scala, java, c, c++, javascript, can't actually think of one that does, though some have **)

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
simonellistonball: I went with POWER to be consistent with SQL and Excel, rather than python, since that's a more likely familiar ground for security and BI people.

cestella:
If we do decide to leave it in @simonellistonball you might want to base it around #650, as you indicated.  I added the 2 arg abstraction.  This should just be:
* Add an entry to MathOperations: `POW( new MathOperation(d -> Math.pow(d[0], d[1]), 2))`
* Follow the example in `MathFunctions` and add the stellar function

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
cestella: hmm, I'm ok with that, I guess.

simonellistonball:
Yes, I think I'll just refactor this around #650 when that's committed to keep the workflow simple.

mattf-horton:
@simonellistonball , you're right, '**' would have been far more appropriate given what most languages do.  And I find your argument convincing that the function form will be more familiar to many.
@cestella , python allows both '**' and "math.exp()".  Java also uses "Math.exp()".  So I recommend we go with EXP().  Did I see in your comment in #650 that you've now added EXP to your PR?

cestella:
@mattf-horton Actually, `**` and `math.exp()` are different;  the first being an arbitrary power function and the later presuming that we're taking a power of `e` (i.e. `e**x == exp(x)` ).  I did add `EXP` (as in `e**x`) in #650

metron-stellar/stellar-common/README.md
```diff
@@ -518,6 +519,13 @@ In the core language functions, we support basic functional programming primitiv
     * dateTime - The datetime as a long representing the milliseconds since unix epoch
   * Returns: The current month (0-based).
 
+### `POWER`
+  * Description: Raises the number to a given power.
+  * Input:
+    * number - the number
+    * power - the exponent to raise the number to
+  * Returns: The power-th power of the number passed in.
```
mattf-horton: Suggest: Returns: The value of number to the given power.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
```diff
@@ -60,4 +60,38 @@ public boolean isInitialized() {
...
+           ,description="Raises the number to a given power"
+           ,params = {
+                  "number - the number",
+                  "power - the exponent to raise the number to"
+                     }
+           , returns="The power-th power of the number passed in."
```
mattf-horton: ditto: "The value of number to the given power."

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/MathFunctionsTest.java
```diff
@@ -44,4 +44,16 @@ public void testAbs() {
     Assert.assertEquals((Double)run("ABS(value)", ImmutableMap.of("value", -10.5)), 10.5, 1e-7);
   }
 
+  @Test
+  public void testPower() {
+    Assert.assertEquals((Double)run("POWER(e,b)", ImmutableMap.of("e", 2.0, "b", 2.0)), 4.0, 1e-7 );
+    Assert.assertTrue(Double.isNaN((Double)run("ABS(value)", ImmutableMap.of("e", Double.NaN, "b", 2.0))));
+    Assert.assertTrue(Double.isNaN((Double)run("ABS(value)", ImmutableMap.of("e", 2.0, "b", Double.NaN))));
+    Assert.assertTrue(Double.isNaN((Double)run("ABS(value)", ImmutableMap.of("e", Double.NaN, "b", Double.NaN))));
+
```
mattf-horton: Were these 3 lines supposed to be POWER(e,b) invocations rather than ABS(value)?

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/MathFunctionsTest.java
```diff
@@ -44,4 +44,16 @@ public void testAbs() {
     Assert.assertEquals((Double)run("ABS(value)", ImmutableMap.of("value", -10.5)), 10.5, 1e-7);
   }
 
+  @Test
+  public void testPower() {
+    Assert.assertEquals((Double)run("POWER(e,b)", ImmutableMap.of("e", 2.0, "b", 2.0)), 4.0, 1e-7 );
```
mattf-horton: Please put expected value first, test result value second, in assertEquals() calls.  Otherwise reported failures look funny.

mattf-horton:
Har, I give up.  POWER it is.  I agree would be good to use @cestella 's new framework, and see above review comments.

PR #657, opened 2017-07-18
METRON-1048 Intermittent Test Failure for SimpleHBaseEnrichmentWriterLoggingTest

nickwallen:
### Problem

This unit test failure seems to occur sporadically.  The same test passed after an inconsequential change forced a re-run of the tests.

```
Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 11.596 sec <<< FAILURE! - in org.apache.metron.writers.SimpleHBaseEnrichmentWriterLoggingTest
shouldWarnOnMissingConfig(org.apache.metron.writers.SimpleHBaseEnrichmentWriterLoggingTest)  Time elapsed: 0.047 sec  <<< FAILURE!
org.mockito.exceptions.verification.TooManyActualInvocations: 
logger.warn(
    "No config object found for key: '{}'",
    "shew.table"
);
Wanted 1 time:
-> at org.apache.metron.writers.SimpleHBaseEnrichmentWriterLoggingTest.shouldWarnOnMissingConfig(SimpleHBaseEnrichmentWriterLoggingTest.java:77)
But was 2 times. Undesired invocation:
-> at org.apache.metron.enrichment.writer.SimpleHbaseEnrichmentWriter$Configurations.get(SimpleHbaseEnrichmentWriter.java:72)

	at org.apache.metron.writers.SimpleHBaseEnrichmentWriterLoggingTest.shouldWarnOnMissingConfig(SimpleHBaseEnrichmentWriterLoggingTest.java:77)
```

### Solution

The test was expecting the call to occur only once, when in-fact it could occur a few times.  This is why it would fail sporadically. This doesn't impact the validity of the test.

### DIFF
--- a/metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/SimpleHBaseEnrichmentWriterLoggingTest.java
+++ b/metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/SimpleHBaseEnrichmentWriterLoggingTest.java
@@ -41,7 +41,6 @@ import java.util.*;
 import static org.junit.Assert.*;
 import static org.mockito.Matchers.*;
 import static org.mockito.Mockito.atLeastOnce;
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.powermock.api.mockito.PowerMockito.*;
 
@@ -75,7 +74,7 @@ public class SimpleHBaseEnrichmentWriterLoggingTest {
   public void shouldWarnOnMissingConfig() {
     SimpleHbaseEnrichmentWriter.Configurations config = getHbaseTableConfig();
     config.get(Collections.emptyMap());
-    verify(logger).warn("No config object found for key: '{}'", config.getKey());
+    verify(logger, atLeastOnce()).warn("No config object found for key: '{}'", config.getKey());
   }
 
   @Test
@@ -99,14 +98,14 @@ public class SimpleHBaseEnrichmentWriterLoggingTest {
     keyTransformer.transform(message);
 
     // Proof the result has been captured
-    verify(logger).debug("Transformed message: '{}'", value);
+    verify(logger, atLeastOnce()).debug("Transformed message: '{}'", value);
   }
 
   @Test
   public void shouldDebugSensorConfig() {
     String sensorName = "someSensor";
     hbaseEnrichmentWriter.configure(sensorName, getMockWriterConfiguration());
-    verify(logger).debug("Sensor: '{}': {Provider: '{}', Converter: '{}'}",
+    verify(logger, atLeastOnce()).debug("Sensor: '{}': {Provider: '{}', Converter: '{}'}",
         sensorName, MockTableProvider.class.getName(), EnrichmentConverter.class.getName());
   }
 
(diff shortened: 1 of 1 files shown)


jjmeyer0:
@nickwallen I noticed this is still failing, but on another test. I was wondering, if you mock the logger after each test would this fix the problem with the test? Only mocking it once way not reset the times the mock was interacted with.

jjmeyer0:
Actually, after looking a bit closer I am not certain that is possible to do in this case.

nickwallen:
Yes, I had go back and correct a few other test cases that were doing the same thing.  Thanks @jjmeyer0

mmiklavc:
One problem I see with this test is that it sets up the mock in a BeforeClass annotated init. That's generally frowned upon for mock tests that use verify() because expectations for the number of invocations is part of the api. A better alternative is to pull that mock out into an @Before. I rather like the idiom I've taken here best:

```
public class PcapCliTest {

  @Mock
  private PcapJob jobRunner;
  @Mock
  private ResultsWriter resultsWriter;
  @Mock
  private Clock clock;

  @Before
  public void setup() {
    MockitoAnnotations.initMocks(this);
  }
```

It keeps the setup lean and requires less boilerplate code (well, when you have multiple mocks, at least). Also, this approach will re-instantiate a new mock for every test, which I think is what we want in this case. It keeps the intent and expecations clear, to wit 1 invocation, not 1..n.

jjmeyer0:
@nickwallen out of curiosity, what do you think about just deleting these tests? I think the logging they are testing is great, but I don't think these are really needed.

mmiklavc:
Found this useful for duplicating the problem locally.
```
# build everything depended on by metron-parsers
cd metron
mvn clean install -pl metron-platform/metron-parsers -am

# change to the parsers project
cd metron-platform/metron-parsers

# keep looping on this specific test until it fails. Using macos "say" command for added fanfare :)
for (( i = 1; ; i++ )); do echo "Attempt $i"; mvn test -o -Dtest=SimpleHBaseEnrichmentWriterLoggingTest; exitcode=$?; if [ $exitcode -ne 0 ]; then say test failed on attempt $i; break; fi; done
```

mmiklavc:
Another issue with this test that complicates things a bit more is that these are static calls which require more than vanilla Mockito. @jjmeyer0 might be right here. Do we really need these tests?

mmiklavc:
@nickwallen ok, here's the 1-line fix (2 with the static import). Ran this 30+ times in a loop with no failures.
```
...
import static org.mockito.Mockito.reset;
...
  @Before
  public void setUp() {
    reset(logger);
    ...
  }
```

This resets the mock counts prior to each test, thereby eliminating the problem. I'm not clear why this isn't consistently failing to begin with, but there it is.

nickwallen:
@mmiklavc I dunno, Mike.  That is not working for me.  I removed all the changes I made and just added the `reset` to `setUp()` and the `shouldDebugWriteOperation` fails consistently.

Feel free to open a separate PR with your fix.  Maybe I did something wrong.
```
org.mockito.exceptions.verification.TooManyActualInvocations: 
logger.debug(
    <Capturing argument>,
    <Capturing argument>,
    <Capturing argument>,
    <Capturing argument>
);
Wanted 1 time:
-> at org.apache.metron.writers.SimpleHBaseEnrichmentWriterLoggingTest.shouldDebugWriteOperation(SimpleHBaseEnrichmentWriterLoggingTest.java:139)
But was 3 times. Undesired invocation:
-> at org.apache.metron.enrichment.writer.SimpleHbaseEnrichmentWriter.getTransformer(SimpleHbaseEnrichmentWriter.java:223)
```

nickwallen:
Honestly, I think we should just remove this test.  It seems unnecessary to test when specific log statements are executed.

merrimanr:
Why not just move line 60 (`logger = mock(Logger.class);`) to the setup method?  Since you're testing interactions with this mock it would be better to have a new one for every test.

merrimanr:
Nevermind, LoggerFactory is static.  For what it's worth, adding `org.mockito.Mockito.reset(logger);` to setUp() worked for me.

nickwallen:
I opened #658 to see if we have support for simply removing this set of tests.  I find them unnecessary.  Please review and vote on #658 whether you agree or disagree.  Thanks

nickwallen:
I am closing this PR in favor of #658.  This will not be merged.

PR #658, opened 2017-07-19
METRON-1048 Removed SimpleHBaseEnrichmentWriterLoggingTest

nickwallen:
Another alternative solution for METRON-1048.  This simply removes the test itself.  This set of tests simply test whether specific log statements are executed.  I find this unnecessary.

This is an alternative to #657 .


merrimanr:
+1

nickwallen:
@mmiklavc What do you think?  Just want to make sure we have consensus from everyone that chimed in on #657.

mmiklavc:
I'm fine with that, +1. I think we have an idea what the problem was so we'll know how to fix it if we ever run into it again.

PR #659, opened 2017-07-19
METRON-1053 Relocate Metron Docker

nickwallen:
Having `metron-docker` at the top-level of the project seems to catch the attention of new users. Some then start using metron-docker to explore/try-out/demo Metron.  The `metron-docker` code that we have is not well-suited for this purpose. It is only really useful for development. It is not regularly tested and maintained like our Vagrant environment.

### Changes

* Moved `metron-docker` to `metron-contrib/metron-docker`.  I based the top-level name on a popular suggestion from the mailing list along with matching the existing top-level naming convention (metron-*).

* Fixed an issue with untar'ing the Kafka, HBase, and Hadoop tarballs.  It seems that the behavior of Docker's `ADD` now is to automatically untar.   That is at least in Docker for Mac version 17.06.0-ce-mac19.  I assume on older versions this was not the case,  I changed this logic to avoid using `ADD` so that it would work for older and newer versions of Docker.

* Updated the documentation to make it clear that `metron-docker` is for development purposes only.

* Also renamed the installation location of the core components.  For example instead of `/opt/hadoop-2.7.3` I changed it to `/opt/hadoop`.  This makes it so we don't have the specific version number for these components duplicated in a few different files.  

### Testing

I launched `metron-docker` based on the instructions in the README.  The entire Docker compose application launched successfully and I validated that I was able to hit all of the web interfaces outlined in the README.


justinleet:
+1 by inspection. Thanks for taking care of this, and I think the readme updates will help clear up a lot of confusion.

nickwallen:
@ottobackwards @JonZeolla @kylerichardson  You guys all chimed in on the original email chain.  I'd like to get your eyes on this. Does this PR work for you?

JonZeolla:
I took a high level look when it first came out - no concerns.  I'm mobile only until Thursday so don't hold this up based on me.

kylerichardson:
Going to run this up now. I'll get my feedback in today.

kylerichardson:
Found a couple of issues when running from a clean source tree. A few quick mods to the pom and one Dockerfile. I'll submit a PR on your branch in the morning.

nickwallen:
Merged in @kylerichardson 's changes.  Thanks!

ottobackwards:
+1, I think this will help cut down what we are seeing on the user list.  Thanks!

kylerichardson:
+1, thanks for taking this on

PR #662, opened 2017-07-24
METRON-1056: Get field types from Elasticsearch

merrimanr:
## Contributor Comments
This PR adds 2 additional search REST endpoints.  The first takes a list of indices as input and returns field to type mappings for each index.  The second also takes a list of indices as input but only returns field to type mappings for the intersection of the common fields across the indices.  These endpoints are needed by the alerts UI and will also be necessary for correctly formatting faceted search results.

This can been tested in full dev.  After spinning up full dev, go the the SearchController section of the Swagger UI.  There the endpoints can be tested against sample data in Elasticsearch.  Passing in "["bro","snort"] to "/api/v1/search/column/metadata" should return type mappings for both indices.  Passing in the same list to "/api/v1/search/column/metadata/common" should only return type mappings for fields that are in both indices.

Reviewers will notice that tests were not added to the unit tests in ElasticsearchDaoTest.  As the original author of that class I found it awkward mocking the various Elasticsearch client classes and it took a non-trivial amount of time to figure out.  I started adding to the test for this PR and ran into similar issues.  I feel like this test will be difficult to maintain, especially since several different people will be contributing to this class, and I don't feel it gives us a good return on investment.  We have 100% test coverage in the integration tests which are much easier to maintain and understand.  Maybe there is an easier way to test Elasticsearch client classes but I vote we remove this and instead rely on the integration tests since they provide complete coverage and are easy to work with.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-rest/README.md
+++ b/metron-interface/metron-rest/README.md
@@ -200,6 +200,8 @@ Request and Response objects are JSON formatted.  The JSON schemas are available
 | [ `DELETE /api/v1/kafka/topic/{name}`](#delete-apiv1kafkatopicname)|
 | [ `GET /api/v1/kafka/topic/{name}/sample`](#get-apiv1kafkatopicnamesample)|
 | [ `GET /api/v1/search/search`](#get-apiv1searchsearch)|
+| [ `GET /api/v1/search/search`](#get-apiv1searchcolumnmetadata)|
+| [ `GET /api/v1/search/search`](#get-apiv1searchcolumnmetadatacommon)|
 | [ `GET /api/v1/sensor/enrichment/config`](#get-apiv1sensorenrichmentconfig)|
 | [ `GET /api/v1/sensor/enrichment/config/list/available/enrichments`](#get-apiv1sensorenrichmentconfiglistavailableenrichments)|
 | [ `GET /api/v1/sensor/enrichment/config/list/available/threat/triage/aggregators`](#get-apiv1sensorenrichmentconfiglistavailablethreattriageaggregators)|
@@ -353,6 +355,20 @@ Request and Response objects are JSON formatted.  The JSON schemas are available
       * searchRequest - Search request
   * Returns:
     * 200 - Search results
+    
+### `GET /api/v1/search/column/metadata`
+  * Description: Get column metadata for each index in the list of indicies
+  * Input:
+      * indices - Indices
+  * Returns:
+    * 200 - Column Metadata
+    
+### `GET /api/v1/search/column/metadata/common`
+  * Description: Get metadata for columns shared by the list of indices
+  * Input:
+      * indices - Indices
+  * Returns:
+    * 200 - Common Column Metadata
 
 ### `GET /api/v1/sensor/enrichment/config`
   * Description: Retrieves all SensorEnrichmentConfigs from Zookeeper
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/SearchController.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/SearchController.java
@@ -50,17 +50,17 @@ public class SearchController {
     return new ResponseEntity<>(searchService.search(searchRequest), HttpStatus.OK);
   }
 
-  @ApiOperation(value = "Get Column Metadata")
+  @ApiOperation(value = "Get column metadata for each index in the list of indices")
   @ApiResponse(message = "Column Metadata", code = 200)
   @RequestMapping(value = "/column/metadata", method = RequestMethod.POST)
-  ResponseEntity<Map<String, Map<String, FieldType>>> getColumnMetadata(@RequestBody List<String> indices) throws RestException {
+  ResponseEntity<Map<String, Map<String, FieldType>>> getColumnMetadata(final @ApiParam(name = "indices", value = "Indices", required = true) @RequestBody List<String> indices) throws RestException {
     return new ResponseEntity<>(searchService.getColumnMetadata(indices), HttpStatus.OK);
   }
 
-  @ApiOperation(value = "Get Common Column Metadata")
+  @ApiOperation(value = "Get metadata for columns shared by the list of indices")
   @ApiResponse(message = "Common Column Metadata", code = 200)
   @RequestMapping(value = "/column/metadata/common", method = RequestMethod.POST)
-  ResponseEntity<Map<String, FieldType>> getCommonColumnMetadata(@RequestBody List<String> indices) throws RestException {
+  ResponseEntity<Map<String, FieldType>> getCommonColumnMetadata(final @ApiParam(name = "indices", value = "Indices", required = true) @RequestBody List<String> indices) throws RestException {
     return new ResponseEntity<>(searchService.getCommonColumnMetadata(indices), HttpStatus.OK);
   }
 }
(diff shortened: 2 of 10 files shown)


ottobackwards:
Awesome @merrimanr, and you speak at all to how this would translate for a Solr implementation conceptually?  Does everything carry over?

merrimanr:
Sure @ottobackwards.  There is a similar api for Solr that will return field types for a list of collections.  When we add Solr as an indexing option we will create a "SolrDao" class that will implement the "IndexDao.getColumnMetadata(List<String> indices)" and "IndexDao.getCommonColumnMetadata(List<String> indices)" methods.

ottobackwards:
Thanks!

metron-interface/metron-rest/README.md
```diff
@@ -353,6 +355,20 @@ Request and Response objects are JSON formatted.  The JSON schemas are available
       * searchRequest - Search request
   * Returns:
     * 200 - Search results
+
```
ottobackwards: This is not a problem for this PR, but rather the documentation in general.  The actual TYPES returned from the rest calls are not documented at all.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/SearchController.java
```diff
@@ -45,4 +49,18 @@
   ResponseEntity<SearchResponse> search(final @ApiParam(name = "searchRequest", value = "Search request", required = true) @RequestBody SearchRequest searchRequest) throws RestException {
     return new ResponseEntity<>(searchService.search(searchRequest), HttpStatus.OK);
   }
+
```
ottobackwards: Is this operation very fast?  or should it be a candidate for an async call?

metron-interface/metron-rest/README.md
ottobackwards: Such as complex types.

This comes through here, because we are creating our own Field Type Enums

ottobackwards: We may want to consider explaining the types we support and return here... but I am not sure it should be inline.
Since the detail is in line with the rest of the document, maybe that can be a follow on

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/FieldType.java
```diff
@@ -0,0 +1,52 @@
...
+  LONG("long"),
+  @JsonProperty("date")
+  DATE("date"),
+  @JsonProperty("float")
+  FLOAT("float"),
+  @JsonProperty("double")
+  DOUBLE("double"),
+  @JsonProperty("boolean")
+  BOOLEAN("boolean"),
+  @JsonProperty("other")
```
ottobackwards: What does other mean?  Is it mapped to string in return values..

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/IndexingDaoIntegrationTest.java
```diff
@@ -27,28 +28,32 @@
 import org.json.simple.parser.ParseException;
 import org.junit.*;
 
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
```
ottobackwards: How do we test unknown

metron-interface/metron-rest/README.md
merrimanr: This isn't intended to be the source of truth for types that we support.  I would think any data type someone may need should be supported in some way.  This is restricted to (mostly) simple types and is driven by the need to format simple values in different ways.  Anything else besides simple types is categorized as "other".  If at any point there is a need to include more types this can easily be extended.  

The Elasticsearch API gives you either a number or string (only string in the case of the REST API) so we need more information about the type in Elasticsearch to format it correctly.  Solr is similar.  I will take another pass at the documentation to make this more obvious.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/SearchController.java
merrimanr: It's a legitimate concern but I think synchronous is ok here because Elasticsearch is intended to provide results in real-time.  The http javascript APIs we use are asynchronous so it shouldn't make a difference in the UI.  If this were a batch operation (Hive, Map Reduce) I would agree that it should be an async call.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/FieldType.java
merrimanr: "other" just means it's not one of the other simple types in the enum.  The Elasticsearch Java API returns the search result as a Map<String, Object> type which is also the return type of the controller so it's up to Jackson to serialize it.  We are not doing anything special to map values to types when returning search results.  So far I haven't seen Jackson have any problems with the types being returned.  From what I can tell Elasticsearch mostly sticks to simple types for the values.

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/IndexingDaoIntegrationTest.java
merrimanr: "location_point" is of type geo_field so it's falls into the "other" category.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/FieldType.java
ottobackwards: At some point, if we are going to abstract these other systems, we need to document our abstractions.  I am not sure where though.  Like stellar, the rest api documentation prob. needs to be broken out with some better detail for specific entry points, including references to outside systems the user is expected to know.

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/IndexingDaoIntegrationTest.java
ottobackwards: Yes

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/FieldType.java
ottobackwards: Just to be clear, I don't think this should hold up this PR, but that we in a general sense need to go into more detail.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/SearchController.java
simonellistonball: Feels like any asynchronicity should be at the service layer anyway for other things to potentially benefit from, of course that would bubble up here, but do we really expect the REST server to have to worry about high load?

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/IndexingDaoIntegrationTest.java
james-sirota: @ottobackwards are you ok with this PR? Can this get a +1 from you?

ottobackwards:
+1

PR #667, opened 2017-07-25
METRON-1061 Add FUZZY_SCORE STELLAR function

ottobackwards:
[Apache Commons Text Similarity](https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/similarity/FuzzyScore.html)

FuzzyScore :
"A matching algorithm that is similar to the searching algorithms implemented in editors such as Sublime Text, TextMate, Atom and others.
One point is given for every matched character. Subsequent matches yield two bonus points. A higher score indicates a higher similarity."

This PR adds FUZZY_SCORE to STELLAR, exposing the Apache Commons FuzzyScore class functionality.

This allows for fuzzy matching of strings to queries, and composing statements based on thresholds related to those scores.

For example:

term = "metron"
query = "metron"

```java
16 == FUZZY_SCORE(term,query,'en')
```
+1 for each match and +2 bonus for the 2nd, 3rd, 4th, 5th, and 6th matches.

This is related to [METRON-1052](https://issues.apache.org/jira/browse/METRON-1052)

### For all changes:

### For code changes:
  ```
  ```

- [] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

-  [x] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via `site-book/target/site/index.html`:

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
<magoo>Oh ProfilerIntegrationTest you've done it again!</magoo>

ottobackwards:
@JonZeolla

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
```diff
@@ -0,0 +1,63 @@
...
+   * Stellar.
+   */
+  public static class FuzzyScoreFunction extends BaseStellarFunction {
+
+    @Override
+    public Object apply(List<Object> list) {
+      if (list.size() < 3) {
+        throw new IllegalStateException("FUZZY_SCORE expects three args: [string, string, string]");
+      }
+      String term = (String) list.get(0);
```
cestella: If there's an exception here (specifically a classcastexception), we're going to exception.  Given the noise in the data that we have, I would expect this to happen.

I have a couple of questions:
* Do we want to return `NaN` or `Infinity` in that case and log at a warn level (or error level)?
* Do we want to throw an exception which could not possibly be caught in the stream?

I, personally, vote for the first, but I'd like to hear other people's impressions.

ottobackwards: 1. The (CAST)list.get(0) is the common pattern in our stellar code.  I believe I have asked before in another pr maybe why we don't use the conversion utils. 
2. I think we want to return 0 for invalid args.

ottobackwards: We really need a pattern or a helper class for stellar for variables, that everyone uses.

ottobackwards: Also, the return should consider the common usage, I imagine something like:

IF (FUZZY_SCORE(fld,qry,'EN') > 4) THEN SET SOME FIELD

cestella: Well, in this case, I don't think we want to use `ConversionUtils` since we probably dont' want to coerce a list into a string for a fuzzy score, but I might be wrong.

I think we probably want to return `0` in the case that someone passes in a wrong type too.

ottobackwards: done

ottobackwards:
@cestella are we all set with the changes?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
```diff
@@ -0,0 +1,70 @@
...
+    public Object apply(List<Object> list) {
+      if (list.size() < 3) {
+        throw new IllegalStateException("FUZZY_SCORE expects three args: [string, string, string]");
+      }
+      Object oterm = list.get(0);
+      Object oquery = list.get(1);
+      Object olang = list.get(2);
+
+      if (!(oterm instanceof String) || !(oquery instanceof String) || !(olang instanceof String)) {
+        return 0;
```
nickwallen: This is an error condition, right?  Should we treat it like you did above and throw an `IllegalStateException`?  Better yet, maybe we should use `IllegalArgumentException`.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
```diff
@@ -0,0 +1,70 @@
...
+      }
+      String term = (String) oterm;
+      String query = (String) oquery;
+      String lang = (String) olang;
+      if (StringUtils.isEmpty(term) || StringUtils.isEmpty(query) || StringUtils.isEmpty(lang)) {
+        return 0;
```
nickwallen: Aren't there are a few different scenarios that should be treated differently here?  
* If the `lang` is an empty string, I would assume that is an error condition.  
* If either the `term` or `query` is empty, then why not just allow the `FuzzyScore` to handle it and return a value?

metron-stellar/stellar-common/README.md
```diff
@@ -411,6 +412,14 @@ In the core language functions, we support basic functional programming primitiv
     * format - string
     * arguments... - object(s)
   * Returns: A formatted string.
+  
+### `FUZZY_SCORE`
+  * Description: Returns the Fuzzy Score which indicates the similarity score between two strings. One point is given for every matched character. Subsequent matches yield two bonus points. A higher score indicates a higher similarity.
+  * Input:
+    * string - The full term that should be matched against.
+    * string - The query that will be matched against a term.
+    * string - The IETF BCP 47 language code to use.
```
nickwallen: Can we document the common language codes here, so I can easily see them when running `?FUZZY_SCORE` in the REPL?  I'll let you define "common languages", but I assume English, Spanish, at least.

nickwallen: You seem to have an affinity for Klingon, so I'm fine with that too.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
nickwallen: Sorry, I am finding this discussion late.  I read through your previous comments and changed my mind about 3 times as I read through.

If we know there is an error condition, I think something like `NaN` would definitely be more helpful than returning 0.  I feel like that too easily masks a problem and might make this a little frustrating to use. 

If there are not even 3 arguments passed in, then throwing an exception works.  That's clearly an error that should be caught.

metron-stellar/stellar-common/README.md
ottobackwards: How about List Lang codes()?

nickwallen: If you're ambitious, sure.  If you do implement a function that lists the lang codes, I would try to use the same namespace.

Otherwise, just the common langs and/or a URL link works for me.

ottobackwards: you mean to the doc?

cestella:
I'm fine with this so far, but will hold for @nickwallen 's +1. :)

metron-stellar/stellar-common/README.md
nickwallen: Right, just to the doc string that gets printed when I run `?FUZZY_SCORE` in the REPL.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
nickwallen: > I think we want to return 0 for invalid args.

Hey @ottobackwards , why do you like returning 0 for invalid args?  Maybe there is some reasoning that I am missing.

To me it seems like I can't distinguish between an error condition and two dissimilar strings.  A 0 would be returned for both.

metron-stellar/stellar-common/README.md
ottobackwards: I did both

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.List;
+import java.util.Locale;
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.text.similarity.FuzzyScore;
+import org.apache.metron.stellar.dsl.BaseStellarFunction;
+import org.apache.metron.stellar.dsl.Stellar;
+
+public class TextFunctions {
+
+  @Stellar(name = "GET_AVAILABLE_LANGUAGE_TAGS",
```
nickwallen: Do the language tags have any use outside of `FUZZY_SCORE`?

If its only use is for `FUZZY_SCORE`, I was thinking we could the same namespace for both functions.
`FUZZY_SCORE` -> namespace = "FUZZY", name = "SCORE"
`FUZZY_LANGS` -> namespace = "FUZZY", name = "LANGS"

Or any other logical naming scheme that ties them together with a namespace.  I mean, we do have this cool namespace feature at our disposal.

ottobackwards: I always forget about the undocumented Namespace feature... I sure will do that.

ottobackwards: done

JonZeolla:
@nickwallen Any additional comments?  This looks good to me, happy to do additional testing if needed but I did spin this up a week back

nickwallen:
Yes, I have some outstanding comments that Otto hasn't responded to yet.  It is in the "fold" which is why they were probably missed.

https://github.com/apache/metron/pull/667#discussion_r133093147

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
ottobackwards: The thinking is that the user is going to write an expression where they want the score between two things.  In an error conditions, logically the score would be 0, and they can still have the correct logical outcome.  It is just more forgiving.

nickwallen: I would think that if I use the wrong lang, I would want to know its an error, rather than mask that error by returning 0.

For example, it might be frustrating to to see the following return 0.  
```
FUZZY_SCORE("metron","metron","eng")
```
Why is this returning 0?  
The terms are exactly the same!  
Oh, wait I used the wrong lang code, it should be 'en'.

ottobackwards: ok, i'll tighten that up

ottobackwards: @nickwallen Check out my latest.  Note that because of the way validate works, we need to return 0 for non-string arguments.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
```diff
@@ -0,0 +1,114 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.metron.stellar.dsl.functions;
+
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedList;
```
nickwallen: Some of these are unused imports which would be nice to clean-up.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/TextFunctionsTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Test(expected = ParseException.class)
+  public void testMissingLanguage() throws Exception {
+    runPredicate("0 == FUZZY_SCORE(metron,'z',klingon)",
+        new DefaultVariableResolver(v -> variableMap.get(v),
+            v -> variableMap.containsKey(v)));
+  }
+
+  @Test()
+  public void testEmptyFirstArg() throws Exception {
```
nickwallen: This test will never fail

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/TextFunctionsTest.java
```diff
@@ -0,0 +1,103 @@
...
+    runPredicate("0 == FUZZY_SCORE(empty,'z',english)",
+        new DefaultVariableResolver(v -> variableMap.get(v), v -> variableMap.containsKey(v)));
+  }
+
+  @Test()
+  public void testEmptyFirstTwoArgs() throws Exception {
+    runPredicate("0 == FUZZY_SCORE(empty,empty,english)",
```
nickwallen: This test will never fail.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/TextFunctionsTest.java
```diff
@@ -0,0 +1,103 @@
...
+    Assert.assertTrue(ret instanceof List);
+    List<String> tags = (List<String>) ret;
+    Assert.assertTrue(tags.size() > 0);
+    Assert.assertTrue(tags.contains("en"));
+    Assert.assertTrue(tags.contains("fr"));
+  }
+
+  @Test()
+  public void testNoMatchStrings() throws Exception {
+    runPredicate("0 == FUZZY_SCORE(metron,'z',english)",
```
nickwallen: This test will never fail.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/TextFunctionsTest.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.apache.metron.stellar.dsl.functions;
+
+import java.util.Arrays;
```
nickwallen: Some of these are unused imports.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
nickwallen: Thanks @ottobackwards. That change looks good.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/TextFunctions.java
ottobackwards: done

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/TextFunctionsTest.java
ottobackwards: done

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/TextFunctionsTest.java
ottobackwards: sorry, I lost the asserts refactoring, done

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/TextFunctionsTest.java
ottobackwards: sorry, I lost the asserts refactoring, done

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/TextFunctionsTest.java
ottobackwards: sorry, I lost the asserts refactoring, done

nickwallen:
+1 Looks good. Thanks for all the hard work, Otto.

PR #675, opened 2017-08-01
METRON-379 STELLAR can differentiate between a value passed as null and a value not passed

ottobackwards:
That being that there IS a variable at all vs. there is a variable, and its value is NULL.

Referenced variables in Stellar statements with the default variable resolver will throw ParseExceptions
if they are not included or passed.  Other variable resolvers can implement this policy as they need.
For example the LambdaExpression's VariableResolver as well as the MapResolver used in enrichments and parsers allow missing variables, since the variable naming in the lambda expression is arbitrary and separate from the original call context.  This means that for StellarFieldTransforms and StellarEnrichment there is no change in behavior.

The context supports this by maintaining a thread local activity type, such that the context can be inspected for a validation vs. a parse operation. Variables are not present during validation and this is allowed.

Tests have been changed or updated ( hopefully ) to thier _intent_ where possible.  If they were testing
for handling a variable that *is* null vs a missing variable etc.

### Implementation
The VariableResolver interface now has a second function -> exists(). This allows for checking if there is a variable at all.  The resolve method would return null for both cases.   A new DefaultVariableResolver class has been created, to help callers who were only passing something like (x) -> hashMap.get(x),  to parse.

The ActivityType enum has been added ( see Context.java ).  It is stored as a thread local var in the Context.  We set this to Parse or Validate depending on the operation. We don't need to check during validation. It is a implemented as a thread local var to support instances where we use a single context in multiple threads.

### Testing
All tests should pass when building the product
Full Dev should run and have data

### Questions for Review

1.  Are there issues here that are not accounted for with upgrading or ????
2. I am not sure that some of the classes and interfaces in DSL shouldn't be in Common ( like VariableResolver ).

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/LambdaExpression.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/LambdaExpression.java
@@ -18,6 +18,7 @@
 
 package org.apache.metron.stellar.common;
 
+import org.apache.metron.stellar.dsl.DefaultVariableResolver;
 import org.apache.metron.stellar.dsl.Token;
 import org.apache.metron.stellar.dsl.VariableResolver;
 
@@ -56,9 +57,9 @@ public class LambdaExpression extends StellarCompiler.Expression {
       lambdaVariables.put(variables.get(i), null);
     }
 
-    VariableResolver variableResolver = variable -> lambdaVariables.getOrDefault(variable
+    VariableResolver variableResolver = new DefaultVariableResolver(variable -> lambdaVariables.getOrDefault(variable
                                                                                 , state.variableResolver.resolve(variable)
-                                                                                );
+                                                                                ), variable -> true);
     StellarCompiler.ExpressionState localState = new StellarCompiler.ExpressionState(
             state.context
           , state.functionResolver
--- a/metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/stellar/WindowLookbackTest.java
+++ b/metron-analytics/metron-profiler-client/src/test/java/org/apache/metron/profiler/client/stellar/WindowLookbackTest.java
@@ -22,6 +22,7 @@ package org.apache.metron.profiler.client.stellar;
 import com.google.common.collect.ImmutableMap;
 import org.apache.commons.lang3.Range;
 import org.apache.metron.stellar.dsl.Context;
+import org.apache.metron.stellar.dsl.DefaultVariableResolver;
 import org.apache.metron.stellar.dsl.ParseException;
 import org.apache.metron.stellar.dsl.functions.resolver.FunctionResolver;
 import org.apache.metron.stellar.dsl.functions.resolver.SimpleFunctionResolver;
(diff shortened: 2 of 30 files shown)


cestella:
In my opinion, I don't think we want this behavior.  I think it may be too severe given how dirty and sparse our data tends to be.

Does this mean that if you have a stellar expression `FOO(bar)` and the field `bar` doesn't exist, will we get an exception?  If so, given that messages come through with certain fields which are optional (i.e. some messages have field `foo` and others might not have a field `foo` depending on the parser), I think this will result in a fair number of errors.

I'd rather leave it up to the Stellar function creator to figure out if the field is able to be optional or not rather than having a blanket policy of throwing an exception.  Consider the function `TO_UPPER`.  I'd rather `TO_UPPER` tolerate missing fields by passing them through (i.e. `TO_UPPER(null) == null`) rather than throw an exception.

One approach to this would be changing the `StellarFunction` interface to a `List<Optional<Object>>` instead of a `List<Object>` and have the variability made explicit.

cestella:
Ok, wait, I think I might've misunderstood.  This `DefaultVariableResolver` is just being used for tests, validation and Lambda functions now, is that correct?  I think I'm ok with that.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/LambdaExpression.java
```diff
@@ -56,9 +57,9 @@ public Object apply(List<Object> variableArgs) {
       lambdaVariables.put(variables.get(i), null);
     }
 
-    VariableResolver variableResolver = variable -> lambdaVariables.getOrDefault(variable
+    VariableResolver variableResolver = new DefaultVariableResolver(variable -> lambdaVariables.getOrDefault(variable
```
cestella: Ok, help me out here, can you give me an example of the impact of this change?  Maybe a `MAP()` function and how it acts before the change and how it would act after the change?

ottobackwards:
Right.  This PR adds the Language Capability to required variables be passed based on the VariableResolver implementation defined.  Which can be defined based on the requirements of the 'host' -> so Enrichments and Parsers use Map, which is forgiving because of the reasons you mention.
Lambdas use their own, which is forgiving, because in a lambda (x) -> DO(x) you can name 'x' whatever you want.

cestella:
Yeah, I'm totally ok with this now.  Really sorry about jumping to the wrong conclusion; I saw the errors in the aftermath of the math PR that got in this morning and jumped to the wrong conclusion about default behavior.

To that end, that `NaN` change you have, shouldn't have been necessary.  We need `NaN` as a language keyword.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/LambdaExpression.java
ottobackwards: The only impact would be in the tests right now.  Since we use the MapResolver in most places ( shell, profiler, enrichments, parser ).

ottobackwards:
I'll create a Jira for the NaN thing, unless you have already coded it?

cestella:
hah no, a JIRA will be necessary ;)

cestella:
+1 by inspection, this is good work, @ottobackwards 

Sorry for jumping to conclusions!

ottobackwards:
https://issues.apache.org/jira/browse/METRON-1079

PR #676, opened 2017-08-01
METRON-1072: Cleanup Maven build warnings

mmiklavc:
https://issues.apache.org/jira/browse/METRON-1072

## Contributor Comments

This one is pretty straightforward. Our build outputs a large amount of avoidable WARNING noise. This PR cleans up quite a bit of low-hanging fruit, including:

- The expression ${parent.version} is deprecated. Please use ${project.parent.version} instead.
- Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
- File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
- The assembly descriptor contains a filesystem-root relative reference, which is not cross platform compatible /bin
- site-book: JAR will be empty - no content was marked for inclusion!

There are also a number of warnings pertaining to overlapping classes found during the shading process. And a number of npm related issues, e.g. "npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue". This is not addressed by this work.

To test this, run full-dev and verify data flows through indexing. The biggest concern would be around the path changes (removing leading slash '/') in the assembly.xml files. In spite of the examples included in https://maven.apache.org/guides/mini/guide-assemblies.html, this results in a potentially platform-dependent build. Excluding leading slashes addresses these warnings.

### For all changes:

### For code changes:
  ```
  ```
 
  ```
  cd site-book
  mvn site
  ```

### DIFF
--- a/metron-analytics/metron-maas-service/src/main/assembly/assembly.xml
+++ b/metron-analytics/metron-maas-service/src/main/assembly/assembly.xml
@@ -20,7 +20,7 @@
   <fileSets>
     <fileSet>
       <directory>${project.basedir}/src/main/config</directory>
-      <outputDirectory>/config</outputDirectory>
+      <outputDirectory>config</outputDirectory>
       <useDefaultExcludes>true</useDefaultExcludes>
       <excludes>
         <exclude>**/*.formatted</exclude>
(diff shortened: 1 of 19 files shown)


ottobackwards:
+1, built with full commit build options, ran full dev, saw data.
Honestly, these errors are not what flood my logs, but it is good to chip away

mmiklavc:
Agreed @ottobackwards - we still have quite a bit of work around the warnings related to shading and relocating jars. Thanks for the review!

PR #681, opened 2017-08-03
METRON-1079 Add NaN as a keyword in STELLAR language

ottobackwards:
Add "NaN" to STELLAR as a keyword.  

Add NaN, and tests to verify that it works in line with the java specification.

## Testing
Build and Tests should work.
Execute statement similar to the added tests in Stellar Shell.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


nickwallen:
Can you share some use cases where you find `NAN` useful to have as a keyword?  What itch is this scratching?

ottobackwards:
@cestella and I were talking about null variables and how NaN was being used, and that we probably needed NaN as a keyword.  So I took a stab at it.

cestella:
Some of our math functions return `NaN` (e.g. `SQRT(-1)`) and being able to do `!= NaN` is a useful thing.

cestella:
Just thinking, we might want +Infinity and -Infinity too as they're part of the floating point standard.

nickwallen:
I see in the unit tests that @ottobackwards intended that any equality with `NaN` to be false.  But then how is that useful for the basic use case of `FUNCTION(x) != NaN`?

```
[Stellar]>>> SQRT(-1)
NaN
```

I would expect the following to return `true`.
```
[Stellar]>>> SQRT(-1) == NaN
false
```

I would expect the following to return `false`.
```
[Stellar]>>> SQRT(-1) != NaN
true
```

nickwallen:
Based on the link that you embedded in the tests, your implementation seems correct.  I assume that these grammar rules are well reasoned and necessary, but it surely seems counter-intuitive to me for this use case.

> The equality operator == returns false if either operand is NaN. 
> The inequality operator != returns true if either operand is NaN (§15.21.1). 
> In particular, x!=x is true if and only if x is NaN, and (x<y) == !(x>=y) will be false if x or y is NaN.

 Am I thinking about this all wrong?  Maybe I just need some lunch to get my blood sugar up. :)

mattf-horton:
Technically this is correct, as for example SQRT(-1) and SQRT(-2) clearly should not show as equal, but by using NaN we (deliberately) lose the info that would allow us to recognize that SQRT(-1)==SQRT(-1).

To detect/validate NaN, we would need an operator IS_NAN(x).

ottobackwards:
We are not 'doing' the math here.  What we are doing is introducing the ability to use NaN in an expression correctly.  Java is doing the math.  The tests are to verify that through stellar, the specified behavior still is adhered to.

mattf-horton:
BTW, +/- Infinity **are** considered "real" values (essentially as tho via rounding) and can be compared with equality and inequality.  NaN is considered a nonsensical or impossible-to-calculate value, not a number at all.  The fact that 1.0/0.0 returns positive Infinity rather than NaN is a mathematical choice, presumably based on limit logic and a preference for continuity. (I didn't write the standard! :-)

mattf-horton:
@ottobackwards , the validator for IS_NAN(x) in Java would be Double.isNaN(double x).

nickwallen:
Thanks for clarifying @mattf-horton .  That makes sense now that I've had lunch.

Based on the original use case of detecting a function returning NaN, it seems what we really need is something like `IS_NAN(x)` rather than a keyword for `NaN`.  Implementing `IS_NAN(x)` doesn't technically require a keyword for `NaN`.

But I can see a potential use for the `NaN` keyword when creating a lambda expression.

cestella:
Yeah, it's true, I suppose having `==` double as a `NaN` check is probably not the right thing due to the transitivity of equality (I'd like to `SQRT(-1) == NaN` to be `true` but `SQRT(-1) == SQRT(-2)` to not be true).  Probably it's better to have an `IS_NAN(x)` like other languages.

mattf-horton:
@nickwallen is right, since they can't be usefully compared, the only use for a literal NaN is if you want to inject that value into a calculation, which as far as I can see is only useful for unit testing.

ottobackwards:
I have added IS_NAN()

justinleet:
@ottobackwards Could you deconflict this?  It seems like semantic conversation died down and hopefully we're at least mostly set since `IS_NAN()` got added.

ottobackwards:
done

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/BasicStellarTest.java
```diff
@@ -251,6 +251,142 @@ public void testNull() {
...
+      Assert.assertTrue(run(query,new HashMap<>()).toString().equals("NaN"));
+    }
+    {
+      String query = "(NaN / NaN) != NaN";
+      Assert.assertTrue(runPredicate(query,new HashMap<>()));
+    }
+    {
+      String query = "NaN / NaN";
+      Assert.assertTrue(run(query,new HashMap<>()).toString().equals("NaN"));
+    }
```
justinleet: Can you add a test case for 0.0/0.0 is Nan?  It's properly calculated based on my playing around, but it's an important enough case that I think it should show up here.

justinleet:
@cestella and @mattf-horton Do we care about the Infinity cases and presumably an `IS_INFINITE()` in this PR, or are we okay with holding off until a follow on is made?

Other than that question, and one test case I'd like added, I'm happy with this.  Thanks for the contribution!

justinleet:
I'm on the side of holding off for an eventual follow-on regarding that question, for the record.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/BasicStellarTest.java
ottobackwards: done

cestella:
Yeah, I'm ok with that.

justinleet:
I'm +1, thanks again for the contribution!

cestella:
me too, +1

PR #643, opened 2017-07-10
METRON-1026: threatintel_taxii_load.sh throws exception

cestella:
## Contributor Comments
The delegation wrapping done to enable Stellar transformations on data as it is imported broke the taxii loader.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  mvn -q clean integration-test install && build_utils/verify_licenses.sh 
  ```

- [x] Have you written or updated unit tests and or integration tests to verify your changes?
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [x] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

### For documentation related changes:
- [x] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via `site-book/target/site/index.html`:

  ```
  cd site-book
  mvn site
  ```

#### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.
It is also recommended that [travis-ci](https://travis-ci.org) is set up for your personal repository such that your branches are built there before submitting a pull request.

PR #599, opened 2017-05-31
METRON-975: Normalize logging and switch to common idiom with SLF4J

mmiklavc:
https://issues.apache.org/jira/browse/METRON-975

We currently have a mixture of logging approaches scattered throughout the codebase. Log4j, slf4j, and commons logging. In addition, there are some deprecated idioms being used to declare loggers (post-jdk 7) and format strings within the logging statements.

This PR handles converting and normalizing Metron's logging over to slf4j in all major metron-platform areas except for tests. It also converts logging in metron-analytics, with the exception of MaaS which is currently leveraging commons logging.

I've made a few types of changes (change varietals?).
1. Normalized logger factory calls to the following `LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());` , per https://www.slf4j.org/faq.html#declaration_pattern
2. Removed `LOG.isDebugEnabled()` in instances where the SLF4J parameterized messages is a better option, e.g. `logger.debug("The entry is {}.", entry);`. I have *not* changed this implementation in instances where more costly operations are being done, like string concatenation derived from parsing or serializing json into string form, per https://www.slf4j.org/faq.html#logging_performance
3. Replaced string concatenation with parameterized nessages. I did this even when a formatted string precedes an exception/throwable. The javadoc wasn't immediately clear about this, but as of SLF4J 1.6 a final exception argument will be treated as an exception rather than as an object for use during parameterization - https://www.slf4j.org/faq.html#paramException

Testing instructions to follow. For starters, I've compared the number of lines in the Travis build output before and after this change and see no notable difference in total number of lines.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
[...]

PR #689, opened 2017-08-10
METRON-1102: Add support for ingesting cybox URI observables from taxii feeds

cestella:
## Contributor Comments
There is value in ingesting URIs from taxii feeds and we should provide support to do so.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/extractor/stix/StixExtractor.java
+++ b/metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/extractor/stix/StixExtractor.java
@@ -19,6 +19,7 @@ package org.apache.metron.dataloads.extractor.stix;
 
 import com.google.common.base.Splitter;
 import org.apache.commons.io.FileUtils;
+import org.apache.log4j.Logger;
 import org.apache.metron.dataloads.extractor.Extractor;
 import org.apache.metron.dataloads.extractor.stix.types.ObjectTypeHandler;
 import org.apache.metron.dataloads.extractor.stix.types.ObjectTypeHandlers;
@@ -38,6 +39,7 @@ import java.util.List;
 import java.util.Map;
 
 public class StixExtractor implements Extractor {
+    private static final Logger LOG = Logger.getLogger(StixExtractor.class);
     Map<String, Object> config;
     @Override
     public Iterable<LookupKV> extract(String line) throws IOException {
@@ -50,10 +52,20 @@ public class StixExtractor implements Extractor {
                 if(props != null) {
                     ObjectTypeHandler handler = ObjectTypeHandlers.getHandlerByInstance(props);
                     if (handler != null) {
+                        if(LOG.isDebugEnabled()) {
+                            LOG.debug("Found " + handler.getTypeClass().getCanonicalName()
+                                    + " for properties " + props.toXMLString()
+                            );
+                        }
                         Iterable<LookupKV> extractions = handler.extract(props, config);
                         for(LookupKV extraction : extractions) {
                             ret.add(extraction);
                         }
+                    }
+                    else if(LOG.isDebugEnabled()) {
+                        LOG.debug("Did not find a handler"
+                                + " for properties " + props.toXMLString() + " of type " + props.getClass()
+                        );
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


cestella:
Testing plan should be the same as [here](https://github.com/apache/metron/pull/643#issuecomment-321415666)

metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/extractor/stix/StixExtractor.java
```diff
@@ -38,6 +39,7 @@
 import java.util.Map;
 
 public class StixExtractor implements Extractor {
+    private static final Logger LOG = Logger.getLogger(StixExtractor.class);
```
simonellistonball: Didn't we change how logging worked in METRON-975? @mmiklavc see also #599

ottobackwards:
So,

I think this is a great addition, but I have some comments.

- Where is the documentation for the version of Stix and the Version of Cybox metron supports?
- How is the extractor factored to handle support for other versions?
- How is the extractor factored to handle JSON if / when stix and cybox change over?
- Can we handle just Cybox?  Should this be factored to support them separately?
- Where is the documentation for the support in this PR?

metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/extractor/stix/StixExtractor.java
cestella: Yeah, this ended up starting live as a patch against an older branch that I just made current.  I changed the logging to use slf4j now.  Good eye, @simonellistonball

cestella:
Ok, so good questions @ottobackwards .  I'll do my best to answer them, but the answer to some of these expands past this PR and to the history of Taxii support for Metron (which was one of the first things we added and thus at a period of time where documentation was scarcer than it is even now ;) )

* `Where is the documentation for the version of Stix and the Version of Cybox metron supports?`
We do not currently document that, the answer is, however, that our support for Stix, cybox and taxii is entirely bound up in the mitre java-stix library.  We use the most current version [released](https://github.com/STIXProject/java-stix/tree/v1.2.0.2), which is 1.2.0.2.

* `How is the extractor factored to handle support for other versions?`
The extractor is handled to support other versions only insomuch as the java-stix library can support multiple versions.  As this is officially supported by the stix project, I think that it's backwards compatible, but there may be nuance here that I'm missing.

* `How is the extractor factored to handle JSON if / when stix and cybox change over?`
The extractor abstraction works at the level of the object model that the java-stix library provides us rather than doing actual parsing (i.e. we implement support for new types by providing a handler that looks for objects of that type as the output of the parse).  If Stix moves to JSON, presumably the library will handle that transparently *or* we'll need another approach.

* `Can we handle just Cybox? Should this be factored to support them separately?`
We can create handlers for anything the java-stix library can parse, but cybox seems to be common and officially supported by the stix project.

* `Where is the documentation for the support in this PR?`
I added the new URI type to the README.md in metron-data-management.  Since that's the scope of this PR, not to document better our taxii/stix/etc support.

Ok, so it's apparent that some of the design decisions around taxii never made it into documentation.  A couple of questions for you:
* Where should that documentation live?
* Are we unhappy enough about having our abstraction bound to the (from what I can tell only) java library provided by the stix project that you'd like to start a discuss thread about developing a better approach to taxii?

Just a note on the second, we chose that because it was the only game in town other than parsing the XML ourselves and it was the officially supported library.  I even looked into that and decided against it as the XML format is extremely complex with lots of referential links that need to get coalesced to handle the blocks of stix that come across.

ottobackwards:
> Can we handle just Cybox

What I mean is, can we import a file that is **ONLY** cybox observables, without Stix.

Stix supports cybox.  But they are separate things.

ottobackwards:
As for where... over in METRON-777 @mmiklavc is talking about a better documentation area... this may fall into that

cestella:
Ah, for the moment we only have the StixExtractor.  We could MAKE a cybox extractor and it could reuse the handler logic, but we haven't done that.

cestella:
Yeah, it seems to me that we might need something like a location for discussion and description of architectural decisions made.

ottobackwards:
I would like to see at least the version of Stix and Cybox that is supported documented.

cestella:
Sure, not a problem.  I noted the versions supported in the documentation and linked to the actual schemata for cybox and stix.

ottobackwards:
+1 by inspection.

PR #694, opened 2017-08-14
METRON-1085: Add REST endpoint to save a user profile for the Alerts UI

merrimanr:
## Contributor Comments
This PR adds an ORM framework to the REST application.  An ORM (object relational mapping) framework provides tools for mapping java objects to relational database tables.  Hibernate is probably the most well-known but is not used here because of license issues.  Eclipselink is used instead and is a suitable replacement because it is Apache license friendly.

This PR also includes a REST endpoint that leverages this feature.  The intention is to give reviewers some context around how and why an ORM framework is needed.  The new endpoint allows a user to persist their searches in a database and also allows an admin user to manage users' searches.  These saved searches are part of an "AlertsProfile" that is used in the alerts UI and will replace the current approach of persisting user data client-side.  Saved searches are highlighted here but other information can (and will) be added as needed.  For example, a user's preferred display columns are also stored in this profile.  

This can be tested in full dev as follows:

1. Navigate to the [alert-controller](http://node1:8082/swagger-ui.html#!/alert-controller) in Swagger and login as "user1/password".
2. Create a new profile with the [save](http://node1:8082/swagger-ui.html#!/alert-controller/saveUsingPOST) function.  The actual values in the profile are not important as long as you can differentiate one profile from another.  This is the profile I used for user 1:
```
{
  "savedSearches": [
    {
      "name": "user 1 search",
      "searchRequest": {
        "facetFields": [
          "string"
        ],
        "from": 0,
        "indices": [
          "string"
        ],
        "query": "string",
        "size": 0
      }
    }
  ],
  "tableColumns": [
    "string"
  ]
}
```
3. Login as "user2/password" (a new incognito window in chrome will work) and create a profile.
4. The [get](http://node1:8082/swagger-ui.html#!/alert-controller/getUsingGET) function should return the profile of the logged in user.
5. Try to execute the [getall](http://node1:8082/swagger-ui.html#!/alert-controller/findAllUsingGET) or [delete](http://node1:8082/swagger-ui.html#!/alert-controller/deleteUsingDELETE) functions with either user1 or user2.  You should get a 403 permission error.
6. Login as "admin/password" in a new window.  You should now be able to execute the getall and delete functions.

There are a couple design choices I would like to highlight and get the community's feedback on:

1. I chose Eclipselink because it is license friendly and used in other Apache projects.  Is this the correct replacement for Hibernate?  
2. The "savedSearches" and "tableColumns" fields are persisted as JSON blobs in a single table instead of a normalized data model.  In my experience with ORM frameworks, it can sometimes be advantageous to store a serialized version in a single column as opposed to representing an entity with multiple tables.  Consider the "tableColumns" field.  A normalized approach would be a separate "TableColumn" table with a foreign key column that references the "AlertsProfile" table.  This is not attractive in this specific case for a couple reasons:  the "TableColumn" table will end up having mostly duplicate data and accessing the tables columns will only happen in one direction, you will never start with a table column and lookup an alerts profile.  The "TableColumn" table will grow over time, lookups will eventually become slower and database tuning will be needed.
3. I took the simpler approach to securing a user's profile.  The "id" on the "AlertsProfile" table is a user's id and the REST app uses the current user's id for lookup.  The only way to get another user's profile is to be logged in as an admin user and list all profiles.  Alternatively, this could have been achieved by managing a list of "AlertsProfile" ACLs but I felt that was overkill.
4. I added a short description and reference to Eclipselink in the REST README.  Is this sufficient documentation?

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/converter/JsonConverter.java
```diff
@@ -0,0 +1,49 @@
...
+
+@Converter
+public class JsonConverter implements AttributeConverter<Object, String> {
+
+  @Override
+  public String convertToDatabaseColumn(Object savedSearches) {
+    try {
+      return JSONUtils.INSTANCE.toJSON(savedSearches, false);
+    } catch (JsonProcessingException e) {
+      e.printStackTrace();
```
cestella: Why are we printing stacktraces?

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/converter/JsonConverter.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+    return null;
+  }
+
+  @Override
+  public Object convertToEntityAttribute(String savedSearches) {
+    try {
+      return JSONUtils.INSTANCE.load(savedSearches, Object.class);
+    } catch (IOException e) {
+      e.printStackTrace();
```
cestella: Same question re stacktraces

dependencies_with_url.csv
```diff
@@ -112,6 +112,7 @@ com.fasterxml.jackson.core:jackson-core:jar:2.6.6:compile,ASLv2,https://github.c
 com.fasterxml.jackson.core:jackson-core:jar:2.7.4:compile,ASLv2,https://github.com/FasterXML/jackson-core
 com.fasterxml.jackson.core:jackson-core:jar:2.8.3:compile,ASLv2,https://github.com/FasterXML/jackson-core
 com.fasterxml.jackson.core:jackson-databind:jar:2.2.3:compile,ASLv2,http://wiki.fasterxml.com/JacksonHome
+com.fasterxml.jackson.core:jackson-databind:jar:2.4.3:compile,ASLv2,http://github.com/FasterXML/jackson
```
cestella: These dependencies tend to be fragile.  Moving up a version can have impact.  Have you run this up on full-dev?

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/converter/JsonConverter.java
merrimanr: You are correct, that was a mistake.  Recent commit changes this to use slf4j Logger.

dependencies_with_url.csv
merrimanr: This version was brought in by eclipselink.  It has been tested on full dev and everything seems to work fine.  If you think this may cause problems down the road, I'm happy to investigate excluding it and staying at the current jackson version in metron-rest.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/AlertsProfileControllerIntegrationTest.java
```diff
@@ -0,0 +1,226 @@
...
+    this.mockMvc.perform(delete(url + "/user2").with(httpBasic(admin, password)))
+        .andExpect(status().isOk());
+
+    this.mockMvc.perform(get(url).with(httpBasic(user1, password)))
+        .andExpect(status().isNotFound());
+
+    this.mockMvc.perform(get(url).with(httpBasic(user2, password)))
+        .andExpect(status().isNotFound());
+
+    this.mockMvc.perform(get(url + "/all").with(httpBasic(admin, password)))
```
ottobackwards: I agree having a huge pack of tests in one method is not preferable, but this is consistent with the rest of metron-rest

mmiklavc: I'm all for consistency, but I'm not sure I see the value in continuing a pattern that is undesirable. I'd agree that we shouldn't fix all the other tests in this PR, but these are new.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/AlertsProfileControllerIntegrationTest.java
```diff
@@ -0,0 +1,226 @@
...
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+import org.springframework.web.context.WebApplicationContext;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles(TEST_PROFILE)
+public class AlertsProfileControllerIntegrationTest {
+
+  /**
+   * { "tableColumns": ["user1_field"], "savedSearches": [ { "name": "user1 search 1",
```
merrimanr: This was an attempt to adhere to the new style guidelines.  This was the only way I could find to get under the line length limit.  This applies to comments you've made in other places as well.  I agree it would be better if it were more readable so I will make those changes.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/AlertsProfileControllerIntegrationTest.java
```diff
@@ -0,0 +1,226 @@
...
+   * { "name": "user1 search 2", "searchRequest": { "from": 10, "indices": ["snort"],
+   * "query": "*", "size": 10 } } ]
+   * }
+   */
+  @Multiline
+  public static String user1ProfileJson;
+
+
+  /**
+   * { "tableColumns": ["user2_field"], "savedSearches": [ { "name": "user2 search 1",
```
merrimanr: Attempting to adhere to style guidelines.  Will revert to a more readable format.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/SearchRequest.java
```diff
@@ -106,4 +107,33 @@ public void setSort(List<SortField> sort) {
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    SearchRequest that = (SearchRequest) o;
+
+    return indices != null ? indices.equals(that.indices) : that.indices == null
```
merrimanr: Attempting to adhere to style guidelines.  Will revert to a more readable format.

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/model/AlertsProfile.java
```diff
@@ -0,0 +1,89 @@
...
+  public List<SavedSearch> getSavedSearches() {
+    return savedSearches;
+  }
+
+  public void setSavedSearches(List<SavedSearch> savedSearches) {
+    this.savedSearches = savedSearches;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
merrimanr: This is used in unit tests to test for equility.

metron-interface/metron-rest-client/src/main/java/org/apache/metron/rest/model/AlertsProfile.java
```diff
@@ -0,0 +1,89 @@
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    AlertsProfile that = (AlertsProfile) o;
+
+    return id != null ? id.equals(that.id)
```
merrimanr: Attempting to adhere to style guidelines. Will revert to a more readable format.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/AlertsProfileControllerIntegrationTest.java
merrimanr: This is not just consistent with the other integration tests in metron-rest, this is consistent with integration tests throughout Metron.  If you look at our integration tests across all modules, the majority of them only have 1 test method.  Integration tests are different than unit tests because they generally are simulating a use case, state changes between the individual tests, and tests depend on state.  Unit tests are different because each test usually sets up it's own state with mocks and test objects.

There are probably some other integration tests that we could break into smaller tests (security tests can be separate because they don't change state) but most of the time we can't.  I'm not sure I could move any of these tests because they all depend on previous endpoints being executed.  For example if I were to move the list users test to a different test it would no longer pass because it depends on the creation/deletion of users in other tests.

The only benefit I can see to doing this is to provide clarity on what is being tested.  If I added comments before each test explaining their purpose would that achieve the same result?

merrimanr:
The latest commit addresses the formatting comments.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/AlertsProfileControllerIntegrationTest.java
mmiklavc: Yeah, minimally we should explain what's going on and what the pre-reqs are for a test case. I agree with us having existing monolithic tests, and this isn't the right place to redesign everything. I prefer not having a large waterfall of dependencies between the steps. Ttehere are side effects between the steps that just aren't immediately obvious. 

Side note - this is mocking the mvc architecture - is it spinning up anything expensive? Would there be much cost in having each test do the necessary setup in an `@Before` method or within the test method itself?

Even better than documentation, I'd prefer splitting off well-named methods that explicitly state what the step is doing, e.g. 
```
@Test
public void add_profile_makes_mike_happy() {
    assertFalse(checkProfileExists(user1, password));
    assertFalse(checkProfileExists(user2, password));
    addUserProfile(user1, password, profile1);
    addUserProfile(user2, password, profile2);
    assertTrue(checkProfileExists(user1, password));
    assertTrue(checkProfileExists(user2, password));
}
```

As it's currently laid out, I find it difficult to know what the tests are doing.

Thoughts?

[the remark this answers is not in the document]
merrimanr: There is a technical issue here and that's annotating multiple methods with @Test.  This changes the lifecycle of the tests and makes it so that each @Test method has to do it's own setup and teardown.  Not a big deal with unit tests because functions are tested in isolation with mock objects but it's harder for integration tests.  This test is not mocking anything, the same in-memory components that our other integration tests use serve as the backend.  We definitely don't want to spin them up/down before each test.  In fact, the REST integration tests only spin them up/down once for all tests.

If we were to organize the cases into different methods but NOT annotate them with @Test would that be ok?  I would rather document the state each test expects rather than recreating it.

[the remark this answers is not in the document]
mmiklavc: Sounds reasonable. That would definitely be an improvement. If there are no mocks here, we should probably change this `this.mockMvc...` to this `this.mvc...` That is misleading.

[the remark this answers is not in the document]
merrimanr: I added more documentation and split the use cases into separate methods.  Let me know if this is clear enough now.

My previous comment wasn't entirely accurate.  The test is mocking the http layer which is what the mockMvc object is for.  The backend is not mocked though and my comments about  not wanting to initialize backend components before each test still stands.

merrimanr:
The latest commit merges in master and also refactors this endpoint into the AlertService and AlertController classes that were recently accepted into master.  I think this will make it cleaner by having all the alert-related functions in a single controller.  

This is ready for another pass @mmiklavc when you have time.  Check out the changes I made to AlertController and let me know if they make sense.

mmiklavc:
Thanks @merrimanr for the contribution, and thanks for making those changes - big improvement. +1

PR #706, opened 2017-08-17
METRON-1119 Bad Out-of-Order Logic for SaltyRowKeyBuilder

nickwallen:
This is a bug that was previously Identified by @mattf-horton 
https://github.com/apache/metron/pull/622#pullrequestreview-51351898

The 'out-of-order' logic for the `SaltyRowKeyBuilder.encode` function does not work.

> If it starts out in the wrong order, say end is 1 and start is 5, won't this pair of statements result in both end and start being equal to the larger, ie 5 ? We need an intermediate variable for a binary swap!

## Pull Request Checklist


cestella:
What's the consequence of this one?  How does the error look like in the wild?

nickwallen:
This is a bug that would be hard to trigger.  That is probably why it has been in the code base for quite some time.  It took @mattf-horton 's sharp eyes to uncover it.

This bug cannot be triggered by a user, it would have to be triggered by a developer using the `SaltyRowKeyBuilder`.  The developer would have to accidentally reverse the start and end when calling the `rowKeys` function.  If that happens, it would cause an infinite loop, which would be tragic.

mattf-horton:
+1, lgtm

nickwallen:
@cestella Any other thoughts on this one?  Otherwise I will get it merged.  

Thanks for the reviews, @mattf-horton and @cestella

cestella:
Nope, +1 by inspection

PR #707, opened 2017-08-18
METRON-1121 Ignore Profile with Bad 'init', 'update' or 'groupBy'

nickwallen:
The Profiler needs to gracefully ignore profiles that have a bad 'init', 'update', or 'groupBy' expression.  Under certain conditions a message may not be applied to all valid profiles, due to a single bad profile.

If the Profiler is running with multiple profiles defined and one of those profiles has a Stellar expression that throws an exception for either the  'init', 'update', or 'groupBy', the current logic will not continue to attempt to apply the message to the remaining valid profiles.  Only valid profiles executed before the bad profile will have the message applied.

## Testing

Unit tests have been added to check for these conditions.  This can be manually tested with either a live Profiler running in a cluster or using the Profiler debugging functions.  

Create two profiles; one that is valid and another that is invalid, like the following.  Ensure that all test messages that are sent through the Profiler, reach the `good-profile` despite the `bad-profile`.

```
 {
   "profiles": [
      {
        "profile": "bad-profile",
        "foreach": "ip_src_addr",
        "init":   { "x": "2 / 0" },
        "update": { "x": "x + 1" },
        "result": "x"
      },
      {
        "profile": "good-profile",
        "foreach": "ip_src_addr",
        "init":   { "x": "0" },
        "update": { "x": "x + 1" },
        "result": "x"
      }
   ]
 }
```

## Pull Request Checklist

### DIFF
--- a/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
+++ b/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
@@ -131,7 +131,6 @@ public class DefaultProfileBuilder implements ProfileBuilder, Serializable {
       assign(definition.getUpdate(), message, "update");
 
     } catch(Exception e) {
-      // mark the profile in error
       LOG.error(format("Unable to apply message to profile: %s", e.getMessage()));
     }
   }
--- a/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
+++ b/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
@@ -130,8 +130,8 @@ public class DefaultProfileBuilder implements ProfileBuilder, Serializable {
 
       assign(definition.getUpdate(), message, "update");
 
-    } catch(Exception e) {
-      LOG.error(format("Unable to apply message to profile: %s", e.getMessage()));
+    } catch(Throwable e) {
+      LOG.error(format("Unable to apply message to profile: %s", e.getMessage()), e);
     }
   }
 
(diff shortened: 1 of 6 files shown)


cestella:
+1 by inspection

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
```diff
@@ -140,33 +144,41 @@ public void apply(JSONObject message) {
...
+      result = Optional.of(new ProfileMeasurement()
+              .withProfileName(profileName)
+              .withEntity(entity)
+              .withGroups(groups)
+              .withPeriod(clock.currentTimeMillis(), periodDurationMillis, TimeUnit.MILLISECONDS)
+              .withProfileValue(profileValue)
+              .withTriageValues(triageValues)
+              .withDefinition(definition));
+
+    } catch(Exception e) {
```
cestella: Should this be a Throwable?

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
```diff
@@ -140,33 +144,41 @@ public void apply(JSONObject message) {
...
+      // if any of the Stellar expressions fail, a measurement should NOT be returned
+      LOG.error(format("Unable to flush profile: %s", e.getMessage()));
```
cestella: Can we pass the exception as the second argument to `LOG.error`

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
```diff
@@ -253,7 +265,8 @@ private void assign(Map<String, String> expressions, Map<String, Object> transie
               .forEach((expr) -> results.add(executor.execute(expr, transientState, Object.class)));
 
     } catch (Throwable e) {
-      String msg = format("Bad '%s' expression: %s, profile=%s, entity=%s", expressionType, e.getMessage(), profileName, entity);
+      String msg = format("Bad '%s' expression: error=%s, profile=%s, entity=%s", expressionType, e.getMessage(), profileName, entity);
+      LOG.error(msg);
```
cestella: Can you pass `e` as the second argument to `LOG.error`?

nickwallen:
Good eye, @cestella .  Latest commit is catching Throwable and logging e's.

cestella:
Yep, +1 by inspection, thanks!

PR #711, opened 2017-08-23
METRON-1127: Add ability to escalate alerts for external consumption

merrimanr:
## Contributor Comments
This PR exposes a way to put alerts (or any message) into a configurable escalation Kafka topic.  In summary:
- Adds a "metron_escalation_topic" parameter to MPack and Spring property file
- Provides a default value and sets up the topic upon MPack installation
- Adds a new AlertController for this and future alert-related functions
- Adds a producer to the Kafka service and controller
- Adds and updates the appropriate unit and integration tests, including making the KafkaControllerIntegrationTest simpler
- Adds documentation to the README

This can be tested in full dev with Swagger by first ensuring the "escalation" topic is automatically created and showing up in the list of topics (use the "/api/v1/kafka/topic" endpoint).  Next use "/api/v1/alert/escalate" endpoint in the alert-controller to escalate an alert, you should receive a 200 response code.  Finally get the latest message from the Kafka "escalation" topic with "/api/v1/kafka/topic/{name}/sample" endpoint.  You should get the same message that you escalated in the previous step.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
@@ -1,0 +1,57 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.metron.rest.service.impl;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.util.Map;
+import org.apache.metron.common.utils.JSONUtils;
+import org.apache.metron.rest.MetronRestConstants;
+import org.apache.metron.rest.RestException;
+import org.apache.metron.rest.service.AlertService;
+import org.apache.metron.rest.service.KafkaService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Service;
+
+/**
+ * The default service layer implementation of {@link AlertService}.
+ *
+ * @see AlertService
+ */
+@Service
+public class AlertServiceImpl implements AlertService {
+
+  private Environment environment;
+  private final KafkaService kafkaService;
+
+  @Autowired
+  public AlertServiceImpl(final KafkaService kafkaService,
+                          final Environment environment) {
+    this.kafkaService = kafkaService;
+    this.environment = environment;
+  }
+
+  @Override
+  public void escalateAlert(Map<String, Object> alert) throws RestException {
+    try {
+      kafkaService.produceMessage(environment.getProperty(MetronRestConstants.KAFKA_TOPICS_ESCALATION_PROPERTY), JSONUtils.INSTANCE.toJSON(alert, false));
+    } catch (JsonProcessingException e) {
+      throw new RestException(e);
+    }
+  }
+}
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
@@ -18,6 +18,7 @@
 package org.apache.metron.rest.service.impl;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import java.util.List;
 import java.util.Map;
 import org.apache.metron.common.utils.JSONUtils;
 import org.apache.metron.rest.MetronRestConstants;
@@ -47,9 +48,13 @@ public class AlertServiceImpl implements AlertService {
   }
 
   @Override
-  public void escalateAlert(Map<String, Object> alert) throws RestException {
+  public void escalateAlerts(List<Map<String, Object>> alerts) throws RestException {
     try {
-      kafkaService.produceMessage(environment.getProperty(MetronRestConstants.KAFKA_TOPICS_ESCALATION_PROPERTY), JSONUtils.INSTANCE.toJSON(alert, false));
+      for (Map<String, Object> alert : alerts) {
+        kafkaService.produceMessage(
+            environment.getProperty(MetronRestConstants.KAFKA_TOPICS_ESCALATION_PROPERTY),
+            JSONUtils.INSTANCE.toJSON(alert, false));
+      }
     } catch (JsonProcessingException e) {
       throw new RestException(e);
     }
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-rest-env.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-rest-env.xml
@@ -109,4 +109,10 @@
             <empty-value-valid>true</empty-value-valid>
         </value-attributes>
     </property>
+    <property>
+        <name>metron_escalation_topic</name>
+        <description>Escalated alerts will be produced to this topic</description>
+        <value>escalation</value>
+        <display-name>Metron escalation topic</display-name>
+    </property>
 </configuration>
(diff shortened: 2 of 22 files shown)


metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+  private Environment environment;
+  private final KafkaService kafkaService;
+
+  @Autowired
+  public AlertServiceImpl(final KafkaService kafkaService,
+                          final Environment environment) {
+    this.kafkaService = kafkaService;
+    this.environment = environment;
+  }
+
```
ottobackwards: Is this doing a single alert at a time?
Does it actually pass all the alert data from the client to the endpoint?

merrimanr: I can adjust it to accept a list of alerts.  I agree that would be better.

Yes the endpoint expects all alert data to be passed in.  I'm trying to avoid an extra ES call (alert data will travel across the network either way).  

But now that you bring it up I'm not sure.  Do we want to make it flexible so that an alert can be annotated/modified before being escalated?  Would fetching an alert from ES be unnecessary since it's assumed the client will already have a copy?  Would we want to ensure that the most recent version in ES is always escalated?  Should we expose both options?

ottobackwards: What I was thinking about is the assumption that the client will have the 'full alert'.
If the client requests alerts based on queries, and specifies the fields for return as a subset of the complete fields, then you are making the assumption that the subset is what will be required for escalation.  These are different use cases.  It is more likely that what will be required is:

- Query from any point to service returned sum subset of fields for alert A.
- User decides to escalate, and has some option to add new fields/data/annotations to alerts  ( best would we to have some 'additional field list + description' per escalation target that we can load into the ui ) to alert.
- User may add same fields/annotations to multiple alerts ( escalate all in query )
- Escalation takes annotations and alert id's, merges the annotations with the full alert (or from a query that returns the fields required by the target?) and sends them

merrimanr: I was coming at it from a transactional perspective and trying to avoid a race condition (alert changes after a client requests it and before it gets escalated).  I think a lot of what you describe can already be done by chaining rest calls together.

ottobackwards: Can you give an example?  I don't see how.  

escalate(List<?> alertId, String patchFields)?

When we get the alerts.... is there a version - so that we will just work with the version we have and even if v+1 goes in it won't matter?

ottobackwards:
This is an important piece of functionality.  I think it requires more than the rest endpoint documentation.   I'm not sure where to put that though.  I almost imagine like recipes for rest.

creating_a_parser_and_starting_it.md
querying_and_escalating_alerts.md

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
simonellistonball: My understanding is that this is for manual escalation, and should just expose everything in the alert (or meta-alert) that is being manually escalated. 

The user would essentially add comments and make modifications before escalating, which would be persisted 

The alert which is pushed onto the queue should be the current alert state, i.e. with all those modifications. This was done elsewhere (sorry, can't remember the ticket number). So the alert data being pushed should come off the cluster side, and it should have all the content already sent from the client by a separate process. 

I think that should achieve pretty much what you're talking about @ottobackwards.

simonellistonball:
I'd say the docs belong with the UI docs, since this is pretty much an endpoint to drive the UI buttons, no?

ottobackwards:
Platform wise, we want both.   The metron configuration UI is the 'default' ui for the project, but the rest api is the main supported interface.  As such they should be treated as separate entities, each documented on their own.  There could be many other rest clients.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
ottobackwards: @simonellistonball I don't think that is what this api does.  We need to clarify.
There is also a difference between the rest api and the ui workflow.  They both have to make sense on their own.

Do we need a discussion thread?  or a document?

merrimanr: Documentation issues aside, are we concerned about a potential race condition?  It sounds like we are.  In that case we would need to retrieve the alerts from ES in the REST service before they are escalated.  Is my understanding correct?

ottobackwards: I wouldn't use the word concerned.  I am interested in the use cases and api design.

nickwallen: > Yes the endpoint expects all alert data to be passed in. I'm trying to avoid an extra ES call (alert data will travel across the network either way).

I think @merrimanr 's implementation makes the most sense right now.  We assume the client has all of the necessary details about the alert that they want to escalate.  Adding extra logic to pull the freshest version from the index (ES) does not make sense to me.

As a SOC analyst using the Alerts UI, if I escalate an alert, I want to know that what I am seeing on my screen is what I actually escalated.  I escalated the alert based on the information that I had in front of me; it is a point-in-time snapshot.

If the alert changes between the time that I loaded the alert on my screen and when I clicked escalate, then that is a condition that can only be correctly recovered from in the UI, not in the REST API.  

In addition, the result of escalation here is that we are pushing data into a Kafka topic.  The only reasonable expectation is that this data is a point-in-time snapshot.  We don't need to jump through hoops to try and sync with any changes in the index.

If the alerts data needs refreshed, then the consumer of the 'escalation' Kafka topic needs to do that.  Whether the data needs refreshed and how fresh that needs to be are decisions that only the consumer of that topic can make.

nickwallen:
I don't see the need for additional docs.  Escalating an alert here means to push the alert data to a Kafka topic.  That is all documented in the README, as far as I can see.  Additional detailed docs about `querying and escalating alerts` would be specific to the Alerts UI or whatever other UI uses this REST API.  Additional docs should go there.

simonellistonball:
@ottobackwards agreed, this is very separate from the management ui (won't touch, or be used by anything in the management ui). Also agreed this is a separate entity, but one that will be used by the alerts ui. Honestly, I'm not sure what else this needs to do. If we want to expand the scope, maybe we can do that as a follow on.

@nickwallen thinking about it, I'm inclined to agree that the path of least surprise for the soc user is the one to go for here, so you're right. The representation passed should be that shown in the UI. The challenge here is things like meta-alerts, where the actual alert to pass may not be fully present on the ui due to paging on its member alerts. (I can certainly see use cases where a meta-alert might contain O(1000s) of alerts).

ottobackwards:
part of the issue is there is no use case for how this is going to be used beyond the general.  How often do analysts that have possibly 1000s of alerts find the one magic alert?

I can imagine a flow in the ui like:

Do some queries and dig around, select some set of alerts from there to escalate, get an escalate dialog where you add some info, fill out some fields required for target routing etc ( so you need per target config unless you are going to automate it with nifi in-between? ) and then send the alerts as a set.

Aside from that, to restate where I am coming from:

* it is my exp. that escalateTheseAlerts is preferable and more practical than escalateThisOneAlert
* it is my exp. that most query api with possibly huge returns, paging etc for dense data objects ( remember the original string etc ) try to return contextually appropriate subsets of the data, across different network zones etc for performance etc, and the assumption that what the UI has is what you need to push to the other system is not a good one.  There is def. a tradeoff between optimizing for server-client ( only send what you need to, LDAP comes to mind for example, you get the props you need ) and not wanting to make other ES/? calls to get the real data.

These things are worth talking about though.

At this point in our REST api, we only have the main readme.  So just the call descriptions.  We don't have examples, more in depth docs per call or anything.  That is just were we are at.  But in the context of the PR I think we can talk about the call patterns, and maybe we should have a usage example.  Since we did not have a DISCUSS on this before hand, the PR is it.

nickwallen:
@ottobackwards Those are all important points to consider.  I just don't know that we need to think through all of that right now.  I am thinking of this PR as a very basic, first pass.  Let's get some really basic functionality in, work with it, learn from it, then iterate a few times.  Let's solve those problems as we go, piece-by-piece.  

@merrimanr @simonellistonball @all Do we have a JIRA for the basic UI changes that would leverage the functionality in this PR?  Maybe if we get that JIRA defined and documented we can better assess whether the functionality that we have here will work for a first pass.  I don't think we need any grand vision or anything; just something stupid simple.

merrimanr:
I create a Jira for this:  https://issues.apache.org/jira/browse/METRON-1147.  It's could probably use some more description but this is a start.

nickwallen:
> The alert or list of alerts should be annotated with the user who escalated the alert(s) and time the alert(s) was escalated.

That JIRA works for me @merrimanr .  So the assumption right now is that the caller (the Alerts UI) would do the annotation that is required?

merrimanr:
I don't think it matters that much but I'm leaning towards keeping the REST layer simple.  The Alerts UI will be responsible for adding other annotations so it's not a stretch to add these too.  Are there any advantages to pushing this responsibility to REST?  I don't have a strong opinion either way.

nickwallen:
I think your current approach works best right now.  Keep it simple.  +1

nickwallen:
@ottobackwards What do you think about this one?  Good enough for a first pass?

ottobackwards:
@nickwallen 
The question of the assumption of how complete the passed data is can wait.
My point about escalateOne vs. escalateMany stands.

nickwallen:
Ok, then I will leave that for @merrimanr to respond to.  In my opinion we can tackle escalate many in a future PR.

merrimanr:
The latest commit addresses the escalateMany request.

ottobackwards:
+1

PR #712, opened 2017-08-23
METRON-1129 Management UI - Package Node Dependencies in RPM

nickwallen:
If you install the Management UI using the RPM, it will not install all of the required Node dependencies.  You have to manually run `npm install ...` to get those dependencies installed.  

This works just fine in an environment that can access the interwebs.  This does not work so well when installing in an offline environment with no interweb access.  If we package all Node dependencies in the RPM itself, this makes it much simpler to install in an offline environment.  And this, of course, will also work in an online environment and removes a manual steps.  So win, win.

### Testing

I tested this change in two ways.
* I built the RPMs and then manually copied them to a raw CentOS box.  I installed Node on the box, but did not install any dependencies with NPM.  I then installed the RPMs and was able to launch the Management UI.  All dependencies were satisfied based on what was contained in the RPM.
* I launched Full Dev, which builds the RPMs and deploys them with the Mpack.  The Management UI and all other core functions worked.

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/management_ui_master.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/management_ui_master.py
@@ -38,7 +38,6 @@ class ManagementUIMaster(Script):
         from params import params
         env.set_params(params)
         self.install_packages(env)
-        Execute('npm --prefix ' + params.metron_home + '/web/expressjs/ install')
 
     def configure(self, env, upgrade_type=None, config_dir=None):
         print 'configure managment_ui'
--- a/metron-deployment/packaging/docker/rpm-docker/Dockerfile
+++ b/metron-deployment/packaging/docker/rpm-docker/Dockerfile
@@ -27,3 +27,7 @@ RUN mv apache-maven-3.2.5 /opt/maven
 RUN ln -s /opt/maven/bin/mvn /usr/bin/mvn
 RUN yum -y install asciidoc rpm-build rpm2cpio tar unzip xmlto zip rpmlint && yum clean all
 WORKDIR /root
+
+# install node so that the node dependencies can be packaged into the RPMs
+RUN curl --silent --location https://rpm.nodesource.com/setup_6.x | bash -
+RUN yum -y install gcc-c++ make nodejs
--- a/metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
+++ b/metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
@@ -17,7 +17,7 @@
 %define timestamp           %(date +%Y%m%d%H%M)
 %define version             %{?_version}%{!?_version:UNKNOWN}
 %define full_version        %{version}%{?_prerelease}
-%define prerelease_fmt      %{?_prerelease:.%{_prerelease}}          
+%define prerelease_fmt      %{?_prerelease:.%{_prerelease}}
 %define vendor_version      %{?_vendor_version}%{!?_vendor_version: UNKNOWN}
 %define url                 http://metron.apache.org/
 %define base_name           metron
@@ -29,6 +29,8 @@
 
 %define metron_root         %{_prefix}/%{base_name}
 %define metron_home         %{metron_root}/%{full_version}
+
+%define _binaries_in_noarch_packages_terminate_build   0
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -90,6 +92,8 @@ tar -xzf %{SOURCE11} -C %{buildroot}%{metron_home}
 
 install %{buildroot}%{metron_home}/bin/metron-rest %{buildroot}/etc/init.d/
 install %{buildroot}%{metron_home}/bin/metron-management-ui %{buildroot}/etc/init.d/
+
+npm install --prefix="%{buildroot}%{metron_home}/web/expressjs" --only=production
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -389,6 +393,8 @@ This package installs the Metron Management UI %{metron_home}
 %dir %{metron_home}/bin
 %dir %{metron_home}/web
 %dir %{metron_home}/web/expressjs
+%dir %{metron_home}/web/expressjs/node_modules
+%dir %{metron_home}/web/expressjs/node_modules/.bin
 %dir %{metron_home}/web/management-ui
 %dir %{metron_home}/web/management-ui/assets
 %dir %{metron_home}/web/management-ui/assets/ace
@@ -399,6 +405,8 @@ This package installs the Metron Management UI %{metron_home}
 %dir %{metron_home}/web/management-ui/license
 %{metron_home}/bin/metron-management-ui
 /etc/init.d/metron-management-ui
+%attr(0755,root,root) %{metron_home}/web/expressjs/node_modules/*
+%attr(0755,root,root) %{metron_home}/web/expressjs/node_modules/.bin/*
 %attr(0755,root,root) %{metron_home}/web/expressjs/server.js
 %attr(0644,root,root) %{metron_home}/web/expressjs/package.json
 %attr(0644,root,root) %{metron_home}/web/management-ui/favicon.ico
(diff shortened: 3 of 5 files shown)


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/management_ui_master.py
```diff
@@ -38,7 +38,6 @@ def install(self, env):
         from params import params
         env.set_params(params)
         self.install_packages(env)
-        Execute('npm --prefix ' + params.metron_home + '/web/expressjs/ install')
```
nickwallen: We don't need the MPack to install the Node dependencies any longer.

metron-deployment/packaging/docker/rpm-docker/Dockerfile
```diff
@@ -27,3 +27,7 @@ RUN mv apache-maven-3.2.5 /opt/maven
 RUN ln -s /opt/maven/bin/mvn /usr/bin/mvn
 RUN yum -y install asciidoc rpm-build rpm2cpio tar unzip xmlto zip rpmlint && yum clean all
 WORKDIR /root
+
+# install node so that the node dependencies can be packaged into the RPMs
+RUN curl --silent --location https://rpm.nodesource.com/setup_6.x | bash -
+RUN yum -y install gcc-c++ make nodejs
```
nickwallen: The Docker container needs to have Node installed.  It is in this container that we fetch and build all of the Node dependencies that get added to the RPM.

metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
```diff
@@ -30,6 +30,8 @@
 %define metron_root         %{_prefix}/%{base_name}
 %define metron_home         %{metron_root}/%{full_version}
 
+%define _binaries_in_noarch_packages_terminate_build   0
```
nickwallen: Some of the Node dependencies have shell scripts.  The RPM builder would complain that we are adding shell scripts that need /bin/sh into a 'noarch' RPM.  I had to turn this check off to get the RPMs to build.  The other alternative would be to make this an arch-specific RPM, but that doesn't seem necessary to me.

metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
```diff
@@ -91,6 +93,9 @@ tar -xzf %{SOURCE11} -C %{buildroot}%{metron_home}
 install %{buildroot}%{metron_home}/bin/metron-rest %{buildroot}/etc/init.d/
 install %{buildroot}%{metron_home}/bin/metron-management-ui %{buildroot}/etc/init.d/
 
+# allows node dependencies to be packaged in the RPMs
+npm install --prefix="%{buildroot}%{metron_home}/web/expressjs" --only=production
```
nickwallen: We package only production dependencies.  We know that all production dependencies are Apache license compatible.  Some of the development dependencies are not Apache license compatible.

metron-deployment/packaging/docker/rpm-docker/build.sh
```diff
@@ -36,7 +36,7 @@ if [ $? -ne 0 ] && [ $OWNER_UID -ne 0 ]; then
 fi
 
 rm -rf SRPMS/ RPMS/ && \
-rpmbuild -v -ba --define "_topdir $(pwd)" --define "_version ${VERSION}" --define "_prerelease ${PRERELEASE}" SPECS/metron.spec && \
+QA_SKIP_BUILD_ROOT=1 rpmbuild -v -ba --define "_topdir $(pwd)" --define "_version ${VERSION}" --define "_prerelease ${PRERELEASE}" SPECS/metron.spec && \
```
nickwallen: This was another annoying issue.  When NPM builds the dependencies, it adds a `package.json` to each dependency.  In this `package.json`, there is a "_where" field that ends up containing the complete build path from where the RPMs are built.  

Based on some research I can find no way to get rid of this absolute path.  And I also know that Node/NPM does not actually use this for anything.  So it is harmless, but triggers the RPM checks, which I had to disable.

https://github.com/npm/npm/issues/10393

Is there a better way?

metron-interface/metron-config/README.md
```diff
@@ -49,12 +49,6 @@ This module provides a user interface for management functions in Metron.
     rpm -ih metron-config-$METRON_VERSION-*.noarch.rpm
     ```
 
-1. Install the [Express](https://expressjs.com/) web framework from the `package.json` file in `$METRON_HOME/web/expressjs`:
-
```
nickwallen: When installing from the RPMs, there is no need to do this manually anymore.

metron-deployment/packaging/docker/rpm-docker/Dockerfile
ottobackwards: Right now, we copy everything we need into the docker rpm environment using assembly and resource copy.  Since we have *just* built the node stuff in the build phase, is there a way we can assemble and resource copy the node dependencies in?  or does this have to be done on the spot regardless.

nickwallen: I actually tried this approach first and ran into some headaches.  Here are the headaches...:)

First, the build host may be different than the installation host.  `npm install ...` sometimes builds/rebuilds dependencies that I assume are customized for the installation environment.  So if I build the RPMs on my Mac, I don't want NPM dependencies built for the Mac to be included in an RPM destined for a CentOS host.

In addition the dependencies built on the build machine are the "development" dependencies and not the "production" dependencies.  We only want the "production" deps to be packaged.

Also, there are two different `package.json` files.  I don't understand why, but there are.  The deps that get built on the build machine are defined by the root `metron-interface/metron-config/package.json` file.  While what we actually need to run the Management UI is defined in `metron-interface/metron-config/scripts/package.json`.

ottobackwards: So, do we know if there are issues between centos 7 and 6 then?

nickwallen: I have only tested it on CentOS 6.

merrimanr:
I tested this in full dev and everything works.  The only issue I'm seeing is during the rpm build.  There are several instances of this error being reported by rpmlint:

"This text file has executable bits set or is located in a path dedicated for
executables, but lacks a shebang and cannot thus be executed.  If the file is
meant to be an executable script, add the shebang, otherwise remove the
executable bits or move the file elsewhere."

Should we change these files to not be executable?

nickwallen:
I took a look through the permissions that get set when running `npm install` manually.  It appears that it is kind of a mixed bag.  Each module kind of does its own thing.  I could try to either...
* Not set permissions in the RPM spec, which would hopefully take whatever perms are set by NPM when building the RPMs.
* Or set to 644 to match some of the other web assets that we package in the RPM.

nickwallen:
If I don't set permissions, then those warnings still occur as some of the packages have files that violate this rpmlint check.

If I set to 644, then I get the following warnings.
```
metron-config.noarch: E: non-standard-dir-perm /usr/metron/0.4.1/web/expressjs/node_modules/isobject 0644L
A standard directory should have permission set to 0755. If you get this
message, it means that you have wrong directory permissions in some dirs
included in your package.
```

Exhausting those two ideas, I am not sure there is a good way to work around this.  Other thoughts @merrimanr ?

merrimanr:
I think suppressing the warnings is good enough for now.

JonZeolla:
@nickwallen do you need any additional testing on this?  I would love to see it get into our upcoming release.  From what I can tell, the only thing outstanding is the suppression, correct?

nickwallen:
I have found that I cannot suppress the warnings.  At this point I'd propose that we accept this in it's current state.  Although I am totally open to alternative suggestions.

nickwallen:
Yes the more testing the better, Jon.

JonZeolla:
I'll be unable to test until tonight but I will spin things up asap.

JonZeolla:
+1, spun up full dev and combed through the changes; all appears to be good.  Similarly did some `rpmlint`ing:
```
# rpmlint metron-config-0.4.1-201709020023.noarch.rpm  | grep -v script-without-shebang | grep -v hidden-file-or-dir
metron-config.noarch: I: enchant-dictionary-not-found en_US
metron-config.noarch: W: no-documentation
metron-config.noarch: E: zero-length /usr/metron/0.4.1/web/expressjs/node_modules/mime/.npmignore
metron-config.noarch: W: non-standard-dir-in-usr metron
metron-config.noarch: E: subsys-not-used /etc/init.d/metron-management-ui
metron-config.noarch: W: incoherent-init-script-name metron-management-ui ('metron-config', 'metron-configd')
```

After a bit of digging into the above `W` and `E`s, I think we should address the `subsys-not-used` as a very simple follow-on, but otherwise this looks good to me.

ottobackwards:
Is there a jira for these errors?  If not, can someone who has been working on the errors capture and create a jira please?

ottobackwards:
and reference it back here

nickwallen:
https://issues.apache.org/jira/browse/METRON-1155

ottobackwards:
thanks @nickwallen!

metron-deployment/packaging/docker/rpm-docker/build.sh
juanjoDiaz: Just saw this by chance. I can recommend using https://www.npmjs.com/package/removeNPMAbsolutePaths which is already use by other Apache projects like Cordova to remove the absolute paths created by NPM.

PR #717, opened 2017-08-25
METRON-1134: Allow parser command line options to be specified in the zookeeper parser config.

cestella:
## Contributor Comments
It would be useful to maintain sensible defaults for various properties in the construction of parser topologies within the sensor parser config in zookeeper. This, along with a UI, would allow people to manage defaults per sensor in a single place. CLI should take precedent and override anything stored in zookeeper.

Manual testing via fulldev should focus on ensuring the following conditions hold for the properties able to be specified in the config:
* Specify the property in the config.
* Start parser and ensure that the property is preserved.
* Specify the property via the CLI param
* Start parser and ensure that the CLI param is honored

The properties able to be specified in the config are as follows: 
* `spoutParallelism` : The kafka spout parallelism (default to `1`).  This can be overridden on the command line.
* `spoutNumTasks` : The number of tasks for the spout (default to `1`). This can be overridden on the command line.
* `parserParallelism` : The parser bolt parallelism (default to `1`). This can be overridden on the command line.
* `parserNumTasks` : The number of tasks for the parser bolt (default to `1`). This can be overridden on the command line.
* `errorWriterParallelism` : The error writer bolt parallelism (default to `1`). This can be overridden on the command line.
* `errorWriterNumTasks` : The number of tasks for the error writer bolt (default to `1`). This can be overridden on the command line.
* `spoutConfig` : A map representing a custom spout config. This can be overridden on the command line.
* `securityProtocol` : The security protocol to use for reading from kafka.  This can be overridden on the command line and also specified in the spout config via the `security.protocol` key.  If both are specified, then they are merged and the CLI will take precedence.
* `stormConfig` : The storm config to use.  This can be overridden on the command line.  If both are specified, they are merged with CLI properties taking precedence.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/SensorParserConfigControllerIntegrationTest.java
+++ b/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/SensorParserConfigControllerIntegrationTest.java
@@ -19,6 +19,7 @@ package org.apache.metron.rest.controller;
 
 import org.adrianwalker.multilinestring.Multiline;
 import org.apache.commons.io.FileUtils;
+import org.apache.metron.common.configuration.SensorParserConfig;
 import org.apache.metron.rest.MetronRestConstants;
 import org.apache.metron.rest.service.GrokService;
 import org.apache.metron.rest.service.SensorParserConfigService;
@@ -37,6 +38,7 @@ import org.springframework.web.context.WebApplicationContext;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Method;
 
 import static org.apache.metron.rest.MetronRestConstants.TEST_PROFILE;
 import static org.hamcrest.Matchers.hasSize;
@@ -196,11 +198,17 @@ public class SensorParserConfigControllerIntegrationTest {
     cleanFileSystem();
     this.sensorParserConfigService.delete("broTest");
     this.sensorParserConfigService.delete("squidTest");
-
+    Method[] method = SensorParserConfig.class.getMethods();
+    int numFields = 0;
+    for(Method m : method) {
+      if(m.getName().startsWith("set")) {
+        numFields++;
+      }
+    }
     this.mockMvc.perform(post(sensorParserConfigUrl).with(httpBasic(user, password)).with(csrf()).contentType(MediaType.parseMediaType("application/json;charset=UTF-8")).content(squidJson))
             .andExpect(status().isCreated())
             .andExpect(content().contentType(MediaType.parseMediaType("application/json;charset=UTF-8")))
-            .andExpect(jsonPath("$.*", hasSize(10)))
+            .andExpect(jsonPath("$.*", hasSize(numFields)))
             .andExpect(jsonPath("$.parserClassName").value("org.apache.metron.parsers.GrokParser"))
             .andExpect(jsonPath("$.sensorTopic").value("squidTest"))
             .andExpect(jsonPath("$.parserConfig.grokPath").value("target/patterns/squidTest"))
@@ -215,7 +223,7 @@ public class SensorParserConfigControllerIntegrationTest {
     this.mockMvc.perform(get(sensorParserConfigUrl + "/squidTest").with(httpBasic(user,password)))
             .andExpect(status().isOk())
             .andExpect(content().contentType(MediaType.parseMediaType("application/json;charset=UTF-8")))
-            .andExpect(jsonPath("$.*", hasSize(10)))
+            .andExpect(jsonPath("$.*", hasSize(numFields)))
             .andExpect(jsonPath("$.parserClassName").value("org.apache.metron.parsers.GrokParser"))
             .andExpect(jsonPath("$.sensorTopic").value("squidTest"))
             .andExpect(jsonPath("$.parserConfig.grokPath").value("target/patterns/squidTest"))
@@ -244,7 +252,7 @@ public class SensorParserConfigControllerIntegrationTest {
     this.mockMvc.perform(post(sensorParserConfigUrl).with(httpBasic(user, password)).with(csrf()).contentType(MediaType.parseMediaType("application/json;charset=UTF-8")).content(broJson))
             .andExpect(status().isCreated())
             .andExpect(content().contentType(MediaType.parseMediaType("application/json;charset=UTF-8")))
-            .andExpect(jsonPath("$.*", hasSize(10)))
+            .andExpect(jsonPath("$.*", hasSize(numFields)))
             .andExpect(jsonPath("$.parserClassName").value("org.apache.metron.parsers.bro.BasicBroParser"))
             .andExpect(jsonPath("$.sensorTopic").value("broTest"))
             .andExpect(jsonPath("$.readMetadata").value("true"))
@@ -254,7 +262,7 @@ public class SensorParserConfigControllerIntegrationTest {
     this.mockMvc.perform(post(sensorParserConfigUrl).with(httpBasic(user, password)).with(csrf()).contentType(MediaType.parseMediaType("application/json;charset=UTF-8")).content(broJson))
             .andExpect(status().isOk())
             .andExpect(content().contentType(MediaType.parseMediaType("application/json;charset=UTF-8")))
-            .andExpect(jsonPath("$.*", hasSize(10)))
+            .andExpect(jsonPath("$.*", hasSize(numFields)))
             .andExpect(jsonPath("$.parserClassName").value("org.apache.metron.parsers.bro.BasicBroParser"))
             .andExpect(jsonPath("$.sensorTopic").value("broTest"))
             .andExpect(jsonPath("$.readMetadata").value("true"))
--- a/metron-platform/metron-management/src/test/java/org/apache/metron/management/ConfigurationFunctionsTest.java
+++ b/metron-platform/metron-management/src/test/java/org/apache/metron/management/ConfigurationFunctionsTest.java
@@ -81,7 +81,15 @@ public class ConfigurationFunctionsTest {
       "parserConfig" : { },
       "fieldTransformations" : [ ],
       "readMetadata":false,
-      "mergeMetadata":false
+      "mergeMetadata":false,
+      "parserParallelism" : 1,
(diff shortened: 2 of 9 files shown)


ottobackwards:
#post-777

cestella:

# Preliminaries

Set an environment variable to indicate `METRON_HOME`:
* `export METRON_HOME=/usr/metron/0.4.1` 

# Deploy the dummy parser
* Edit `$METRON_HOME/config/zookeeper/parsers/dummy.json`:
```
{
  "sensorTopic":"dummy"
}
```
* Create the dummy kafka topic:
  `/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper node1:2181 --create --topic dummy --partitions 1 --replication-factor 1`
* Persist config changes: `$METRON_HOME/bin/zk_load_configs.sh -m PUSH -i $METRON_HOME/config/zookeeper -z node1:2181`
* Start via `$METRON_HOME/bin/start_parser_topology.sh -k node1:6667 -z node1:2181 -s dummy`

# Send dummy data through
* Edit `~/msg.json` with the following content:
```
{ "guid" : "guid0", "sensor.type" : "dummy", "timestamp" : 100 }
```
* Send `msg.json` through to kafka via `cat ~/msg.json | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic dummy`

## Test Case: Update via patch
* Patch the message in ES and create a new field 'project' by executing
  the following:
```
curl -u user:password -X PATCH --header 'Content-Type: application/json' --header 'Accept: */*' -d '{
  "uuid" : "guid0",
    "sensorType" : "dummy",
    "patch" : [
    {
      "op": "add"
        , "path": "/project"
        , "value": "metron"
    }
  ]
```
* Validate that the message has a field 'project':
```
curl -XPOST 'http://localhost:9200/dummy*/_search?pretty' -d '
{
  "_source" : [ "project" ]
}
'
```

## Test Case: Update via replace 
* Replace the message in ES and create a couple of modifications:
  * new field `new_field` == "brand new"
  * modified `timestamp` == 7
Execute the following:
```
curl -u user:password -X POST --header 'Content-Type: application/json' --header 'Accept: */*' -d '{
     "uuid" : "guid0",
     "sensorType" : "dummy",
     "replacement" : {
       "source:type": "dummy",
       "guid" : "guid0",
       "new_field" : "brand new",
       "timestamp" : 7
      }
   }' 'http://node1:8082/api/v1/update/replace'
```
* Validate that the message has a field 'project':
```
curl -XPOST 'http://localhost:9200/dummy*/_search?pretty' -d '
{
  "_source" : [ "new_field", "timestamp" ]
}
'
```
## Test Case: Validate that the changes persist to HBase
* Execute `echo "scan 'metron_update'" | hbase shell` 
* Ensure that 
  * There is one row: `guid0`
  * There are 2 columns 
  * The values associated with the columns correspond to the two modifications done in the tests above

ottobackwards:
Does the UI not support this then?

ottobackwards:
Or Rest?

cestella:
The UI and REST do allow you to override (they construct CLI params for starting parsers), but there is no way to save these params between starts.  This would allow people to specify the configs in zookeeper so they would persist between parser starts.

ottobackwards:
So we need a ui jira for this?

cestella:
Yeah, we would need a UI JIRA to have people set this for the sensors.  It provides some value by itself and it should not be a regression, but in order to change it in the UI, it'd need a hook.

cestella:
Also, to my knowledge, the other value it provides is being able to set these configs in such a way that ambari starting parsers actually is configurable (just not configurable directly from the ambari config).  Essentially this allows users to specify their own defaults rather than take the defaults in code.

ottobackwards:
We will need to integrate this in with 1136, maybe we should make it 'part' of it.

ottobackwards:
I need to think of how this need to integrate with the archetype and extension installer

cestella:
So, correct me if I'm wrong, but it should be independent of the archetype and extensions.  They define the operation of the parser, this would be entirely specifying the configs for the storm topology.  It should interact with the archetype and extensions the same way that the CLI options interact with it.

ottobackwards:
Maybe I'm miss understanding.  Does this introduce per parser configuration?

cestella:
Per parser instance config.  It's not defaulting config for all instances of a parser (i.e. not defaults for all Grok parsers, but you can specify defaults for the squid parser in the parser config in zookeeper).

cestella:
To my knowledge, even in a post-777 world, you're still specifying this via the CLI (which is the only way to specify it now in Metron).

ottobackwards:
maybe a little better explanation of 'how' we are doing this will help unfog me?

cestella:
Ok, I think maybe a better explanation is in order then.  Currently, when you start a topology, you can specify certain properties of the topology via the CLI.  These are things like the spout parallelism, number of executors, etc.  Originally these were added to the CLI because they were the things that you needed to adjust to scale a given parser.  What this PR is about is giving you the alternative to specify these topological or structural properties in the parser config in zookeeper.  This enables the admin to specify a config setup for a given instance of a parser in zookeeper without having to specify the CLI options.  

Now, it bears mentioning that this is also backwards compatible.  For these configuration options mentioned in this PR's description, the order of precedence is:
* CLI
* Zookeeper
which is to say that properties specified in the CLI will be honored first.

In a post-777 world, you would still have these CLI parameters for the basic topological properties (likely) and the config is still stored in zookeeper, so there really should be no conflict between this and what's going on in 777, I think.

Now, as to the 'how', well, when starting the parser, in the CLI code, we pull the zookeeper config and figure out the parameters to `ParserTopologyBuilder.build(...)` which gets invoked to construct the parser topology, like it did prior to this PR.

mmiklavc:
Yeah, we definitely need this.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/SensorParserConfigControllerIntegrationTest.java
```diff
@@ -196,11 +198,17 @@ public void test() throws Exception {
     cleanFileSystem();
     this.sensorParserConfigService.delete("broTest");
     this.sensorParserConfigService.delete("squidTest");
-
+    Method[] method = SensorParserConfig.class.getMethods();
+    int numFields = 0;
+    for(Method m : method) {
+      if(m.getName().startsWith("set")) {
```
ottobackwards: we should have a reflection utils method for getting these maybe.  Setters would also have to be checked for having one parameter and void return I think

ottobackwards:
I am +1 on this.  Though I think it should be documented in the tuning guide as well at some point.

merrimanr:
I tested this in full dev and everything worked as advertised.  The only major issue I see is that the num workers property is missing from SensorParserConfig.  I feel like this is one of the most important settings for a topology.  You could set it with "topology.workers" in the "stormConfig" property but it's not obvious and I think this setting should be a first class citizen like the others.

For me, the use of the ValueSupplier functional interface made it more difficult to understand when reviewing.  The code seemed more verbose and complex and I wasn't able to work out the benefit of using it in this case.  It's is a very minor style issue and is just my (probably worthless) opinion.  I would not hold up the PR over something like this, just some honest feedback.

Awesome work.  If we can promote num workers to a getter method in SensorParserConfig then I'm confident giving a +1.

cestella:
@merrimanr Thanks for mentioning num workers.  I just missed it when going through; I intended to add all relevant things to override, so I'm going to add num workers and num ackers too.  I'll add them now.  

The intent of the use of suppliers here is to keep the logic for parameter overrides in the CLI without needing the CLI to necessarily have to know how to pull in the sensor parser configs (we currently keep that functionality in the ParserTopology class.  I believe this separation of concerns has some value, so I wanted to provide a way where we could keep the logic in the CLI without having to interact with zookeeper and move logic around, thus moving from "here's the value" to "here's a way to get the value" type semantics.  I could see how that might be confusing since it's creating some indirection where there was no indirection before.  Perhaps I can add some comments to that effect justifying the reasons and it'll be clearer?

cestella:
Alright, I've added both a comment about the supplier abstraction and support for num workers and num ackers.

metron-platform/metron-management/src/test/java/org/apache/metron/management/ConfigurationFunctionsTest.java
```diff
@@ -81,7 +81,15 @@ private static void pushConfigs(String inputPath) throws Exception {
       "parserConfig" : { },
       "fieldTransformations" : [ ],
       "readMetadata":false,
-      "mergeMetadata":false
+      "mergeMetadata":false,
```
merrimanr: should numWorkers and numAckers be in this list?

metron-platform/metron-parsers/README.md
```diff
@@ -103,6 +103,15 @@ then it is assumed to be a regex and will match any topic matching the pattern (
 * `mergeMetadata` : Boolean indicating whether to merge metadata with the message or not (`false` by default).  See below for a discussion about metadata.
 * `parserConfig` : A JSON Map representing the parser implementation specific configuration.
 * `fieldTransformations` : An array of complex objects representing the transformations to be done on the message generated from the parser before writing out to the kafka topic.
+* `spoutParallelism` : The kafka spout parallelism (default to `1`).  This can be overridden on the command line.
```
merrimanr: should numWorkers and numAckers be in this list?

merrimanr:
Thanks for the writeup on ValueSupplier.  I think you went above and beyond documenting it and I appreciate that.  it makes sense now that I have my head around it :)

I noticed the numWorkers and numAckers were missing in a couple places.  The test is probably not necessary but I think we want it in the README.

metron-platform/metron-parsers/README.md
cestella: Yep, fixed.

cestella:
Ok, added the workers and ackers in the README.  Take a look and let me know if you see anything else.

merrimanr:
Looks good +1

PR #718, opened 2017-08-25
METRON-1135: Make Stellar field transformations handle nulls properly

cestella:
## Contributor Comments
As it stands, stellar field transformations disallow users to set variables to null (and thus remove the field). This should be corrected. For instance,
```
"fieldTransformations" : [
        { "transformation" : "STELLAR"
        ,"output" : [ "new_field", "new_field2"]
        ,"config" : {
          "new_field" : "old_field"
         ,"new_field2" : "old_field2"
         ,"old_field" : "null"
         ,"old_field2" : "null"
                    }
        }
]
```
should rename "old_field" to "new_field and "old_field2" to "old_field2" The current behavior is to keep all 4 fields and to not change old_field or old_field2 at all.

Manual testing:
* Create a sensor with the field transformation above.
* Send data through with `old_field` set to a value
* Ensure that `old_field` does not exist in any message in the index for your sensor and it is replaced with `new_field`.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-parsers/README.md
+++ b/metron-platform/metron-parsers/README.md
@@ -212,6 +212,49 @@ into `{ "protocol" : "TCP", "source.type" : "bro", ...}`
 * `STELLAR` : This transformation executes a set of transformations
   expressed as [Stellar Language](../metron-common) statements.
 
+### Assignment to `null`
+
+If, in your field transformation, you assign a field to `null`, the field will be removed.
+You can use this capability to rename variables.
+
+Consider this example:
+```
+ "fieldTransformations" : [
+         { "transformation" : "STELLAR"
+         ,"output" : [ "new_field", "old_field"]
+         ,"config" : {
+           "new_field" : "old_field"
+          ,"old_field" : "null"
+                     }
+         }
+ ]
+```
+This would set `new_field` to the value of `old_field` and remove `old_field`.
+
+### Warning: Transforming the same field twice
+
+Currently, the stellar expressions are expressed in the form of a map where the keys define
+the fields and the values define the Stellar expressions.  You order the expression evaluation
+in the `output` field.  A consequence of this choice to store the assignments as a map is that
+the same field cannot appear in the map as a key twice.
+
+For instance, the following will not function as expected:
+```
+ "fieldTransformations" : [
+         { "transformation" : "STELLAR"
+         ,"output" : [ "new_field"]
+         ,"config" : {
+           "new_field" : "TO_UPPER(field1)"
+          ,"new_field" : "TO_LOWER(new_field)"
+                     }
+         }
+ ]
+```
+
+In the above example, the last instance of `new_field` will win and `TO_LOWER(new_field)` will be evaluated
+while `TO_UPPER(field1)` will be skipped.
+
+### Example
 Consider the following sensor parser config to add three new fields to a
 message:
 * `utc_timestamp` : The unix epoch timestamp based on the `timestamp` field, a `dc` field which is the data center the message comes from and a `dc2tz` map mapping data centers to timezones
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/StellarTransformation.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/StellarTransformation.java
@@ -55,6 +55,11 @@ public class StellarTransformation implements FieldTransformation {
               intermediateVariables.put(oField, o);
             }
           }
+          else {
+            if(outputs.contains(oField)) {
+              ret.put(oField, null);
+            }
+          }
         }
         catch(Exception ex) {
           throw new IllegalStateException( "Unable to process transformation: " + transformObj.toString()
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/StellarTransformation.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/StellarTransformation.java
@@ -57,7 +57,13 @@ public class StellarTransformation implements FieldTransformation {
           }
           else {
             if(outputs.contains(oField)) {
-              ret.put(oField, null);
+              ret.put(oField, o);
+            }
+            if( o != null ) {
+              intermediateVariables.put(oField, o);
+            } else {
+              // remove here, in case there are other statements
+              intermediateVariables.remove(oField);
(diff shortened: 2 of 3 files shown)


ottobackwards:
Would it not be better to have different transformers to specify the behavior here?  What is here is correct, but what if the old behavior was fine?  What if they don't want the fields removed?

cestella:
So, the problem with the old way was that it's confusing and just wrong. Perhaps an example would help here.  For the message
```
{
  "title" : "otto is great"
}
```
If I wanted to rename title to `better_title`, I'd expect the following to work:
```
{ 
  "transformation" : "STELLAR"
 ,"output" : [ "better_title", "title"]
 ,"config" : {
          "better_title" : "title"
         ,"title" : "null"
                    }
}
```
At the end of this, I'd expect either `title` to be gone *or* at the very least, for it to be set to `null` as that is what I directed it to do, but in actuality, it'll still be `otto is great`.  In fact, the output would be:
```
{
  "title" : "otto is great",
  "better_title" : "otto is great"
}
```

This becomes even MORE insidious if we start doing things like returning null from Stellar functions.

ottobackwards:
if(outputs.contains(oField)) {
              ret.put(oField, o);
            }
            if( o != null ) {
              intermediateVariables.put(oField, o);
            }

cestella:
@ottobackwards good suggestion.  I've adopted it.

ottobackwards:
Ok, the question here is this.  I have 3 statements in the transform

{
  "title" : "casey is stellar"
}

{ 
  "transformation" : "STELLAR"
 ,"output" : [ "better_title", "title"]
 ,"config" : {
          "better_title" : "title"
         ,"title" : "SOME_UDF_THAT_CAN_RET_NULL()"
         ,"title" : "if title == null then GET_RANDOM_TITLE()
                    }
}

If the second function results in the removal of title... what do we do?  What was the intent of that null ( from a UDF maybe ), we don't know if it was to remove or not.

cestella:
Ok, I think I see the point of confusion.  Here's the issue; in every other place where we use stellar (i.e. enrichments) assignment to null implies a variable removal.  In field transformations, assignment to null is noop.  I'm just trying to make them consistent.  We could make both noop, but then we'd remove the ability to have temporary variables in enrichment, which has proven very useful.  Also, if we keep the current functionality, the example above, I'd argue, yields confusing results.

That being said, I think this is probably the WRONG behavior in both cases.  What I'd propose, instead, is support within stellar to be able to modify the variables.  I'd propose:
* `REMOVE('field')` would remove the field
* `^=` instead of `:=` would create a temporary variable so you would not need to remember to remove.  e.g. `foo ^= TO_UPPER('casey')` would not be persisted to the message in either field transformations or enrichments.

I think the current functionality is a bit kludgey, but I think as a stopgap we should at least make them do the same thing and I think removal is the most sensible choice.

cestella:
Currently you cannot have two statements in stellar field transformations with the same variable, so 
```
{
"transformation" : "STELLAR"
,"output" : [ "better_title", "title"]
,"config" : {
"better_title" : "title"
,"title" : "SOME_UDF_THAT_CAN_RET_NULL()"
,"title" : "if title == null then GET_RANDOM_TITLE()
}
}
```
would evaluate only the last `title` expression.  It's a problem with our use of maps and why in enrichment we transitioned to lists of strings as an alternative 😢

cestella:
This goes to another deficiency, field transformations don't act like enrichment.  YOu have to indicate the output fields to order the statements of evaluation.  This is ridiculous; we should allow an array of stellar expressions which are evaluated and incrementally update the message, like we have in enrichments.

ottobackwards:
Should there be a documentation change for this, like a 

> note about field transformations and null
> 1. sorry
> 2. here is how it works

cestella:
@ottobackwards Definitely there should and I'll make it as part of this PR.

ottobackwards:
ok, I'm +1 pending the doc change.  Thanks for explaining.

cestella:
Ok, I updated the docs with a summary of what we talked about and some discussion of gotchas.  Hope that is sufficient; if not, let me know.

ottobackwards:
+1, thanks for the contribution!

metron-platform/metron-parsers/README.md
```diff
@@ -212,6 +212,49 @@ into `{ "protocol" : "TCP", "source.type" : "bro", ...}`
...
+This would set `new_field` to the value of `old_field` and remove `old_field`.
+
+### Warning: Transforming the same field twice
+
+Currently, the stellar expressions are expressed in the form of a map where the keys define
+the fields and the values define the Stellar expressions.  You order the expression evaluation
+in the `output` field.  A consequence of this choice to store the assignments as a map is that
+the same field cannot appear in the map as a key twice.
+
+For instance, the following will not function as expected:
```
justinleet: I think this should be reworded. It's not that it doesn't function as expected; that would imply a bug or an improvement that needs to happen.

justinleet: Sigh, I don't read well, it sounds like this actually doesn't function as expected based on other expectations, and this is fine until we fix it.

justinleet:
Forgive my lack of understanding (possibly remembering?) of some of the underlying Stellar assumptions, but it seems like we mix up the meaning of actual null and String null.  It may not be entirely reasonable, but I would expect "null" to be an actual string null.

For example:
```
          "better_title" : "title"
         ,"title" : "null"
```
There's no formatting or any other distinction between "title" and "null" that would imply null is treated as a special case in string handling.

JSON has an actual null value that's not string null, and it seems like if we're defining transforms in JSON that's how it should be handled, e.g.
```
          "better_title" : "title"
         ,"title" : null
```

ottobackwards:
@justinleet, the value side of the json is a string, that is a Stellar statement.  So you are not passing in a null statement, you are passing in a Stellar statement that evaluates to null.  In this case via the null keyword.

If you wanted to pass a string that said null, you would have to have:
```
,"title" : "'null'"
```

ottobackwards:
```

[Stellar]>>> foo := null
[Stellar]>>> foo
[Stellar]>>> foo := 'null'
[Stellar]>>> foo
null
[Stellar]>>>
```

justinleet:
@ottobackwards Thanks for the refresher on that.  I'm +1

PR #721, opened 2017-08-29
METRON-1138 Improve Error Message with Bad Profile Expression

nickwallen:
I have found that it can be very natural for a new Profiler user to attempt to reference a field defined in a telemetry message, in a Profile's `groupBy` expression.  Unfortunately, this does not work as the 'groupBy' expression runs during a flush where there is no telemetry message available.  Only specific fields including duration, result, profile, start, end, entity and any variables defined in the profile itself are available.  This behavior is documented in the README, but should be more accessible when an error occurs.

For example, this seems like a Profile that should work.  But unfortunately there is no `timestamp` variable defined when the `groupBy` expression is executed.
```
{
      "profile": "calender-effects",
      "onlyif":  "exists(ip_src_addr) and exists(timestamp)",
      "foreach": "ip_src_addr",
      "init":    { "count": 0 },
      "result":  "count",
      "groupBy": ["DAY_OF_WEEK(timestamp)"]
    }
```

When this expression fails the error message looks-like the following, which is not exactly enlightening.

```
org.apache.metron.stellar.dsl.ParseException: Bad 'groupBy' expression: Unexpected type: 
expected=Object, 
actual=null, 
expression=DAY_OF_WEEK(timestamp), 
profile=calender-effects, 
entity=10.0.0.1 
```

With this change, additional information is provided when any Profile expression fails. 
* The variables that are in-scope are logged.
* The exact expression that failed is logged.  Previously, for the 'init' and 'update' expressions it wasn't clear which expression failed, if the user defined multiple.

This should make it much easier to debug for a user.  The same error looks-like the following with the code change.  Notice the new `variables-available` which defines the variables that were in-scope when the expression was executed.

```
ERROR DefaultProfileBuilder:189 - ...
org.apache.metron.stellar.dsl.ParseException: Bad 'groupBy' expression: 
error='Unexpected type: expected=Object, actual=null, expression=DAY_OF_WEEK(timestamp)', expr='DAY_OF_WEEK(timestamp)', 
profile='calendar-effects', 
entity='10.0.0.1', 
variables-available='[duration, result, profile, start, x, end, entity]'
```

### DIFF
--- a/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
+++ b/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultProfileBuilder.java
@@ -27,9 +27,11 @@ import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.apache.commons.collections4.ListUtils;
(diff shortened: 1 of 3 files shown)


cestella:
+1 by inspection, this is really great.

PR #727, opened 2017-08-31
METRON-1146: Add ability to parse JSON string into JSONObject for stellar

anandsubbu:
## Contributor Comments
This pull request allows the ability to parse a given JSON string using Jackson and return a JSON object. Here is an example:
`MAP_GET( 'bar', PARSE_JSON_STRING(  '{ "foo" : 1, "bar" : 2}' )`  would yield `2`.

Another example would be:
`MAP_GET( 'foo1', GET_FIRST( PARSE_JSON_STRING('[{ "foo1":"a", "bar1":"b"}, { "foo2":"c", "bar2":"d"}]'))` ==> would output `a`

## Pull Request Checklist

### For all changes:

### For code changes:
- [] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/StringFunctionsTest.java
```diff
@@ -449,4 +450,68 @@ public void testCountMatches() throws Exception {
     Assert.assertTrue(thrown);
 
   }
+
```
ottobackwards: Would it be possible for you to use the 
```java
@multiline
```
Annotation for defining your json strings instead of inline concatenation and escaping?  It makes it more readable.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
```diff
@@ -450,4 +454,40 @@ public Object apply(List<Object> strings) {
...
 
+  @Stellar( name = "PARSE_JSON_STRING"
+          , description = "Returns a JSON object for the specified JSON string"
+          , params = {
+            "str - the JSON String to convert, may be null"
+          }
+          , returns = "an Object containing the parsed JSON string"
+  )
+  public static class ParseJsonString extends BaseStellarFunction {
+
```
ottobackwards: So we are just returning whatever we get back from Jackson to the caller.  Meaning that they have to understand the content of the json in order to write the expression correctly.  This needs to be documented if it stays like this.

I am not sure I like that.  Can you give the use cases or circumstances that where you imagine this being used?  What are the bounds?

cestella: So, the usecase is having an enrichment which stores a JSON blob in HBase (or some other store) that you want to marshall into a Java object to interact with via stellar.  Some people would prefer to have their data in a readable format rather than serialized via kryo (like we do in the profiler) or restricted to a certain format (like we do in enrichments).  This opens up those use-cases.

That JSON blob may be a map or a list or a primitive; the idea is to let jackson handle the parsing of it, so the bounds are what JSON can represent.  The knowledge of what is returned would be necessary for the caller, though, as you say.

anandsubbu: Updated README

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/StringFunctionsTest.java
anandsubbu: Thanks @ottobackwards . Have now used multiline strings.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
ottobackwards: What do you guys think of having PARSE_TO_MAP, PARSE_TO_XXXX functions to have **some** type safety?

cestella: I don't mind having them as convenience functions in addition to the generic one.

ottobackwards: I think it will make more sense to authors, and have less of a surprise factor.

- you need to understand your data
- use the right function
- get good error checking and verification

anandsubbu:
@ottobackwards and @cestella please have a look at the latest commit.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
```diff
@@ -550,4 +548,90 @@ else if(var.length() == 0) {
...
+        throw new IllegalArgumentException("[JSON_TO_MAP] incorrect arguments. Usage: JSON_TO_MAP <JSON String>");
+      }
+      String var = (strings.get(0) == null) ? null : (String) strings.get(0);
+      if (var == null) {
+        return null;
+      } else if (var.length() == 0) {
+        return var;
+      } else {
+        ObjectMapper objectMapper = new ObjectMapper();
+
```
ottobackwards: I have a question here.  If you just do the load, and the string is invalid, would you not get the same exception?  What is the difference.  If the string is always valid, and that is the most common case, then we are always doing 2* the processing.  That doesn't seem right in a high volume stream system

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
```diff
@@ -550,4 +548,90 @@ else if(var.length() == 0) {
...
+        try {
+          objectMapper.readTree((String) strings.get(0));
+        } catch (JsonProcessingException ex) {
+          throw new ParseException("Valid JSON string not supplied", ex);
+        } catch (IOException e) {
+          e.printStackTrace();
+        }
+
+        // Return parsed JSON Object as a HashMap
+        try {
```
ottobackwards: variables = JSONUtils.INSTANCE.load(String, new TypeReference<Map<String, Object>>() 

Use TypeReference to determine the type returned, not Object for this case

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
```diff
@@ -550,4 +548,90 @@ else if(var.length() == 0) {
...
+
+        // First parse and check if input is valid JSON string
+        try {
+          objectMapper.readTree((String) strings.get(0));
+        } catch (JsonProcessingException ex) {
+          throw new ParseException("Valid JSON string not supplied", ex);
+        } catch (IOException e) {
+          e.printStackTrace();
+        }
+
```
ottobackwards: I am not sure about using TypeReference here, Map<String,Object> is the natural type, I think it would be List<Object> ?

What type of list do you get when you just cast?

ottobackwards: Ok, I see from your tests it is list object:

SO variables = JSONUtils.INSTANCE.load(new FileInputStream(variablesFile.get()), new TypeReference<List<Object>() {

nickwallen:
I have not reviewed the code, but I definitely like the functionality @anandsubbu .  What do you think about simplifying the name to `TO_JSON` or something like that?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
anandsubbu: Yup, that makes sense @ottobackwards . Cleaned up now.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
```diff
@@ -550,4 +548,90 @@ else if(var.length() == 0) {
...
+          , returns = "a MAP object containing the parsed JSON string"
+  )
+  public static class JsonToMap extends BaseStellarFunction {
+
+    @Override
+    public Object apply(List<Object> strings) {
+
+      if (strings == null || strings.size() == 0) {
+        throw new IllegalArgumentException("[JSON_TO_MAP] incorrect arguments. Usage: JSON_TO_MAP <JSON String>");
+      }
```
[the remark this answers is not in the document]
anandsubbu: Added check.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
```diff
@@ -506,29 +506,27 @@ public Object apply(List<Object> strings) {
...
-  public static class ParseJsonString extends BaseStellarFunction {
+  public static class JsonParse extends BaseStellarFunction {
 
     @Override
     public Object apply(List<Object> strings) {
 
       if (strings == null || strings.size() == 0) {
-        throw new IllegalArgumentException("[PARSE_JSON_STRING] incorrect arguments. Usage: PARSE_JSON_STRING <String>");
+        throw new IllegalArgumentException("[JSON_PARSE] incorrect arguments. Usage: JSON_PARSE <String>");
       }
```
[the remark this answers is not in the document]
anandsubbu: Added check

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
anandsubbu: Sure, done!

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/StringFunctions.java
anandsubbu: Sure, done!

anandsubbu:
Hi @nickwallen thank you. Sure, I see value in simplifying the naming. How does `TO_JSON_OBJECT`, `TO_JSON_MAP` and `TO_JSON_LIST` sound?

anandsubbu:
Thanks very much @ottobackwards ! 

@nickwallen , pushed latest commit with the functions renamed. Please have a look.

ottobackwards:
+1 by inspection

cestella:
+1 by inspection, great job!

PR #731, opened 2017-09-03
METRON-1136  Refactor for ParserExtensionConfig and Rest get all call

ottobackwards:
This a PR for some cleanup on the 1136 Feature Branch

- return a list from GET ( getAllParsers ) and not a map.
- fix names in Parser Extension Configs to be consistant with extension/extensions
- add a field just called identifier, to be used instead of assembly name
- differentiate identifier/assemblyname

## 

### DIFF
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/ParserExtensionController.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/ParserExtensionController.java
@@ -22,6 +22,7 @@ import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
+import java.util.List;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
 import org.apache.metron.common.configuration.extensions.ParserExtensionConfig;
@@ -51,7 +52,7 @@ public class ParserExtensionController {
   DeferredResult<ResponseEntity<Void>> install(@ApiParam(name = "extensionTgz", value = "Metron Parser Extension tar.gz", required = true) @RequestParam("extensionTgz") MultipartFile extensionTgz) throws RestException {
     DeferredResult<ResponseEntity<Void>> result = new DeferredResult<>();
 
-    String extensionName = extensionService.formatPackageName(extensionTgz.getOriginalFilename());
+    String extensionName = extensionService.formatExtensionIdentifier(extensionTgz.getOriginalFilename());
     try {
       if (extensionService.findOneParserExtension(extensionName) != null) {
         result.setResult(new ResponseEntity<Void>(HttpStatus.FORBIDDEN));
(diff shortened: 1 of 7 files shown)


metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/ParserExtensionController.java
```diff
@@ -90,8 +91,8 @@
   @ApiOperation(value = "Retrieves all ParserExtensionConfigs from Zookeeper")
   @ApiResponse(message = "Returns all ParserExtensionConfigs", code = 200)
   @RequestMapping(method = RequestMethod.GET)
-  ResponseEntity<Map<String, ParserExtensionConfig>> findAll() throws RestException {
-    return new ResponseEntity<Map<String, ParserExtensionConfig>>(extensionService.getAllParserExtensions(), HttpStatus.OK);
+  ResponseEntity<List<ParserExtensionConfig>> findAll() throws RestException {
```
nickwallen: What is the logic for changing this from a map to a list?

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/ExtensionServiceImpl.java
```diff
@@ -528,20 +527,26 @@ private void rollBackPatternsFromHdfs(InstallContext context) throws Exception{
...
     config.setExtensionAssemblyName(context.extensionPackageName.get());
     config.setDefaultParserConfigs(context.defaultParserConfigs.get());
-    config.setDefaultEnrichementConfigs(context.defaultEnrichmentConfigs.get());
+    config.setDefaultEnrichmentConfigs(context.defaultEnrichmentConfigs.get());
     config.setDefaultIndexingConfigs(context.defaultIndexingConfigs.get());
     if(context.defaultElasticSearchTemplates.isPresent()) {
       config.setDefaultElasticSearchTemplates(context.defaultElasticSearchTemplates.get());
     }
-    ConfigurationsUtils.writeParserExtensionConfigToZookeeper(context.extensionPackageName.get(),config.toJSON().getBytes(), client);
+    ConfigurationsUtils.writeParserExtensionConfigToZookeeper(context.extensionIdentifier.get(),config.toJSON().getBytes(), client);
```
nickwallen: What is the logic for changing package name to identifier?

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/ParserExtensionController.java
ottobackwards: I have another branch where I am doing work on extending the management ui to manage extensions.  I had a lot of trouble consuming the maps.  The map is the configurations with a field pulled out as the key, so making it a list made it work, similar to the getAll for parser configurations, and really didn't lose anything.

TL;DR
I am not good enough at angular to consume the maps correctly so I refactored.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/ExtensionServiceImpl.java
ottobackwards: While working with the config, the fact that I had picked a field to use as the identifier, but was also something else seemed confusing.  Making the identifier explicit, and also making that field what it should be ( the assemblyName/packageName ) without truncating it seems to me to make it more understandable.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/ParserExtensionController.java
nickwallen: Makes sense.  Its cleaner this way.  (Even if the real reason is that you aren't good enough with Angular.  Ha)

nickwallen:
+1 Thanks, Otto

anandsubbu:
Hi @ottobackwards , I guess you wanted to name the PR title as METRON-1136 instead of METRON-1146?

ottobackwards:
I just got done doing that!  Luckily the scripts that I use have you verify everything, and I caught the different message :)

PR #737, opened 2017-09-08
METRON-1161: Add ability to edit parser command line options in the management UI

merrimanr:
## Contributor Comments
This PR is a follow up to https://github.com/apache/metron/pull/717 and exposes the new Storm parser topology options in the Management UI.  This can be verified in full dev by navigating to the Management UI and editing any parser.  There should be a new "Storm Settings" pane in the primary panel that can be expanded and exposes input controls for all the new parser topology properties.  Editing these should successfully change the parser configs in zookeeper.

I also performed a slight refactor of the Raw JSON editor to make it easier to maintain as we add more parser config properties.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-interface/metron-config/src/app/sensors/sensor-parser-config-readonly/sensor-parser-config-readonly.component.spec.ts
```diff
@@ -301,7 +301,7 @@ describe('Component: SensorParserConfigReadonly', () => {
   }));
 
   it('should have metadata defined ', async(() => {
```
ottobackwards: Why is this '32'?  Maybe a comment for a maintainer down the road?

metron-interface/metron-config/src/app/sensors/sensor-parser-config/sensor-parser-config.component.ts
```diff
@@ -409,12 +403,17 @@ export class SensorParserConfigComponent implements OnInit {
...
     this.showRawJson = (pane ===  Pane.RAWJSON) ? visibilty : false;
     this.showThreatTriage = (pane ===  Pane.THREATTRIAGE) ? visibilty : false;
+    this.showStormSettings = (pane === Pane.STORMSETTINGS) ? visibilty : false;
   }
 
   onRawJsonChanged(): void {
     this.sensorFieldSchema.createFieldSchemaRows();
   }
 
+  onStormSettingsChanged(): void {
```
ottobackwards: Is this a debug artifact?

nickwallen:
@merrimanr This will be uber useful.  Here are my initial reactions focusing on usability.  I am totally open to knocking some of these out as separate PRs, should that be a better approach.

### (1) Double Save Required

A user is required to do a "double save" for the changes to hit Zookeeper.  When I click "Save" on the "Storm Settings" panel, it doesn't actually save anything yet.  I then have to click "Save" again on the parser edit panel.  Once I do this "double save", I can then see the settings take effect in Zookeeper.  

This is counter-intuitive to me.  A button that saves "Save" should save things.

### (2) Unsaved Changes Indicator

When the user changes a field in the "Storm Settings" panel, it would be nice to have an indication that there are unsaved changes.  This "unsaved change indicator" should stick around until we actually push the changes to Zookeeper.  I should not be able to exit the panel without a warning, if I have unsaved changes.  Right now, I can close the panel and lose all of my changes.  

In this case, I think this is important because the "Save" and "Cancel" buttons show up "below the fold" and are not visible unless the user scrolls down.  On my first pass, I found myself making changes on the panel, clicking the X in the top-right corner because this is the only transition that is visible.  I then found that my changes were not saved.

### (3) What Now?

Once the user saves changes on this panel, I think we need to provide some guidance to the user as to what to do next.  As a user I am left wondering; 
* What just happened?  
* How do I know that his action was successful?  
* What do I need to do now?

### (4) Warning about Persistence

We should have a warning that indicates under what conditions these settings will not persist.   Right now all we do is dump them in Zk.  So if Zk is restarted, we lose them, right?  Whatever does actually happen should be made clear to the user.

### (5) Reset to Defaults

It would be nice to have a "Reset to Defaults" button of some sort.  I wanted to go back to what I had initially.  I tried to do this by erasing the value in each field and then saving.  All this did was set each value to 0, which is not the "reset to defaults" that I expected.

### (6) Blank Fields

The `NUM_WORKERS` and `NUM_ACKERS` are blank, unlike all of the other configuration fields.

metron-interface/metron-config/src/app/sensors/sensor-parser-config-readonly/sensor-parser-config-readonly.component.spec.ts
merrimanr: Sure I will add a comment here.

metron-interface/metron-config/src/app/sensors/sensor-parser-config/sensor-parser-config.component.ts
merrimanr: Yes this was a mistake.  I think we should leave this function in the event we need it in the future but I will definitely remove this logging.

merrimanr:
Just pushed out a commit that addresses @ottobackwards comments and fixes the NUM_WORKERS and NUM_ACKERS issue.  I also added a warning that appears when a storm setting changes, telling the user that the topology needs to be restarted.  It may seem redundant that the same message appears for each field but we might want a different message in the future once we expose a way to rebalance using zookeeper values.  At that point this pattern would make sense and I don't think it necessarily looks too bad now.

@nickwallen I also started a discussion on the other points you brought up.  Hopefully whatever conclusion we come to there can be a follow-on JIRA.

merrimanr:
Changing numWorkers and numAckers default values from null to 1 had the unintended consequences of breaking the ParserTopologyCliTest.  This is because they can no longer be overriden by setting the topology.workers and topology.acker.executors properties in stormConfig.  I believe this is the correct behaviour because numWorkers and numAckers are top-level properties and should take precedence.  If this is not correct then I would argue that we shouldn't even have them.  Can you verify this @cestella?

cestella:
So, I do believe the default values for num workers and ackers should be taken from the storm config if unspecified by us.  Providing defaults when storm already has defaults for these properties seems wrong IMO.  I think that's what you are asking.

merrimanr:
I moved the default values to the storm config and changed getNumWorkers and getNumAckers to return the values in the storm config if numWorkers/numAckers is null.  Is this what you were thinking?

I had to set the defaults in SensorParserConfig because I could not find a way to get these defaults from the Storm Java API.  Is there a way and I just missed it?

cestella:
Wait, I'm confused, the current behavior is to default to the storm config.  Why did you pull that into the SensorParserConfig?

The issue with what you've done here is that when you save out the configs into zookeeper, you are serializing the object with the values set (including those defaults).  Now, if someone restarts the topology, those are the values set regardless of what is set in the storm config.

Can we please just solve this by some hover-over text indicating if the field is not set, then the storm defaults are used?  Maybe even indicating what the default currently is.  I'd have honestly made those values Optional if I could've (serialization issues caused that not to be the case).

I guess what I'm saying is that, no, I don't like this approach. :)

merrimanr:
Ok I reverted the SensorParserConfig class so that defaults are not provided for numWorkers and numAckers.  I also added help text to make it clear to the user what happens when these are not set.  Let me know how this looks.

cestella:
+1 by inspection.  This looks good to me.  Great job!

PR #690, opened 2017-08-11
METRON-1091 Package STELLAR shell as stand alone

ottobackwards:
~~This PR adds metron-deployment/packaging/archive, and the metron-stellar-shell module.~~
This PR adds a second archive product ( ${}-stand-alone.tar.gz ) to stellar-common.

This module packages a stellar shell environment that allows running stellar on a workstation that is not a dev/build machine, nor is a cluster node.

The idea is that a metron 'admin' may want to run stellar commands on his local machine.

Remote management ( config_get etc ) and profile would be useful, but have not been included.  I do not believe it is realistic for the expected user to open firewall ports or have the 'confs' setup to have these functions work as they would on a metron cluster node.

This use case would seem to require REST support of some kind.  Maybe a /metron-interface/metron-rest-stellar module.  Or maybe when extensions come to stellar we can just do it as an extension.

Also a problem, is that there are management functions that bring in hadoop dependencies mixed in with other functions like SHELL_EDIT() that may be useful on their own.

## Testing
- package
~~- take /metron-deployment/packaging/archive/metron-stellar-shell/target/metron-stellar-shell-0.4.1-archive.tar.gz and unpack it in some directory~~
 take metron-stellar/stellar-common/target/stellar-common-0.4.1-stand-alone.tar.gz and unpack it in some directory
- run it
- `stellar it up

## Questions for reviewers
- Where should the doc for this go? 
- What NOTICE or other Files should be in the deployed env?
- Do we want to post this as a separate download?
- Above questions wrt port opening and configuration

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
[...]

PR #740, opened 2017-09-08
METRON-1167 Define Session Specific Global Configuration Values in the REPL

nickwallen:
Many Stellar functions accept configuration values from the Global configuration.  When using the REPL you can load the global configuration values by launching the REPL with the -z option, which loads the global configuration stored in Zookeeper.

The only way to modify the global configuration within the REPL currently is to do the following steps.
1. Retrieve the global configuration with `conf := CONFIG_GET("global")`
2. Alter that global configuration by modifying the JSON contained in the `conf` variable.
3. Push the new global configuration using `CONFIG_PUT("global", conf)`
4. Close and then reopen the REPL.  Without restarting the REPL the new global configuration is not loaded.

I want a way to do this directly in the REPL, without restarting it, and also in a way that does not modify the persisted global configuration in Zookeeper.  I may be monkeying with something in the REPL, but I don't want to break one of my live Metron topologies.

### Test Drive

1. Check-out this PR and build Metron. 
     ```
     mvn clean install -DskipTests -T1C
     ```

1. Then from the root of the Metron source directory, launch the REPL.
    ```
    $ mvn exec:java \
        -Dexec.mainClass="org.apache.metron.stellar.common.shell.StellarShell" \
        -pl metron-stellar/stellar-common/
    ```

1. Then `%define`, `%undefine`, and `%globals` to your heart's content.  There are no globals right now.
    ```
    [Stellar]>>> %globals
    {}
    ```
1.  A regular assignment (my first pass at this) no longer works as expected.
    ```
    [Stellar]>>> %define foo=bar
    [!] %define expected assignment expression
    ```
1. Define a simple global configuration value.  The result of the expression gets echoed just like a normal assignment expression.
    ```
    [Stellar]>>> %define foo := "bar"
    bar
    ```
1.  Define a more complex global configuration value.
    ```
    [Stellar]>>> %define baz := { "k1": "v1" }
    {k1=v1}
    ```
1. See how the globals changed.
    ```
    [Stellar]>>> %globals
    {foo=bar, baz={k1=v1}}
    ```
1. Undefine a global value.
    ```
    [Stellar]>>> %undefine foo
    [Stellar]>>> %globals
    {baz={k1=v1}}
    [Stellar]>>> %undefine baz
    [Stellar]>>> %globals
    {}
    ```
1.  Also works for my simple use case.
    ```
    [Stellar]>>> %define bootstrap.servers := "node1:6667"
    node1:6667
    [Stellar]>>> %globals
    {bootstrap.servers=node1:6667}
    ```
**Edit:** Altered the 'Test Drive' instructions based on syntax change.

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/StellarExecutor.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/StellarExecutor.java
@@ -92,9 +92,10 @@ public class StellarExecutor {
   }
 
   /**
-   * prefix tree index of autocompletes
+   * Prefix tree index of auto-completes.
    */
   private PatriciaTrie<AutoCompleteType> autocompleteIndex;
+
   /**
    * The variables known by Stellar.
    */
@@ -118,7 +119,9 @@ public class StellarExecutor {
   private Console console;
 
   public enum OperationType {
-    DOC,MAGIC,NORMAL;
+    DOC
+    , MAGIC
+    , NORMAL
   }
 
   public interface AutoCompleteTransformation {
@@ -138,6 +141,7 @@ public class StellarExecutor {
     , VARIABLE((type, key) -> key )
     , TOKEN((type, key) -> key)
     ;
+
     AutoCompleteTransformation transform;
     AutoCompleteType(AutoCompleteTransformation transform) {
       this.transform = transform;
@@ -147,7 +151,6 @@ public class StellarExecutor {
     public String transform(OperationType type, String key) {
       return transform.transform(type, key);
     }
-
   }
 
   /**
@@ -178,18 +181,17 @@ public class StellarExecutor {
 
     // asynchronously update the index with function names found from a classpath scan.
     new Thread( () -> {
-        Iterable<StellarFunctionInfo> functions = functionResolver.getFunctionInfo();
-        indexLock.writeLock().lock();
-        try {
-          for(StellarFunctionInfo info: functions) {
-            String functionName = info.getName();
-            autocompleteIndex.put(functionName, AutoCompleteType.FUNCTION);
-          }
+      Iterable<StellarFunctionInfo> functions = functionResolver.getFunctionInfo();
+      indexLock.writeLock().lock();
+      try {
+        for(StellarFunctionInfo info: functions) {
+          String functionName = info.getName();
+          autocompleteIndex.put(functionName, AutoCompleteType.FUNCTION);
         }
-          finally {
-            System.out.println("Functions loaded, you may refer to functions now...");
-            indexLock.writeLock().unlock();
-          }
+      } finally {
+        System.out.println("Functions loaded, you may refer to functions now...");
+        indexLock.writeLock().unlock();
+      }
     }).start();
   }
 
@@ -203,6 +205,9 @@ public class StellarExecutor {
     index.put("quit", AutoCompleteType.TOKEN);
     index.put(StellarShell.MAGIC_FUNCTIONS, AutoCompleteType.FUNCTION);
     index.put(StellarShell.MAGIC_VARS, AutoCompleteType.FUNCTION);
+    index.put(StellarShell.MAGIC_GLOBALS, AutoCompleteType.FUNCTION);
+    index.put(StellarShell.MAGIC_DEFINE, AutoCompleteType.FUNCTION);
+    index.put(StellarShell.MAGIC_UNDEFINE, AutoCompleteType.FUNCTION);
     return new PatriciaTrie<>(index);
   }
 
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/StellarShell.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/StellarShell.java
@@ -30,10 +30,9 @@ import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.PosixParser;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.metron.stellar.dsl.Context;
-import org.apache.metron.stellar.dsl.StellarFunctionInfo;
 import org.apache.metron.stellar.common.StellarAssignment;
 import org.apache.metron.stellar.common.utils.JSONUtils;
+import org.apache.metron.stellar.dsl.StellarFunctionInfo;
 import org.jboss.aesh.complete.CompleteOperation;
 import org.jboss.aesh.complete.Completion;
 import org.jboss.aesh.console.AeshConsoleCallback;
@@ -51,12 +50,17 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
+
+import static org.apache.metron.stellar.dsl.Context.Capabilities.GLOBAL_CONFIG;
 
 /**
  * A REPL environment for Stellar.
@@ -90,6 +94,9 @@ public class StellarShell extends AeshConsoleCallback implements Completion {
   public static final String MAGIC_VARS = MAGIC_PREFIX + "vars";
   public static final String DOC_PREFIX = "?";
   public static final String STELLAR_PROPERTIES_FILENAME = "stellar.properties";
+  public static final String MAGIC_GLOBALS = MAGIC_PREFIX + "globals";
+  public static final String MAGIC_DEFINE = MAGIC_PREFIX + "define";
+  public static final String MAGIC_UNDEFINE = MAGIC_PREFIX + "undefine";
 
   private StellarExecutor executor;
 
@@ -241,7 +248,7 @@ public class StellarShell extends AeshConsoleCallback implements Completion {
     // welcome message and print globals
     writeLine(WELCOME);
     executor.getContext()
-            .getCapability(Context.Capabilities.GLOBAL_CONFIG, false)
+            .getCapability(GLOBAL_CONFIG, false)
             .ifPresent(conf -> writeLine(conf.toString()));
 
     console.start();
@@ -287,19 +294,19 @@ public class StellarShell extends AeshConsoleCallback implements Completion {
    * Executes a magic expression.
    * @param rawExpression The expression to execute.
    */
-  private void handleMagic( String rawExpression) {
-    String[] expression = rawExpression.trim().split(" ");
+  private void handleMagic(String rawExpression) {
+    String[] expression = rawExpression.trim().split("\\s+");
 
     String command = expression[0];
-    if(MAGIC_FUNCTIONS.equals(command)) {
+    if (MAGIC_FUNCTIONS.equals(command)) {
 
       // if '%functions FOO' then show only functions that contain 'FOO'
       Predicate<String> nameFilter = (name -> true);
-      if(expression.length > 1) {
+      if (expression.length > 1) {
         nameFilter = (name -> name.contains(expression[1]));
       }
 
-      // list available functions
+      // '%functions' -> list all functions in scope
       String functions = StreamSupport
               .stream(executor.getFunctionResolver().getFunctionInfo().spliterator(), false)
               .map(info -> String.format("%s", info.getName()))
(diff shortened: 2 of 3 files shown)


metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/StellarShell.java
```diff
@@ -308,17 +315,68 @@ private void handleMagic( String rawExpression) {
...
-      // list all variables
+      // '%vars' -> list all variables in scope
       executor.getVariables()
-              .forEach((k,v) -> writeLine(String.format("%s = %s", k, v)));
+              .forEach((k, v) -> writeLine(String.format("%s = %s", k, v)));
+
+    } else if (MAGIC_GLOBALS.equals(command)) {
+
+      // '%globals' -> list all globals in scope
+      Map<String, Object> globals = Collections.emptyMap();
```
ottobackwards: I don't understand why we check for capabilities here, but not when setting.  It looks like I can define as much as I want, but only see if the capability is there?  Why do we only guard here?

ottobackwards: Why wouldn't we getOrCreate here as well?

nickwallen: It is just not strictly needed since if I start out with an empty map and there is no global config in Zk, then we are just left with the empty map, so we're good.  So functionally getOrCreate is not needed here.

That being said, I can see how it would bring a nice sense of symmetry to just use getOrCreate in this case also.  Let me try that out real quick.

nickwallen: Yep, it is prettier this way.  Thanks  @ottobackwards

ottobackwards: When would you not get an empty map?  It looks like if you are not using ZK you always get an empty map.  I must be missing something

ottobackwards: Never mind, calling getOrCreate will put the capability in. I see. Sorry

nickwallen: Yep, exactly.  No worries.  I'm glad you're looking at it closely.

ottobackwards: Nick. If I understand correctly, if there is no ZK, then this turns on the GLOBAL_CONFIG flag, and sets whatever properties.

Is there any risk that other functions that check for GLOBAL_CONFIG can see the flag, but assuming the full config is present, when in this case only what the user has set is present?

cestella:
How would we, say, specify a global config value as a map (e.g. foo={'bar' : 'blah'})?  global config is also exposed as variables stellar in the parser and enrichments, so this could very well be a use-case.  complex global config is also used in the stellar config IIRC.

cestella:
I think I'd be ok without that capability if we had an ability to pass in a global config as a JSON blob, like we do the initial set of variables.  What do you think?

nickwallen:
Yes, I think it would be useful to pass in a global config as a command line argument.  Maybe something like...
```
bin/stellar -g config/zookeeper/global.json
```

I think we could tackle that as a follow-on.

cestella:
so, what happens if we modify a field via `%define` from a stellar shell started with `-z` ?

nickwallen:
And in terms of the complex types I can see the use case definitely.  In terms of the REPL and the changes in this PR, a `%define` does the same as when GLOBAL_CONFIG gets created with a `-z` for complex types.  In either case, the complex type gets loaded as a string. 

For example, I launched Full Dev and created a GLOBAL_CONFIG that looks-like this.
```
{
    ...
    "foo": "{'bar' : 'blah'}"
}
```
I then launched the REPL in a debug session with a -z.  The value for the key "foo" is a string.

nickwallen:
Oops.  Maybe I am wrong on the complex type thing.  I just noticed I had quotes around the {}.  Let me try that again, maybe I'm wrong.

cestella:
yeah, we can definitely have complex types in the global config; that gets serialized as a `Map<String, Object>` and jackson will happily load a map of maps or lists.  The idea originally was that the global config could, for instance, hold static reference data that was super small, which could include maps.

nickwallen:
Yep, you are right.  How could I doubt you?  For your complex type example a %define would load the value as a string, and the -z would load it as a map.  So the behavior is different. But the reason it gets loaded as a map is because it is a map in JSON, not anything to do with Stellar itself.  

So I am a little corn-fused about how to match the same functionality with %define.  I'd have to treat it as JSON?  The other thought I had would be to treat it as a Stellar expression and just execute it, but that would actually be different than `-z`.

cestella:
So, one way to do this is to allow users to refer to stellar fields:
```
my_complex_field := { 'blah' : 7 }
%define nicks_new_global_field=&my_complex_field
```

Another option to support this is to just insist that `%define` is a string and you have a `replace_global` that takes a field as an argument and the field contains the new global config (in JSON form) and replaces the global config:
```
new_global_config := SHELL_EDIT()
%replace_global new_global_config
```

cestella:
you could also make `%define nicks_new_global_field := my_complex_field`  That makes somewhat more sense since stellar assignment is `:=` and property assignment is `=`

nickwallen:
Those options seem to add a bit more complexity than I'd like.  How about we just restrict %define to strings only for now.  If you need to use complex types, then you can still do the "CONFIG_GET, CONFIG_PUT, restart REPL" dance that I described in the PR description.

nickwallen:
"a bit more complexity than I'd like" may just need to be read as "Nick is lazy", but I'll let you be the judge

ottobackwards:
Can't we turn everything passed in into json and use the json patch stuff to "build out' the configuration?  Why not have the globals as a map<string,object> all the time?

ottobackwards:
I don't know if you saw my question, and it is hidden now so I'll restate:

What about functions that check for the GLOBAL_ flag, and may make assumptions that do not hold with a non-z partial configuration?

ottobackwards:
Even if they don't now, I believe that is the intent of the feature

nickwallen:
> Can't we turn everything passed in into json and use the json patch stuff to "build out' the configuration? Why not have the globals as a map<string,object> all the time?

It still is always a map<string, object>, that has not changed. When I said "How about we just restrict %define to strings only for now.", that just means that a `%define` drops into the map<string, object>, values of strings without trying to interpret the value and create a complex type.

The main itch that I'm trying to scratch here (which is maybe the 80% case?) is to just set a simple property value.
```
%define bootstrap.servers=node1:6667
KAFKA_GET("indexing", 10)
```

If we treat `%define` arguments as JSON (so it behaves just like the `-z`) then it would have to look something like this, which is way more quotes and parens than I'd like to deal with.
```
%define { "bootstrap.servers" : "node1:6667" }
KAFKA_GET("indexing", 10)
```

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/StellarShell.java
nickwallen: I do not see a problem here.  Functions need whatever GLOBAL_CONFIG settings they need.  This PR does not make it harder or less likely for functions to get the values they need.  This PR makes it easier for a user to provide the settings, or alter the settings that a function needs.  IMHO

nickwallen:
I am re-reading through everyone's comments just to make sure I haven't misjudged all the feedback that I have received so far (thanks to all for feedback).  

> you could also make %define nicks_new_global_field := my_complex_field That makes somewhat more sense since stellar assignment is := and property assignment is =

I think I misjudged this one @cestella .  I like this for a number of reasons.
* This is still very simple for the user.  
* This is more consistent with the existing assignment construct.
* This makes it far easier to build complex types for a global %define; its easier to build a Stellar map, than a JSON string where the quotes need escaped in Stellar.

So the simple case becomes this.
```
%define bootstrap.servers: = node1:6667
```

And the more complex case becomes this.
```
value := { "foo": "bar" }
%define bootstrap.servers := value
```

nickwallen:
The latest commit implements the previous suggestion.  I like it... I like it a lot.  See the updated PR description which walks through the syntax.   Let me know what you guys think.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/StellarExecutor.java
```diff
@@ -203,6 +205,9 @@ public StellarExecutor(String zookeeperUrl, Console console, Properties properti
     index.put("quit", AutoCompleteType.TOKEN);
     index.put(StellarShell.MAGIC_FUNCTIONS, AutoCompleteType.FUNCTION);
     index.put(StellarShell.MAGIC_VARS, AutoCompleteType.FUNCTION);
+    index.put(StellarShell.MAGIC_GLOBALS, AutoCompleteType.FUNCTION);
+    index.put(StellarShell.MAGIC_DEFINE, AutoCompleteType.FUNCTION);
+    index.put(StellarShell.MAGIC_UNDEFINE, AutoCompleteType.FUNCTION);
```
nickwallen: I also made sure the new magics are added to the auto-complete feature.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/StellarShell.java
```diff
@@ -287,39 +294,131 @@ private void handleStellar(String expression) {
...
+   */
+  public void handleMagicDefine(String expression) {
+
+    // grab the expression in '%define <assign-expression>'
+    String assignExpr = StringUtils.trimToEmpty(expression.substring(MAGIC_DEFINE.length()));
+    if (assignExpr.length() > 0) {
+
+      // the expression must be an assignment
+      if(StellarAssignment.isAssignment(assignExpr)) {
+        StellarAssignment expr = StellarAssignment.from(assignExpr);
```
nickwallen: Here I leveraged the same assignment code that is used elsewhere in the REPL.

cestella:
YES!  I love it.  We can create a follow-on for a `-g` option to initialize the global config, but I like this approach.  +1 by inspection, but I want to make sure @ottobackwards is cool with it as well before we go to master.

ottobackwards:
thanks @cestella.  I am waiting on an answer to the question on other functions that look for the GLOBAL_CONFIG flag.   See above, a couple of times.

nickwallen:
> What about functions that check for the GLOBAL_ flag, and may make assumptions that do not hold with a non-z partial configuration?

 I thought I had responded to this, but I must not have.  Can you give an example?  I am not seeing a problem.

ottobackwards:
So the GLOBAL_CONFIG is used when setting up stellar in a few places, and used in a couple of functions to determine what they return.  And may be used by other functions in the future both Metron's or UDFs.

I am asking about co-opting this flag, and if we have thought about any unwanted consequences from this now or in the future, based on the intent of the flag.

Is the scope of that flag only the shell?  Will the functions see it? If they do, and they expect a field to be set in the config, but it is not because the shell only set one property that the user did set, what will the behavior be?

We may be able to say "Well, right now non of the functions that depend on that flag and specific properties will be called from the shell".  But what keeps someone from writing a UDF that depends on that flag from breaking in the shell because they don't understand that we do this?

nickwallen:
What flag am I co-opting?

ottobackwards:
I am not sure I know how else to explain it.

ottobackwards:
I am +1 on this.

While I think it is confusing to conceptually have a GLOBAL_CONF wrt the zookeeper global.js, good documentation and community feedback from use will get us there.

I think UDF authors, when they 'guard' with the GLOBAL_CONF profile are not just requiring that there **is** any config at all, they really always want the GLOBAL_CONF **AND** some amount of the properties set there. Having them have to do secondary validation of config property presence doesn't seem right.


I would like to see a jira for follow up documentation as part of the stellar UDF author's guide.

Thanks for working on this and taking the time to answer my questions @nickwallen

nickwallen:
This is a replay of an offline conversation....  

> Sorry for the confusion on my part @ottobackwards . I know much of the innards of Stellar is only documented in the code (if that), so let's level set on that.  
> 
> A "StellarExecutor" is responsible for running Stellar code for the REPL.  The executor can maintain additional "Context" that provides resources for the Stellar code that it is executing.  Each of these resources in the "Context" are called "Capabilities".   For example, a Zookeeper Client is available as a Capability called "ZOOKEEPER_CLIENT".  There is also the "GLOBAL_CONFIG" capability that exposes Metron's global configuration to Stellar code.
> 
> These capabilities, like the GLOBAL_CONFIG, are exposed to Stellar code in the same way when running either in the REPL or in a topology.  A Stellar function can be written that fails if the GLOBAL_CONFIG is not available.  It can also be written to use acceptable defaults, if a GLOBAL_CONFIG does not exist.
> 
> Previously the only way to create the GLOBAL_CONFIG in the REPL, was to have a Zookeeper service running with the Metron config deployed and then launch the REPL with `bin/stellar -z node1:2181".  Without doing that there was no GLOBAL_CONFIG.  Any function that was written to expect a GLOBAL_CONFIG, was not usable in the REPL, unless you launched and configured a Zk instance.
> 
> This PR provides the ability to...
> (1) Create a GLOBAL_CONFIG manually in the REPL without the need for Zk and 
> (2) Modify a GLOBAL_CONFIG manually within a single session, without modifying Zk.
> 
> There really is no such thing as an "incompletely formed" GLOBAL_CONFIG.  The GLOBAL_CONFIG is just a key/value store.  It contains whatever it contains.  But if there is a missing value in the GLOBAL_CONFIG that your function needs, then this change only makes it far simpler to add that missing value to the GLOBAL_CONFIG. 
> 
> And this is something you would need per your PR #690.  If you expect to run the REPL disconnected from a Metron cluster, then you need a way to create a GLOBAL_CONFIG without a Zk instance.

ottobackwards:
"There really is no such thing as an "incompletely formed" GLOBAL_CONFIG. The GLOBAL_CONFIG is just a key/value store. It contains whatever it contains. But if there is a missing value in the GLOBAL_CONFIG that your function needs, then this change only makes it far simpler to add that missing value to the GLOBAL_CONFIG."

Just a clarification.  Let's say the users here are the UDF developer, and the metron user.
They don't know each other, never met, never talked nothing.  The User just installed 'bob's stellar lib'.

Bob's stellar lib needs one of the properties from the metron global.js to work, and is guarded.
User just runs the shell without zk and wants to use the function.

It passes the guard, and he can call and use the function, but it blows up because it expects a property to be there, but it is not because it is running like this.
User not happy, has no idea why it didn't work, has to track everything down...

By the time it get's back to Bob, he makes the point that 'hey, i guard for the config, when do you have a metron config that is not a metron config?  what is that flag for?'

ottobackwards:
- So we need to document what the flag actually means
- We need UDF writers to understand the flag and how to guard for missing properties
- Need the users to understand that the functions that require the global config will need THEM to define them somehow to use in this mode, in the function documentation maybe
- Maybe part of the annotation should be required profile and properties: ?

ottobackwards:
I might be thinking a little too far down the road, sorry

nickwallen:
I see this feedback as related to the design/use of global properties in general, versus anything that this PR does.  I'd suggest opening a discuss thread on this IMHO.

Nothing wrong with thinking too far down the road.  Thanks @ottobackwards 

Specific to this PR, I will take the two +1s on this as good enough to merge.  Please hollar, if I am wrong.

ottobackwards:
I agree, you are good to go

PR #741, opened 2017-09-08
METRON-1153 HDFS HdfsWriter never recovers from exceptions

justinleet:
## Contributor Comments
Added a try-catch around the actual write that will rotate the file and try again if there's a stream closed underneath it.  Added two unit tests, one that ensures things flow through nicely to a single file with a double write and one that ensures that things flow to two files if the channel is closed underneath for whatever reason (done by just calling `closeOutputFile()` when outside of the normal flow).

It's not a perfect solution, but it should alleviate any transient pain and let us know if the problem is deeper if it keeps showing up.

Also added a missing set to a field from the constructor which probably wasn't helping things since I happened to notice it.  It's a one line change, so it seemed excessive to create a separate PR.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/SourceHandler.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/SourceHandler.java
@@ -75,7 +75,8 @@ public class SourceHandler {
       try {
         out.write(bytes);
       } catch (IOException writeException) {
-        // Hope it's a transient issue, rotate our output file, and carry on.
+        // Hope it's a transient issue, rotate our output file, and try again.
+        // If it's not transient, the second attempt's exception will bubble up.
         LOG.warn("IOException while writing output. Attempting to rotate file and continue",
             writeException);
         rotateOutputFile();
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/SourceHandler.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/SourceHandler.java
@@ -75,12 +75,16 @@ public class SourceHandler {
       try {
         out.write(bytes);
       } catch (IOException writeException) {
-        // Hope it's a transient issue, rotate our output file, and try again.
-        // If it's not transient, the second attempt's exception will bubble up.
-        LOG.warn("IOException while writing output. Attempting to rotate file and continue",
-            writeException);
-        rotateOutputFile();
-        out.write(bytes);
+        LOG.warn("IOException while writing output", writeException);
+        // If the stream is closed, attempt to rotate the file and try again, hoping it's transient
+        if (writeException.getMessage().contains("Stream Closed")) {
+          LOG.warn("Output Stream was closed. Attempting to rotate file and continue");
+          rotateOutputFile();
+          // If this write fails, the exception will be allowed to bubble up.
+          out.write(bytes);
+        } else {
+          throw writeException;
+        }
       }
       this.offset += bytes.length;
 
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/SourceHandler.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/SourceHandler.java
@@ -64,6 +64,7 @@ public class SourceHandler {
     this.rotationPolicy = rotationPolicy;
     this.syncPolicy = syncPolicy;
     this.fileNameFormat = fileNameFormat;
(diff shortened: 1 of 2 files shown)


justinleet:
@ottobackwards From the ticket description, it looks like you thought there was a more involved refactoring for doing something like this.  Are you good with this sort of solution (at least as an initial step)?

justinleet:
Adjusted to only catch the "Stream Closed" exception, not the general IOException.  If it's stream closed, attempt the rotate and write, otherwise rethrow

cestella:
This looks good to me as a stopgap.  I'd like to understand eventually why this happens, but it appears very intermittent, so this isn't a bad solution at least as a stopgap.

+1 by inspection.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/SourceHandler.java
```diff
@@ -64,19 +64,34 @@ public SourceHandler(List<RotationAction> rotationActions
...
+      try {
+        out.write(bytes);
+      } catch (IOException writeException) {
+        LOG.warn("IOException while writing output", writeException);
+        // If the stream is closed, attempt to rotate the file and try again, hoping it's transient
+        if (writeException.getMessage().contains("Stream Closed")) {
+          LOG.warn("Output Stream was closed. Attempting to rotate file and continue");
+          rotateOutputFile();
+          // If this write fails, the exception will be allowed to bubble up.
+          out.write(bytes);
```
ottobackwards: If there is a non-recoverable error, you are in the same position as before.
I was thinking of having a 'failed' flag ( similar to what hadoop has in it's screams where they track closed explicitly )

ottobackwards: If there is a non-recoverable error, you are in the same position as before.
I was thinking of having a 'failed' flag ( similar to what hadoop has in it's screams where they track closed explicitly )

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/hdfs/SourceHandler.java
```diff
@@ -64,19 +64,34 @@ public SourceHandler(List<RotationAction> rotationActions
...
   protected void handle(JSONObject message, String sensor, WriterConfiguration config, SyncPolicyCreator syncPolicyCreator) throws IOException {
     byte[] bytes = (message.toJSONString() + "\n").getBytes();
     synchronized (this.writeLock) {
-      out.write(bytes);
+      try {
+        out.write(bytes);
+      } catch (IOException writeException) {
+        LOG.warn("IOException while writing output", writeException);
+        // If the stream is closed, attempt to rotate the file and try again, hoping it's transient
+        if (writeException.getMessage().contains("Stream Closed")) {
```
ottobackwards: Using string compare on the message seems kind of thin.  And there may be other exceptions that come up that we want to handle.

Part of what I hoped would come from the discuss thread was a discussion about what was recoverable and what was not recoverable, and a better way to handle both cases.

ottobackwards:
@justinleet My question with regards to the necessity for a refactor came from wanting to handle the exception in the HDFSWriter, where the exceptions where being caught at the time.

The fact that the tuples are 'unpaired' from their messages, and that we handle them all together seems to me to be problematic if you want to take per message/sourcehandle action.

I think your approach removes the immediate problem with that, although having both the hdfs writer and the source handler itself take actions and split the 'ownership' doesn't feel quite right to me.

@cestella It would be good to know why this is happening, but in truth, any persistent network connection in a long lived 'forever' type application needs to guard against these kinds of errors.

For reliability reasons, and where we want to get to, we require a more clear documentation of the failure and recovery states of the writers, esp. the hdfs as we are batching.  We also need to understand all the ways the stream can fail, to the extent that that is possible.

I am not sure that limiting it to this one case is enough,  there will still be many possible ways to end up with a very unclear situation, that the writer is failing continuously and but the source handler is not removed from 'service'.  Users, as we have seen on the list will be left to track through system to work their way to this problem.  

Failure to store is a critical failure of the system.  Esp. in systems where there are data retention rules or SLA's on data loss.  Thus we need in addition to the handling of this, an alerting strategy.

While this fix ( questions pending ) is an improvement of the symptom, it does not address the higher level issue or severity of this problem.

cestella:
I absolutely agree, this does not address the higher level issue or severity.  This is merely a stopgap to fix something that we are seeing in the field and on the list.  I made a JIRA to capture this: https://issues.apache.org/jira/browse/METRON-1170

PR #747, opened 2017-09-11
METRON-1175 [FEATURE-BRANCH] allow lazy loading of BundleSystem with BundleSystem interface and holder pattern

ottobackwards:
The VFS system doesn't play well with Spring as a bean. When initialized at startup, it doesn't seem to get the right Configuration, and cannot see all of hdfs.
In order to be both a bean used by multiple services in REST and actually work, we should implement a holder pattern with on demand creation of bundle system.
To do this, BundleSystem should be an interface, with multiple implementations. The Builder class should be able to create the correct implementation.

This PR creates the BundleSystem interface, the DefaultBundleSystem, OnDemandBundleSystem, BundleSystemType, BundleSystemBuilder classes required to implement this.

It allows introduces this to the REST service.
This is the base capability to allow the REST service to use the BundleSystem as a bean.
That will be done in the UI PR.

This PR blocks development of the metron config ui for extensions and parsers.

## Testing
Testing should look for a lack of regression
- build and tests run
- full dev -> data in kibana, sensors ui works in management

- [x] Is there a JIRA ticket associated with this PR? 

### For code changes:
- [n/a] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
  ```
  ```

- [n/a] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 


ottobackwards:
Will be integrated with bundles pr

PR #680, opened 2017-08-03
METRON-1078: Metron Indexing fails to stop during "Stop All Services"

dlyle65535:
## Contributor Comments
There is a corner-case where sometimes the Storm REST UI is stopped before the Metron topologies when 'STOP ALL' is selected from Ambari. This PR orders STORM_REST_API-STOP after the Metron topologies STOP.

To verify, spin up full-dev and select 'Stop All' from the UI. Verify that 'Storm UI Server Stop' comes after all the the Metron topologies.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [X] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [X] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [X] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [X] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [X] Have you included steps or a guide to how the change may be verified and tested manually?
- [X] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  mvn -q clean integration-test install && build_utils/verify_licenses.sh 
  ```

- [N/A] Have you written or updated unit tests and or integration tests to verify your changes?
- [N/A] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [X] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

#### Note:
Please ensure that once the PR is submitted, you check travis-ci for build issues and submit an update to your PR as soon as possible.
It is also recommended that [travis-ci](https://travis-ci.org) is set up for your personal repository such that your branches are built there before submitting a pull request.

PR #660, opened 2017-07-19
METRON-1054 Ambari Mpack Attempts to Kill Topologies That Are Not Run…

nickwallen:
The Ambari Metron MPack attempts to kill topologies that are not running.  It also attempts to start topologies that are already running.  It reports these conditions as errors.  The MPack should not report errors when attempting to stop an already stopped topology nor when starting an already started topology.

### Changes

The Ambari MPack management scripts for the Parsers, Enrichment, and Indexing were all updated so that they only attempt to start topologies that are not running and only attempt to stop topologies that are running.

### Testing

This can be tested in the Full Dev environment using the following steps.

#### Indexing
1. Launch Full Dev and ensure all Metron components are running. 
1. `storm kill indexing`
1.  In Ambari, go to http://node1:8080/#/main/services/METRON/summary and click Service Actions > Restart All.
1. This action should complete successfully.  It should have stopped, then started the Parsers and Enrichment, but it should have only started Indexing, since Indexing was not running.

#### Enrichment
1. Ensure that all Metron components are running.
1. `storm kill enrichment`
1. In Ambari, go to http://node1:8080/#/main/services/METRON/summary and click Service Actions > Restart All.
1. This action should complete successfully.  It should have stopped, then started the Parsers and Indexing, but it should have only started Enrichment, since Enrichment was not running.

#### Parsers
1. Ensure that all Metron components are running.
1. `storm kill bro`
1.  In Ambari, go to the Hosts page at http://node1:8080/#/main/hosts/node1/summary.
1. Find the "Metron Parsers" service and click Start on the adjacent button.
1. This action should complete successfully.  It should have stopped, then started Snort.  It should have only started Bro, since Bro was not running.

PR #748, opened 2017-09-11
METRON-1177 Stale running topologies seen post-kerberization and cause exceptions

nickwallen:
[METRON-1177](https://issues.apache.org/jira/browse/METRON-1177)

### Problem

After running the Ambari Kerberization process on a cluster where Metron was installed with the MPack, often times the Kerberization process would complete successfully, but the running Metron topologies were stale and had not been restarted properly after all Kerberization steps completed.  In other cases, the Metron service check would fail when Ambari began restarting all cluster services.

One clue that this has occurred is that when querying Storm using the Thrift API to check on topology status after kerberization would result in the following error.
```
AuthorizationException(msg:getTopologyInfo on topology snort is not authorized)
```

### Solution

* All Metron services have to be started before performing a Metron service check.
* All external dependencies like Storm, HBase, Kafka, etc must complete their service check before performing the service check of a Metron service having those dependencies.
* Added Storm as a start dependency for the Metron Profiler.
* Metron Profiler has to be stopped before Storm is stopped.

### Testing

This was tested by launching the Full Dev environment, kerberizing the environment, and then monitoring the order in which each of the service start, stop and status check actions occurred.  I was not able to replicate the failure condition with this fix.


mmiklavc:
Oh man, after looking at this it dawns on me that a dependency tree analyzer for the Ambari role command order would be extremely useful.

+1 by inspection. Thanks for fixing this @nickwallen.

nickwallen:
Thanks. We are still testing so I'll hold off on committing it.  I think we found some other items to address.

nickwallen:
I committed another change.  Everything seems to be working with this additional fix.  Would like @anandsubbu to add his experiences working with this patch.

This problem started with #660 .  I added a conditional to make sure that a topology is actually running before trying to stop it and similar logic for starting topologies.  This introduced a subtle dependency in that the "is running" conditional depends on the Storm UI/API.  If the Storm UI/API is not running, then it assumes the topology is already stopped.  This is only a problem if the Storm UI/API is stopped before Metron components.

We then attempted a fix in #680.  There we added dependencies to `STORM_REST_API-STOP` to ensure that all Metron topologies are stopped prior to the Storm UI.  This is necessary since the Metron MPack uses the Storm API to check the status of topologies before shutting them down.  

Unfortunately, this did not fix the problem, at least not in all cases. We found instances where the Metron MPack was still unable to stop topologies before Kerberization because the Storm UI/API had already been stopped.  

After some sleuthing, it seems that the status checks depend on a process running on port 8744. 
 I found that the storm component that listens on port 8744 is actually called `STORM_UI_SERVER` by Ambari.  So it is the `STORM_UI_SERVER` component that we need to add the dependencies to.  And this is what I have added in the latest commit. 

I am assuming that this is needed in addition to `STORM_REST_API`, but I am not completely sure. 
 So I left the existing `STORM_REST_API` dependencies.

anandsubbu:
Hi @nickwallen , the latest of you patch works just perfect. I can see that the STORM_UI_SERVER waits until all of the metron topologies are stopped (killed i.e.) and then shuts down. This causes all of the topologies to come up gracefully during the kerberization process. Thank you for the fix! 

+1 (non-binding).

nickwallen:
@mmiklavc Let me know if you can re-up on your previous +1.  I added another commit since you originally looked at it. It should be ready to go now.  Thanks for the review.

mmiklavc:
+1. Thanks for the thorough explanation and post-mortem on this!! That was extremely helpful.

PR #749, opened 2017-09-12
METRON-1176 Rest support for setting permissions on files written by HdfsService

ottobackwards:
This PR adds support to the HdfsController and HdfsService to optional set the permissions on files written to hdfs.

> note: this pr is in support of METRON-1136 feature branch, and the installation and management of extensions.  Since it is generally applicable, it is being done against master and will be then pulled into
the feature branch. It is blocking the feature branch.

Three new parameters, one each for user, group, and other mode are passed, in the form of symbolic permission strings ('rwx','r-x','r--').

If **any** of the three are missing, no permission override will be executed.
If **any** of the three are not valid, and exception will be thrown.

##Testing
[ ] Build, tests should run
[ ] Full Dev, Swagger, use the HdfsController post, and set the permissions, don't set the permissions, set invalid permissions
[ ] Verify files, for example a good place to write to is "/apps/metron/patterns/NAME"
[ ] In the metron ui, testing grok patterns should work

```bash
> vagrant ssh
> sudo su -
> hdfs dfs -ls /apps/metron/patterns

```

- [na] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
  ```
  ```

### DIFF
--- a/metron-interface/metron-rest/README.md
+++ b/metron-interface/metron-rest/README.md
@@ -300,6 +300,9 @@ Request and Response objects are JSON formatted.  The JSON schemas are available
   * Input:
     * path - Path to HDFS file
     * contents - File contents
+    * userMode - [optional] symbolic permission string for user portion of the permissions to be set on the file written. For example 'rwx' or read, write, execute. The symbol '-' is used to exclude that permission such as 'rw-' for read, write, no execute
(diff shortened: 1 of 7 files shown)


merrimanr:
I spun this up in full dev and it worked as expected.  Thanks for the contribution.  +1

PR #755, opened 2017-09-13
METRON-1185: Stellar REPL does not work on a kerberized cluster when calling functions interacting with HBase

cestella:
## Contributor Comments

Because we put all the jars on the classpath from `$METRON_HOME/lib` via a glob, if we get the wrong order, we get classpath issues around Stopwatch (i.e. more recent guava may sneak onto the classpath first). This is non-deterministic and can cause issues randomly. Instead of a glob, we should choose one of the uber jars with all the metron Stellar functions baked in and allow users to provide any extra directories we should include via an env variable. Also, we are not pulling the full hbase classpath, so in a kerberized cluster we do not get all of the configs on the classpath, so we can't find always find tgts.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/main/scripts/stellar
+++ b/metron-stellar/stellar-common/src/main/scripts/stellar
@@ -28,7 +28,8 @@ elif [ -e /usr/lib/bigtop-utils/bigtop-detect-javahome ]; then
   . /usr/lib/bigtop-utils/bigtop-detect-javahome
 fi
 
-export HBASE_CONFIGS=/etc/hbase/conf
+export HBASE_CONFIGS=$(hbase classpath)
 export METRON_VERSION=${project.version}
 export METRON_HOME=/usr/metron/$METRON_VERSION
-java $JVMFLAGS -cp "$HBASE_CONFIGS:$METRON_HOME/lib/*" org.apache.metron.stellar.stellar.shell.StellarShell "$@"
+export STELLAR_LIB=$(find $METRON_HOME/lib/ -name metron-enrichment*.jar)
+java $JVMFLAGS -cp "$HBASE_CONFIGS:${CONTRIB_LIB:-$METRON_HOME/contrib-lib}:$STELLAR_LIB" org.apache.metron.stellar.common.shell.StellarShell "$@"
--- a/metron-stellar/stellar-common/src/main/scripts/stellar
+++ b/metron-stellar/stellar-common/src/main/scripts/stellar
@@ -32,4 +32,5 @@ export HBASE_CONFIGS=$(hbase classpath)
 export METRON_VERSION=${project.version}
 export METRON_HOME=/usr/metron/$METRON_VERSION
 export STELLAR_LIB=$(find $METRON_HOME/lib/ -name metron-enrichment*.jar)
-java $JVMFLAGS -cp "$HBASE_CONFIGS:${CONTRIB:-$METRON_HOME/contrib}:$STELLAR_LIB" org.apache.metron.stellar.common.shell.StellarShell "$@"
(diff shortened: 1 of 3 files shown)


metron-stellar/stellar-common/src/main/scripts/stellar
```diff
@@ -28,7 +28,8 @@ elif [ -e /usr/lib/bigtop-utils/bigtop-detect-javahome ]; then
   . /usr/lib/bigtop-utils/bigtop-detect-javahome
 fi
 
-export HBASE_CONFIGS=/etc/hbase/conf
+export HBASE_CONFIGS=$(hbase classpath)
 export METRON_VERSION=${project.version}
 export METRON_HOME=/usr/metron/$METRON_VERSION
-java $JVMFLAGS -cp "$HBASE_CONFIGS:$METRON_HOME/lib/*" org.apache.metron.stellar.stellar.shell.StellarShell "$@"
+export STELLAR_LIB=$(find $METRON_HOME/lib/ -name metron-enrichment*.jar)
```
nickwallen: This will not pick-up any of the Profiler or Management functions, right?

nickwallen: We might need to create a JAR specifically tailored for the REPL.

ottobackwards: So we are just picking the enrichment jar?

ottobackwards:
Is this just a temporary thing?

I think the longer term thing is to split the stellar out into their own modules, an possibly have modules split further by special dependencies. 

I would like to do this with stellar extensions based on bundles.

metron-stellar/stellar-common/src/main/scripts/stellar
cestella: It will pick up the profiler functions, they're included in the enrichment jar (that's actually what motivated this, `PROFILE_GET` didn't work randomly based on the order of the classpath).

We probably should include the management functions directly, though.  I'll adjust.

cestella:
@ottobackwards I'd argue that it SHOULD be a temporary thing.  At the moment things are broken as they are.

ottobackwards:
ok, if and when we look at stellar archetypes and extensions we'll have to keep this in mind

cestella:
Ok, I just want to be completely clear, we're actually choosing *one* of our jars, the parsers jar, because it's an uber jar that has ALL of our stellar functions in Metron in it.  I'm also providing explicit support for the management functions.  This should make our classpath more deterministic, I think, and provide a mechanism that is more flexible for passing in custom jars (the env variable, rather than presuming users would drop things in `$METRON_HOME/lib`.

metron-stellar/stellar-common/src/main/scripts/stellar
cestella: The current commit should support both profiler and the management function at this point.

ottobackwards:
Whatever it takes in the meantime.  When this is plugin based, this all goes away.

cestella:
So, I tested this on a kerberized multi-node cluster and full-dev.  I ensured that I tried with and without the management functions jar and with/without kerberos as well.

justinleet:
I'm +1 with this (as a temporary measure to make things work), assuming @ottobackwards and @nickwallen are as well.

cestella:
@ottobackwards and @nickwallen any issues?

nickwallen:
+1 by inspection.  I agree with @justinleet; its good for a temporary work around.  Thanks for the fix.

ottobackwards:
+1, just for the record

PR #757, opened 2017-09-14
METRON-938: "service metron-rest start <password>" does not work on CentOS 7.

justinleet:
## Contributor Comments
Moved things from an init script to Ambari, since we manage pretty much everything through Ambari anyway.  Hive does something similar. Now, the password comes from Ambari and we're independent of the `service` command and underlying CentOS management in this realm.

A couple properties are pushed to Ambari to make this easier.  Notably, it's a pain in the butt (and maybe not possible at all), to get the version of our Metron install from interrogating Ambari.  As such, it's added an invisible read-only property.  This probably has to be revisited when we support upgrades (in order to make sure that the version increments appropriately).  If this is objectionable, we may want to just hard code it and replace it when we do the overall release increments like we do the other instances.

As a sidenote, it would be really nice to have a version/profile/whatever of full-dev available on CentOS 7.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


justinleet:
As a note, the way I tested this was running it up in a CentOS 7 VM, along with a couple changes to get things working:
* Turning off firewalld during spinup.
* Downgrading python.x86_64 and python-libs.x86_64 to 2.7.5-48.el7 to avoid an SSL error.  Apparently 58 or whatever latest epel has causes entirely unrelated issues with node registration in Ambari.
* Installing Metron + dependencies via Ambari.

I also ran in standard full dev to ensure that it still worked in 6.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
```diff
@@ -76,6 +81,9 @@
 global_json_template = config['configurations']['metron-env']['global-json']
 global_properties_template = config['configurations']['metron-env']['elasticsearch-properties']
 
+# Java
+java64_home = config['hostLevelParams']['java_home']
```
merrimanr: Is this necessary?  This property is also defined in params.py.

justinleet: Nope, I dropped it and the reference to it is changed to just java_home

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_commands.py
```diff
@@ -58,17 +74,94 @@ def init_kafka_acls(self):
 
     def start_rest_application(self):
         Logger.info('Starting REST application')
-        command = format("service metron-rest start {metron_jdbc_password!p}")
-        Execute(command)
+
+        # Build the spring options
+        # the vagrant Spring profile provides configuration values, otherwise configuration is provided by rest_application.yml
+        metron_spring_options = format(" --server.port={metron_rest_port}")
```
merrimanr: This is probably just a style issue and not critically important but would consider moving some of this logic to a start script that lives inside the metron-rest module?  MPack code is notoriously challenging to troubleshoot and having all this in here will make starting the REST app outside of Ambari very tedious.  I have had to do this a lot when troubleshooting REST problems and helping others.  Maybe REST is fairly stable at this point and we shouldn't have to mess with it much though.

This would also keep you from having to add a metron_version property to Ambari because the version would be substituted in the script at build time.  This is how all our other scripts get the version.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/service_advisor.py
```diff
@@ -69,6 +69,10 @@ def getServiceComponentLayoutValidations(self, services, hosts):
             message = "Metron REST must be colocated with an instance of STORM SUPERVISOR"
             items.append({ "type": 'host-component', "level": 'WARN', "message": message, "component-name": 'METRON_REST', "host": metronRESTHost })
 
+        if metronRESTHost not in hbaseClientHosts:
```
merrimanr: Good catch here.  This was a problem someone was going to run into eventually.  Do we also need a check for hdfsClientHosts?  zookeeperClientHosts?  Now that I think about it we also need to ensure REST is installed on the same host as parsers, enrichment and indexing (for topologoy start/stop).  Add a check for metronParsersHost might satisfy all those indirectly.

merrimanr:
Spun this up in full dev and tested several different cases including setting up a MySQL database for authentication.  Worked great.  I made a couple minor suggestions but I think this is pretty close to being ready.  Nice job @justinleet.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_commands.py
justinleet: @merrimanr So I tried doing this, but the problem is that if the Java command gets kicked up from a script, the Java command ends up showing up on in ps, which then goes and gives up spring.datasource.password.  Which is incredibly annoying.

Unless you have any ideas on how to get around that, I think we're stuck in Ambari

justinleet: Ignore the things I say, it would have helped if I'd done stuff right.  PR is updated to primarily run from a script, that's largely a renamed and slightly modified metron-rest (rename is to metron-rest.sh).  Need to update docs to reflect it.

I also killed the metron version variable by doing some shell globbing.  If you have multiple metron-rest jars in the lib directory it could be problematic, but I don't think it's unreasonable to not add random crap to the dirs.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/service_advisor.py
justinleet: Updated the service advisor to check on metronParsersHost

ottobackwards: isn't making everything require everything else on the same node kind of an anti pattern?

justinleet: It's more a shorthand for the fact that REST requires requires basically everything else (HDFS, ZK, HBase, all the topologies, etc.).  We could do all the requirements in specifics if we want (and I'm not necessarily opposed to doing it for REST while I'm already in here), but I'd argue that's a good candidate for a follow on task (since we do it on multiple items).

I'd prefer to do what the Hive service does and just automatically place every required item on the same node automatically (the selector on install automatically selects the same thing for all the components), which then means that this colocation check on something like parsers is more robust.  Having said that, I never quite dug deep enough into how Ambari actually enforces that for Hive.

merrimanr:
Spun it up again in full dev, setup MySQL and everything works as it should.  Thanks so much @justinleet for fixing this!  Turned out great.  +1

justinleet:
Merged in master for extremely minor deconflict of `rest_commands.py` for imports added on the same line.

PR #759, opened 2017-09-14
METRON-1187 Indexing/Profiler Kafka ACL Groups Not Setup Correctly

nickwallen:
The Profiler MPack mistakenly uses the wrong flag/guard file to indicate that the Kafka ACL group has been setup.  Whichever component (either Profiler or Indexing) that is executed first will complete successfully.  The component to run next will not perform this setup task because the duplicated flag/guard file indicates that the setup was already completed successfully.

Most of the changes in this PR enhance the existing logging to help in the future when tracking down an issue like this.  The actual fix is a one line.

### Testing
To test this change, you need to kerberize and environment and then ensure that both the Profiler and Indexing topologies are successfully running and consuming data after kerberization.  

- [x] Tested in Full Dev
- [x] Tested in multi-node cluster

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
@@ -169,12 +169,13 @@ class IndexingCommands:
                                       self.__params.metron_keytab_path,
                                       self.__params.metron_principal_name,
                                       execute_user=self.__params.metron_user)
+
             start_cmd_template = """{0}/bin/start_elasticsearch_topology.sh \
-                                    -s {1} \
-                                    -z {2}"""
+                                        -s {1} \
+                                        -z {2}"""
             start_cmd = start_cmd_template.format(self.__params.metron_home,
                                                   self.__indexing_topology,
-                                                  self.__params.zookeeper_quorum),
+                                                  self.__params.zookeeper_quorum)
             Execute(start_cmd, user=self.__params.metron_user, tries=3, try_sleep=5, logoutput=True)
 
         else:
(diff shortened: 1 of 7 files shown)


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
```diff
@@ -273,7 +272,7 @@
 profiler_hbase_table = config['configurations']['metron-profiler-env']['profiler_hbase_table']
 profiler_hbase_cf = config['configurations']['metron-profiler-env']['profiler_hbase_cf']
 profiler_configured_flag_file = status_params.profiler_configured_flag_file
-profiler_acl_configured_flag_file = status_params.indexing_acl_configured_flag_file
+profiler_acl_configured_flag_file = status_params.profiler_acl_configured_flag_file
```
nickwallen: This is the crux of the problem.  A simple copy-paste error.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, params):
         self.__acl_configured = os.path.isfile(self.__params.indexing_acl_configured_flag_file)
         self.__hbase_configured = os.path.isfile(self.__params.indexing_hbase_configured_flag_file)
         self.__hbase_acl_configured = os.path.isfile(self.__params.indexing_hbase_acl_configured_flag_file)
+        self.__hdfs_perm_configured = os.path.isfile(self.__params.indexing_hdfs_perm_configured_flag_file)
```
nickwallen: This is another issue that I found.  I am not sure what the impact is, but seems like we would just always set the HDFS perms.

justinleet:
+1 by inspection.  This is a good catch to have made, thanks!

nickwallen:
Thanks to @anandsubbu and @justinleet for the assistance on this one.  This was tough to track down.

anandsubbu:
+1 (non-binding)

Indeed, this was a tricky one. Thank you @nickwallen for your root cause and fix!! I spun up a 12 node cluster with this patch, kerberized the cluster and confirmed that all of the topologies come up fine (esp. indexing and profiler).

PR #763, opened 2017-09-15
METRON-1190: Fix Meta Alert Type handling in calculation of scores

justinleet:
## Contributor Comments
Migrating the create method to not pass anything as an object array and instead to use a list.  At that point, it should all be one type going in.  Still need to spin up in fulldev to make sure nothing odd happens.

Rather than handle both types as noted in the JIRA, I opted to make the inputs all end up being the same (list instead of array). We control both constructions of input in the DAO, so there's no reason to dela with both.

To test, make sure both create and patch work correctly.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
+++ b/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
@@ -277,7 +277,7 @@ public class ElasticsearchMetaAlertDao implements MetaAlertDao {
     }
     metaSource.put(ALERT_FIELD, alertList.toArray());
 
-    // Add any meta fields and score calculation.
+    // Add any meta fields
     String guid = UUID.randomUUID().toString();
     metaSource.put(Constants.GUID, guid);
     metaSource.put(Constants.Fields.TIMESTAMP.getName(), System.currentTimeMillis());
--- a/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
+++ b/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
@@ -362,8 +362,18 @@ public class ElasticsearchMetaAlertDao implements MetaAlertDao {
    */
   @SuppressWarnings("unchecked")
   protected MetaScores calculateMetaScores(Document document) {
-    List<Object> alertsRaw = ((List<Object>) document.getDocument().get(ALERT_FIELD));
-    if (alertsRaw == null || alertsRaw.isEmpty()) {
+    Object rawField = document.getDocument().get(ALERT_FIELD);
+    List<Object> alertsRaw;
+    // Can come in as Object[] from create or List<Object> from update.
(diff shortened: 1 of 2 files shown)


justinleet:
The relevant method that things go into is `calculateMetaScores`.  Didn't mention that in the initial setup.  This is used by `handleMetaUpdate` and `update` (in the case of an update to a metalert itself), but they were both passing in different field types for the alert field.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
```diff
@@ -275,13 +275,13 @@ protected Document buildCreateDocument(MultiGetResponse multiGetResponse, List<S
         alertList.add(response.getSource());
       }
     }
-    metaSource.put(ALERT_FIELD, alertList.toArray());
+    metaSource.put(ALERT_FIELD, alertList);
```
nickwallen: @justinleet  "The alerts field can actually come as two different types List<Object> vs Object[] during calculation of meta score fields."


Your changes here seem simple enough.  Before we used arrays, now using Lists.  But I am not following how this solves the problem you describe in JIRA.  

Where is the code that would handle/transform/standardize items that come in as either Array or List?

justinleet: If you look at the first two commits, they were originally to do the JIRA as noted (handle both).  The solution of the last commit is to just not use arrays at all. In actuality though, both constructions of the alert field are controlled by the DAO itself, so there's no reason to use arrays at all.

Basically, this PR solves the underlying problem (The two paths used different types), rather than the symptom (That two different types went into the same method).

justinleet: Updated the PR description to make the reasoning more obvious to anyone glancing there first.

justinleet:
Correction to what I said earlier: the update side isn't directly actually controlled by the `ElasticsearchMetaAlertDao` itself, it's controlled by the `IndexDao`'s building of the document that is done in `patch` and passed to the `update` method.  The fix to `create` ensures that what the `ElasticsearchMetaAlertDao` (which extends `IndexDao`) is consistent with the typing that the parent class does.

nickwallen:
+1 LGTM based on what you have explained.

nickwallen:
Outside of this PR, we might need to better doc the expected behavior for meta-alerts (or maybe I just need to find the existing docs for that.)

nickwallen:
It was hard for me to grok how `ElasticsearchDAO` and `ElasticsearchMetaAlertDAO` differ and how the `ElasticsearchMetaAlertDAO` uses another `IndexDAO` under-the-hood.  I think more cow bell (I mean javadocs) would help here.  But that is all outside the scope of the work you've done here.

justinleet:
@nickwallen Some of that is provided in metron-indexing/README.md, but I'm definitely not opposed to fleshing that stuff out more.

nickwallen:
Thanks for the pointer.  Duh, I should have looked there.  I'll read up on that

PR #768, opened 2017-09-21
Metron 1123: Add group by option using faceted search capabilities of metron-rest-api

iraghumitra:
## Contributor Comments
This PR extends the Metron Alerts GUI capabilities to do group by on Alerts Data

The PR is on top of METRON-1189

The group by buttons appear on top of the table. When user selects a group, the table view changes to a tree view with the values of the groups as the root of the tree. The individual alerts would be shown as the leaf nodes of the tree.

Tree view has all the following features

- Search will search through all the groups
- Users can click on the data  inside the table to search
- Users can select and table inside tree to view the details
- Users can escalate the alerts from the table inside the tree
- Tree view is refreshed automatically depending on the polling setting in the settings
- A cumulative score and number of alerts  in a group is shown next to group name
- Table inside a group can be paginated
- The column names for the alerts table in the tree can be configured via configure column pane

The table config settings is moved next to settings on the GUI as it makes more meaning there.

image
image
image

## Testing	

E2E test cases are available to test groups functionality. From root of the project 

```
cd metron-interface/metron-alerts
sh ./scripts/start-server-for-e2e.sh
```

In othere shell run the test cases

```
npm run e2e
```

All the features listed in contributor comments can be verified manually 

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
I would like to review this, can you resolve conflicts?

iraghumitra:
@ottobackwards Done please have a look ...

ottobackwards:
I don't understand how this is not picking up the management pack integration for alerts?  I don't see it in ambari

iraghumitra:
@ottobackwards can u check it now I merged master again looks like mpack is in now.

ottobackwards:
I cannot get this to vagrant up:

----> MANY OF THESE ----->
metron-config.noarch: E: script-without-shebang /usr/metron/0.4.1/web/expressjs/node_modules/repeat-string/package.json
This text file has executable bits set or is located in a path dedicated for
executables, but lacks a shebang and cannot thus be executed.  If the file is
meant to be an executable script, add the shebang, otherwise remove the
executable bits or move the file elsewhere.

metron-config.noarch: E: script-without-shebang /usr/metron/0.4.1/web/expressjs/node_modules/lodash/fp/pullAll.js
This text file has executable bits set or is located in a path dedicated for
executables, but lacks a shebang and cannot thus be executed.  If the file is
meant to be an executable script, add the shebang, otherwise remove the
executable bits or move the file elsewhere.

metron-config.noarch: E: script-without-shebang /usr/metron/0.4.1/web/expressjs/node_modules/lodash/_LazyWrapper.js
This text file has executable bits set or is located in a path dedicated for
executables, but lacks a shebang and cannot thus be executed.  If the file is
meant to be an executable script, add the shebang, otherwise remove the
executable bits or move the file elsewhere.

------
snip
------
(none): E: error while reading SRPMS/metron: 'SRPMS/metron'
14 packages and 1 specfiles checked; 1822 errors, 90 warnings.
[INFO]
[INFO] --- maven-resources-plugin:3.0.1:copy-resources (copy-rpms-target) @ metron-rpm ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 14 resources
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] metron-deployment .................................. SUCCESS [  0.672 s]
[INFO] metron-rpm ......................................... SUCCESS [07:45 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 07:45 min
[INFO] Finished at: 2017-09-29T18:10:52-04:00
[INFO] Final Memory: 15M/220M
[INFO] ------------------------------------------------------------------------
ottofowler@Winterfell [18:10:52] [~/tmp/metron-pr-768/metron-deployment] [pr-768]

ottobackwards:
Let me be more clear, sorry.

Vagrant up doesn't work.
If I do 
```bash
% metron-deployment> mvn package -DskipTests -P build-rpms
```
I get these errors in the RPM's.
Any idea?

iraghumitra:
Looks like the problem is not with the build but with starting metron-config ui on the VM. Are you using fedora instead of centos by any chance?. 
AFAIU some script is missing shell invocation. Can you try manually installing the rpm with verbose options that might give us some more help.
I just tried a full-dev-setup on my dev machine it is working fine for me.

ottobackwards:
All of this happens in docker

iraghumitra:
@ottobackwards If you dont mind can you give me exact steps to simulate the installation issue. I seem to miss something here.

ottobackwards:
I run this standard [checkout-pr](https://github.com/ottobackwards/Metron-and-Nifi-Scripts/blob/master/metron/checkout-pr) script, to download in to ~/tmp.

cd into the directory created 's metron-deployment/vagrant/full-dev-platform
vagrant up

ottobackwards:
I am going to try again

ottobackwards:
It looks like it is working now.  Maybe this error was an artifact or manifestation of the dependency issues that 'resolved' themselves.  I'll get going

ottobackwards:
What does the ungroup button **do** when you hit it and it is still in table view?

ottobackwards:
The cut off text for the enrichment, both in the group and the column header will make it unusable if there is more than one enrichment I think
screen shot 2017-10-03 at 10 40 51
screen shot 2017-10-03 at 10 40 43

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/MetaAlertStatus.java
```diff
@@ -31,4 +31,4 @@
   public String getStatusString() {
     return statusString;
   }
-}
+}
```
merrimanr: Was this intentional?

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/metaalert/MetaAlertCreateResponse.java
```diff
@@ -28,4 +28,4 @@ public boolean isCreated() {
   public void setCreated(boolean created) {
     this.created = created;
   }
-}
+}
```
merrimanr: Was this intentional?

metron-interface/metron-alerts/e2e/alerts-list/tree-view/tree-view.e2e-spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      'firstSubGroupIdCol': ['07b29c29-9...ff19eaa888', '7cd91565-1...de5be54a6e', 'ca5bde58-a...f3a88d2df4',
+        '5d6faf83-8...b88a407647', 'e2883424-f...79bb8b0606']
+    };
+
+    page.selectGroup('host');
+    page.selectGroup('enrichments:geo:ip_dst_addr:country');
+    expect(page.getActiveGroups()).toEqualBcoz(['host', 'enrichments:geo:ip_dst_addr:country'], 'two groups should be selected');
+
+    expect(page.getDashGroupValues('runlove.us')).toEqualBcoz(dashRow_runLoveUs.dashRow,
+                                                              'Dash Group Values should be present for runlove.us');
```
merrimanr: Shouldn't each expect message be different?  Looks like there are 3 with the exact same message.

merrimanr:
I spun this up in full dev and it's working pretty well.  The only functional issue I could find was that the bulk actions are disabled when alerts are selected in group/tree view.  This is likely due to the fact that tree view inherits from list view and selectedAlerts are part of list view.

Other than that I think this is pretty close.  We should add an e2e test for bulk actions in tree view (would have caught the issue above).  Also is there a test for group reordering (drag and drop)?

iraghumitra:
@ottobackwards Good to hear the build issue is out of the way I was going nowhere in figuring out the build failure.
 - The ungroup button does nothing when alerts are not grouped. I feel we can disable the ungroup button when no groups are selected. That might be less confusing
 - Most of the field names in ES are too long for an aesthetic display, we have an option to rename filed's names to make them look meaningful. We can increase the number of chars but I feel there is no comprehensive way to make them look meaningful. I would suggest using renaming field names in UI. If you have a better suggestion I am open for that too.

ottobackwards:
I am not sure.  For multi field names "foo:bar:baz" maybe just "baz".  or maybe like exceptions "f:b:baz".  But that may not be great either.  Just wanted to bring up the point for discussion, not a blocker.

iraghumitra:
@merrimanr I will check the escalation actions, I guess I missed something there.

merrimanr:
With respect to displaying field names, I think it's going to be tough coming up with a strategy that solves every case.  This will likely depend on a user's preference and is subjective anyways.  I agree with @iraghumitra, exposing a way to rename fields solves it.

ottobackwards:
I think we should disable the button if there are no groups.  I *does* do something, I don't know if it is resorting or something else, but the table refreshes and seems to have different content.

metron-interface/metron-alerts/e2e/alerts-list/tree-view/tree-view.e2e-spec.ts
iraghumitra: I am just holding the expected values in a map for easy understanding. 
The actual values that we are expecting are dashRow, firstSubGroup, firstSubGroupIdCol which are present in the map named dashRow_runLoveUs.

iraghumitra:
@ottobackwards & @merrimanr Fixed the following 
- Disabling the ungroup button when no groups are selected
- Fixed issue with select all alerts in tree view

merrimanr:
Thanks @iraghumitra the bulk actions are working now.  Just need to address the e2e test suggestions and I think this will be good to go.

iraghumitra:
@merrimanr oops missed the comment thanks for letting me know again. I will add a test case for tree bulk actions. Looks like e2e for drag-drop is going to be tough I tried a bit but could not get it to work. The dragula community says it is going to be tough having protractor cases for it too you can check it [here](https://github.com/valor-software/ng2-dragula/issues/720)

iraghumitra:
@merrimanr Added e2e tests for escalation actions in tree view

metron-interface/metron-alerts/src/app/alerts/alerts-list/tree-view/tree-view.component.ts
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  getAlerts(selectedGroup: TreeGroupData): Subscription {
+    let searchRequest = new SearchRequest();
+    searchRequest.query = this.createQuery(selectedGroup);
+    searchRequest.from = selectedGroup.pagingData.from;
+    searchRequest.size = selectedGroup.pagingData.size;
+    searchRequest.sort = selectedGroup.sortField ? [selectedGroup.sortField] : [];
+
+    return this.searchGroup(selectedGroup, searchRequest);
```
nickwallen: @iraghumitra I noticed that when grouping there are cases when the two locations where we display total counts will differ.  See the attached screenshots.  I underlined in yellow the counts that I am talking about.

screen shot 2017-10-09 at 3 26 49 pm

screen shot 2017-10-09 at 3 23 09 pm

It seems that the data used to build the total count in the group header is stale in comparison with the data used to display the individual alerts. We initially call `/group` to display the group header.  When a user expands the group we issue a `/search`, which retrieves fresher data.  Data that likely has a different total from what is currently displayed in the group header.

Can we update the count in the group header somewhere?  Maybe around here?

nickwallen:
When grouping, we are implicitly filtering any alerts that do not have a value defined for the 'group by' field.  

For example, in this screenshot there are roughly 44k alarms.  I have grouped by host and all my groups add up to about 17k.  The other 27k alarms are actually from Snort where the 'host' field is not defined.

screen shot 2017-10-09 at 4 34 37 pm

Would it be better to treat undefined, blank or missing fields as their own group?  So in this example, we would have another group that has no value and would have an alert count of roughly 27k.

The advantages I see with this...
* My group subtotals will always add up to the total number of alarms, which eliminates the mystery of missing alarms.
* Digging into alarms where a field is not defined or blank is a reasonable thing that a user would want to do.

This may very well be something to fix on the backend.

justinleet:
To handle the missing grouping aggregation. https://www.elastic.co/guide/en/elasticsearch/reference/2.3/search-aggregations-bucket-missing-aggregation.html

I *think* this is a backend change (probably just to add a param to the request and carry it through)

metron-interface/metron-alerts/src/app/alerts/alerts-list/tree-view/tree-view.component.ts
iraghumitra: @nickwallen that's a really nice catch, fixed it

iraghumitra:
@nickwallen I kind of agree with @justinleet here. The rest api is should handle empty values in a better way.  
Though we can circumvent the empty values issue in groups by assuming all the missing records have an empty value for the selected group but the current search api doesn't support searching on empty value ie we cannot fire a search query that says 'give me all rows where host is empty'. Hence we will not be able to show the individual alerts under the group. This need to be handled in rest.
Do you think this is a must-have for this PR or can we handle it in a follow-up PR ?.

nickwallen:
Thanks for the thoughts and links @justinleet and @iraghumitra . 

> @iraghumitra: but the current search api doesn't support searching on empty value ie we cannot fire a search query that says 'give me all rows where host is empty'

Right.  We would have to change the backend to do something like this.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-exists-query.html#_literal_missing_literal_query

> @iraghumitra: Do you think this is a must-have for this PR or can we handle it in a follow-up PR ?

Based on the level of work to get this done, I think this is definitely a separate effort.  I will create a JIRA for this.

I will run up your latest code.  It's looking really solid.

metron-interface/metron-alerts/src/app/alerts/alerts-list/tree-view/tree-view.component.ts
merrimanr: Do we have a test case for this?  I didn't see one in the commit that addressed this issue.

nickwallen:
https://issues.apache.org/jira/browse/METRON-1242

metron-interface/metron-alerts/src/app/alerts/alerts-list/tree-view/tree-view.component.ts
nickwallen: I confirmed manually that the latest does indeed fix this issue.

@merrimanr does have a good point.  A quick unit test of this would be great.

nickwallen: Actually, I spoke too soon.  @iraghumitra Your fix works great for a single group. But if you are grouping by multiple fields, we still get in a state where the group headers are showing stale data.  

In this screen shot, I am grouping by source-type, ip_src_addr, then country.  
* The count for the "country" group was correctly updated as soon as I expanded the header.
* The count for the "ip_src_addr" group was NOT updated.  It still displays the stale count of 1568.
* The count for the "source:type" group also was correctly updated.

screen shot 2017-10-12 at 3 10 40 pm

nickwallen: I am also noticing another issue with the group header counts that may or may not be related.  

In this example, I am grouping by 'source:type' then 'host'.  Since 'host' is specific to bro, there should be no alerts under the "snort" group.  If I expand the Snort group, I correctly see no alerts. 

The Snort group count is reported as 17,300 whereas I would expect that to be 0.  The count of 17,300 is tallying how many Snort alerts there are.  But adding a sub-group like 'host' can filter what the user will see when drilling down in the alerts.  

IMHO We should always expect the group count to reflect how many alerts the user will see when drilling down.  

screen shot 2017-10-12 at 3 41 29 pm

iraghumitra: @nickwallen  Again the problem is with how we are handling null values in the rest api.  The host name for all the alerts is emty so we are not seeing any alerts there.

nickwallen: @iraghumitra That doesn't explain this one, right?  https://github.com/apache/metron/pull/768#discussion_r144387064

merrimanr: I believe this issue is similar to https://github.com/apache/metron/pull/768#issuecomment-335281376 and should be resolved in the same follow on PR.  Do you agree @nickwallen?

nickwallen: Yes, that makes sense to me. @merrimanr

nickwallen:
@iraghumitra In case you missed it, there are some new merge conflicts based on one of your other PRs being merged into master today.

PR #769, opened 2017-09-21
METRON-1198: Pycapa - No such configuration property: "sasl.kerberos.principal"

nickwallen:
When running Pycapa in a Kerberized environment, but without a version of Librdkafka built with SASL support, it can produce error messages that look-like the following.

```
KafkaError{code=_INVALID_ARG,val=-186,str="No such configuration property: "sasl.kerberos.principal""}
```

This can happen when a user accidentally installs multiple version of Librdkafka and the version that the Python interpreter links to is the one without SASL support.  I am going to update the README to doc this specific condition.

### DIFF
--- a/metron-sensors/pycapa/README.md
+++ b/metron-sensors/pycapa/README.md
@@ -280,10 +280,20 @@ The probe can be used in a Kerberized environment.  Follow these additional step
 FAQs
 ====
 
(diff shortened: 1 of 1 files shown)


justinleet:
+1 by inspection. Thanks!

PR #779, opened 2017-09-28
METRON-1218: Metron REST should return better error messages

merrimanr:
## Contributor Comments
This PR improves the error handling in the REST app and adds logging to failed searches in the ElasticsearchDao.  This can be tested in full dev by executing a search against REST with a value for "from" that is larger than the index.  For example:
```
{
  "from": 1000000000,
  "indices": [
    "bro"
  ],
  "query": "*",
  "size": 0
}
```

Before it would just return:
```
{
  "responseCode": 500,
  "message": "Could not execute search",
  "fullMessage": "Could not execute search"
}
```

but after this change it should instead include the root cause:
```
{
  "responseCode": 500,
  "message": "Could not execute search",
  "fullMessage": "Result window is too large, from + size must be less than or equal to: [10000] but was [100000005]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level parameter."
}
```

There should also be the full stacktrace in /var/log/metron/metron-rest.log.  

This change should improve all error messages that are returned from REST and are not limited to this case.  I also added a unit test for the REST exception handler class.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
@@ -35,7 +35,7 @@ public class RestExceptionHandler extends ResponseEntityExceptionHandler {
   @ResponseBody
   ResponseEntity<?> handleControllerException(HttpServletRequest request, Throwable ex) {
     HttpStatus status = getStatus(request);
-    return new ResponseEntity<>(new RestError(status.value(), ex.getMessage(), ex.getMessage()), status);
+    return new ResponseEntity<>(new RestError(status.value(), ex.getMessage(), getFullMessage(ex)), status);
   }
 
   private HttpStatus getStatus(HttpServletRequest request) {
@@ -44,5 +44,15 @@ public class RestExceptionHandler extends ResponseEntityExceptionHandler {
       return HttpStatus.INTERNAL_SERVER_ERROR;
     }
     return HttpStatus.valueOf(statusCode);
+  }
+
+  private String getFullMessage(Throwable ex) {
+    String fullMessage = ex.getMessage();
+    Throwable cause = ex.getCause();
+    while(cause != null) {
+      fullMessage = cause.getMessage();
+      cause = cause.getCause();
+    }
+    return fullMessage;
   }
 }
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
@@ -36,7 +36,7 @@ public class RestExceptionHandler extends ResponseEntityExceptionHandler {
   @ResponseBody
   ResponseEntity<?> handleControllerException(HttpServletRequest request, Throwable ex) {
     HttpStatus status = getStatus(request);
-    return new ResponseEntity<>(new RestError(status.value(), ex.getMessage(), ExceptionUtils.getStackTrace(ex)), status);
+    return new ResponseEntity<>(new RestError(status.value(), ex.getMessage(), ExceptionUtils.getRootCauseMessage(ex)), status);
   }
 
   private HttpStatus getStatus(HttpServletRequest request) {
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
@@ -17,6 +17,7 @@
  */
 package org.apache.metron.rest.controller;
 
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.metron.rest.RestException;
 import org.apache.metron.rest.model.RestError;
 import org.springframework.http.HttpStatus;
@@ -35,7 +36,7 @@ public class RestExceptionHandler extends ResponseEntityExceptionHandler {
   @ResponseBody
   ResponseEntity<?> handleControllerException(HttpServletRequest request, Throwable ex) {
     HttpStatus status = getStatus(request);
-    return new ResponseEntity<>(new RestError(status.value(), ex.getMessage(), getFullMessage(ex)), status);
+    return new ResponseEntity<>(new RestError(status.value(), ex.getMessage(), ExceptionUtils.getStackTrace(ex)), status);
   }
 
   private HttpStatus getStatus(HttpServletRequest request) {
@@ -44,15 +45,5 @@ public class RestExceptionHandler extends ResponseEntityExceptionHandler {
       return HttpStatus.INTERNAL_SERVER_ERROR;
     }
     return HttpStatus.valueOf(statusCode);
-  }
-
-  private String getFullMessage(Throwable ex) {
-    String fullMessage = ex.getMessage();
-    Throwable cause = ex.getCause();
-    while(cause != null) {
-      fullMessage = cause.getMessage();
-      cause = cause.getCause();
-    }
-    return fullMessage;
   }
 }
--- a/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/RestExceptionHandlerTest.java
+++ b/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/RestExceptionHandlerTest.java
@@ -46,7 +46,7 @@ public class RestExceptionHandlerTest {
     assertEquals(HttpStatus.UNAUTHORIZED, responseEntity.getStatusCode());
     RestError actualRestError = (RestError) responseEntity.getBody();
     assertEquals("unauthorized", actualRestError.getMessage());
-    assertEquals(ExceptionUtils.getFullStackTrace(throwable), actualRestError.getFullMessage());
+    assertEquals("RuntimeException: unauthorized", actualRestError.getFullMessage());
     assertEquals(401, actualRestError.getResponseCode());
   }
 
@@ -68,7 +68,7 @@ public class RestExceptionHandlerTest {
     assertEquals(HttpStatus.INTERNAL_SERVER_ERROR, responseEntity.getStatusCode());
     RestError actualRestError = (RestError) responseEntity.getBody();
     assertEquals("some error", actualRestError.getMessage());
-    assertEquals(ExceptionUtils.getFullStackTrace(throwable), actualRestError.getFullMessage());
+    assertEquals("RuntimeException: some root cause", actualRestError.getFullMessage());
     assertEquals(500, actualRestError.getResponseCode());
   }
 }
--- a/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/RestExceptionHandlerTest.java
+++ b/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/RestExceptionHandlerTest.java
@@ -1,0 +1,81 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements.  See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership.  The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with the License.  You may obtain
+ * a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.metron.rest.controller;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import javax.servlet.http.HttpServletRequest;
+import org.apache.metron.rest.model.RestError;
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+public class RestExceptionHandlerTest {
+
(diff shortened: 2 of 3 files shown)


cestella:
+1 this one bit me today and this fixes it; great job!

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
```diff
@@ -45,4 +45,14 @@ private HttpStatus getStatus(HttpServletRequest request) {
...
   }
+
+  private String getFullMessage(Throwable ex) {
+    String fullMessage = ex.getMessage();
+    Throwable cause = ex.getCause();
+    while(cause != null) {
+      fullMessage = cause.getMessage();
+      cause = cause.getCause();
+    }
+    return fullMessage;
```
cestella: Could we also get the stacktrace?  Maybe a call to `ExceptionUtils.getStackTrace(ex)` appended?

ottobackwards: Maybe be a security risk to show stack in ui or rest errors?  Could we control it by profile or something?

simonellistonball: Perhaps a good alternative would be to show the error to the endpoint (the UI) and at least log the exception in the metron-rest.log

merrimanr: The stack trace is not shown in the UI, only the root cause.  If a user is not authenticated, they wouldn't be see this error anyways, they would get a 401.

@simonellistonball are you suggesting we move the logging call to the exception handler instead of the offending class?  In this case it's very likely an error would get logged twice.

merrimanr: @cestella I initially tried to include the whole stack trace in the response but I couldn't find a good way to format it.  Multi-line statements just don't work well in JSON so I settled on the root cause because it's usually the most important piece of information anyways.  The whole stack trace is logged, just not returned in the JSON response.  Any ideas on how to do this better?

ottobackwards: I think the way you describe is the correct way

simonellistonball: @merrimanr I'm suggesting we log it somewhere, right now there is no apparent indication in the rest logs, so people end up hunting. When I hit the missing alter: nested template the other day, it took a PCAP of the transport protocol to find out what the actual root error was.

merrimanr: @simonellistonball an error was not being logged before but is now in the PR.  Check out the ElasticsearchDao change.

cestella: Hmm, in the situation that I was in last night while debugging, I was using this patch and the exception was not logged in metron-rest. I was hunting the same bug as Simon.  I ended up adjusting this to include the stack trace and that's how I eventually figured it out.  I'm not entirely certain WHY the error logging wasn't showing up.

Since this has happened to 2 of us now, could you validate that when you try to use the alerts-ui and use a sensor that does not have the `alert` nested template (any new sensor), you get an exception in the log?  It should be something like:
```
Caused by: org.apache.metron.indexing.dao.search.InvalidSearchException: Could not execute search
        at org.apache.metron.elasticsearch.dao.ElasticsearchDao.search(ElasticsearchDao.java:170)
        at org.apache.metron.elasticsearch.dao.ElasticsearchMetaAlertDao.search(ElasticsearchMetaAlertDao.java:190)
        at org.apache.metron.rest.service.impl.SearchServiceImpl.search(SearchServiceImpl.java:53)
        ... 88 more
Caused by: Failed to execute phase [query_fetch], all shards failed; shardFailures {[S6cJSn5nSmOpd7Lm2_OF9g][cowrie_index_2017.09.29.05][0]: RemoteTransportException[[node1][192.168.66.121:9300][indices:data/read/search[phase/query+fetch]]]; nested: SearchParseException[failed to parse search source [{\"from\":0,\"size\":25,\"query\":{\"constant_score\":{\"filter\":{\"bool\":{\"should\":[{\"query_string\":{\"query\":\"*\"}},{\"bool\":{\"must\":[{\"term\":{\"status\":\"active\"}},{\"nested\":{\"query\":{\"query_string\":{\"query\":\"*\"}},\"path\":\"alert\"}}]}}]}}}},\"_source\":{\"includes\":[],\"excludes\":[]},\"sort\":[{\"timestamp\":{\"order\":\"desc\"}}],\"track_scores\":true}]]; nested: QueryParsingException[[nested] failed to find nested object under path [alert]]; }
        at org.elasticsearch.action.search.AbstractSearchAsyncAction.onFirstPhaseResult(AbstractSearchAsyncAction.java:206)
        at org.elasticsearch.action.search.AbstractSearchAsyncAction$1.onFailure(AbstractSearchAsyncAction.java:152)
        at org.elasticsearch.action.ActionListenerResponseHandler.handleException(ActionListenerResponseHandler.java:46)
        at org.elasticsearch.transport.TransportService$DirectResponseChannel.processException(TransportService.java:855)
        at org.elasticsearch.transport.TransportService$DirectResponseChannel.sendResponse(TransportService.java:833)
        at org.elasticsearch.transport.TransportService$4.onFailure(TransportService.java:387)
        at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:39)
```
If you can get it to work, then I'm +1.  I absolutely couldn't get it to work with this PR, but I might've been doing something weird and I worry there's something weird with the logging level (which I think is at INFO) on metron-rest by default.

I still, personally, would rather see a stack trace in the error in the UI (since I'd prefer to *get* a stack trace from users for diagnosing issues and that's where they'll be looking).  I take @ottobackwards 's point about security, but I still think it's valid and valuable.  For the moment, I'd just append the multiline string and we can sort it out when debugging.  Mind you, this would be in addition to in the rest logs.

I guess the net-net of it is that I want it logged and validated at LEAST one place and I'd be ok if it were 2 (one to the user and one to the system logs).

cestella: Also, yes, I realize that the stack trace shown is right in the path of your `LOG.error` call, but I absolutely did *not* see anything in the log.

merrimanr: The log level is indeed INFO and it should not be.  That regression was introduced when we added ${HBASE_HOME}/conf to the REST classpath because that directory contains a log4j.properties file with logging set to INFO.  I'm planning on fixing that in a separate PR.  I'm not sure that's the cause here though because I'm able to produce an error in the logs with the same type of error (InvalidSearchException).

Are you sure you have the code in this PR deployed properly?  Your stack trace says the exception was thrown at line 170 but in this PR the exception is thrown at line 176.

ottobackwards: I'm ok with having it the UI, but it must be configurable or governed by the production profile.
You can have the customer switch and reproduce to get the screen shot ( although, when would you NOT want the whole log in practice? ).

cestella: Yeah it’s could be a bad merge.  If you saw that in the log, then I’m ok with this.  My +1 stands.  Great job.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
```diff
@@ -45,4 +45,14 @@ private HttpStatus getStatus(HttpServletRequest request) {
     }
     return HttpStatus.valueOf(statusCode);
   }
+
+  private String getFullMessage(Throwable ex) {
+    String fullMessage = ex.getMessage();
+    Throwable cause = ex.getCause();
```
ottobackwards: This is a possible recursion point, where  ex.getCause() == ex .
Can we use the commons ExceptionUtils.getRootCause()?  If not we need to guard against this.

cestella: Yeah, I think this method is just `ExceptionUtils.getRootCause.getMessage()`, isn't it?

ottobackwards: I think so.  The get cause recursion is an oldie but a goodie.

ottobackwards: This is a blocker imo.

merrimanr:
The latest commit replaces the getFullMessage call with ExceptionUtils.getStackTrace.  Hopefully that addresses the recursion issue.  As for security @ottobackwards, I'm not sure what the requirements are.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
```diff
@@ -35,7 +36,7 @@
   @ResponseBody
   ResponseEntity<?> handleControllerException(HttpServletRequest request, Throwable ex) {
     HttpStatus status = getStatus(request);
-    return new ResponseEntity<>(new RestError(status.value(), ex.getMessage(), getFullMessage(ex)), status);
+    return new ResponseEntity<>(new RestError(status.value(), ex.getMessage(), ExceptionUtils.getStackTrace(ex)), status);
```
ottobackwards: Why change to stack trace from message?
ExceptionUtils.getRootCauseMessage() ?

merrimanr: Because @cestella requested it earlier.

ottobackwards: ok, but I thought you wanted to show the root cause to the user, **and** have the stack trace that casey wanted.  Not just the stack trace.  What is the user going to see now?

merrimanr: I would prefer just root cause but I had the impression the full stack trace was wanted.  Doesn't make that much difference to me so we just need to agree on what it should be.

cestella: Well, the stack trace is generally more helpful in debugging than just the root cause, so yeah, I kinda wanted the stack trace along with the root cause.  I'm willing to be argued with, but as a dev, I'd like to know line numbers of failures along with why.

merrimanr: Are you ok with having the full stack trace returned @ottobackwards?

ottobackwards: My preference would be -> root cause only.  Stack trace if profile is not production.  Stack trace to log.

I will not hold this back for that, but if this gets audited, that is what is going to have to be done later.

ottobackwards: I mean root cause ALWAYS, stack trace if profile allows

merrimanr: My vote would be root cause in the REST response, full stack trace in the logs.  I would also vote to skip the separate profile for stack trace.  Turning that on and restarting REST would be more effort than just looking at the log.  Doubt anyone would use it.

I think this is a minor issue.  I would be good with either.

ottobackwards: I agree with you @merrimanr

cestella: I think I can live with just the root cause, but I'd like to know how exposing the stack trace is a security issue first.  Can you clarify the reasoning behind it, @ottobackwards ?  It's not that I disbelieve you, but I'd like to better understand because we currently have stack traces in logs all over the place.

ottobackwards: https://software-security.sans.org/blog/2010/08/11/security-misconfigurations-java-webxml-files

ottobackwards: https://www.owasp.org/index.php/Error_Handling#Vulnerable_Patterns_for_Error_Handling

ottobackwards: http://www.creativebloq.com/web-design/website-security-tips-protect-your-site-7122853

etc etc.

ottobackwards: It is not in logs that is the problem, it is if it is in the UI.  I think that it would apply to rest too... but I am not sure on that

merrimanr:
The most recent commit changes fullMessage to use ExceptionUtils.getRootCauseMessage.

ottobackwards:
+1

cestella:
+1, great job

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/controller/RestExceptionHandler.java
cestella: That's fine; I'm with it :)  Thanks for the clarity, @ottobackwards

PR #788, opened 2017-10-05
METRON-1223: Support for adding comments to alerts

iraghumitra:
## Contributor Comments
This PR adds support for adding arbitrary text notes to an alert. This can be achieved by using METRON-1051.The comments along with comment contain username and timestamp of the comment

image

image

## Testing

- Login to alerts-ui
- Select an alert this will open the alerts detail pane
- Select the comments icon on the pane
- Add comment in the textarea and click on 'ADD COMMENT'
- If you hover over a comment the delete  comment icon would be avialble to delete the comment
- E2E tests are available to test the add, delete comments

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.ts
```diff
@@ -133,6 +173,40 @@ export class AlertDetailsComponent implements OnInit {
...
+    if (this.alertCommentsWrapper[index].alertComment.comment.length > 25 ) {
+      commentText += ' \'' + this.alertCommentsWrapper[index].alertComment.comment.substr(0, 25) + '...\'';
+    } else {
+      commentText += ' \'' + this.alertCommentsWrapper[index].alertComment.comment + '\'';
+    }
+
+    this.metronDialogBox.showConfirmationMessage(commentText).subscribe(response => {
+      if (response) {
+        this.alertCommentsWrapper.splice(index, 1);
+        this.patchAlert(new Patch('add', '/comments', this.alertCommentsWrapper.map(alertsWrapper => alertsWrapper.alertComment)));
```
justinleet: I'm concerned that this will actually cause problems when searching comments in Elasticsearch.  Keep in mind this is theoretical, since I'm not at all an ES expert.

Without comments this being a [Nested Object](https://www.elastic.co/guide/en/elasticsearch/guide/current/nested-objects.html), this ends up being subject to cross object mapping (see  [Object Arrays](https://www.elastic.co/guide/en/elasticsearch/guide/current/complex-core-fields.html#object-arrays)).  In fact, the actual example ES gives is on comment fields. To the best of my knowledge, this only becomes an issue when you want to do queries involving the actual comment fields (e.g. give me all of a particular user's comments between two times).  For the case of just saving off and reading back comments, it should be fine; I'm worried it'll bite us later and it won't at all be obvious why.

We might want/need to actually make this a real nested field in ES, which unfortunately probably involves backend work and some maintenance of mappings / templates to ensure the field exists (which is a problem meta alerts also contend with).

simonellistonball: In the short term, we could just consider this an unanalysed field, and not have comments searchable, no? 

In terms of doing it properly, we already have the nested field issue with the meta-alerts concept, so reusing the same coping mechanisms seems viable.

justinleet: @simonellistonball I think an `"index" : "no"` field would work? I think?

I think that still involves mapping updates since that's not the default and would need to be on every sensor (or we'll end up with analyzed or raw or whatever the default is fields). Which loops back around to backend changes. My concern with not doing a more complete solution is you'd have to reindex anything with a non null comments field, which is potentially a pain (and I don't know how much work that is, so it's also possible it's not at all a pain).

Using the meta alert coping mechanisms is still backend change, though.  That primarily gets handled by the DAOs.

justinleet: Thinking through this a bit more, my major concern of making it nonsearchable is that if we ever want to make those comments searchable, we'd probably have to eat a reindex of every sensor. Existing mappings can't be changed, you have to create a new index and migrate the data. That sounds awful. And I definitely can see cases where you'd want to search those comments (give me every comment that person X made in the last month.  Give me every comment by person X regarding sensor Y. and so on).

If we make the field nested from the beginning, that does mean that the current patch calls need to be replace calls where the entire doc is constructed by the UI.  Alternatively, we eat the backend work and make a comments endpoint that accepts a GUID and has add comment / delete comment and the backend constructs the new doc and ships it out.

simonellistonball: @justinleet makes sense, the goal is to get something in fast and iterate, but it sounds like getting the field nested is not going to be much more work than the alerts work we already have to do, so can piggyback on that and minimise the upgrade effort. There is I believe some separate work going on making reindexing easier, which may help here (though we should also ensure that that work is DAO aware when reindexing commented stuff from HDFS)

james-sirota: What I would recommend is just to document around potential issues.  I think what we have here is fine

justinleet: I'm good with that, as long as we document it

iraghumitra: @justinleet I captured this limitation in the caveats section  of the alerts readme

PR #793, opened 2017-10-09
METRON-1226 Searching Can Errantly Query the Wrong Indices

nickwallen:
The ElasticsearchDAO can errantly query the wrong indices under certain conditions. This can occur when there are at least two different sensors that have been configured with a similar name prefix.

For example, consider the case when the user has created two different sensors; "snort" and "snort_v2".  When the user attempts to query the "snort" index in Elasticsearch, the query generated by the ElasticsearchDAO will hit both of these indices which is unexpected.

## Changes

Previously the `ElasticsearchDao` would append a simple `*` to the index name.  If the user had created a sensor "test" and another sensor "test_v2", then the wildcard logic would result in a pattern of `test*` when the user searches against the first sensor.  This pattern would unexpectedly also include data from the "test_v2" sensor.

The code was changed so that after wildcarding the index, it results in a pattern like `test_index*`.  Since "_index" is automatically appended by Metron, it can be used as a delimiter.  This will ensure that only the proper indices are searched.

**UPDATE:**  The name of the meta-alerts index was also changed from `metaalerts` to `metaalert_index`.  The new wildcard strategy would not work with the previous meta-alerts naming convention.  The simple way to address this was to rename the meta-alerts index.  This new name also more closely matches the way we name other indices.

## Testing

To manually test this specific condition, do the following in a Metron environment.
1. Configure a sensor called "test".
2. Allow telemetry to be consumed and indexed in Elasticsearch.
3. Create a second sensor called "test_v2".
4. Allow telemetry to be consumed and indexed in Elasticsearch.
5. Pass in a simple query such as the following.  Ensure that only data from the first sensor "test" is returned.  No data should be returned from "test_v2".
    ```
      {
       "indices": ["test"],
       "query": "*",
       "from": 4,
       "size": 3,
      "sort": [
        {
          "field": "timestamp",
         "sortOrder": "desc"
        }
      }
    ```
**UPDATE:**  The Alerts UI should also be tested to ensure there are no regressions.  I was able to perform queries and use the Alerts UI as before.

## Pull Request Checklist


cestella:
Does this cause any issues in the alerts UI?

nickwallen:
@cestella Not that I am aware of. I launched Full Dev and was able to submit queries and use the Alerts UI as before.  I will add that to the "Test" section for completeness.

cestella:
cool, just making sure :)

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
```diff
@@ -154,9 +134,9 @@ protected SearchResponse search(SearchRequest searchRequest, QueryBuilder queryB
             .size(searchRequest.getSize())
             .from(searchRequest.getFrom())
             .query(queryBuilder)
-
             .trackScores(true);
-    searchRequest.getSort().forEach(sortField -> searchSourceBuilder.sort(sortField.getField(), getElasticsearchSortOrder(sortField.getSortOrder())));Optional<List<String>> fields = searchRequest.getFields();
+    searchRequest.getSort().forEach(sortField -> searchSourceBuilder.sort(sortField.getField(), getElasticsearchSortOrder(sortField.getSortOrder())));
```
cestella: haha, thanks for cleaning this up. :)

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/utils/ElasticsearchUtils.java
```diff
@@ -45,15 +49,30 @@ public static SimpleDateFormat getIndexFormat(Map<String, Object> globalConfig)
...
+    indexName = indexName + getIndexDelimiter() + "_" + indexPostfix;
     return indexName;
   }
 
+  /**
+   * Returns the delimiter that is appended to the user-defined index name to separate
+   * the index's date postfix.  For example, if the user-defined index name is 'bro'
+   * and the delimiter is '_index_', then one likely index name is 'bro_index_2017.10.03.19'.
+   */
+  public static String getIndexDelimiter() {
```
cestella: I'm more curious than opposed to this, but is there a strong reason why this isn't just a `public static final String`?

nickwallen: Honestly, I have no idea.  I went down a dangerous rabbit hole making tons of changes, then backed myself out to just what is in this PR.  It might have been for changes that I backed out.

Making it a `static final String` certainly would be more expected.  I'll fix it.

mmiklavc:
This is looking pretty good by inspection. Good minor refactorings and javadoc additions as well. I'd like to run it up in full dev when I get a moment.

nickwallen:
@cestella @mmiklavc How is this looking?

cestella:
I'm +1 by inspection but will hold it until @mmiklavc gives his.

mmiklavc:
+1 by inspection

PR #799, opened 2017-10-13
METRON-1249 Improve Metron MPack service checks

nickwallen:
This PR enhances the Metron 'Service Check' functionality in the MPack.  The Service Check is an easy way for a user to know if their Metron cluster is healthy.  The Service Check can be run manually from within Ambari.  It is also automatically executed after kerberization.

In the current version of the Service Check, healthy means the Parser and Indexing topologies are running.  This PR enhances that to validate all of the install actions that occur across each of the Metron services. These checks include the following.
* Kafka topics, user permissions, group permissions
* HBase tables, column families, and user permissions
* HDFS resources like the grok patterns and geo database
* Ensures all Metron topologies are running
* Ensures the web-based resources are responding

I added considerable logging so that if a check does fail, a user will have a reasonable chance to understand why.  Ambari doesn't give me an easy way to tell a user "hey, this is the problem!", so the user still has to go through the output of the Service Check in the Operations Panel to know why the Service Check failed.

## Testing

* I manually tested each of the checks by, for example, deleting a Kafka topic then running the Service Check.  This can be repeated for all of the different types of checks that I outlined above.
* I tested the Service Check on a fresh deployment of Full Dev.
* I then kerberized Full Dev and again validated the Service Check

## Pull Request Checklist

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/enrichment_commands.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/enrichment_commands.py
@@ -18,6 +18,7 @@ limitations under the License.
 import os
 import time
 from datetime import datetime
+from resource_management.core.exceptions import Fail
 from resource_management.core.logger import Logger
 from resource_management.core.resources.system import Execute, File
 
@@ -46,6 +47,12 @@ class EnrichmentCommands:
         self.__hbase_configured = os.path.isfile(self.__params.enrichment_hbase_configured_flag_file)
         self.__hbase_acl_configured = os.path.isfile(self.__params.enrichment_hbase_acl_configured_flag_file)
         self.__geo_configured = os.path.isfile(self.__params.enrichment_geo_configured_flag_file)
+
+    def __get_topics(self):
+        return [self.__enrichment_topic, self.__params.enrichment_error_topic]
+
+    def __get_kafka_acl_groups(self):
+        return [self.__enrichment_topic]
 
     def is_kafka_configured(self):
         return self.__kafka_configured
@@ -105,15 +112,15 @@ class EnrichmentCommands:
     def init_kafka_topics(self):
         Logger.info('Creating Kafka topics for enrichment')
         # All errors go to indexing topics, so create it here if it's not already
-        metron_service.init_kafka_topics(self.__params, [self.__enrichment_topic, self.__params.enrichment_error_topic])
+        metron_service.init_kafka_topics(self.__params, self.__get_topics())
         self.set_kafka_configured()
 
     def init_kafka_acls(self):
         Logger.info('Creating Kafka ACls for enrichment')
+        metron_service.init_kafka_acls(self.__params, self.__get_topics())
+
         # Enrichment topic names matches group
-        metron_service.init_kafka_acls(self.__params,
-                                       [self.__enrichment_topic, self.__params.enrichment_error_topic],
-                                       [self.__enrichment_topic])
+        metron_service.init_kafka_acl_groups(self.__params, self.__get_kafka_acl_groups())
 
         self.set_kafka_acl_configured()
 
@@ -182,6 +189,7 @@ class EnrichmentCommands:
                   self.__params.hbase_keytab_path,
                   self.__params.hbase_principal_name,
                   execute_user=self.__params.hbase_user)
+
         cmd = "echo \"create '{0}','{1}'\" | hbase shell -n"
         add_enrichment_cmd = cmd.format(self.__params.enrichment_hbase_table, self.__params.enrichment_hbase_cf)
         Execute(add_enrichment_cmd,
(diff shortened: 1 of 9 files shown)


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/enrichment_commands.py
```diff
@@ -232,3 +241,48 @@ def set_hbase_acls(self):
...
         Logger.info("Done setting HBase ACLs")
         self.set_hbase_acl_configured()
+
+    def service_check(self, env):
+        """
+        Performs a service check for Enrichment.
+        :param env: Environment
+        """
+        Logger.info("Checking for Geo database")
+        metron_service.check_hdfs_file_exists(self.__params, self.__params.geoip_hdfs_dir + "/GeoLite2-City.mmdb.gz")
```
cestella: You know, honestly, the better approach here, unfortunately, is to pull the config from the global config and ensure that the file exists.  What if the user renames the GeoLite database to something other than `GeoLite2-City.mmdb.gz`?  Can we make that a follow-on JIRA at least?

nickwallen: I agree.  I don't like hard-coding "GeoLite2-City.mmdb.gz" here.  I just didn't see anywhere in the Mpack where we had that value parameterized already.

Where in the code do we have it parameterized?  I'd love to fix this.

nickwallen: Ok, I see it now under a global properties key "geo.hdfs.file".  There is nothing in Ambari MPack for it, which might complicate using it.  I am just thinking through, if we would need to first introduce it as an Ambari-managed configuration value.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_service.py
```diff
@@ -201,15 +200,123 @@ def init_kafka_topics(params, topics):
...
+          execute_user=params.metron_user)
+
+  cmd = """{0}/kafka-topics.sh \
+    --zookeeper {1} \
+    --list | \
+    awk 'BEGIN {{cnt=0;}} /{2}/ {{cnt++}} END {{if (cnt > 0) {{exit 0}} else {{exit 1}}}}'"""
+  for topic in topics:
+    Logger.info("Checking existence of Kafka topic '{0}'".format(topic))
+    try:
+      Execute(
```
cestella: Any chance we could make a function rather than cutting and pasting?  Something like:
```
def exec( cmd, user_as, fail_msg ) :
try:
  Execute(cmd,
                tries=3,
                try_sleep=5,
                logoutput=True,
                path='/usr/sbin:/sbin:/usr/local/bin:/bin:/usr/bin',
                user=user_as)
    except:
        raise Fail(fail_msg)
```

cestella:
I really like this.  Modulo a couple of very minor nits, I'm +1 by inspection.

cestella:
Yes, that's right.  In my opinion, though, we should probably stop managing parsers in ambari and just focus on the management UI to have one place to manage all parsers.

ottobackwards:
untangling ambari and zookeeper would be required.  If we want ( as I think we need to ) to be able to manage the parsers from the ui / rest, but also have the ambari service management still work ( restart all effected services ).  I'm not sure if is more than having ambari read and write to zookeeper for the all_parsers list or not

ottobackwards:
But I am taking what you are saying about managing parsers to mean 'managing configured sensors is not done in ambari, but managing the services is'....

cestella:
Really, what I'd like to see is a status on the parsers in the management UI that indicates that those parsers should be running all the time and a REST call to validate that they are running or not.  I'd then expect ambari to call the REST call and fail the service check if any of the installed sensors that are marked as running aren't running.  

What we have now is that list in Ambari and ambari managing to start them and stop them.  I'd prefer to delegate teh starting and stopping of sensors to the management UI and JUST have ambari interact with the REST API to indicate whether the current state is nominal.  That, however, is really more of a topic for a discuss thread, though.

nickwallen:
@ottobackwards Do you have any concerns that need addressed in this PR?

ottobackwards:
@nickwallen no, the problem exists before already

nickwallen:
@ottobackwards Thanks.  Your points definitely warrant a discussion.  Just wanted to make sure we're green lighted on this PR.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/enrichment_commands.py
cestella: Yeah, we just have the HDFS directory, which is what you're using.  For this PR, I think it's good.  I'd hate to make the perfect the enemy of the good.

Going forward, what would be cool is if we could execute a stellar script via the REPL and if it fails, fail the service check.  Since the REPL can interact with global config parameters and has the capability to validate HDFS, that would be a clean way to do this.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_service.py
nickwallen: I refactored this a bit and added some pydocs.

PR #814, opened 2017-10-24
METRON-1277 Add match statement to Stellar language 

ottobackwards:
as discussed on the dev list:  [Stellar support for switch/case style conditionals](http://mail-archives.apache.org/mod_mbox/metron-dev/201710.mbox/%3cCAPpQHK0J4bpVMpPZnQpZeu2JtbuxHW8YHeNG0LMbbL_vw+x3kQ@mail.gmail.com%3e)

This PR adds 'match' capability to the stellar language.  This is very similar to the Scala match feature.

FROM THE README: 
### Stellar Language Match Expression
Stellar provides the capability to write match expressions, which are similar to switch statements commonly found in c like languages, but more like
Scala's match.

The syntax is:
* `match{ logical_expression1 => evaluation expression1, logical_expression2 => evaluation_expression2, default => default_expression}` 

Where:

* `logical_expression` is a Stellar expression that evaluates to true or false.  For instance `var > 0` or `var > 0 AND var2 == 'foo'` or `IF ... THEN ... ELSE` 
* `evaluation_expression` is any Stellar Expression
* `default` is a required default return value, should no logical expression match

> default is required 

> Lambda expressions are supported, but they must be no argument lambdas such as `() -> STATEMENT`

* Only the first clause that evaluates to true will be executed.

## Review items

* more test cases
* help with issue with MAP() function
* correctness of short circuit / grammar

This PR does not
* add the aliasing of long variables
* support variable arg lambda

## Testing

from the Stellar shell execute various statements involving match such as: 

```bash
ttofowler@Winterfell [13:01:16] [~/src/apache/forks/metron/metron-stellar/stellar-common] [stellar_match]
-> % mvn exec:java \
   -Dexec.mainClass="org.apache.metron.stellar.common.shell.StellarShell" -Dexec.args="-l src/test/resources/log4j.properties"
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Building stellar-common 0.4.1
[INFO] ------------------------------------------------------------------------
[INFO]
[INFO] --- exec-maven-plugin:1.6.0:java (default-cli) @ stellar-common ---
Stellar, Go!
[Stellar]>>> foo := 1
[Stellar]>>> match{foo == 1 : TO_UPPER('ok')}
OK
[Stellar]>>>
```

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```


metron-stellar/stellar-common/README.md
```diff
@@ -100,6 +102,28 @@ In the core language functions, we support basic functional programming primitiv
...
+Stellar provides the capability to write match expressions, which are similar to switch statements commonly found in c like languages, but more like
+Scala's match.
+
+The syntax is:
+* `match{ logical_expression1 : evaluation expression1, logical_expression2 : evaluation_expression2` : A match expression with no default
+* `match{ logical_expression1 : evaluation expression1, logical_expression2 : evaluation_expression2, default : default_expression}` : A match expression with a default expression
+
+Where:
+
+* `logical_expression` is a Stellar expression that evaluates to true or false.  For instance `var > 0` or `var > 0 AND var2 == 'foo'`
```
jjmeyer0: It seems as if the below stellar always throws an exception. Is it by design that `default` clauses are always evaluated? If so can you mention that here?

```
foo := 500
match{ foo < 100 : THROW('oops'), foo > 200 : 'ok', default : THROW('exception thrown') }
```

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/StellarCompiler.java
```diff
@@ -105,59 +107,77 @@ public Object apply(ExpressionState state) {
...
-              if(token.getUnderlyingType() == EndConditional.class) {
+              if (token.getUnderlyingType() == EndConditional.class) {
                 break;
               }
             }
             skipElse = false;
           }
+          if (skipMatchClauses && (token.getUnderlyingType() == MatchClauseEnd.class
+              || token.getUnderlyingType() == MatchClauseCheckExpr.class)) {
+            while (it.hasNext()) {
```
jjmeyer0: This will only skip one match clause after a true one is found. See below test for a stellar statement that fails with a `ParseException`.

```java
@Test
public void moreThanTwoConsecutiveTrueCasesWillResultInFailure() {
    Assert.assertEquals("ok",  run("match{ foo < 100 : THROW('oops'), foo > 200 : 'ok', foo > 300 : 'ok', default : 'works' }",
            new HashMap() {{
                put("foo", 500);
            }}));
}
```

metron-stellar/stellar-common/src/main/antlr4/org/apache/metron/stellar/common/generated/Stellar.g4
```diff
@@ -272,4 +278,30 @@ lambda_variable:
...
+
+match_clause_action :
+  match_clause_action_expr #MatchClauseAction
+  ;
 
+match_clause_action_expr :
+  transformation_expr
+  ;
+
+match_clause_check :
```
jjmeyer0: Is this extra expression needed? Could we get rid of it? Instead of this just have what's below?

```
match_clause_action :
  transformation_expr #MatchClauseAction
  ;

match_clause_check :
  logical_expr #MatchClauseCheckExpr
  ;
```

metron-stellar/stellar-common/README.md
```diff
@@ -100,6 +102,28 @@ In the core language functions, we support basic functional programming primitiv
 * `FILTER` - Filters a list by a predicate in the form of a lambda expression.  For instance `FILTER([ 'foo', 'bar'], (x ) -> x == 'foo' )` returns `[ 'foo' ]`
 * `REDUCE` - Applies a function over a list of input.  For instance `REDUCE([ 1, 2, 3], (sum, x) -> sum + x, 0 )` returns `6`
 
+### Stellar Language Match Expression
+
+Stellar provides the capability to write match expressions, which are similar to switch statements commonly found in c like languages, but more like
+Scala's match.
```
jjmeyer0: I'd be a bit hesitant to say this is similar to Scala's match. Our match statement doesn't really support pattern matching.

jjmeyer0:
Also, I don't think there is an issue with the map function directly. It's actually an issue with the logical_expression `var1` and using/not using a `default` branch. It seems like the stack isn't properly being managed. For example things are being pushed (eg. when a logical_expr is just `var1`) when they shouldn't and popped  when they shouldn't be (eg. when there is no default?). I haven't pin pointed it directly yet, but I hope this helps somewhat. Below are a few example tests that I have created. 

```java
// Here there is a null on the stack
@Test
public void test1FailsWithMoreOnStack() {
    Assert.assertEquals("a",  run("match{ foo : 'a' }",
            new HashMap() {{
                put("foo", true);
            }}));
}
```

// Here it's an empty stack when we expect 'a' i believe.
```java
@Test
public void test1FailsEmptyStack() {
    Assert.assertEquals("a",  run("match{ foo == true : 'a' }",
            new HashMap() {{
                put("foo", true);
            }}));
}
```

```java
// This is *not* properly handled
@Test
public void test1Passes() {
    Assert.assertEquals("a",  run("match{ foo : 'a', default: 'b' }",
            new HashMap() {{
                put("foo", true);
            }}));
}
```

```java
// This is properly handled
@Test
public void test1Passes() {
    Assert.assertEquals("a",  run("match{ foo == true : 'a', default: 'b' }",
            new HashMap() {{
                put("foo", true);
            }}));
}
```

```java
// Example where map works
@Test
public void workingMatchWithMap() {
   Assert.assertEquals(Arrays.asList("OK", "HAHA"),  run("match{ foo > 100 : THROW('oops'), foo > 200 : THROW('oh no'), foo >= 50 : MAP(['ok', 'haha'], (a) -> TO_UPPER(a)), default: 'a' }",
          new HashMap() {{
            put("foo", 50);
          }}));
}
```

metron-stellar/stellar-common/README.md
```diff
@@ -100,6 +102,28 @@ In the core language functions, we support basic functional programming primitiv
...
+
+The syntax is:
+* `match{ logical_expression1 : evaluation expression1, logical_expression2 : evaluation_expression2` : A match expression with no default
```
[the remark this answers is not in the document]
ottobackwards: oops, thanks

metron-stellar/stellar-common/README.md
ottobackwards:
```bash

[Stellar]>>> foo := 500
[Stellar]>>> match{ foo < 100 : THROW('oops'), foo > 200 : 'ok', default : THROW('exception thrown') }
ok
[Stellar]>>>
```

I don't see that.

Also, if you check the last test in TestMatch you will see I have a test for this kind of thing.

metron-stellar/stellar-common/src/main/antlr4/org/apache/metron/stellar/common/generated/Stellar.g4
```diff
@@ -67,6 +67,10 @@ ELSE : 'ELSE' | 'else';
 NULL : 'null' | 'NULL';
 NAN : 'NaN';
 
+AS : 'as' | 'AS';
```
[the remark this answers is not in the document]
ottobackwards: Actually, AS was for the match(longVarName as shortName){ x > 5 : shortName is available }
case, but since I'm not doing that, i should remove it.

metron-stellar/stellar-common/src/main/antlr4/org/apache/metron/stellar/common/generated/Stellar.g4
ottobackwards: I'll try that again.  I don't remember the specific reason why I ended up with it factored like this.  But this is what ended up working ;)

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/MatchTest.java
```diff
@@ -0,0 +1,202 @@
...
+        }}));
+
+  }
+
+  @Test
+  @SuppressWarnings("unchecked")
+  @Ignore
+  public void testMatchMAPEvaluation() {
+
+    // NOTE: THIS IS BROKEN RIGHT NOW.
```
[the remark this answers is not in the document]
ottobackwards: I actually wanted to leave this in so @cestella might help me figure out why it isn't working with map.  How about I leave it for now, and you hold your plus +1 until it is removed.  I would got to commit with this, but I want it there for the review.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/MatchTest.java
```diff
@@ -0,0 +1,202 @@
...
+      caught = true;
+    }
+    Assert.assertTrue(caught);
+
+  }
+
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void testShortCircut() {
```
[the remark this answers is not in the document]
ottobackwards: thanks

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/MatchTest.java
```diff
@@ -0,0 +1,202 @@
...
+              put("foo", 0);
+              put("bar", "yes");
+            }}));
+  }
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void testMatchErrorNoDefault() {
+
+    boolean caught = false;
```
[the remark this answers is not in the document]
ottobackwards: Good catch.  When I was developing, I just had one big test function that I was adding test after test too, so I used this to catch the exception and keep going.  When I did the PR I refactored the tests, but forgot to clean this up.  Thanks!

metron-stellar/stellar-common/README.md
ottobackwards: Do you have any suggestions for a better way to describe it?  I'm not super happy with this either, I'll take suggestions ;)

ottobackwards:
Thanks, I added those tests to the test class, and these fail:
```java
@Test
  public void testVariableOnlyNoDefault() {
    Assert.assertEquals("a",  run("match{ foo : 'a' }",
        new HashMap() {{
          put("foo", true);
        }}));
  }

  @Test
  public void testVariableEqualsCheckNoDefault() {
    Assert.assertEquals("a",  run("match{ foo == true : 'a' }",
        new HashMap() {{
          put("foo", true);
        }}));
  }

  @Test
  public void testVariableOnlyCheckWithDefault() {
    Assert.assertEquals("a",  run("match{ foo : 'a', default: 'b' }",
        new HashMap() {{
          put("foo", true);
        }}));
  }
```

I will take a look.  I feel that I have similar tests to these that are working, so it is strange.

ottobackwards:
@jjmeyer0 thanks for the tests.  The issue is that we call validate and then parse.  When we call validate, we resolve all vars to NULL.  This means that the clauses in these failures are false.  As stated in the readme, something in the match must be true or you get an error.  In this case, we have an issue where validate is forcing us into that condition and we don't have a default, so it is an error.
I'll keep looking.  
If you have any thoughts, or maybe @cestella  does, I'm all ears.

ottobackwards:
So for example:

```java
 /*
          curr is the current value on the stack.  This is the non-deferred actual evaluation for this expression
          and with the current context.
           */
          Token<?> curr = instanceDeque.peek();
          if (curr != null && curr.getValue() != null && curr.getValue() instanceof Boolean
              && ShortCircuitOp.class.isAssignableFrom(token.getUnderlyingType())) {
            //if we have a boolean as the current value and the next non-contextual token is a short circuit op
            //then we need to short circuit possibly
            if (token.getUnderlyingType() == BooleanArg.class) {
```

with this function: 

```java
 @Test
  public void testVariableOnlyCheckWithDefault() {
    Assert.assertEquals("a",  run("match{ foo : 'a', default : 'b' }",
        new HashMap() {{
          put("foo", true);
        }}));
  }
```

When we get to that point, because of the validation of vars to null `curr` IS null.  So even though token IS a short circuit token, we do not process it as we should.

ottobackwards:
Because it is possible in validate() for a boolean to be null instead of true or false.

ottobackwards:
@jjmeyer0 @cestella : Ok.
So, I added some code, where it is possible context wise to guard against the case where there is a single variable as the check, and no default.  This resolves the test issues, save for one, where there is no default and we are using == true.  This is more difficult.

I feel as if all of these things are wrong, because the validation by executing with null variables is itself logically wrong.  We have discussed this before, but this is really evident here.  At least I feel it is.  

I think that we should use compilation instead of validation.  We can decide to make it a separate option when running tests, or to make it the way we do all tests etc.  But if compilation does what I think it does, then I think it is more correct.

I would like to do that, and remove the work around I have introduced here ( basically detecting that we are validating and have a single var that is null because of validation ) since changing validation would be the real complete answer.

ottobackwards:
for example: 
```java
 @Test
  public void testValidation() {
    Object value = run("IF x == null THEN THROW('it cannot be null') ELSE 'it is ok'", new HashMap(){{
      put("x","something");
    }} );
  }
```
This is a valid test, but it fails on validation not parse.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/StellarCompiler.java
ottobackwards: This is another issue with validate

metron-stellar/stellar-common/src/main/antlr4/org/apache/metron/stellar/common/generated/Stellar.g4
ottobackwards: Done

ottobackwards:
OK, the validation issues are a big big deal.  I have changed the match functionality so that they work in that context, supporting guard expressions and requiring the default operation.

1. Require default statement
2. change the action separator from : to => so that...
3. we can use conditional_exp and logical_exp ( can check for EXISTS() with IF etc )

I still have to do the hack for the match{ var1 => 'ok' , default => 'notOk' }
since the lone null as logical doesn't work as stated earlier.

So the new syntax is

match { logical_or_conditional_expr => transform_exp , default => transform_exp }

ottobackwards:
the issue with MAP() is resolved as well

jjmeyer0:
@ottobackwards something is still going on with this. I'm seeing the following behavior:

```bash
[Stellar]>>> foo := 500
[Stellar]>>> match{ foo > 100 => ['oops'], foo > 200 => ['oh no'], foo >= 500 => MAP(['ok', 'haha'], (a) -> TO_UPPER(a)), default => ['a']}
[!] Invalid parse, found [OK, HAHA]
org.apache.metron.stellar.dsl.ParseException: Invalid parse, found [OK, HAHA]
        at org.apache.metron.stellar.common.StellarCompiler$Expression.apply(StellarCompiler.java:210)
        at org.apache.metron.stellar.common.BaseStellarProcessor.parse(BaseStellarProcessor.java:152)
        at org.apache.metron.stellar.common.shell.StellarExecutor.execute(StellarExecutor.java:292)
        at org.apache.metron.stellar.common.shell.StellarShell.handleStellar(StellarShell.java:282)
        at org.apache.metron.stellar.common.shell.StellarShell.execute(StellarShell.java:514)
        at org.jboss.aesh.console.AeshProcess.run(AeshProcess.java:53)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
[Stellar]>>> 
```

metron-stellar/stellar-common/README.md
jjmeyer0: @ottobackwards sure. Right now, I think our match statement syntax is just a nicer syntax for if then else blocks. I would just state something along those lines. I know in the future it will be more.

ottobackwards:
@jjmeyer0 I have fixed the issue.  I had lost short circuit to the end of all clauses, fixing that fixes the case.  I added tests for your issues.

I think I'm going to refactor the tests, but you should be able to check where it is in the mean time.
Thanks for the report!

ottobackwards:
That fix also fixes the issue where IF THEN ELSE in the action clause was not working

ottobackwards:
Bump?

metron-stellar/stellar-common/src/main/antlr4/org/apache/metron/stellar/common/generated/Stellar.g4
```diff
@@ -254,6 +259,7 @@ identifier_operand :
   | EXISTS LPAREN IDENTIFIER RPAREN #ExistsFunc
   | LPAREN conditional_expr RPAREN #condExpr_paren
   | functions #func
+  | DEFAULT #Default
```
jjmeyer0: I just noticed this. Why do we have default here? This would allow default to be used in some unintended places. Is there some reasoning behind this?

ottobackwards:
Great catch!  Moved it out.

jjmeyer0:
@ottobackwards this looks good to me now. I re-built everything, ran the tests, and played around in stellar shell. The only question I'm still not sure about is the one that deals with backwards compatibility. I *believe* with the new keywords (`default`, `match`) we can potentially break expressions with this change. @ottobackwards and @cestella do you have an opinion on this?

ottobackwards:
Great!  I am not sure on that.  @cestella what do you think?

ottobackwards:
Maybe we need a tool that tests stellar scripts or configurations that the user can run and report?
And we can maintain that per release?

ottobackwards:
Ok, I'll update the upgrading doc.

ottobackwards:
Ok, please have a look at the doc. I'll not that we have not added any notes previously for this type of thing.

Also: [METRON-1339: Stellar shellShould have a way to validate deployed functions](https://issues.apache.org/jira/browse/METRON-1339) has been created.
You are welcome to comment. 

I may have a PR for that soon

jjmeyer0:
once travis build completes successfully this looks good +1.

ottobackwards:
@cestella any last words before I pull the trigger on this?  Any comment?

cestella:
Yeah, let me run this up once more in the REPL and take a final look at the docs, but I've been monitoring and I like what I see so far.

ottobackwards:
PR: https://github.com/apache/metron/pull/856  adds capability to use the stellar shell to validate stellar statements at rest out of ZK

cestella:
Alright, +1 otto, you did great work here.  I'm very, very impressed.  Thanks to @jjmeyer0 for the careful review and assistance.  Open source dev at its finest.

cestella:
actually, hold on there, If ound one more bug:
```
[Stellar]>>> match { is_alert == null => null, is_alert => 'alert', default => 'nah' }
[!] null
java.lang.NullPointerException
	at org.apache.metron.stellar.common.StellarCompiler.lambda$exitMatchClauseAction$19(StellarCompiler.java:752)
	at org.apache.metron.stellar.common.StellarCompiler$Expression.apply(StellarCompiler.java:190)
	at org.apache.metron.stellar.common.BaseStellarProcessor.parse(BaseStellarProcessor.java:152)
	at org.apache.metron.stellar.common.shell.StellarExecutor.execute(StellarExecutor.java:292)
	at org.apache.metron.stellar.common.shell.StellarShell.handleStellar(StellarShell.java:282)
	at org.apache.metron.stellar.common.shell.StellarShell.execute(StellarShell.java:514)
	at org.jboss.aesh.console.AeshProcess.run(AeshProcess.java:53)
```

ottobackwards:
Ok,  what was is_alert in this test?

cestella:
sorry, it's not set, so it's null.

ottobackwards:
Ok, I'll check it out after i do my full dev test on validate

ottobackwards:
@cestella  new test and fix is in

cestella:
Ok, my +1 stands

PR #817, opened 2017-10-26
METRON-1283: Install Elasticsearch template as a part of the mpack startup scripts

anandsubbu:
## Contributor Comments

For a Metron multi-node deployment using mpack, the Elasticsearch template is required to be installed manually post-setup. These templates are required for the proper working of, for e.g. the Alerts UI.

In the event that these templates are not installed, and if data is ingested, these would not be shown in the Alerts UI, since there would be missing fields without the template files (E.g. snort alert indices are not displayed in the Alerts UI, since it is missing the "alerts" field from the mapping). In such a case, one needs to install the templates, delete all indices for the given parser and re-ingest data again into the parser for it to appear in the Alerts UI.

Further, the indices from all the parsers will have to be deleted and re-ingested again which could be a tedious job in the event that this step was missed out by chance. I have also seen other ill-effects from having stale indices for parsers that was created before template install.
While documenting the template installation is a good practice, nothing would more failsafe than installing the template as a part of the mpack startup scripts itself.

I have added the ES template install step just before starting the Indexing topology, since I found this to be the closest related step when this could be done. Do let me know if this needs to be moved to a different point in the startup, and I can make the change.

**Testing Done**
* Built mpack with the changes
* Deployed a 12-node Metron setup using the mpack
* Validated that the start Indexing topology step also installs the Elasticsearch template files.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
@@ -80,6 +80,8 @@ class Indexing(Script):
         from params import params
         env.set_params(params)
         self.configure(env)
+        # Install elasticsearch templates
+        self.elasticsearch_template_install(env)
         commands = IndexingCommands(params)
         commands.start_indexing_topology(env)
 
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
@@ -37,8 +37,6 @@ class Indexing(Script):
         from params import params
         env.set_params(params)
         self.install_packages(env)
-        # Install elasticsearch templates
-        self.elasticsearch_template_install(env)
 
     def configure(self, env, upgrade_type=None, config_dir=None):
         from params import params
@@ -84,6 +82,13 @@ class Indexing(Script):
         self.configure(env)
         commands = IndexingCommands(params)
         commands.start_indexing_topology(env)
+        # Install elasticsearch templates
+        try:
+            if not commands.is_elasticsearch_template_installed():
+                self.elasticsearch_template_install(env)
+                commands.set_elasticsearch_template_installed()
+        except:
+            Logger.warning("WARNING: Elasticsearch templates could not be installed. The Elasticsearch service is possibly not running")
 
     def stop(self, env, upgrade_type=None):
         from params import params
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
@@ -37,6 +37,8 @@ class Indexing(Script):
         from params import params
         env.set_params(params)
         self.install_packages(env)
+        # Install elasticsearch templates
+        self.elasticsearch_template_install(env)
 
     def configure(self, env, upgrade_type=None, config_dir=None):
         from params import params
@@ -80,8 +82,6 @@ class Indexing(Script):
         from params import params
         env.set_params(params)
         self.configure(env)
-        # Install elasticsearch templates
-        self.elasticsearch_template_install(env)
         commands = IndexingCommands(params)
         commands.start_indexing_topology(env)
 
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
@@ -80,15 +80,15 @@ class Indexing(Script):
         from params import params
         env.set_params(params)
         self.configure(env)
-        commands = IndexingCommands(params)
-        commands.start_indexing_topology(env)
         # Install elasticsearch templates
         try:
             if not commands.is_elasticsearch_template_installed():
                 self.elasticsearch_template_install(env)
                 commands.set_elasticsearch_template_installed()
         except:
-            Logger.warning("WARNING: Elasticsearch templates could not be installed. The Elasticsearch service is possibly not running")
+            Logger.warning("WARNING: Elasticsearch templates could not be installed. The Elasticsearch service is probably down.")
+        commands = IndexingCommands(params)
+        commands.start_indexing_topology(env)
 
     def stop(self, env, upgrade_type=None):
         from params import params
--- a/metron-platform/metron-elasticsearch/README.md
+++ b/metron-platform/metron-elasticsearch/README.md
@@ -86,8 +86,13 @@ rm ${SENSOR}.template
 
 The stock set of Elasticsearch templates for bro, snort, yaf, error index and meta index are installed automatically during the first time install and startup of Metron Indexing service.
 
-It is possible that Elasticsearch service is not available when the Metron Indexing Service startup, in that case the Elasticsearch template will not be installed. For such a scenario, an Admin can install the template manually from the Ambari UI by following the below flow:
+It is possible that Elasticsearch service is not available when the Metron Indexing Service startup, in that case the Elasticsearch template will not be installed. 
 
+For such a scenario, an Admin can have the template installed in two ways:
+
+_Method 1_ - Manually from the Ambari UI by following the flow:
 Ambari UI -> Services -> Metron -> Service Actions -> Elasticsearch Template Install
+
+_Method 2_ - Stop the Metron Indexing service, and start it again from Ambari UI.
 
 > Note: If you have made any customization to your index templates, then installing Elasticsearch templates afresh will lead to overwriting your existing changes. Please exercise caution.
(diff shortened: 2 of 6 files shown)


ottobackwards:
The templates are installed from ansible right now for the vagrant builds. Wouldn't we need to remove that step?

Have you tested full_dev ?

nickwallen:
Hi @anandsubbu - Just a few issues that came to mind. These may or may not be problems, so I just wanted to open them for discussion.

1. This will reinstall the same templates each time indexing is started.  This may unexpectedly overwrite changes made by the user to their index templates.  

Many of the other install actions use a file as an indicator so that it only performs the action once.  Maybe you could do that too?

2. If Elasticsearch is not running, then the Indexing topology will fail to start.  While normally this is probably OK, it might not be a dependency that we want to introduce.  What if someone has configured their topology to not write to ES and is using an alternative indexing destination?

What if you made it so that the topology could still start, even if Elasticsearch is not available?  Maybe a "best effort" to install the templates, but if it fails, move on.  

3.

anandsubbu:
Hi @ottobackwards , made the change to remove the ES template installation from 'load_web_template' role for vagrant deployments. Fired up full-dev with the change and it comes up fine. Was able to launch and view alerts in the UI.

Hi @nickwallen - sure, what you are saying makes sense to me. To address point 1, I have moved the ES template installation step into the Indexing "Install" method, so this would be called only once. This would also not meddle with users making customization to the templates subsequently. 

Point 2 is partly addressed with the recent commit in that it takes care of ES service not running scenario. However I am not sure about how we need to proceed with a user _not_ using Elasticsearch at all. If I have understand right, this is something that would require changes and considerations to packaging and installation, not limited to this pull request alone. Please correct me if I am missing something.

anandsubbu:
Restarting travis

nickwallen:
> Hi @nickwallen - sure, what you are saying makes sense to me. To address point 1, I have moved the ES template installation step into the Indexing "Install" method,

Did this work for you?  I would think that when the Indexing Install/Configure method runs, Elasticsearch is most likely not running.

anandsubbu:
You're right @nickwallen , it wouldn't work. I reworked on the patch and here are the latest changes:
- moved the ES template installation back to the Indexing "start" method
- use a flag file to ensure the template installation runs only once. And so any start/restart of Indexing topology does not get affected.
- enclosed the template installation in a try/except block so that we warn and move on in the event of ES service not running.

I ran the following test scenarios to validate the changes:

**Full dev deployment**

- [x] Verified that the Indexing Start step also installs the ES templates. 
- [x] Verified that the flag files are created under $METRON_HOME/config
- [x] Restarted Indexing service and verified that template install is not called.

**12-node deployment**

- [x] Verified that the Indexing Start step also installs the ES templates. 
- [x] Verified that the flag files are created under $METRON_HOME/config
- [x] Restarted Indexing service and verified that template install is not called.
- [x] Stopped the ES service manually before the Indexing service starts and confirmed that the Indexing service startup proceeds successfully with a warning message indicating the template could not be installed

ottobackwards:
If the indexing start doesn't install the templates because ES is down or something, will they be installed if restarted and ES is available, or is it one shot, pass or fail.

anandsubbu:
If the install fails the first time due to ES being down, then we WARN and move on. If ES service comes up later, the admin needs to install the templates subsequently by using 'Ambari -> Metron -> Service Actions -> Elasticsearch Template Install' option. Alternatively s/he can also stop the Indexing service and start again so the template will be installed (Note that calling "restart" on the Indexing service will not trigger the template installation)

ottobackwards:
How are we going to document all this?

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py
```diff
@@ -82,6 +82,13 @@ def start(self, env, upgrade_type=None):
         self.configure(env)
         commands = IndexingCommands(params)
         commands.start_indexing_topology(env)
+        # Install elasticsearch templates
+        try:
+            if not commands.is_elasticsearch_template_installed():
+                self.elasticsearch_template_install(env)
+                commands.set_elasticsearch_template_installed()
```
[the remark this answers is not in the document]
anandsubbu: Thank you @nickwallen . Sure, done!

nickwallen:
+1 Thanks for fixing this @anandsubbu!  This takes out a manual step that if missed can cause some ugly, unexpected problems.

This is contingent on getting @ottobackwards buy-in before merging too.

anandsubbu:
Hi @ottobackwards , let me know what you think.

ottobackwards:
I think we need to document this somewhere.  I am not sure where though.

anandsubbu:
I have added this to the Elasticsearch README in my recent commit. Please see here:
https://github.com/apache/metron/pull/817/files#diff-325aea0d364d12c8637eef347ebcfca6

The preceding section to the template installation talks about the update required to the templates. I hence found this to be a better option than hosting it in the Metron Indexing README.

metron-platform/metron-elasticsearch/README.md
```diff
@@ -81,3 +81,13 @@ curl -XPUT "http://${ELASTICSEARCH}:9200/${SENSOR}_index*/_mapping/${SENSOR}_doc
 '
 rm ${SENSOR}.template
 ```
+
+## Installing Elasticsearch Templates
+
+The stock set of Elasticsearch templates for bro, snort, yaf, error index and meta index are installed automatically during the first time install and startup of Metron Indexing service.
+
```
ottobackwards: Instead of saying 'in the rare case', it might be better to say:

"It is possible that Elasticsearch is not available when the Metron Indexing Service starts the Elasticsearch topology, in that case the Elasticsearch templates will not be installed."

Also, this doesn't document the scenarios you have been describing fully.

anandsubbu: Reworded per your suggestion. 

I believe the README section covers the current behavior with my change. If you are talking about the scenarios noted in the PR description, then these are captured in the existing ES README under the section [Using Metron with Elasticsearch 2.x](https://github.com/apache/metron/tree/master/metron-platform/metron-elasticsearch#using-metron-with-elasticsearch-2x). Please let me know which specific scenario you are referring to and I would be happy to include them.

ottobackwards: https://github.com/apache/metron/pull/817#issuecomment-340269215
https://github.com/apache/metron/pull/817#pullrequestreview-72703358

My impression from these comments is that automatic installation could still possibly happen.  That is what I'm referring to as being missing.  I may be confused though.

anandsubbu: Ah okay, this comment and review is the same scenario described in the README as:

> It is possible that Elasticsearch service is not available when the Metron Indexing Service startup, in that case the Elasticsearch template will not be installed. For such a scenario, an Admin can install the template manually from the Ambari UI by following the below flow:
> 
> Ambari UI -> Services -> Metron -> Service Actions -> Elasticsearch Template Install

ottobackwards: So I am def. confused.  I don't see how "It will try again next time X happens" as the same as Go and do it manually in ambari

anandsubbu: Okay, let me try to explain more using examples:

**Scenario 1 - Happy Path**
* Fresh install
* ES service up and running
* When the Indexing service comes up, it also installs the ES templates
* Admin can start ingesting into sensors, launch alerts UI and everything works

**Scenario 2 - ES service down**
* Fresh install
* For some reason, the ES service is down when the Indexing service is coming up
* We log a warning message in the Ambari install logs, and the Indexing service starts up fine.
* Once the ES service issue is resolved, the Admin needs to install the ES templates manually before s/he can start ingesting into sensors. This can be done in two ways:
1) Using the Ambari UI -> Services -> Metron -> Service Actions -> Elasticsearch Template Install
2) By stopping the Metron Indexing service from Ambari UI, and starting it again so that it can trigger the piece of code to install the template.

Now, from a documentation perspective, point 2 above is counter intuitive IMHO, since it would not make sense to ask the Admin to stop/start Indexing service in order to have the ES template installed. I have hence documented only the first option--which is also the same way it is done presently.

Let me know if this helps clarify.

ottobackwards: But, if that is the way it works, and is what will happen, you should documented it.  If it doesn't make sense enough to document, then change the behavior.

We have people all over the user's list with issues understanding what is going on with things.  

"Oh, the doc doesn't say it but we *do* try to install automatically next time or if you .... " isn't right.

anandsubbu: Fair enough, added both methods in the latest doc change.

metron-platform/metron-elasticsearch/README.md
```diff
@@ -86,8 +86,13 @@ rm ${SENSOR}.template
...
-It is possible that Elasticsearch service is not available when the Metron Indexing Service startup, in that case the Elasticsearch template will not be installed. For such a scenario, an Admin can install the template manually from the Ambari UI by following the below flow:
+It is possible that Elasticsearch service is not available when the Metron Indexing Service startup, in that case the Elasticsearch template will not be installed. 
 
+For such a scenario, an Admin can have the template installed in two ways:
+
+_Method 1_ - Manually from the Ambari UI by following the flow:
 Ambari UI -> Services -> Metron -> Service Actions -> Elasticsearch Template Install
 
+_Method 2_ - Stop the Metron Indexing service, and start it again from Ambari UI.
+
```
ottobackwards: I am not trying to nit, but this should expressly call out that we will attempt to install the templates each START until successful.

-------

_Method 2_ - Stop the Metron Indexing service, and start it again from Ambari UI.

> Note:  The Metron Indexing service tracks if it has successfully installed the Elasticsearch templates, and will attempt to do so each time it is Started until successful.

anandsubbu: Sure, added this as well @ottobackwards

ottobackwards:
+1, thanks!

PR #827, opened 2017-11-03
METRON-1294: IP addresses are not formatted correctly in facet and group results

merrimanr:
## Contributor Comments
By default, values in facet and group by results are not formatted correctly in some cases (IP addresses being one) so the ElasticsearchDao must correctly format them instead.  To do this it must know the type of each facet/group column to know how to correctly format them.  The root cause of this issue is that the ElasticsearchDao was retrieving "common" column metadata, or columns that exist in all indices.  The addition of the "metaalert" index broke this because it is missing most of the columns that are declared in alert indices.  This PR corrects this by retrieving all column metadata across indices so that a missing column in one index does not affect the results.

The http://localhost:8080/swagger-ui.html#!/search-controller/getColumnMetadataUsingPOST changed slightly in that it now just returns a single map with all columns/types included and no longer segments columns/types by index.  This interface more closely matches how this function is used in the alerts UI and eliminates the need for the client side code to flatten the result.  Column metadata can still be retrieved for an index by just passing in a single index to that endpoint.  The case of columns existing in multiple indices with different types is solved by just excluding them from the results.

This can be tested in full dev by executing a search (either with the Alerts UI or Swagger) and including an ip address field in the facet list.  The results should now include correctly formatted IP addresses in the facetCounts section of the results.  A group query can be verified by including an IP address in the list of group fields.

I adjusted the SearchControllerIntegrationTest to account for the new interface and modified cases in the SearchIntegrationTest to catch these issues in the future.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/InMemoryDao.java
```diff
@@ -221,12 +222,24 @@ public void batchUpdate(Map<Document, Optional<String>> updates) throws IOExcept
...
+    Map<String, FieldType> indexColumnMetadata = new HashMap<>();
     for(String index: indices) {
-      columnMetadata.put(index, new HashMap<>(COLUMN_METADATA.get(index)));
+      Map<String, FieldType> columnMetadata = COLUMN_METADATA.get(index);
+      for (Entry entry: columnMetadata.entrySet()) {
+        String field = (String) entry.getKey();
+        FieldType type = (FieldType) entry.getValue();
+        if (indexColumnMetadata.containsKey(field)) {
+          if (!type.equals(indexColumnMetadata.get(field))) {
+            indexColumnMetadata.remove(field);
```
nickwallen: >  The case of columns existing in multiple indices with different types is solved by just excluding them from the results.

Why did you choose to exclude these fields?  

What happens If the user adds a new sensor that triggers this condition?  They add a sensor that has the same name, but different type.  Wouldn't the user no longer be able to see that field in the UI?   That seems like it would be confusing for the user to understand what has happened.

merrimanr: I agree with you that it would be confusing but I'm not sure what the correct behavior should be.  Should we include the field but just set the type to OTHER?  This is how the ElasticsearchDao treats fields it doesn't have type information for but so it might be better to explicitly state this in the column metadata endpoint response.

nickwallen: > I agree with you that it would be confusing but I'm not sure what the correct behavior should be. 

Right.  I don't think there is an obviously correct answer here. 

Whichever approach we do take, I think we should have a big fat, `LOG.error` here logging the duplicate field names, the index names, and what the impact of this is.

> Should we include the field but just set the type to OTHER? This is how the ElasticsearchDao treats fields it doesn't have type information for but so it might be better to explicitly state this in the column metadata endpoint response.

Do you know how the UI behaves if we were to do this?  I think your idea definitely sounds reasonable, but I'd want to see how the UI behaves with the logic defined like this.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/SearchControllerIntegrationTest.java
```diff
@@ -236,17 +236,14 @@ public void test() throws Exception {
...
-            .andExpect(jsonPath("$.*", hasSize(2)))
-            .andExpect(jsonPath("$.bro.common_string_field").value("string"))
-            .andExpect(jsonPath("$.bro.common_integer_field").value("integer"))
-            .andExpect(jsonPath("$.bro.bro_field").value("boolean"))
-            .andExpect(jsonPath("$.bro.duplicate_field").value("date"))
-            .andExpect(jsonPath("$.snort.common_string_field").value("string"))
-            .andExpect(jsonPath("$.snort.common_integer_field").value("integer"))
-            .andExpect(jsonPath("$.snort.snort_field").value("double"))
-            .andExpect(jsonPath("$.snort.duplicate_field").value("long"));
+        .andExpect(status().isOk())
```
nickwallen: Does the `SearchControllerIntegrationTest` actually exercise the ElasticsearchDAO?  I think this test will only drive the `InMemoryDao`. 

Please correct me if I am wrong, I am basing this off memory, but I don't believe it actually uses the ElasticsearchDAO.  If I am right, we probably need to update the integration tests that drive the ElasticsearchDAO.

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/InMemoryDao.java
nickwallen: And actually my comments here are tagged on the `InMemoryDao`.  I should have started this conversation on the `ElasticsearchDao`, not here.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
```diff
@@ -388,22 +388,24 @@ protected UpdateRequest buildUpdateRequest(Document update, String sensorType, S
...
       Iterator<String> mappingIterator = mapping.keysIt();
       while(mappingIterator.hasNext()) {
         MappingMetaData mappingMetaData = mapping.get(mappingIterator.next());
         Map<String, Map<String, String>> map = (Map<String, Map<String, String>>) mappingMetaData.getSourceAsMap().get("properties");
         for(String field: map.keySet()) {
-          indexColumnMetadata.put(field, elasticsearchSearchTypeMap.getOrDefault(map.get(field).get("type"), FieldType.OTHER));
+          FieldType type = elasticsearchSearchTypeMap.getOrDefault(map.get(field).get("type"), FieldType.OTHER);
+          if (indexColumnMetadata.containsKey(field)) {
+            if (!type.equals(indexColumnMetadata.get(field))) {
+              indexColumnMetadata.remove(field);
```
nickwallen: > Whichever approach we do take, I think we should have a big fat, LOG.error here logging the duplicate field names, the index names, and what the impact of this is.

Per the other conversation, this is where we actually need the log statement.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/SearchControllerIntegrationTest.java
nickwallen: And I should add, you did update the `SearchIntegrationTest` to test your changes to `ElasticsearchDAO`.  So I am happy with the level of testing you have.

The test changes here really only test the `InMemoryDao`, I think.  I find the purpose of some of the tests here slightly confusing because they are not really testing the `ElasticsearchDAO`, but that is a side issue that doesn't matter all that much.

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/InMemoryDao.java
merrimanr: As far as I know the only place the UI uses this endpoint directly is when it presents a list of fields a user can select to display in the UI.  I think we would want it to show up there so including it with a type set to OTHER makes sense to me. 

If you're ok with it I will proceed with this change and add a big fat `LOG.error` call.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/SearchControllerIntegrationTest.java
merrimanr: Yes the `SearchControllerIntegrationTest` does not actually exercise the ElasticsearchDAO and I find this confusing as well.  Really all the `SearchControllerIntegrationTest` tests is that serialization/deserialization happens correctly (which is still valuable).

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/InMemoryDao.java
nickwallen: > As far as I know the only place the UI uses this endpoint directly is when it presents a list of fields a user can select to display in the UI.

It is also used for sorting and faceting (but that is internal to the ElasticsearchDAO.)

nickwallen: > If you're ok with it I will proceed with this change and add a big fat LOG.error call.

Yes, let's give it a go.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/controller/SearchControllerIntegrationTest.java
cestella: Just some historical color to this discussion.  metron-rest's tests do not assume ES, rather the ES Dao tests are in metron-elasticsearch.  The idea was that metron-rest should have a runtime dependency on the index being used, rather than a compile-time dependency.  The thought was that the testing for the controller and such should assume an implementation of the DAO that conforms to the interface allowing the ES DAO tests to ensure that it's conforming to the expectations of the interface independently.

nickwallen:
I submitted a PR against branch that adds a specific test case in the integration tests for this defect.  I validated that this test fails in master, but passes with your fix.  Feel free to include in your PR.

nickwallen:
How does `getCommonColumnMetadata` behave when we have a mismatched field (aka a field named the same in two different indices, but with different types)?  Do we need similar mismatch handling?

merrimanr:
I don't think that method is even used anymore but I will add the mismatch handling just in case.

nickwallen:
Even better then.  Can we just get rid of it?

merrimanr:
Actually the more I think about it, I would not consider a field to be common between indices when the types mismatch.  Hard to say what the right answer is because there is no real requirement driving this.

merrimanr:
Was commenting while your comment posted.  I think your latest suggestion is the right answer :)

JonZeolla:
I didn't get myself intimately familiar with this PR, but I wanted to mention that assuming two fields with the same name but different types between indexes are not the same may not always hold.  For instance, in bro there are a couple of field names that overlap between bro logs, and so the type in the bro index needs to be one that can commonly work (usually string), but if there was no overlap it would be a different type.  [Example](https://github.com/JonZeolla/metron/blob/master/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/files/bro_index.template#L131-L134).  Not even positive this is relevant (especially given `getCommonColumnMetadata` may just get removed), but in case it is I wanted to mention it.

justinleet:
I tried hitting the `/api/v1/search/column/metadata` endpoint on fulldev with `["madeupindex"]`; e.g.

curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '["madeupindex"]' 'http://node1:8082/api/v1/search/column/metadata'

I would expect this to return no results, because the index doesn't exist, but instead I get back a lot of fields.

justinleet:
For the search, I ran,
```
  "facetFields": [
    "ip_src_addr"
  ],
  "fields": [
    "guid", "ip_src_addr"
  ],
  "from": 0,
  "indices": [
    "bro"
  ],
  "query": "*",
  "size": 5
```

and returned 

```
...// Search results
  "facetCounts": {
    "ip_src_addr": {
      "192.168.138.158": 5861,
      "192.168.66.1": 2439
    }
  }
```

@merrimanr The problem before was that those weren't in string format right?  This is what's expected with fix?

merrimanr:
@JonZeolla I think we should just remove it for now and I've done that in the latest commit.  I'm not entirely clear on what we should expect in this case and we don't need it right now anyways.  Better to implement it the right way once we are clear on how it should behave in my opinion.

Good catch @justinleet!  I did not think of this test case and to my surprise the Elasticsearch API returns ALL field mappings if you pass in an empty array of indices.  I will fix this and add a test case.  You are correct in that we should expect correctly formatted ip addresses in "facetCounts".  Before this PR they would have been in long format when you include an index that does not have a mapping for the ip_src_addr field.

merrimanr:
The latest commit fixes the bug @justinleet found and adds some test cases to cover it.  

I also removed an "ignoredIndices" variable from the ElasticsearchDao class.  This was being used to exclude certain indices but I believe we've moved towards a white list approach for specifying indices so this no longer makes sense.  I can revert if anyone disagrees.

justinleet:
I ran a request giving sensors:
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '["snort", "bro"]' 'http://node1:8082/api/v1/search/column/metadata'
```
which returns fine
```
{
  "TTLs": "double",
  "bro_timestamp": "string",
  "enrichments:geo:ip_dst_addr:location_point": "other",
  "sha256": "string",
  "enrichmentjoinbolt:joiner:ts": "date",
  "certificate:version": "integer",
...
}
```

but giving the actual indices returns nothing, e.g.
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '["bro_index_2017.11.17.14"]' 'http://node1:8082/api/v1/search/column/metadata'
```

I assume it's intentional that indices don't actually return data, which I'm fine with, but we need to rename things from indices to sensor or something.

nickwallen:
@justinleet I think that's a good find.  I'd suggest we fix this issue on a subsequent PR after we get this and #832 merged.  On #832, the refactoring will make it much easier to unit test a fix for this condition.  Let me know if you think that works or if you think its critical on this PR.

justinleet:
@nickwallen The parameter name is preexisting and I'm fine with leaving that.  Would you be okay with changing the descriptions in the README and annotations?  That should be a nonintrusive change that doesn't require spinning everything back up.

@merrimanr Do you have any objections to doing that as a middle ground?

justinleet:
Great, thanks for the update.  +1

PR #831, opened 2017-11-06
METRON-1302: Split up Indexing Topology into batch and random access sections

cestella:
## Contributor Comments
Currently we have the indexing topology handle writing to both random access indices (e.g. elasticsearch) as well as batch write indices (e.g. hdfs). We should split these up and configure them separately.

In terms of manual testing, full-dev should continue to function in that:
* Elasticsearch should have data written to it
* HDFS should have data written to it
* The indexing configuration should have separate subsections for random access and batch index writers

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
So we are basically saying, that each indexing strategy/endpoint should be it's own topology?

cestella:
Yep, so they can be tuned entirely independently.

mraliagha:
That sounds great. We had a huge headache to find good tuning parameters for "indexing" topology.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
```diff
@@ -162,34 +164,77 @@ def start_indexing_topology(self, env):
...
-            Logger.info('Indexing topology already running')
+            Logger.info('Batch Indexing topology already running')
 
-        Logger.info('Finished starting indexing topology')
+        Logger.info('Finished starting batch indexing topology')
 
-    def stop_indexing_topology(self, env):
-        Logger.info('Stopping ' + self.__indexing_topology)
+    def start_random_access_indexing_topology(self, env):
+        Logger.info('Starting ' + self.__random_access_indexing_topology)
```
nickwallen: First off, I think we definitely need to make this happen.  Each index destination is going to have very different performance characteristics that need to be tuned in isolation.  I think this is a step in the right direction.

As I read this we have effectively hard-coded two indexing topologies; random access and batch. This is definitely the most logical way to get to separate topologies based on our existing code base.  But I am wondering if we might think about this in a slightly different way.

What I really like about indexing is that we have the idea of multiple, independent destinations.  For example, my indexing configuration could look like this.
```
{
   "elasticsearch": {
      "index": "foo",
      "enabled" : true 
    },
   "hdfs": {
      "index": "foo",
      "batchSize": 1,
      "enabled" : true
    }
}
```

What if we introduced logic that consumes the indexing configuration, determines that it needs to launch 2 topologies in this case, and then launches those 2 separate topologies?  If I had 3 destinations configured, then it would launch 3 topologies; one for each destination?

I can definitely see the extra complexity in doing this.  You have to make sure the user can independently configure each of the topologies.  You have to respond to configuration changes made by the user.  And probably a few other complications.

But these are already complications that we need to deal with in Parsing.  A user can define 1 to N Parsing topologies.  It seems like if we can solve these challenges for Parsing, we can do the same for Indexing.

Anywho, I can totally see this PR as a near-term solution to the immediate problem, which might lead towards a longer-term solution like I propose.  I just wanted to see if anyone had related thoughts.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-indexing-env.xml
```diff
@@ -73,18 +98,31 @@
         <display-name>Indexing Update Column Family</display-name>
     </property>
     <property>
-        <name>indexing_workers</name>
+        <name>ra_indexing_workers</name>
+        <description>Number of Indexing Topology Workers</description>
```
nickwallen: With the name and description the same for the random access and batch topologies; can the user distinguish which is which?

metron-platform/metron-indexing/src/main/scripts/start_hdfs_topology.sh
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+METRON_VERSION=${project.version}
+METRON_HOME=/usr/metron/$METRON_VERSION
+TOPOLOGY_JAR=metron-elasticsearch-$METRON_VERSION-uber.jar
+storm jar $METRON_HOME/lib/$TOPOLOGY_JAR org.apache.storm.flux.Flux --remote $METRON_HOME/flux/indexing/batch/remote.yaml --filter $METRON_HOME/config/hdfs.properties
```
nickwallen: In the MPack we refer to this as the "batch" topology.  But in this script we refer to it as the "HDFS" topology.  We should probably stick with one nomenclature.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-indexing-env.xml
nickwallen: Actually, I see the display name is different.  Doh.  But here we are using the terms "Elasticsearch" and "HDFS", rather than "Random Access" and "Batch".  We should probably stick with one nomenclature.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
ottobackwards: I like this a lot.  I would like to see the indexing be like the parses, where you start with a name parameter.  I think the flux files make it difficult though.  I would be +1 exploring this approach though, maybe a feature branch?

metron-platform/metron-indexing/src/main/scripts/start_hdfs_topology.sh
ottobackwards: I almost think that if it is going to be hdfs, we should call it hdfs.  If we have multiple topologies based on the batch indexer, each will end up with a unique name anyways ( S3 indexer etc ).

nickwallen: Very true.  I think you're right, @ottobackwards

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-indexing-env.xml
cestella: That sounds good.  Which one would you guys prefer:
* display names in the mpack of "Elasticsearch" and "HDFS"
* variable prefixes of `es` and `hdfs`

I think I'd prefer the first as I really have been trying to set us up to a world where we can make things pluggable (this is the first baby step).

cestella: I should add, or some other 3rd option that I haven't anticipated. ;)

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
cestella: Definitely, @nickwallen I think we're thinking the same thing.  This is literally just the first baby step toward a broader vision of pluggable writers.  I think we probably want to start here with a discuss thread.  Where I was thinking of going next in here was merging the flux files into one and pulling the configs from zookeeper, which would be a step towards your vision.

metron-platform/metron-indexing/src/main/scripts/start_hdfs_topology.sh
cestella: Yeah, I prefer to keep the scripts `hdfs` and `elasticsearch`

cestella:
I think I managed to address the issues here.  Is there anything else outstanding that I missed?  If not, then bump.

ottobackwards:
The batch v. hdfs stuff still confuses me, I thought we decided on a different name?

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/integration/IndexingIntegrationTest.java
```diff
@@ -197,9 +140,7 @@ public void test() throws Exception {
       //assert that our input docs are equivalent to the output docs, converting the input docs keys based
       // on the field name converter
       assertInputDocsMatchOutputs(inputDocs, docs, getFieldNameConverter());
-      assertInputDocsMatchOutputs(inputDocs, readDocsFromDisk(hdfsDir), x -> x);
-    } catch(Throwable e) {
-      e.printStackTrace();
+      //assertInputDocsMatchOutputs(inputDocs, readDocsFromDisk(hdfsDir), x -> x);
```
merrimanr: Is this comment intentional?

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/templates/hdfs.properties.j2
```diff
@@ -0,0 +1,44 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
merrimanr: Should this file be git ignored since it is copied in at build time?

merrimanr:
I ran this up in full dev and everything worked as advertised.  I only noticed a couple minor issues and left comments for those.

I also am a little confused by the ra/batch vs es/hdfs issue.  I still see places (Ambari config parameter names, Ambari MPack scripts, Flux file paths, Flux properties, Storm topology names, etc) where the prefixes are ra/batch and not es/hdfs.  Reading through the PR comments I'm still not clear on what approach we decided on but I think consistency would be good.

nickwallen:
> I also am a little confused by the ra/batch vs es/hdfs issue.

I think the general consensus was to be explicit and use "hdfs" and "elasticsearch".  @cestella Let us know when you've made this change and are happy with it.

cestella:
Ok, I thought the general consensus was different.  As per my comment:
```
That sounds good. Which one would you guys prefer:

* display names in the mpack of "Elasticsearch" and "HDFS"
* variable prefixes of es and hdfs
I think I'd prefer the first as I really have been trying to set us up to a world where we can make things pluggable (this is the first baby step).
```

Nobody commented on that, so I went with adjusting the display names rather than variable prefixes.  The reason why I preferred to do that was that when we do a better job of including Solr or other indices, prefixing the variables `elasticsearch` will be confusing.  The flux files are generic to the writer type for the random access indices.  I can understand the confusion for the user given we only really support HDFS and elasticsearch right now, which is why I changed the display names.

Can we recap the reasons against this approach?

mmiklavc:
> I went with adjusting the display names rather than variable prefixes. 
@cestella I agree with this approach. On the front-end, make it specific to the technology for the administrator. In back, where things are pluggable, I think the more general/abstract prefixes you proposed make sense. This way the back end doesn't have to change all the time when we introduce a new batch and/or ra system.

mmiklavc:
I'm looking at this deeper and thinking through how status_params.py works with the config files. I'm looking for a dynamic way to manage the endpoints, in the future (not for this PR). For instance, we currently have this to load the config (in master)

```
# common-services/METRON/CURRENT/package/scripts/params/params_linux.py
# Indexing
indexing_kafka_start = config['configurations']['metron-indexing-env']['indexing_kafka_start']
```
This PR currently sets `ra_indexing_kafka_start` as the property name and `Elasticsearch Indexing Offset` as the display name. I think this is fine, but I'm unclear how best to manage 1..n indexing endpoints in later PR's because I don't know that Ambari (yet) offers a way to have dynamic config based on the type of service you choose to install. Maybe we can leverage a dropdown option as part of metron-env that we then leverage to dynamically choose the indexing config type you selected. So metron-env has the following:

```
<property>
        <name>ra_indexing_framework</name>
        <description>How you like to index for RA, friend?</description>
        <value>Elasticsearch</value>
        <display-name>Indexing Framework for Random Access</display-name>
        <value-attributes>
            <type>value-list</type>
            <entries>
                <entry>
                    <value>Elasticsearch</value>
                </entry>
                <entry>
                    <value>Solr</value>
                </entry>
            </entries>
            <selection-cardinality>1</selection-cardinality>
        </value-attributes>
    </property>
```

and the earlier snippet above looks something like:

```
# common-services/METRON/CURRENT/package/scripts/params/params_linux.py
# Indexing
indexing_kafka_start = config['configurations'][ra_indexing_framework + '-env']['indexing_kafka_start']
```

Not sure if you can do that, just thinking out loud. Also, this handles the ability to plug in multiple providers from which you can select only one, but it does not handle the 1..n providers scenario simultaneously, if there were such a need.

nickwallen:
> @cestella Nobody commented on that, so I went with adjusting the display names rather than variable prefixes...

I think there is some confusion as to what exactly you mean here.  You never responded to @ottobackwards 's last [comment](https://github.com/apache/metron/pull/831#issuecomment-351589022).  I think that comment indicates some clarification is needed.

I'm just looking for you to say "hey, this PR is how I want it and its ready for review."  Because hey, there is no better clarifier than code.

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/integration/IndexingIntegrationTest.java
cestella: Nope

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/templates/hdfs.properties.j2
cestella: Yep

cestella:
@nickwallen Whoops, sorry, I totally misunderstood.  This PR *is* how *I* would like it.   I'm just going to copy this here to justify my position:
```
The reason why I preferred to do that was that when we do a better job of including Solr or other indices, prefixing the variables elasticsearch will be confusing. The flux files are generic to the writer type for the random access indices. I can understand the confusion for the user given we only really support HDFS and elasticsearch right now, which is why I changed the display names and kept the command line clients named as they were before. Eventually, when we provide actual proper extension points to the indices, we should genericize the display names and scripts IMO.
```

I hope that clarifies my position on the matter (or at least why I made it the way it is).  If there are other thoughts, I'd like to be able to discuss them.  @mmiklavc chimed in in support, but I'd like to hear differing opinions if my argument is not persuasive.

ottobackwards:
So, all of these parameters with the more generic names will be uniformly applicable to all possible backends?

We will never have more than one of a given type at any time?  ( s3/hdfs or something?)

cestella:
@ottobackwards Not all of them, some of them are specific to the specific implementations (many aren't though), but I wasn't ready to remove them.  Ultimately, when we provide support for pluggable indices, we'll want to move those configs to the zookeeper config.

mmiklavc:
The code itself looks good to me. I'm running this up in full dev now and will report back. Fingers crossed that the extra slot is not a problem with the extra mem required for ES 5.

mmiklavc:
Unless there are other things I should test, I think it looks pretty good. Data is coming through to ES + Kibana (with the new ES 5.6.2 code, no less!), the topologies appear as expected and without error in Storm, data is being written to HDFS, and the menu options appear correct in Ambari. +1 home skillet.

image

image

And data in HDFS

```
[root@node1 indexing]# hdfs dfs -ls -R /apps/metron/indexing/indexed
drwxrwxr-x   - storm hadoop          0 2018-01-08 21:00 /apps/metron/indexing/indexed/bro
-rw-r--r--   1 storm hadoop    2984217 2018-01-08 21:00 /apps/metron/indexing/indexed/bro/enrichment-hdfsIndexingBolt-2-0-1515445254997.json
drwxrwxr-x   - storm hadoop          0 2018-01-08 21:10 /apps/metron/indexing/indexed/error
-rw-r--r--   1 storm hadoop    6670648 2018-01-08 21:10 /apps/metron/indexing/indexed/error/enrichment-hdfsIndexingBolt-2-0-1515445818471.json
drwxrwxr-x   - storm hadoop          0 2018-01-08 21:01 /apps/metron/indexing/indexed/snort
-rw-r--r--   1 storm hadoop    2374289 2018-01-08 21:01 /apps/metron/indexing/indexed/snort/enrichment-hdfsIndexingBolt-2-0-1515445279421.json
```

mmiklavc:
One additional note, I think we may be reaching the limits of what can reasonably fit on 1 box memory-wise. After about 40 minutes, the REST api and Elasticsearch went down. I recall this sort of thing happening independent of this PR, but just thought I'd mention it as something we could look at down the pike.

cestella:
So, this is sitting with one +1 and I think I've addressed the comments, but I'd like to make sure that my explanations are sufficient and persuasive.  Any comments here?

nickwallen:
+1 Looks good @cestella

PR #825, opened 2017-11-02
METRON-1290: Only first 10 alerts are update when a MetaAlert status is changed to inactive

merrimanr:
## Contributor Comments
This PR fixes a small bug in the ElasticsearchMetaAlertDao that only updates the first 10 alerts in a metaalert when that metaalert status is changed to inactive.  This is due to the fact that the ElasticsearchMetaAlertDao needs to lookup up the child alerts and a search created with `elasticsearchDao.getClient().prepareSearch()` defaults to result set size to 10.  This can be reproduced in full dev:
- turn off the sensor stubs so that the number of alerts in ES is fixed
- take note of the number of alerts in the index
- create a metaalert with more than 10 alerts
- the number of alerts should decrease to: # of alerts added - 1 metaalert
- change the metaalert status to inactive 
- the number of alerts will now be less than before where the count should be the same

I also added an embarrassingly complex unit test to cover this situation.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  mvn -q clean integration-test install && build_utils/verify_licenses.sh 
  ```

- [x] Have you written or updated unit tests and or integration tests to verify your changes?
[...]

PR #832, opened 2017-11-06
METRON-1301 Sorting on Triage Score Unexpectedly Filters Some Recods

nickwallen:
UPDATE:  To help avoid causing too much pain with merge conflicts, I'd like to see the following PRs resolved before we consider merging this.
- [x]  #824
- [x]  #825
- [x]  #827

## Changes

* Adds logic to sort 'missing' fields (fields that are not present in all of the underlying indices.)  For example, threat triage score is only present in the records that have been triaged.  Without this PR, sorting on the threat triage score would unexpectedly filter the search results.
    * When sorted descending, missing fields are sorted last.
    * When sorted ascending, missing fields are sorted first.

* Adds validation and error checking of the response received from Elasticsearch.  Previously there was little to no validation of the response.  This also handles partial shard failures; errors that occur only on a subset of the data.  Without this change some errors that occurred within Elasticsearch were not visible.

* Updates to the index templates ensure that the threat triage fields like threat triage overall score and the threat triage rules are mapped to a correct type.  These fields were not mapped correctly.

* Adds additional logging in the ElasticsearchDAO so that errors can be traced and debugged.

* Refactored the ElasticsearchDAO and extracted some logic into two new interfaces.  This was done so that I could write tests for the new logic.  It was not possible to specifically test this change without refactoring due to limitations with mocking.
    * The column metadata logic is now contained in `ColumnMetadataDao` and `ElasticsearchColumnMetadataDao`.  Without pulling this logic out, it was not possible to test the new sorting logic.
    * The logic for submitting a request to Elasticsearch and validating the response is now contained within `ElasticsearchSearchSubmitter` class.  Without refactoring this it was not possible to test the  new response handling logic.

## Testing

1. Launch Full Dev
2. Allow some alerts to accumulate so that when you open the Alerts UI there are a fair number of alerts to work with.
3. By default it will sort by timestamp.  Take note of the number of records returned.  
4. Click on the "Score" field to sort the results by threat triage score.
5. Note the total number of records returned after sorting.  This should be equal to or greater than what you saw before.
6. The highest scores should be sorted first.  Records that do not contain a threat triage score should be sorted last.
7. Click on the "Score" field to reverse the sort order.  Records without a threat triage score should now be sorted first.

## Pull Request Checklist


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/files/bro_index.template
```diff
@@ -98,7 +98,7 @@
           "mapping": {
             "type": "float"
           },
-          "match": "threat.triage.rules:*:score",
+          "match": "threat:triage:*score",
```
nickwallen: Prior to this change, the overall threat triage score field `threat:triage:score` was not mapped to a `float` as we had assumed.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchColumnMetadataDao.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.apache.metron.elasticsearch.utils.ElasticsearchUtils.INDEX_NAME_DELIMITER;
+
+/**
+ * Responsible for retrieving column-level metadata for Elasticsearch search indices.
+ */
+public class ElasticsearchColumnMetadataDao implements ColumnMetadataDao {
```
nickwallen: Extracting this logic into its own implementation class made it possible for me to test the new sort logic.  Just using mocks was not possible due to some of the key methods in the Elasticsearch API being declared final.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
```diff
@@ -98,78 +121,153 @@ public ElasticsearchDao() {
...
+              .orElse(FieldType.OTHER);
+
+      // sort order - if ASC, then missing values sorted last.  Otherwise, missing values sorted first
+      org.elasticsearch.search.sort.SortOrder sortOrder = getElasticsearchSortOrder(sortField.getSortOrder());
+      String missingSortOrder;
+      if(sortOrder == org.elasticsearch.search.sort.SortOrder.DESC) {
+        missingSortOrder = SORT_MISSING_LAST;
+      } else {
+        missingSortOrder = SORT_MISSING_FIRST;
+      }
```
nickwallen: Here is the additional logic to handle sorting of 'missing' fields.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
```diff
@@ -375,79 +523,14 @@ protected UpdateRequest buildUpdateRequest(Document update, String sensorType, S
         .upsert(indexRequest);
   }
 
-  @SuppressWarnings("unchecked")
   @Override
   public Map<String, Map<String, FieldType>> getColumnMetadata(List<String> indices) throws IOException {
```
nickwallen: I did not remove the `getColumnMetadata` from the `IndexDao` interface.  I simply refer to the implementation provided by the `ColumnMetadataDao` instance.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchSearchSubmitter.java
```diff
@@ -0,0 +1,138 @@
...
+import org.elasticsearch.rest.RestStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.invoke.MethodHandles;
+
+/**
+ * Responsible for submitting searches to Elasticsearch.
+ */
+public class ElasticsearchSearchSubmitter {
```
nickwallen: This class contains all the logic for submitting requests, along with validation and error checking of the Elasticsearch responses.  Refactoring this out allows me to easily test the new logic.  This also makes sure that the same error checking logic is applied to both `/search` and `/group` requests.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchSearchSubmitter.java
```diff
@@ -0,0 +1,138 @@
...
+      String msg = String.format(
+              "Failed to execute search; error='%s', search='%s'",
+              ExceptionUtils.getRootCauseMessage(e),
+              ElasticsearchUtils.toJSON(request));
+      LOG.error(msg, e);
+      throw new InvalidSearchException(msg, e);
+    }
+
+    // check for shard failures
+    if(esResponse.getFailedShards() > 0) {
```
nickwallen: This is the logic that will ensure similar, subtle errors that occur in Elasticsearch will not be masked going forward.

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/SearchIntegrationTest.java
```diff
@@ -138,6 +137,46 @@
...
+   * "sort": [
+   *    {
+   *      "field": "threat:triage:score",
+   *      "sortOrder": "asc"
+   *    }
+   *  ]
+   * }
+   */
+  @Multiline
+  public static String sortAscendingWithMissingFields;
```
nickwallen: I added two new queries to the integration tests to exercise this defect.

nickwallen:
I want to get all of @merrimanr 's fixes in (#824, #825 #827 ) before we consider merging this one.  There will be some merge conflicts and I can take care of sorting through those.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchSearchSubmitter.java
```diff
@@ -0,0 +1,138 @@
...
+    /*
+     * shard failures are only logged.  the search itself is not failed.  this approach
+     * assumes that a user is interested in partial search results, even if the
+     * entire search result set cannot be produced.
+     *
+     * for example, assume the user adds an additional sensor and the telemetry
+     * is indexed into a new search index.  if that search index is misconfigured,
+     * it can result in partial shard failures.  rather than failing the entire search,
+     * we log the error and allow the results to be returned from shards that
+     * are correctly configured.
```
nickwallen: Hopefully this explanation makes sense.  When there are partial shard failures, we do not fail fast.  We log the problem and return partial results to the user.  I am torn as to whether this is the right approach.

iraghumitra:
@nickwallen really liked the way you annotated the PR with your comments. You made it really easy to understand and saved a lot of time for me. As you rightly pointed out I will wait till all other dependent PR's to come in before giving this a  spin.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/files/bro_index.template
cestella: great catch.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchColumnMetadataDao.java
```diff
@@ -0,0 +1,192 @@
...
+    fieldTypeMap.put("float", FieldType.FLOAT);
+    fieldTypeMap.put("double", FieldType.DOUBLE);
+    fieldTypeMap.put("boolean", FieldType.BOOLEAN);
+    elasticsearchTypeMap = Collections.unmodifiableMap(fieldTypeMap);
+  }
+
+  private transient AdminClient adminClient;
+  private List<String> ignoredIndices;
+
+  public ElasticsearchColumnMetadataDao(AdminClient adminClient, List<String> ignoredIndices) {
```
cestella: Small nit, can we make `ignoredIndices` a `Set`

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
```diff
@@ -234,26 +366,43 @@ public synchronized void init(AccessConfig config) {
     if(this.client == null) {
       this.client = ElasticsearchUtils.getClient(config.getGlobalConfigSupplier().get(), config.getOptionalSettings());
       this.accessConfig = config;
+      this.columnMetadataDao = new ElasticsearchColumnMetadataDao(this.client.admin(), Collections.singletonList(".kibana"));
```
cestella: While I agree with the inclusion of `.kibana` here, what are the consequences for those users with other, non-metron, indices?  Should we allow users to specify this list in the REST application.yml?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/utils/ElasticsearchUtils.java
```diff
@@ -179,4 +186,46 @@ else if(ipObj instanceof List) {
...
     throw new IllegalStateException("Unable to read the elasticsearch ips, expected es.ip to be either a list of strings, a string hostname or a host:port string");
   }
+
+  /**
+   * Converts an Elasticsearch SearchRequest to JSON.
+   * @param esRequest The search request.
+   * @return The JSON representation of the SearchRequest.
+   */
+  public static String toJSON(org.elasticsearch.action.search.SearchRequest esRequest) {
+    String json = "null";
```
cestella: Should we instead return back an `Optional<String>` rather than the string representation of null on a failure here?  Alternatively, maybe throw an exception so the calling function can handle the exception how it likes?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
nickwallen: Yes, I agree with your feedback.  I was just trying to refactor the column metadata logic and minimize other changes.  So in this case `.kibana` was already hard-coded in ElasticsearchDao.

I would be totally open to making this improvement though.  I was just trying to walk the line of how much should I change when refactoring?  Considering that, what do you think?

nickwallen: Ok, man I have been really bad about reading and comprehension today.  I see in your review comment feedback that you already addressed this point. Sorry.

I think having this hardcoded is just not good.  The change seems simple enough.  I can probably take care of this no problem.

cestella: I'm almost always in favor of having hard coded things passed in via config files.  It gets us out of jams and almost every time I've convinced myself that it's not necessary, it totally is.  How convinced are you that we are only ever going to need one ignored index?

If you have doubts, then I'd probably add something to the `application.yml` which gets set in the AccessConfig object when we set up the index in the IndexConfig from `metron-rest`.

cestella: Given that it's a pretty simple change and requires no new machinery to added, I think it is probably a good fit to add to this PR, but I might be on the wrong side of that line more than I'm on the right side, so take my word with a grain of salt. ;)

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/utils/ElasticsearchUtils.java
nickwallen: Yes, definitely this is kind of weird.  This is taken care of in the latest commit.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchColumnMetadataDao.java
nickwallen: Done.  See latest.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
nickwallen: The latest has the change that makes the ignore indices configurable.

nickwallen:
I just merged in the last bits from master.  This is ready for review.

justinleet:
Spun this up on full dev, and saw that sorting worked as expected in the UI.  This worked both when no groupings were selected and also when drilling down (e.g. grouping by ip_dst_addr and country).

I also like the refactorings a lot.  It's good stuff that makes life easier and more testable.

+1, assuming @cestella is happy with the changes you made on his comments.

cestella:
+1 by inspection.  Looks great!

merrimanr:
+1 worked as advertised.  Thanks @nickwallen!

nickwallen:
Thanks for all the reviews, guys.  Going to commit now.

PR #845, opened 2017-11-20
METRON-1321 Metaalert Threat Score Type Does Not Match Sensor Indices

nickwallen:
After creating Metaalerts in the Alerts UI, I am unable to sort by threat triage score.  This only occurs AFTER creating a Metaalert.  

The exception that is logged is shown in the issue.  When opening the Developer view in Chrome, I can see the API responds with an error.  An error message is logged under `/var/log/metron/metron-rest.log`.

## What happened?

An overall threat score is added to a Metaalert whenever it has any child alerts.  This overall threat score is a summary of the threat scores from each of the child alerts.  

This field is also named the same as the threat score fields from the sensor indices; 'threat:triage:score'.  The type of this field must be the same as the 'threat:triage:score' fields from each of the sensor indices.  Otherwise the Alerts UI cannot properly sort alerts and metaalerts in the same table/view in the UI.

## Changes

I updated the `ElasticsearchMetaalertDao` so that when it creates the overall threat score, it is added as a Float to match the type expected from the other sensor indices.  

I also updated the unit tests to specifically check for this condition, along with other Javadoc improvements to clarify things.

## Testing

1. Spin up Full Dev with this PR
2. Run the Alerts UI from #803.  This enables the creation of Metaalerts in the Alerts UI.  The following will execute the Alerts UI and use `node1` (created in step 1) as the backend.
    ```
    cd metron-interface/metron-alerts/
    npm install
    ./scripts/start-dev.sh
    ``` 
3. Open up the Alerts UI; http://localhost:4201/
4. Sort by triage score and assure no errors occur
5. Create a Metaalert.
6. Sort by triage score and assure no errors occur

## Pull Request Checklist


nickwallen:
I am still testing this in Full Dev.  Will respond once I verify this completely.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
```diff
@@ -614,8 +625,15 @@ protected void calculateMetaScores(Document metaAlert) {
       }
       metaScores = new MetaScores(scores);
     }
+
+    // add a summary (max, min, avg, count, sum) of all the threat scores from the child alerts
     metaAlert.getDocument().putAll(metaScores.getMetaScores());
-    metaAlert.getDocument().put(threatTriageField, metaScores.getMetaScores().get(threatSort));
+
+    // the overall threat score for the metaalert; either max, min, avg, count or sum of all child scores
```
ottobackwards: why isn't it a float to start with?  isn't that the real issue?

justinleet: The calculations were done as Double and given to ES.  However, there's no definition of the field in ES (It just used automatic mapping), so it was given the ES double.

justinleet:
Could we also add the threat score to the metaalert template, to match the other templates?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
nickwallen: The `ElasticsearchMetaAlertDao` adds an overall threat score to the Metaalert.  The overall threat score can be any one of the following summary aggregations of the child alerts; sum, min, max, count, average, or median.  These summary values are calculated in `MetaScores` and result in Double values.  Since the other sensor indices currently define `threat:triage:score` as a float, this solution just casts this to a Double to match those.

I think an alternative way to solve this is to just make the `threat:triage:score` in each of the sensor indices a Double as you mentioned.  I think your approach seems a little cleaner to me.  Although, I am not sure if there are other down sides I am not thinking about.  Can anyone else think of a problem with this approach? @justinleet @merrimanr ?

ottobackwards: I would not hold up the PR for this point if it is too much.  Just seems that we are coding around something else.

nickwallen: I think it is really just a matter of what we'd expect a user to define as scores for their threat triage rules.  Are they really going to define values greater than a 32-bit float?  Or values that sum to greater than a 32-bit float?

I think it is a good point to discuss, Otto.  I'd suggest we go with the float approach now, as it minimizes the scope of change in this PR.  But we can revisit whether a double should be used after we migrate to ES 5.x.

nickwallen:
I ran this up according to my testing instructions and it addresses the problem.  Please take a look-see.

justinleet:
+1 by inspection, assuming @ottobackwards is good.

Thanks for expanding the comments out, it's definitely helpful.

ottobackwards:
+1

PR #849, opened 2017-11-27
METRON-1334 Add C++11 Compliance Check to 'platform-info.sh'

nickwallen:
Some of the module dependencies for the Management and Alerts UI must be built natively on the host.  This requires a C/C++ compiler.  In addition, some of the dependencies require a C++11 compliant compiler.  This is causing problems for users who attempt to build Metron on a system with an older version of GCC, like CentOS 6.

Not having a C++11 compliant compiler can cause some non-obvious error messages when the build fails.  This adds a version check for GCC and also a C++11 compliance check.  The compiler itself must be on the user's PATH, which is what the Node modules also require.

If the compiler is C++11 compliant, you will see the following.
```
--
g++
g++ (GCC) 5.3.1 20160406 (Red Hat 5.3.1-6)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

--
Compiler is C++11 compliant
```

If the compiler is NOT C++11 compliant, you will see the following.
```
--
g++
g++ (GCC) 4.4.7 20120313 (Red Hat 4.4.7-18)
Copyright (C) 2010 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

--
cc1plus: error: unrecognized command line option "-std=c++11"
Warning: Compiler is NOT C++11 compliant
```

This was tested manually on Mac OS X, CentOS 6, CentOS 7, and Ubuntu Trusty 64.

## Pull Request Checklist

### DIFF
--- a/metron-deployment/scripts/platform-info.sh
+++ b/metron-deployment/scripts/platform-info.sh
@@ -75,14 +75,14 @@ npm --version
 
 # C++ compiler
 echo "--"
-echo "g++"
-g++ --version
+if [[ $(command -v g++) && $(g++ --version 2>/dev/null) ]]; then
+  g++ --version
 
-# C++11 compliant compiler
-echo "--"
-OBJFILE=/tmp/test
-CPPFILE=/tmp/test.cpp
-cat > $CPPFILE <<- EOM
+  # check C++11 compliance
+  echo "--"
+  OBJFILE=/tmp/test
+  CPPFILE=/tmp/test.cpp
+  cat > $CPPFILE <<- EOM
 #include <iostream>
 using namespace std;
 int main() {
@@ -90,13 +90,18 @@ int main() {
     return 0;
 }
 EOM
-g++ -std=c++11 $CPPFILE -o $OBJFILE
-if [ $? -eq 0 ]; then
-    echo "Compiler is C++11 compliant"
+  g++ -std=c++11 $CPPFILE -o $OBJFILE &>/dev/null
+  if [ $? -eq 0 ]; then
+      echo "Compiler is C++11 compliant"
+  else
+      echo "Warning: Compiler is NOT C++11 compliant"
+  fi
+  rm -f $CPPFILE $OBJFILE
+elif [[ $(command -v g++) ]]; then
+  echo "Warning: g++ not properly configured"
 else
-    echo "Warning: Compiler is NOT C++11 compliant"
+  echo "Warning: g++ not found"
 fi
-rm -f $CPPFILE $OBJFILE
 
 # operating system
 echo "--"
--- a/metron-deployment/scripts/platform-info.sh
+++ b/metron-deployment/scripts/platform-info.sh
@@ -73,6 +73,28 @@ echo "--"
 echo "npm"
 npm --version
 
+# C++ compiler
+echo "--"
+echo "g++"
+g++ --version
+
+# C++11 compliant compiler
+echo "--"
+CPPFILE=/tmp/test.cpp
+cat > $CPPFILE <<- EOM
+#include <iostream>
+using namespace std;
+int main() {
+    cout << "Hello World!" << endl;
+    return 0;
+}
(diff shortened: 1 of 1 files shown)


metron-deployment/scripts/platform-info.sh
```diff
@@ -73,6 +73,30 @@ echo "--"
...
+    return 0;
+}
+EOM
+g++ -std=c++11 $CPPFILE -o $OBJFILE
+if [ $? -eq 0 ]; then
+    echo "Compiler is C++11 compliant"
+else
+    echo "Warning: Compiler is NOT C++11 compliant"
+fi
+rm -f $CPPFILE $OBJFILE
```
ottobackwards: A standard way in the language is to check for the value of __cplusplus

201402L  is CPP 14
201103L is CPP 11

I don't have a preference, just throwing this out there.

ottobackwards: Do we have to account for platform_info on a non-dev machine?  If there is no g++ at all?

nickwallen: > Do we have to account for platform_info on a non-dev machine? If there is no g++ at all?

The platform-info script is there to tell us whether a machine can be used to deploy metron.  Right now, you have to build Metron to deploy it, so the script checks all dev dependencies.

If g++ is not in the path, then the script will tell us that.  And it would carry on its merry way.  That's the behavior that I would expect.

Maybe I am not following you. Can you describe the scenario you are thinking of more?

nickwallen: > A standard way in the language is to check for the value of __cplusplus

Thanks, I saw that.  I didn't think it was worth the trouble.  IMHO

ottobackwards: i agree about it not being worth the trouble.

as far as g++, I would expect us to check (pseudo)

if c++ in path then
    build test program
else
    echo No Cpp, you are in trouble!
fi

Maybe I'm mis-reading this, if there is no cpp, it will still try to build etc won't it?

ottobackwards: OK, I ran it, I did mis-understand.  sorry.

nickwallen: No worries.  Glad to have a second set of eyes on it.

ottobackwards:
+1, Tested on OSX and on centos 6 with and without compliant g++ installed
Thanks Nick!

JonZeolla:
Ran on a purposefully misconfigured macOS system.  Is this the expected output?
```
--
g++
xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun
--
xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun
Warning: Compiler is NOT C++11 compliant
--
```

ottobackwards:
I think that is valid

nickwallen:
> Is this the expected output?

I agree with @ottobackwards . That looks like what I would expect to see.  

@JonZeolla I assume since you asked, it is not what you would expect.  What would prefer to see here?

JonZeolla:
Instead of trying to explaining my position, I threw together a quick PR to illustrate my thoughts.  Feel free to take another approach.  I tested it with `g++` configured, unconfigured, and not in my PATH on macOS, as well as on a CentOS 7 with and without gcc installed, and an Ubuntu 14.04 box with an old version of gcc installed (not C++11 compliant).  All worked as expected.

JonZeolla:
+1 great stuff @nickwallen

PR #851, opened 2017-11-28
METRON-1336 Patching Can Result in Bad Configuration

nickwallen:
The following problems are addressed in this PR.

* A patch can be constructed that when applied creates an invalid configuration.  The invalid configuration is not discovered until attempting to "pull" it back out of Zookeeper.  In all cases, the result of applying a patch needs to be validated before pushing it to Zookeeper.

* The Profiler configuration can only be pushed when pushing all configurations at once.  Attempting to push just the Profiler configuration alone, fails silently.

* I do not believe that validation was occurring in all cases; in particular when pushing a specific configuration type (like pushing "Squid", PARSING alone.)  Validation was occurring in other cases, like when all configuration elements are pushed at once.

* Added unit tests to validate that a 'bad' patch cannot be applied.

* Added a good number of unit tests to ensure invalid configurations cannot be pushed under all circumstances (parser, enrichment, indexing, profiler, "push all").

* Added a test to ensure that the Profiler configuration can be pushed independently.

## Manual Testing

1. Launch Full Dev.

1. Environment definitions that will be used later.
    ```
    export METRON_HOME=/usr/metron/0.4.2
    export ZOOKEEPER=node1:2181
    ```

1. Dump all existing configurations.
    ```
    $METRON_HOME/bin/zk_load_configs.sh -z $ZOOKEEPER -m DUMP
    ```

1. Dump just the Profiler configuration.  There should be none.
    ```
    $METRON_HOME/bin/zk_load_configs.sh -z $ZOOKEEPER -m DUMP -c PROFILER
    ```

1. Create a Profiler configuration on disk.
    ```
    $ cat $METRON_HOME/config/profiler.json
    {
      "profiles": [
        {
          "profile": "hello-world",
          "foreach": "ip_src_addr",
          "init":    { "count": "0" },
          "result":  "count"
        }
      ]
    }
    ```
    
1. Push the Profiler configuration.  
    ```
    $METRON_HOME/bin/zk_load_configs.sh -z $ZOOKEEPER -m PUSH -c PROFILER -i $METRON_HOME/config
    ```

1. Dump just the Profiler configuration.  It should be defined now.
    ```
    $METRON_HOME/bin/zk_load_configs.sh -z $ZOOKEEPER -m DUMP -c PROFILER
    ```

1. Create a patch for the Profiler configuration.
    ```
    $ cat profile.patch
    [
    	{
    		"op": "add",
    		"path": "/profiles/0/profile",
    		"value": "changed-profile-name"
    	}
    ]
    ```

1. Apply the patch.
    ```
    $METRON_HOME/bin/zk_load_configs.sh -z $ZOOKEEPER -m PATCH -c PROFILER -pf profile.patch
    ```

1. Dump the Profiler configuration and ensure that the name of the Profile was changed.
    ```
    [root@node1 ~]# $METRON_HOME/bin/zk_load_configs.sh -z $ZOOKEEPER -m DUMP -c PROFILER
      PROFILER Config: profiler
      {
        "profiles" : [ {
          "profile" : "changed-profile-name",
          "onlyif" : "exists(ip_src_addr)",
          "foreach" : "ip_src_addr",
          "init" : {
            "count" : "0"
          },
          "update" : {
            "count" : "count + 1"
          },
          "result" : "count"
        } ]
      }
    ```
  
1. Create a patch that would make the Profiler configuration invalid. 
    ```
    [root@node1 ~]# cat bad.patch
      [
      	{
      		"op": "add",
      		"path": "/profiles/0/invalid",
      		"value": "22"
      	}
      ]
    ```

1.  Apply the bad patch.   The script should terminate with an error.
    ``` 
    [root@node1 ~]# $METRON_HOME/bin/zk_load_configs.sh -z $ZOOKEEPER -m PATCH -c PROFILER -pf bad.patch
    2017-11-28 17:37:17 ERROR ConfigurationManager:314 - Unable to apply patch to Zookeeper config
    java.lang.RuntimeException: Unable to load {
      "profiles" : [ {
        "profile" : "changed-profile-name",
        "onlyif" : "exists(ip_src_addr)",
        "foreach" : "ip_src_addr",
        "init" : {
          "count" : "0"
        },
        "update" : {
          "count" : "count + 1"
        },
        "result" : "count",
        "invalid" : "22"
      } ]
    }
    	at org.apache.metron.common.configuration.ConfigurationType.lambda$static$4(ConfigurationType.java:68)
    	at org.apache.metron.common.configuration.ConfigurationType.deserialize(ConfigurationType.java:93)
    	at org.apache.metron.common.configuration.ConfigurationsUtils.applyConfigPatchToZookeeper(ConfigurationsUtils.java:621)
    	at org.apache.metron.common.cli.ConfigurationManager.patch(ConfigurationManager.java:307)
    	at org.apache.metron.common.cli.ConfigurationManager.run(ConfigurationManager.java:285)
    	at org.apache.metron.common.cli.ConfigurationManager.run(ConfigurationManager.java:244)
    	at org.apache.metron.common.cli.ConfigurationManager.main(ConfigurationManager.java:360)
    Caused by: org.apache.metron.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "invalid" (class org.apache.metron.common.configuration.profiler.ProfileConfig), not marked as ignorable (8 known properties: "init", "expires", "groupBy", "profile", "foreach", "result", "onlyif", "update"])
     at [Source: {
      "profiles" : [ {
        "profile" : "changed-profile-name",
        "onlyif" : "exists(ip_src_addr)",
        "foreach" : "ip_src_addr",
        "init" : {
          "count" : "0"
        },
        "update" : {
          "count" : "count + 1"
        },
        "result" : "count",
        "invalid" : "22"
      } ]
      ...
    ```

## Pull Request Checklist


metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
```diff
@@ -156,7 +156,7 @@ public static void writeConfigToZookeeper(ConfigurationType configType,Optional<
   }
 
   private static String getConfigZKPath(ConfigurationType configType, Optional<String> configName) {
-    String pathSuffix = configName.isPresent() && configType != GLOBAL ? "/" + configName : "";
+    String pathSuffix = configName.isPresent() && configType != GLOBAL ? "/" + configName.get() : "";
```
cestella: how did this ever work before?  I'm surprised we didn't get the wrong result because configName is optional.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
```diff
@@ -343,25 +343,57 @@ public static void uploadConfigsToZookeeper(String rootFilePath, CuratorFramewor
...
-      case PARSER: // intentional pass-through
-      case ENRICHMENT: // intentional pass-through
-      case INDEXING:
-        Map<String, byte[]> sensorIndexingConfigs = readSensorConfigsFromFile(rootFilePath, type,
-            configName);
-        for (String sensorType : sensorIndexingConfigs.keySet()) {
-          writeConfigToZookeeper(type, configName, sensorIndexingConfigs.get(sensorType), client);
+
+      case PARSER: {
+        Map<String, byte[]> configs = readSensorConfigsFromFile(rootFilePath, PARSER, configName);
```
cestella: These cases look like they are cut and pasted which seems like code smell to me and might be a maintenance issue.  Can we extract the common code for Parser, Enrichment, and Indexing into a separate function that is called here?  Perhaps something like:
```
void writeSensorConfigs(ConfigurationType type, Optional<String> configName, BiFunction<String, byte[], Void> callback) {
   Map<String, byte[]> configs = readSensorConfigsFromFile(rootFilePath, type, configName);
  for(String sensorType : configs.keySet()) {
    byte[] configData = configs.get(sensorType);
    callback.apply(sensorType, configData);
  }
}
```
which could be called from this case via `writeSensorConfigs(PARSER, configName, (sensorType, configData) -> writeSensorParserConfigToZookeeper(sensorType, configData, client));`

Take the above as a very rough suggestion and you can feel to abstract it however you wish.

cestella: The other option here is to add a new method called `writeSensorConfigToZookeeper(ConfigurationType type, String sensorType, byte[] configData, CuratorFramework client)` that calls the appropriate `writeSensorXConfigToZookeeper` call.

cestella: Oh, an even better option (IMO) would be to add a `writeSensorConfigToZookeeper(String sensorType, byte[] configData, CuratorFramework client)` method to `ConfigurationType`.  Anyway, I'll stop ;)

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
nickwallen: Yeah, this confused me too.  There are a lot of paths through ConfigurationUtils.  Some paths worked, others didn't.

cestella: We need to refactor `ConfigurationUtils` so badly that it hurts.

mmiklavc: I think that's because we handle global config in a special way. This whole class needs refactored, and I'd have done more with the original patch PR had it not blown up scope.

cestella: Yeah, this is not one of those things that can be done as a side-effect of another PR.  I considered refactoring it as part of the zookeeper refactoring that I did earlier.  It needs more direct attention and refactoring.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
nickwallen: Yes, I had the same thought and tried for a bit to refactor it.  I landed on this because the various other ways to do this either (1) seemed more complex and less obvious as to what we are actually doing here or (2) lead down a path of heavy refactoring of ConfigurationUtils.  Both of which i wanted to avoid

nickwallen: Your first suggestion with the callback just seems less obvious and clear to me (IMHO).

I'll try and think through your 2nd and 3rd suggestions, but (at least right now) I'm not seeing something that doesn't add more complexity to ConfigurationUtils.  

But clearly I could be totally wrong here.  Let me think on it.

cestella: Sounds good.  Just so I'm clear about my preferred approach (option 3 there), I went ahead and implemented it in a [PR](https://github.com/nickwallen/metron/pull/12) against your PR so you can see what I was talking about.

mmiklavc: I remember struggling through this class when applying the patching mechanism, so I want to start by saying I'm a big fan of this work and conversation around how to best improve and simplify this class. @cestella I agree with @nickwallen about option 1. Compared to your PR on this PR (sounds like an XZibit meme...) for option 3, I think it's much less intuitive to read. I took a look through it and that would be my preference for the way to go. It's classic strategy pattern through composition, and it eliminates the duplication we see below. I tend to agree with the comments on that PR about eventually pulling those implementations into their own classes. That would make the code much more readable. And I think it probably puts us in a better place for cleaning up the config utils class or removing/renaming it altogether. The only part of this that is still a little bit of a smell to me is that the global config and profiler throw unsupported operation exceptions. I think this is a big enough improvement that I'd be willing to let that slide for now, but ideally the global config would exist similarly to the other config. e.g. instead of this:

- config/zookeeper/
  - global.json
  - profiler.json
  - parsers/
    - ...
  - enrichment/
    - ...
  - indexing/
    - ...
  - .... etc.

You'd give the global config and profiler a directory as well, which would enable a consistent layout on the local FS and in Zookeeper, and a consistent interface for these commands. 

- config/zookeeper/
  - global/
    - global.json
  - profiler/
    - profiler.json
  - parsers/
    - ...

Then this
```
public interface ConfigurationOperations {
...
void writeSensorConfigToZookeeper(String sensorType, byte[] configData, CuratorFramework client) throws Exception;
...
```

could become this

```
public interface ConfigurationOperations {
...
void writeConfigToZookeeper(String type, byte[] configData, CuratorFramework client) throws Exception;
```

I'm not sure it's desirable or not, but you could then theoretically provide multiple configs in global/ and profiler/, which is consistent with the parser/, enrichment/, and indexing/ configs.

mmiklavc: One more thought - the code for option 3 does add more code than the copy/paste, but in spite of that I think it puts us closer to being able to delete a lot of the redundant and/or dead code in config utils. So I think it's a net positive.

mmiklavc: @cestella I submitted a PR against your PR on this PR - https://github.com/cestella/incubator-metron/pull/8

cestella: @mmiklavc I'm ok with that modification if we decide to go this route.

nickwallen: I like the refactoring, guys.  It's looking food.  Can't believe we have a PR of a PR of this PR.  One more level and I think we hit Inception's limbo.

I'd be open to including those changes here or even just creating a separate PR for the refactoring work as a follow-on to this.  That refactoring adds a good bit of code in itself.  Maybe we could even tackle some further ConfigurationUtils clean-up in that same PR.

But If we want @cestella and @mmiklavc 's refactoring to go in with this PR, then we need @cestella to do the first "kick" and merge @mmiklavc's PR.

mmiklavc: Haha, I know. I couldn't resist seeing if GitHub actually allowed this sort of thing, and well, it was actually pretty straightforward. I don't care if we pull it all together or submit them separately. It sounds like we're all in agreement that the refactoring is a good idea, so if it was split them Casey's interface refactoring would go in immediately following with the necessary votes. I'll leave it to you both to decide.

cestella: The meta-PR is merged

nickwallen:
@mmiklavc @cestella The kick is complete. All PRs merged.  Ready for review.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
```diff
@@ -343,25 +343,42 @@ public static void uploadConfigsToZookeeper(String rootFilePath, CuratorFramewor
    * @param type config type to upload configs for
    * @param configName specific config under the specified config type
    */
-  public static void uploadConfigsToZookeeper(String rootFilePath, CuratorFramework client,
-      ConfigurationType type, Optional<String> configName) throws Exception {
+  public static void uploadConfigsToZookeeper(
+          String rootFilePath,
```
ottobackwards: When reading from file, the API should explicitly require file Paths as opposed to just strings.
It makes the api methods easier to understand when there are a bunch of similar methods.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
```diff
@@ -572,15 +590,22 @@ public static void applyConfigPatchToZookeeper(ConfigurationType configurationTy
...
+
     byte[] configData = readConfigBytesFromZookeeper(configurationType, configName, client);
     JsonNode source = JSONUtils.INSTANCE.readTree(configData);
     JsonNode patch = JSONUtils.INSTANCE.readTree(patchData);
     JsonNode patchedConfig = JSONUtils.INSTANCE.applyPatch(patch, source);
-    writeConfigToZookeeper(configurationType, configName,
-        JSONUtils.INSTANCE.toJSONPretty(patchedConfig), client);
+    byte[] prettyPatchedConfig = JSONUtils.INSTANCE.toJSONPretty(patchedConfig);
+
+    // ensure the patch produces a valid result; otherwise exception thrown during deserialization
```
ottobackwards: Can we reword this?  It needs to be clear that some things cannot be deserialized if they have an invalid configuration.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
nickwallen: Sure, but this method signature already existed.  Do I need to fix that on this PR?

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
nickwallen: I'm not sure exactly what you're looking for.  Can you suggest some specific text?

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
ottobackwards: No.  I guess I miss read the ambition of this pr, given the meta-pr and all ;)

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
ottobackwards:
```java
//Some configurations have side effects to deserialization, and may throw exceptions from 
//validation that may happen during the deserialization.  This does not meant that the //deserialization has failed due to invalid serialization.
```
?

ottobackwards: Never mind Nick, I didn't read this correctly, and I have the "serialization mixed with logical validation" stuff upfront in my head because of my work over in https://github.com/apache/metron/pull/856.

Sorry

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
nickwallen: Yeah I get it. ;). It wasn't my preference to include all of that here, but I bend to make people happy.

ottobackwards: I am familiar with that ;)

ottobackwards: I'll open a jira

ottobackwards: METRON-1342

nickwallen:
Bump.  Anything else we need on this one?

cestella:
I'm comfortable with this.  +1 by inspection

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java
mmiklavc: Heh, well our meta-PR was to address a code change that resulted in duplication, not just a style preference. ;)

mmiklavc:
+1 by inspection. Nice work Nick.

nickwallen:
Thanks guys.  Nice work on the meta-PRs.  I like how that turned out.

PR #853, opened 2017-11-28
METRON-1337: List of facets should not be hardcoded

merrimanr:
## Contributor Comments
This PR makes the list of facet fields in the Alerts UI configurable.  There is now a "search.facet.fields" setting in https://github.com/apache/metron/blob/master/metron-interface/metron-rest/src/main/resources/application.yml that is a comma-separate list of fields to be used as facets.  Originally a comment was made that the "host" field wasn't commonly used so I removed that from the default list.

I can think of two options for exposing this configuration: 
- through an AlertProfile that is specific to each user, stored in a relational database and can be edited through REST
- a setting that is exposed in Ambari

I chose to include the first option in this PR to get the conversation going.  Is one of these preferable?  The AlertProfile approach allows this setting to be changed at runtime and each user has their own list of facet fields.  But it is not versioned like it would if it were in Ambari.  Do we prefer one over the other?  Do we want both with Ambari being the default when an AlertProfile doesn't exist for a user?  Are there other options I'm not thinking of?

This works similar to how default search indices work:  it is managed in the REST layer and can be overriden by including facet fields in a search request.  However it seemed useful to allow a way to explicitly NOT include facets in a query so it works slightly different than indices.  A missing facetFields property in the request will use the defaults while an empty array will disable facets.  A missing indices property or an empty array will use the default indices.  Is this the correct behavior?  

This has been tested in full dev and the UI e2e tests pass when run in isolation.  There is currently an effort to stabilize the e2e tests as a follow on to https://github.com/apache/metron/pull/803 so I did not try to solve that here.  

I will add some documentation around configuring the facet field list and facetFields behavior in a search request once we come to a consensus on the solution.

Another issue I would like to point out.  When I added facetFields to the AlertProfile object it required a database update because that new field needed to be added to the database.  This would become an issue if someone were upgrading from a previous version.  Is this acceptable if we document it for future upgrades?  Is a relational database the right solution or should we consider a more flexible storage option?

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
Does ambari own the application.yaml currently?

merrimanr:
Each environment will have it's own application.yaml.  Full dev has one, our testing environment has one, Ambari ships one etc.  There is also a base application.yaml that has the defaults (default facets are included here).  Does that make sense?

ottobackwards:
What I get to was, the setting should be exposed in ambari if ambari manages the config.

merrimanr:
Ambari does not manage this config.  It is only included in the base application.yml as a default setting.  I don't feel like this setting should be in Ambari for a couple reasons:  changing it requires a restart and the list of facets cannot be user specific.

justinleet:
@merrimanr Can you deconflict this?

merrimanr:
Done

justinleet:
To respond to the questions in the description (and maybe kick off conversation, especially if anyone disagrees) + add my own thoughts.  In no particular order.

@merrimanr Let me know if I'm misunderstanding anything as well, I'm digging into this for the first time

1. This feels like a user specific config.  Having a default that each user gets if they don't do anything seems appropriate, but I'd expect different people to want to be looking at different things.
1. The DB feels like a reasonable place for this to live.  We're already storing some per user type data (e.g. recent searches) in the DB, right?
1. It would be nice to have (and this would be follow-on), a really easy UI way to adjust the facet fields.  Especially, if it's that adjustable on the fly, I don't think we need versioning on the user level facet configs.
1. IMO, having the defaults be versioned is nice, but not mandatory (we don't version anything in that file, do we?)
1. Re: the upgrade issue.  We generally don't support upgrades, but having the upgrading.md have a note along with at least a couple of the basic commands to add that field would be nice.
1. I'd have a slight inclination towards empty facet array being default and missing being disable.  It's more intuitive for me that if I leave it out of the request, it's not present.

merrimanr:
Here are my thoughts on your responses.

1.  I think we're in agreement.  This is how this PR currently works.
2. No, saved searches are currently stored client-side.  I think the plan was to store these server-side but it hasn't been implemented yet.  The only thing we store in a relational DB at this point are REST user credentials.  This would change if we switch to an LDAP approach for managing authentication (I think we should) and nothing we be stored in a DB at that point.
3. Agreed.
4. Agreed.  I don't see much value in versioning a user's displayable facet fields.
5. My point here is that a more flexible store for things like this (user profile settings) may eliminate this upgrade complexity altogether.
6. I'm good with that.

cestella:
Just ducking in here, @merrimanr is this ready for review?  Specifically, the responses to @justinleet 's questions have all been factored into the current code for this PR, right?  If yes, @justinleet would you agree or is there more work to be done to react to your comments?

justinleet:
We're good on 1-4.

For point 5.  I'd still like to see a note in the upgrading, even if it gets removed when the more flexible store is added. I don't know when we'll get around to using a different store, and if it ends up not being this release, we'll need to add this anyway.

I think point 6 still needs to be addressed.

merrimanr:
I will address 6 shortly.  For 5, should we explore a more flexible store in this PR?  Or at least validate that an RDBMS is the right choice?  I think this is something we should tackle now as it will likely affect future work as well.  Does it make sense to start a discuss thread on this topic?

mmiklavc:
I don't see anything that should be problematic in ES 5.6.2, but can you confirm @merrimanr?

mmiklavc:
Bump @merrimanr

merrimanr:
The latest commit switches the persistence for storing user settings to HBase rather than a RDBMS as discussed on the dev list.  Instead of fields being stored in RDBMS columns, the user settings object is now serialized with Jackson and stored in HBase as a byte[].  This required several changes including:

- Added code to setup HBase in the MPack, following the update table pattern in the indexing MPack scripts as an example
- Added default table/cf values to REST properties
- Added a Config class to the REST app to setup an HBase client
- Added a Service class to the REST app that manages user settings in HBase
- Added tests for all new classes and updated existing tests
- Updated the REST README

I also refactored some areas to make things clearer and easier to understand:

- Renamed AlertProfile to UserSettings and moved the endpoints to the UserController.  The thinking is that we may not want this to be limited to just the Alerts UI.
- Refactored the REST MPack scripts to more closely align with other components in regards to setting up Kafka, HBase, etc.
- Removed the Optional fields from a couple model classes.  This was causing issues with Jackson and I don't believe it benefits us enough to have to deal with that complexity.

There are a couple of design issues to consider when reviewing this.  The implementation in the REST app is specific to this use case.  I considered trying to make it more generic for future HBase use cases but decided to keep it simple for now.   Instead of trying to predict what those use cases look like and choose a pattern that works, I decided to leave that to whoever implements a new use case in the future.  I also considered making this generic to all of Metron but again, decided to keep it simple.  Should these settings be limited to just REST/UIs?  Any thoughts on this?

This has been tested in full dev and all tests are passing.  In full dev, navigate to the UserController and use the various endpoints to save and retrieve settings for the currently logged in user.  You should also be able to log in as the admin user and see all user settings and delete an individual user's settings.  I also tested evolving the user settings model by adding new fields and it worked without issue.  Users with existing settings just return null for new fields.

ottobackwards:
First, nice work.
Second, I have this question:

Since we will have some unknown number of rest clients, that may want to save 'configuration' associated with a user, and not just Alerts UI, should it not be factored more generically?

getUserSettingsForType()

And stored as user, type, settings byte[]  or some such?

We may also want to store the version so that it is user, type, settingsFmtVersion, settings.

We might also want to consider storing as Protobufs, which are more supporting changes.

ottobackwards:
Of course re-reading your comment, I see you specifically didn't account for these, but I think there are some basic tweaks to allow for this that don't presuppose far flung 20% use cases.

merrimanr:
I think it's a fair question @ottobackwards.  Anything that might affect how the HBase table is laid out should be worked out now or we're back to altering tables during upgrades.  Currently the row key is the user name, the column family is hardcoded (although configurable), and the column qualifier is hardcoded.  Would you add the type to the row key or store different client settings in different columns?

We could add versioning but that is not trivial and will increase the scope of this PR.  I can take it on here if we decide we want it but this PR is already large.  Follow on maybe?  Either way works for me.  If we decide a follow on is better I would make an effort to do it right away before we do a release.

Jackson is the standard in Metron for serialization/deserialization.  I think we should have a very compelling reason to introduce something new.  Happy to discuss.

ottobackwards:
I am not an expert in hbase, I cannot say how i'd implement it.
The version thing we can leave out, it will be just as good or bad as what we have in zookeeper right?

I think the use case is:

For any given User ( at least ) we need the ability to store currently known and future unknown configurations as bytes, with configuration type identifier lookup.  This should be done such that new configuration types can be added by 'putting' config bytes with a new identifier.

I think that gets us a long way.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/rest_commands.py
```diff
@@ -179,6 +203,35 @@ def status_rest_application(self, env):
...
+
+    def set_hbase_acls(self):
+        Logger.info("Setting HBase ACLs")
+        if self.__params.security_enabled:
+            kinit(self.__params.kinit_path_local,
+                  self.__params.hbase_keytab_path,
+                  self.__params.hbase_principal_name,
+                  execute_user=self.__params.hbase_user)
+
+        cmd = "echo \"grant '{0}', 'RW', '{1}'\" | hbase shell -n"
```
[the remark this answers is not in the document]
merrimanr: Good catch.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/SearchRequest.java
```diff
@@ -101,16 +99,16 @@ public void setSort(List<SortField> sort) {
...
   }
 
   public void setFields(List<String> fields) {
     this.fields = fields;
   }
 
-  public Optional<List<String>> getFacetFields() {
-    return facetFields == null || facetFields.size() == 0 ? Optional.empty() : Optional.of(facetFields);
+  public List<String> getFacetFields() {
+    return facetFields;
```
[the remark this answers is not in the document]
merrimanr: Because null and an empty array mean 2 different things.  See point 6 here:  https://github.com/apache/metron/pull/853#issuecomment-356350839.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/config/HBaseConfig.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.context.annotation.Profile;
+import org.springframework.core.env.Environment;
+
+import java.io.IOException;
+
+import static org.apache.metron.rest.MetronRestConstants.TEST_PROFILE;
+import static org.apache.metron.rest.repository.UserSettingsRepository.USER_SETTINGS_HBASE_TABLE;
+
+@Configuration
+@Profile("!" + TEST_PROFILE)
```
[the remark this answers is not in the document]
merrimanr: We don't want to use this Configuration if tests are running.  There is a separate TestConfig that setups up mock, inmemory components, etc.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
```diff
@@ -37,15 +47,21 @@
 @Service
 public class AlertServiceImpl implements AlertService {
```
ottobackwards: can we make the type metron_alert?  alert seems a little generic

ottobackwards:
This looks great, one small comment from review

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertServiceImpl.java
merrimanr: Sure I can do that.  Since these settings are specific to the Alerts UI, should we include "ui" too?  Should the REST service also follow a similar pattern?

ottobackwards: That seems reasonable

merrimanr: Name is changed in the latest commit.  Let me know what you think.

ottobackwards: I think that is great.  At some point, if we are going to do this a few times we should have a standard naming convention ( separators at least or something).  But that isn't a thing for this PR

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/SearchRequest.java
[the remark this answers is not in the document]
mmiklavc: Ok, sounds reasonable

mmiklavc:
@merrimanr this looks much better. I'm +1 via inspection pending any further feedback from @ottobackwards.

merrimanr:
Any other feedback @ottobackwards or is this ready to go?

ottobackwards:
Sorry, I'll try to get back to this today

PR #854, opened 2017-11-29
Experimental Improvements - Feedback Only - Do Not Merge

nickwallen:
This is a PR to get feedback on some minor improvements that I was experimenting with.  If the community finds any of this useful, I will open a proper PR and JIRA for this work.   All feedback welcome; positive or negative.

Note: This PR should not be merged.

### Faster CI Builds

The first piece here is that I leveraged [Travis' Matrix functionality](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) so that the unit tests, integration tests, web unit tests, and license check runs in parallel rather than sequentially.

Benefit 1: Prior to this change the CI builds take anywhere between 30 - 34 minutes.  After this change the last 4 test builds have taken 24m 49s, 26m 20s, 23m 40s, and 24m 31s.  This is not huge, but seems to be an incremental improvement nonetheless.  

Benefit 2: This also helps when in the future we need need to add additional validation/functionality to the CI build.  Any additional functionality will easily run in parallel and not unnecessarily extend the overall build time.

### Automate Maven Installation

There is a tool called [Maven Wrapper](https://github.com/takari/maven-wrapper) that will automatically install Maven for a user.  And also importantly,  it will install the exact correct version that we specify.  I found this [mentioned in the Travis documentation](https://docs.travis-ci.com/user/languages/java/), so I assume it has a large enough following.

My thought is that anything we can do to reduce friction for a new user to work with Metron is a good thing.  At the very least this simplifies prerequisite installation and can prevent one source of errors.  

Instead of installing Maven manually and then running `mvn clean install`, a user can just run `./mvn clean install`. This will automatically install the correct version of Maven in Windows, Mac, or Linux before building Metron.  This also works correctly in Travis.

I should also mention this is backwards compatible.  If you want to use your own mvn, then feel free to continue on with `mvn clean install`.

The ./mvn and ./mvn.cmd files were generated by me simply running `mvn -N io.takari:maven:wrapper`.  We would run the same sort of command again to update these scripts should we need to.


ottobackwards:
Nick, this is awesome.  I like it a lot.  I don't see any downside it this.  
Do you plan on people using this locally as well?

nickwallen:
Yes, I would just get in the habit of running `./mvn clean install -DskipTests` rather than what we normally do.

justinleet:
Casey and I investigated this previously during one of the times we were running into the upper time limit.

The main reason we ultimately chose not to do this was because we share the build queue with other Apache projects and kicking up multiple builds (at a period in time was backing up in the queues) seemed like it could cause problems for other projects.

ottobackwards:
based on @justinleet 's comment, where do we stand with this

PR #855, opened 2017-11-29
METRON-1338 Excluding retry files from RAT check

nickwallen:
When Vagrant fails, it generates a *.retry file. These files are ignored by Git, but will unnecessarily fail a build because of the Rat check. This has annoyed me to no end.

### For all changes:
  ```


cestella:
+1 by inspection

PR #856, opened 2017-11-30
METRON-1339 Stellar Shell functionality to verify stored stellar statements 

ottobackwards:
This will allow users to check their deployed statements, say after upgrade, when they are at rest ( and would fail on use ).
In other words, they were valid when stored, but are not now because of stellar changes, such as new keywords.

The interface `StellarConfiguredStatementReporter`, which is `@IndexSubclasses` ( ClassIndex) marked, allows the shell to discover reporters that can provide statements for validation.  This discovery allows de-coupling of stellar and 'hosts' that know about the location of the stored statements, and the configuration structure details.

> We do mention the configurations in the shell output at this time.

`metron-common` implements this interface, and can run through visiting all the configurations.

A management function was added ` VALIDATE_STELLAR_RULE_CONFIGS()`
When executed, the shell 

- discovers the reporters through class index 
- visits the reports, with callbacks for visits or errors
- per visit ( which is called for a specific stellar statement ) the statement is compiled and errors reported
- if the entire config fails ( threat triage stellar errors fail on deserialize so we don't get to do ANY enrichment visits in that case ) the error callback handles that

I'm getting this out there, still a couple of things todo:

[x] ~~full dev run. I have been testing with stellar external to full dev iteratively~~
[x] ~~readme~~
[x] ~~steps to test~~
[x] ~~unit test~~
[x] ~~ThreatTriage Rule Reason~~

## Testing
- deploy full dev
- edit the squid parser transformation(s) such that the stellar would not compile, such as adding a dangling  `=` in zookeeper
```json
{ 
"parserClassName": "org.apache.metron.parsers.GrokParser", 
"sensorTopic": "squid", 
"parserConfig": { 
"grokPath": "/patterns/squid", 
"patternLabel": "SQUID_DELIMITED", 
"timestampField": "timestamp" 
}, 
"fieldTransformations" : [ 
{ 
"transformation" : "STELLAR" 
,"output" : [ "full_hostname", "domain_without_subdomains" ] 
,"config" : { 
"full_hostname" : "URL_TO_HOST(url) =" 
,"domain_without_subdomains" : "DOMAIN_REMOVE_SUBDOMAINS(full_hostname)" 
} 
} 
] 
}

```

- edit the snort threat triage rules in it's enrichment config in zookeeper ( here with an extra `)` )

```json
{ 
"enrichment" : { 
"fieldMap": 
{ 
"geo": ["ip_dst_addr", "ip_src_addr"], 
"host": ["host"] 
} 
}, 
"threatIntel" : { 
"fieldMap": 
{ 
"hbaseThreatIntel": ["ip_src_addr", "ip_dst_addr"] 
}, 
"fieldToTypeMap": 
{ 
"ip_src_addr" : ["malicious_ip"], 
"ip_dst_addr" : ["malicious_ip"] 
}, 
"triageConfig" : { 
"riskLevelRules" : [ 
{ 
"rule" : "not(IN_SUBNET(ip_dst_addr, '192.168.0.0/24')) )", 
"score" : 10 
} 
], 
"aggregator" : "MAX" 
} 
} 
} 
```

## Working with zookeeper
I am not a zk cli maestro, so I took the easy way out and used [ZK-WEB](https://github.com/qiuxiafei/zk-web).
Following the readme instructions it was very simple to clone, edit the config for full dev, and run from source.  If you log in with the creds in the config you can edit the nodes.

## Results
When you run the function, it will report the failed stellar statements.
Examples with and without :
> NOTE passing a wrap value may be necessary for better visibility as shown

```bash
[Stellar]>>> VALIDATE_STELLAR_RULE_CONFIGS()
╔═══════════════════════════════════════════════════════════════════════════════════════════╤═══════════════════════════════════════════════╤═══════╤═══════╗
║ PATH                                                                                      │ RULE                                          │ ERROR │ VALID ║
╠═══════════════════════════════════════════════════════════════════════════════════════════╪═══════════════════════════════════════════════╪═══════╪═══════╣
║ Apache Metron/PARSER/squid/fieldTransformations/0/Field Mapping/full_hostname             │ URL_TO_HOST(url)                              │       │ true  ║
╟───────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────┼───────┼───────╢
║ Apache Metron/PARSER/squid/fieldTransformations/0/Field Mapping/domain_without_subdomains │ DOMAIN_REMOVE_SUBDOMAINS(full_hostname)       │       │ true  ║
╟───────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────┼───────┼───────╢
║ Apache Metron/ENRICHMENT/snort/threatIntel/triageConfig/riskLevelRules/0/rule             │ not(IN_SUBNET(ip_dst_addr, '192.168.0.0/24')) │       │ true  ║
╚═══════════════════════════════════════════════════════════════════════════════════════════╧═══════════════════════════════════════════════╧═══════╧═══════╝

[Stellar]>>>

[Stellar]>>> VALIDATE_STELLAR_RULE_CONFIGS(40)
╔════════════════════════════════════════════════════════════════════════╤══════════════════════════════════════════╤═══════════════════════════════════════╤═══════╗
║ PATH                                                                   │ RULE                                     │ ERROR                                 │ VALID ║
╠════════════════════════════════════════════════════════════════════════╪══════════════════════════════════════════╪═══════════════════════════════════════╪═══════╣
║ Apache                                                                 │ URL_TO_HOST(url)=                        │ Syntax error @ 1:16 token recognition │ false ║
║ Metron/PARSER/squid/fieldTransformations/0/Field                       │                                          │ error at: '='                         │       ║
║ Mapping/full_hostname                                                  │                                          │                                       │       ║
╟────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────┼───────────────────────────────────────┼───────╢
║ Apache                                                                 │ DOMAIN_REMOVE_SUBDOMAINS(full_hostname)= │ Syntax error @ 1:39 token recognition │ false ║
║ Metron/PARSER/squid/fieldTransformations/0/Field                       │                                          │ error at: '='                         │       ║
║ Mapping/domain_without_subdomains                                      │                                          │                                       │       ║
╟────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────┼───────────────────────────────────────┼───────╢
║ Apache                                                                 │ not(IN_SUBNET(ip_dst_addr,               │                                       │ true  ║
║ Metron/ENRICHMENT/snort/threatIntel/triageConfig/riskLevelRules/0/rule │ '192.168.0.0/24'))                       │                                       │       ║
╚════════════════════════════════════════════════════════════════════════╧══════════════════════════════════════════╧═══════════════════════════════════════╧═══════╝

```

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```


cestella:
Any chance we can add a `VALIDATE(str, type)` function to the stellar management functions where str is the json blob string for the config and the type is the type of config?  Generally the goal is to disallow invalid stellar to get pushed to zookeeper via `zk_load_utils.sh`, so I suspect a function would be more useful in the situation where you're constructing a config in the REPL via management functions and want to validate it before pushing it.

cestella:
Also, it might be useful for `%validate_configured_expressions` to take a file path so you can validate a set of configs on disk (again, if it gets to zookeeper, zk_load_utils.sh should fail if it's invalid)

simonellistonball:
@cestella I would say that proposed validate function has to be very much in a namespace. It feels like a name that would be much more useful for a function replacing our current approach to global validation in the future than config validation, other than that it sounds like a good idea.

ottobackwards:
So, the scenario here is checking things that *were* valid when uploaded, but have been invalidated by external changes ( language changes ).  I would like to keep the magic specific.

I think the functionality for the management functions is valid, but can we do that as a separate Jira/PR?  I'll do it, I just want to keep this tight.  If you create the jira and assign it to me that would be super.

I would do the files on disk using the management functions as well.

So we just need to think of the stellar interface for calling
`VALIDATE`  with a string, and with a file path.  Also saying what configuration type it is.

Does that make sense?

ottobackwards:
@simonellistonball, yes, the namespace should be part of the jira and interface design

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/StellarStatementReporter.java
```diff
@@ -0,0 +1,166 @@
...
+    try {
+      List<RiskLevelRule> riskLevelRules = threatTriageConfig.getRiskLevelRules();
+      riskLevelRules.forEach((r) -> {
+        String name = r.getName();
+        if (org.apache.commons.lang.StringUtils.isEmpty(name)) {
+          name = "(default)";
+        }
+        List<String> names = Arrays
+            .asList(getName(), ENRICHMENT.toString(), topicName, "THREAT_TRIAGE", name);
+        visitor.visit(names, r.getRule());
```
nickwallen: I like what you are doing with this PR.

There is just one downside that bugs me. How do we ensure that as our configuration evolves over time that we also update this to ensure it gets validated?  

We have Stellar expressions as configurations all over the place; parsing, enrichment, triage, indexing, and profiler.  I feel like any/all of these will evolve and change over time.  I would definitely forget to come here and update this to make sure it gets validated.

As a motivating example, right here you are only validating the `rule` field.  But triage also has the `reason` field that is a Stellar expression.  That would need to be validated also.  That was something we added later and likely the scenario where I would forget to add that expression to your validation logic.

nickwallen: And let me follow-on with one proposal that might help address this.  There are probably other (better?) ways to solve this, but here is one approach to chew on.

All of our configuration gets deserialized from JSON into POJOs; EnrichmentConfig, ProfilerConfig, etc.  What if we had an annotation that we marked the fields that are required to be valid Stellar expressions?  Based on the annotation we can then validate each of those fields.

Since the annotation is directly in the configuration classes, it is less likely I'm going to forget that annotation.  And it is also remains decoupled, which is a good benefit of your current approach.

ottobackwards: I need to think about this, but I will add the reason check

nickwallen: This also makes it simple to validate the configuration, no matter where it rests.  

To validate the configuration in Zk, you just read all the configs in from Zk, deserialize into POJOs, iterate all the fields looking for the annotation, if you find the annotation, then run your validation logic on the expression.

To validate the configuration on the file system, you just read it all in from the FS, then repeat the same steps I described in the Zk scenario.

ottobackwards: So - if I'm understanding you correctly, as it applies to what I am doing ->
when visiting configs, instead of explicitly validating fields, we would want to 'visit' all the members per pojo by attribute.  We would only need to worry about tracking the pojos but not the fields ( we need to keep the context ).  Am I understanding you correctly?

ottobackwards: When I say track pojo's I mean the reporter would still 'know' about the Configuration object, but would discover what fields where stellar.  Know the object but not the fields.

ottobackwards: you would either have to 'know' about fields that were other nested types or do a recursive desert through the pojo though

nickwallen: > you would either have to 'know' about fields that were other nested types or do a recursive desert through the pojo though

Apache Commons has a `FieldUtils.getFieldsListWithAnnotation(...)`.  Once you get a 'configuration' class, then you just use that to get all the fields that have been marked with the annotation.

nickwallen: > We would only need to worry about tracking the pojos but not the fields ( we need to keep the context ). Am I understanding you correctly?

Yes, exactly.  You can use the functionality in `ConfigurationUtils` and just do something like....
``` 
config = getProfilerConfig(...)
validateStellar(config)

config = getSensorConfig(...)
validateStellar(config)

config = getEnrichmentConfig(..)
validateStellar(config)

...
```

Or something to that effect.

ottobackwards: does profiler have stellar?  ( sorry to have to ask ).

cestella:
@simonellistonball Agree to the namespace idea.  My bad :)

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/StellarStatementReporter.java
ottobackwards: True, but FieldUtils does a class, I don't think it recurses.
So we would still need to do some work

nickwallen: > does profiler have stellar? ( sorry to have to ask ).

No problem. :)  Yes, it is mostly Stellar.  The docs are pretty good on that one, but I can outline specifically which ones if that helps.

nickwallen: > True, but FieldUtils does a class, I don't think it recurses.  So we would still need to do some work

I think the logic would be something like...
* get the configuration object
* find what class the configuration object is
* use `FieldUtils..getFieldsListWithAnnotation` to get all the fields that are Stellar
* then for those fields that are Stellar, get the value from the configuration object 
* that value is the stellar expression.  validate it.

ottobackwards: right, for SensorEnrichmentConfig, I would have to know about the 'inner' config objects and do them to, just as I do now.

nickwallen: Gotcha.  Understand your point now.  Seems like we should be able to handle that in a generic way for any configuration object.

ottobackwards: @nickwallen, checking the reason field....  Because the rules are checked on deserialize, if they are wrong, the whole config fails ( the entire Enrichment ).  You *could* say that the checks that I do have are more than I need, and that rather than add reason, I should just get rid of the rules stuff all together.  What do you think?

ottobackwards: @nickwallen you know what would work better maybe?  if we could attribute the JSON.

nickwallen: > because the rules are checked on deserialize, if they are wrong, the whole config fails ( the entire Enrichment )

  Only the syntax is checked on deserialization now.  If I add a field that we don't expect, we will catch that.  But if I use invalid Stellar; example use a variable named 'default', we will NOT catch that.  That's what your work comes in.  

I am not sure that I am understanding your suggestion though.  Please correct me if this is a miss.

nickwallen: > you know what would work better maybe? if we could attribute the JSON.

Maybe not?  Users/admins of the system create the JSON.  I don't want them to annotate which field is Stellar. I want the developers, the creators of the extension points, to define which fields are Stellar.

ottobackwards: Nick, when the ThreatTriage is loaded from json, the call to setRiskLevelRules() gets called by jackson, which validates the stellar, which will throw the exception at load time, before I can do the check myself.

So I have code to verify the rules, but the fact is if they are bad stellar, they will throw on load before that code will execute.

nickwallen: Sure, I see in that case validation does exist.  And this highlights an alternative approach.  Add custom code like that to each of the POJOs to validate any Stellar fields.  Personally, I like the annotation based approach.  

But whichever way we do it, we'd probably want to see it done the same across the board.  There are many other places where no validation of Stellar fields happen, right?

ottobackwards: I think we eventually want to get to that point, with a consistent, comprehensive mechanism to validate stellar statements ( short of compiling ;) ).

ottobackwards:
I am glad for the interest in this PR, and that it seems to have sparked some great ideas for continuing on.  

What I would like to do is line it up as follows

1. This PR with it's current scope and focus
2. New Jira and PR(s) for the Stellar Functions/Namespace that @cestella  and @simonellistonball  mentioned
3. Some research and possible prototyping of the Attributed approach @nickwallen has suggested ( which I agree with )

Over the course of that work, and other work identified through working and reviewing it, I will iteratively refactor to a common code and reusable approach.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/StellarStatementReporter.java
ottobackwards: I *think* I know how to do it and not have it be terrible.  The question is if it should be in this PR or after

ottobackwards: I think I can do it such that the stellar-common holds the code for doing the traversals as well as the interfaces and callbacks, and the *reporter* just provides the configs.

I'm thinking it through, but now that I'm thinking of it I will be miserable until I do it so I'll do it.

ottobackwards:
Per the conversation above, i'm going to take a stab at the attributed approach.
I think the Stellar Functions should be a separate Jira.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/StellarStatementReporter.java
ottobackwards: I think you are really going to like this @nickwallen

ottobackwards: but if you complain about scope or the review size @nickwallen I will be very cross

ottobackwards:
https://github.com/apache/metron/pull/856/commits/65278a67a07f1c4c23ab2d95ebb6de92e1cac731 introduces conceptually what @nickwallen  and I have been discussing.
I need to think about reworking the description.

From the commit ->
----------------
Refactor based on review and inspiration from review.
Although the original implementation was functional, it required maintainence to keep current.
The suggested 'best state' was to have it be possible, maybe through annotations, for the validation
system to be able to handle any config, regarless or composition using annotations.
That would leave it up to the implementor to propertly annotate thier configurations, and allow for support of new fields.

This is an implementation of that.

I have refactored the implemenations and details, but kept the discovery and mechanics ( loading and visitation ) somewhat the same.
Hopefully keeping the good and reworking to a more sustainable solution.

Several annotations where created to marks ceratin stellar configruation objects or scenarios.
A holder object, to hold the configuration object, but knows how to process the annotations and run the visitation was added.
This holder object and the annotations have parameters and handling for several special scenarios, such as 2x nested maps.

This implementation should facilitate follow on work to validate files and streams and blobs by using implementing the StellarValidator interface
and re-using the holder concept ( replacing the providers )

ottobackwards:
stellar_validate

ottobackwards:
stellar_validate

ottobackwards:
stellar_validate

ottobackwards:
@cestella @simonellistonball 
With the new implementation, doing the blob or file check should be a piece of cake.... would you prefer it as part of this or as a new issue?

ottobackwards:
@nickwallen @mattf-horton I think we can use the annotation approach to resolve [METRON-989](https://issues.apache.org/jira/browse/METRON-989) as well.

Thoughts?

ottobackwards:
Closing to test build in travis

ottobackwards:
Bump?

ottobackwards:
@nickwallen any feedback, does the annotated approach match what you imagined?

nickwallen:
@ottobackwards Yes, definitely.  I like it at a 50k foot level.  The only thing that struck me was the need for the different annotation types.  But I haven't had a chance to dig into it yet.

ottobackwards:
@nickwallen yeah, we need to cover a bunch a scenarios

nickwallen:
Why is this validation process driven by a %magic command?  

Magics were made for functionality that cannot be implemented directly within a Stellar execution environment.  Often for answering questions 'about' the execution environment itself.  For example `%vars` or `%functions` or `%globals` all tell us about the Stellar execution environment.

This logic doesn't seem like a good fit for a %magic, unless there is a limitation that I am not understanding.  This could be implemented, I would argue more simply, as a regular Stellar function.

ottobackwards:
It did not seem appropriate to me for this to be a stellar function, and %magic is the other way to execute things from the shell.  At the time at least.

ottobackwards:
In other words, a stellar function that called the stellar compilation stuff, did not seem correct.

ottobackwards:
Do you feel strongly that this should be a Function?  @cestella ?  I'm not opposed to changing it if you are.  I would like to here some more feedback

nickwallen:
If that question is to me too, yes, I feel strongly it should be a function.  A function that is part of the management functions. That was also suggested previously [here](https://github.com/apache/metron/pull/856#issuecomment-348208274).

nickwallen:
The justification that you mentioned just doesn't seem strong enough to me.  Unless there is more that I am missing.

IMHO We should only use magic commands for things that can't be accomplished in the language using the preferred extension mechanism; aka defining Stellar functions.

ottobackwards:
Ok, I'll change it.  Feels a little crossing the streams, but we'll see

ottobackwards:
I don't think it should be in management necessarily though.

ottobackwards:
@nickwallen I have refactored to a function.
see updated PR description

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/StellarConfiguredStatementContainer.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.metron.stellar.common.utils.validation;
+
+/**
+ * StellarConfiguredStatementProviders are used provide stellar statements
+ * and the context around those statements to the caller
+ */
+public interface StellarConfiguredStatementContainer {
```
nickwallen: `ExpressionConfigurationHolder` is an implementation of this interface. That being said, I don't understand the point of this interface.  

In all the code that I see, you use the implementation class `ExpressionConfigurationHolder` rather than this interface.  For example, in `StellarZookeeperBasedValidator` and other places.  

We should either use the interface or get rid of it.

metron-platform/metron-management/README.md
```diff
@@ -276,6 +278,13 @@ The functions are split roughly into a few sections:
     * aggregatorConfig - Optional config for aggregator
   * Returns: The String representation of the enrichment config
 
+### Validation Functions
+* `VALIDATE_STELLAR_RULE_CONFIGS`
```
nickwallen: This name confuses me, why Rules?  `VALIDATE_STELLAR` seems simple and to the point to me.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/StellarZookeeperBasedValidator.java
```diff
@@ -0,0 +1,106 @@
...
+    this.client = client;
+  }
+
+
+  @Override
+  public Iterable<ValidationResult> validate(Optional<LineWriter> writer) {
+    // discover all the StellarConfigurationProvider
+    Set<StellarConfigurationProvider> providerSet = new HashSet<>();
+
+    for (Class<?> c : ClassIndex.getSubclasses(StellarConfigurationProvider.class,
```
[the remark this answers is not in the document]
nickwallen: If I understand this correctly, it seems that the `StellarConfigurationProvider` interface allows us to extend where configuration values get pulled in from.  In your current default implementation `ConfigurationProvider` you reach out to Zookeeper to pull in the config.

If I wanted to validate configuration located on a file system, I would just create a `FilesystemConfigurationProvider` implementation of this interface.

The decision as to whether I want to validate the config in Zookeeper, on the file system or both, needs to be user driven.  A user should make that decision based on how they call your new Stellar function.  

Based on your discovery logic here, just having a `FilesystemConfigurationProvider` on the classpath (or any other implementation) will cause the configuration in the file system to get validated.  We don't want that to happen.  We want the user to control that behavior.

So I don't think we really need this discovery logic, which nicely simplifies things.  I think we could just alter the `StellarValidater` interface to make this relationship simpler and more straight forward.  The `StellarConfigurationProvider` just gets passed in.

```
StellarValidator {
   ...
   Iterable<ValidationResult> validate(StellarConfigurationProvider provider);
   ...
}
```

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/StellarZookeeperBasedValidator.java
```diff
@@ -0,0 +1,106 @@
...
+  public StellarZookeeperBasedValidator(CuratorFramework client) throws NullArgumentException {
+    if (client == null) {
+      throw new NullArgumentException("client");
+    }
+    this.client = client;
+  }
+
+
+  @Override
+  public Iterable<ValidationResult> validate(Optional<LineWriter> writer) {
```
nickwallen: Why do we need a 'writer'?  It doesn't make sense to me why we need it nor do we use it.  I think it could be removed from the method signature completely in `StellarValidator`.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationProvider.java
```diff
@@ -0,0 +1,144 @@
...
+import org.apache.metron.stellar.common.utils.validation.ExpressionConfigurationHolder;
+import org.apache.metron.stellar.common.utils.validation.StellarConfigurationProvider;
+import org.apache.metron.stellar.common.utils.validation.StellarConfiguredStatementContainer.ErrorConsumer;
+import org.apache.zookeeper.KeeperException.NoNodeException;
+
+/**
+ * {@code ConfigurationProvider} is used to report all of the configured / deployed Stellar statements in
+ * the system.
+ */
+public class ConfigurationProvider implements StellarConfigurationProvider {
```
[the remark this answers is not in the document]
ottobackwards: That is a good idea

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationProvider.java
```diff
@@ -0,0 +1,144 @@
...
+    return "Apache Metron";
+  }
+
+  @Override
+  public List<ExpressionConfigurationHolder> provideConfigurations(CuratorFramework client,
+      ErrorConsumer errorConsumer) {
+    List<ExpressionConfigurationHolder> holders = new LinkedList<>();
+    visitParserConfigs(client, holders, errorConsumer);
+    visitEnrichmentConfigs(client, holders, errorConsumer);
+    visitProfilerConfigs(client, holders, errorConsumer);
```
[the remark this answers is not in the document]
ottobackwards: There is no stellar in indexing is there?  It is not in the readme

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationProvider.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  private void visitParserConfigs(CuratorFramework client,
+      List<ExpressionConfigurationHolder> holders, ErrorConsumer errorConsumer) {
+    List<String> children = null;
+
+    try {
+      children = client.getChildren().forPath(PARSER.getZookeeperRoot());
+    } catch (Exception nne) {
+      return;
```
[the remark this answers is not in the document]
ottobackwards: Will do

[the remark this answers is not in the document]
ottobackwards: will do

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationProvider.java
```diff
@@ -0,0 +1,144 @@
...
+
+  @SuppressWarnings("unchecked")
+  private void visitEnrichmentConfigs(CuratorFramework client,
+      List<ExpressionConfigurationHolder> holders, ErrorConsumer errorConsumer) {
+    List<String> children = null;
+
+    try {
+      children = client.getChildren().forPath(ENRICHMENT.getZookeeperRoot());
+    } catch (Exception nne) {
+      return;
```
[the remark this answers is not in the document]
ottobackwards: Will do

metron-platform/metron-management/README.md
ottobackwards: Is you search the tree for 'rules' you will see that we call them rule or rules in various places.  It 'seemed like the thing to do'.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationProvider.java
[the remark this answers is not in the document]
nickwallen: I *think* you're right.  Would suggest just a comment to clarify that point and maybe help prompt us should that change
```
// indexing contains no stellar to validate
```

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/StellarConfiguredStatementContainer.java
ottobackwards: The issue is not using the interface, I'll address that.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/StellarZookeeperBasedValidator.java
ottobackwards: Sorry, I was going to take this out, this is from the prior shell based integration

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/StellarZookeeperBasedValidator.java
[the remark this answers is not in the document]
ottobackwards: The idea is that not only can we not know the details of the classes that hold the rules, but also, that stellar may be hosted by other things than metron, that 'know' how to provide those configurations.  

The problem with this isn't the discovery per se, but in that it is not correct given it's purpose and the implementation

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/annotations/StellarExpressionMap.java
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * A map may be a StellarExpressionMap based on the type
+   * of another field, this is that field's name.
+   *
+   * {@code} qualify_with_field_type} is the type of that field
+   *
+   * @return Field Name or empty String
+   */
+  String qualify_with_field() default "";
```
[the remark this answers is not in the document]
ottobackwards: oops

ottobackwards:
Refactored based on feedback for some things, based on making what I was trying for more correct in others.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/StellarZookeeperBasedValidator.java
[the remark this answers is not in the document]
nickwallen: I am not following your explanation of why we need the discovery logic.  Can you try to explain it again?  In the latest commits, I still see the discovery logic in StellarSimpleValidator.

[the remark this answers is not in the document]
ottobackwards: I will try to think of a different way to put it

metron-platform/metron-management/README.md
nickwallen: There is reference to "rules" because that was the first use for Stellar.  @cestella called them rules when he first implemented.  But that is ancient history.  We use Stellar everywhere now.  Its long ago outgrown that.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/validation/StellarZookeeperBasedValidator.java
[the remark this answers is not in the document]
ottobackwards: OK, I think I understand where you are coming from.  In the end, there will be more functions for validation.   Those which take in a passed in object, like CONFIG_GET() -> obj, EDIT -> obj, VALIDATE_CONFIG(obj).  Those functions will be user driven, ie.  __**Validate this thing I gave you__**.

These functions are __**Validate everything the system knows about**.  And are born out of the introduction of possible errors through language changes in already deployed stellar rules.

That is why we need discovery.  The "job" of this function at the moment is to find everything and make sure it is still OK.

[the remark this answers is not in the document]
nickwallen: I think you're not understanding my point.   More specifically... What happens if another implementation gets added to the classpath?  

> Based on your discovery logic here, just having a FilesystemConfigurationProvider on the classpath (or any other implementation) will cause the configuration in the file system to get validated. We don't want that to happen. We want the user to control that behavior.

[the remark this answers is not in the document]
ottobackwards: They would automatically be picked up.

I'm confused.  This PR as stated is to help validation of stellar configurations, where ever they may be.  it is not for the user to hunt and peck and pick a configuration.  So that is kind of the point.

Although there have been comments about doing configuration validation with specific objects or files passed in, that is not the thrust of this pr, even if this pr enables it.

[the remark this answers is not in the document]
nickwallen: But there is no need to discover those, right?  Why have the extra discovery logic when we don't need it?  

It is also more confusing from a user's perspective.  What I have on my class path decides what is going to get validated.  It is often very hard to control what gets added to a classpath.  That might lead to unexpected behavior.

If you just removed the discovery logic and instantiated the `StellarConfigurationProvider`'s that you want to use directly...

1. the functionality would work the same way, all the time, no matter what is on my class path.
1. the implementation would be simpler.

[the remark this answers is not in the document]
ottobackwards: You mean couple it to metron?  We don't want to do that anymore.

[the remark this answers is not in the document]
ottobackwards: Maybe that is where we are missing each other?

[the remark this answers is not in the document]
nickwallen: > You mean couple it to metron? We don't want to do that anymore.

I don't see your justification.  Maybe another reviewer will understand this better.  As is, I am a +0 on this.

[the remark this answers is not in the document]
ottobackwards: https://issues.apache.org/jira/browse/METRON-989
https://issues.apache.org/jira/browse/METRON-876

@mattf-horton @cestella 

I have implemented this as to not increase the amount of tie-in between stellar and metron, and support future non-metron configuration sources. 

Thus, the sources of the configuration are discovered and not coupled.  I believe this is in the spirit if not the letter of the design discussions that we have had.

Can you take a look?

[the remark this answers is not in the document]
cestella: It is true that Stellar should exist, as much as possible, independent from Metron; that was the aim of the effort to move stellar out of metron-common and into its own top level component in the project.  I'll look closer at this PR and the (rather long, but seemingly coherent...so congrats ;) comment thread and weigh in later.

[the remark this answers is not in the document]
ottobackwards: @cestella ping

ottobackwards:
Resolved conflicts, in the event we ever get around to this

ottobackwards:
This has gone from a small thing to at least 'say' we have a way to check if we broke all your stellar stuff after upgrade, to stretching it based on feedback which was a mistake, to a run around to abandoned review status.

I'm going shelve this

ottobackwards:
The next attempt at this, if there is one, should start off with some sort of consensus first. And some agreement on initial scope.  This PR would have been smaller and less ambitious, if that were true.  Or at least everyone would have been on the same page as to the 'why' of certain things.
Lessons to learn :)

ottobackwards:
It has been long enough that **I** don't even like this PR any more.

-1

PR #861, opened 2017-12-07
METRON-1341 Implemented SELECT transformer to project fields from parser

simonellistonball:
## Contributor Comments
This is a simple PR to add FieldTransformation capabilities to the parsers allowing basic projection. There are definitely better ways to implement this as reflected in the comments, but those would need a wide ranging refactor of FieldTransformations, so I've decided to shelve that for now. 

Steps to check:
1. Create a sensor with the fieldTransformation set per
```
{
...
    "fieldTransformations" : [
          {
            "output" : ["field1", "field2" ] 
          , "transformation" : "SELECT"
          }
                      ]
}
```
Pass in data with fields field1, field2 and some other fields. 

Ensure that the data coming out of that parser only contains field1, field2, and the timestamp and original_string fields.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/SelectTransformation.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/SelectTransformation.java
@@ -29,7 +29,7 @@ import org.apache.metron.stellar.dsl.Context;
 
 public class SelectTransformation implements FieldTransformation {
 
-	private static final List<String> systemFields = Arrays.asList("timestamp", "original_string");
+	private static final List<String> systemFields = Arrays.asList("timestamp", "original_string", "source.type");
 
 	@Override
 	public Map<String, Object> map(Map<String, Object> input, List<String> outputField,
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/SelectTransformation.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/SelectTransformation.java
@@ -18,6 +18,7 @@
 
 package org.apache.metron.common.field.transformation;
 
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -28,6 +29,8 @@ import org.apache.metron.stellar.dsl.Context;
 
 public class SelectTransformation implements FieldTransformation {
 
+	private static final List<String> systemFields = Arrays.asList("timestamp", "original_string");
+
 	@Override
 	public Map<String, Object> map(Map<String, Object> input, List<String> outputField,
 			LinkedHashMap<String, Object> fieldMappingConfig, Context context, Map<String, Object>... sensorConfig) {
@@ -37,13 +40,15 @@ public class SelectTransformation implements FieldTransformation {
 
 		// note, this cannot be implemented with streams because HashMap merge guards
 		// against null values
-		
-		HashMap<String, Object> output = new HashMap<String,Object>();
+
+		HashMap<String, Object> output = new HashMap<String, Object>();
 		for (Entry<String, Object> e : input.entrySet()) {
 			if (outputField.contains(e.getKey())) {
 				output.put(e.getKey(), e.getValue());
 			} else {
-				output.put(e.getKey(), null);
+				if (!systemFields.contains(e.getKey())) {
+					output.put(e.getKey(), null);
+				}
 			}
 		}
 		return output;
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/SelectTransformation.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/SelectTransformation.java
@@ -1,0 +1,52 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.metron.common.field.transformation;
+
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import org.apache.metron.stellar.dsl.Context;
+
+public class SelectTransformation implements FieldTransformation {
+
+	@Override
+	public Map<String, Object> map(Map<String, Object> input, List<String> outputField,
+			LinkedHashMap<String, Object> fieldMappingConfig, Context context, Map<String, Object>... sensorConfig) {
+		// note that we have to set the value to null for any field not in the output
+		// list, since FieldTransformer will otherwise put them back again from the
+		// original output.
+
+		// note, this cannot be implemented with streams because HashMap merge guards
+		// against null values
+		
+		HashMap<String, Object> output = new HashMap<String,Object>();
+		for (Entry<String, Object> e : input.entrySet()) {
+			if (outputField.contains(e.getKey())) {
+				output.put(e.getKey(), e.getValue());
+			} else {
+				output.put(e.getKey(), null);
+			}
+		}
+		return output;
+	}
+
+}
--- a/metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/SelectTransformationTest.java
+++ b/metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/SelectTransformationTest.java
@@ -71,5 +71,26 @@ public class SelectTransformationTest {
 		Assert.assertFalse(input.containsKey("field3"));
 		Assert.assertEquals(2, input.size());
 	}
+	
+	@Test
+	public void testPreserveSystemFields() throws Exception { 
+		SensorParserConfig sensorConfig = SensorParserConfig.fromBytes(Bytes.toBytes(selectSingleFieldConfig));
+		FieldTransformer handler = Iterables.getFirst(sensorConfig.getFieldTransformations(), null);
+		JSONObject input = new JSONObject(new HashMap<String, Object>() {
+			{
+				put("timestamp", 12345);
+				put("original_string", "foo,bar");
+				put("field1", "foo");
+				put("field2", "bar");
+			}
+		});
+		handler.transformAndUpdate(input, Context.EMPTY_CONTEXT());
+		
+		Assert.assertTrue(input.containsKey("timestamp"));
+		Assert.assertTrue(input.containsKey("original_string"));
+		Assert.assertTrue(input.containsKey("field1"));
+		Assert.assertFalse(input.containsKey("field2"));
+		Assert.assertEquals(1, input.size());
+	}
 
 }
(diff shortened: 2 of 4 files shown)


ottobackwards:
Thanks Simon. Before a code review, a couple of questions:

I see a check next to steps to verify manually, but I don't see the steps.

This is also missing documentation.

metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/SelectTransformation.java
```diff
@@ -0,0 +1,52 @@
...
+package org.apache.metron.common.field.transformation;
+
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import org.apache.metron.stellar.dsl.Context;
+
```
ottobackwards: Missing Javadoc.  The purpose and an example output of this transformation would be nice

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/SelectTransformationTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.metron.common.configuration.FieldTransformer;
+import org.apache.metron.common.configuration.SensorParserConfig;
+import org.apache.metron.stellar.dsl.Context;
+import org.json.simple.JSONObject;
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.google.common.collect.Iterables;
+
```
ottobackwards: Do we at a test where we have multiple transformers?
STELLAR , SELECT etc?

simonellistonball: that feels a lot more like an integration test, or a broader test than belongs in this one piece. Otherwise every single test of transformations is going to end up with combinatorial complexity.

simonellistonball:
Somehow I knew you were going to say something about docs.... will add.

metron-platform/metron-parsers/README.md
```diff
@@ -216,6 +216,23 @@ whenever `field2` exists and whose corresponding equal to 'foo':
...
+          {
+            "output" : ["field1", "field2" ] 
+          , "transformation" : "SELECT"
+          }
+                      ]
+}
+```
+
+when applied to a message containing keys field1, field2 and field3, will only output the first two.
+
```
ottobackwards: How does this relate to REMOVE?  If I have more than one transformation should this be last or does it not matter?

simonellistonball:
It suddenly occurs to me that we should probably whitelist the original_string and timestamp fields, so that these are always kept by this transformation. Does that make sense?

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/SelectTransformationTest.java
ottobackwards: or a FieldTransformer ( level up ) test more likely.

metron-platform/metron-parsers/README.md
simonellistonball: This is the opposite of remove.

simonellistonball: And order doesn't really matter, you might have this first, or last, or sandwiched between two stellar blocks. That said, I would expect the most common use case to be have this one at the end of a set.

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/SelectTransformationTest.java
simonellistonball: yeah, makes sense, I suspect on a different ticket.

ottobackwards:
That is an excellent catch @simonellistonball, a test to go with it too ;)

metron-platform/metron-common/src/main/java/org/apache/metron/common/field/transformation/SelectTransformation.java
simonellistonball: I've added to relevant readme (in parsers docs) which would seem a more likely place than javadoc (if we want to add javadoc, which we don't publish anyway, we should also look at adding this to everything in this area, which feels out of scope). Not even the stellar version gets javadoc!

metron-platform/metron-parsers/README.md
ottobackwards: If a user has a stellar transform (existing ) that takes field X as input, but doesn't want X as output, and puts the SELECT in the wrong order, could they break their existing transform because X now missing?

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/SelectTransformationTest.java
ottobackwards: That test would belong in this ticket

metron-platform/metron-parsers/README.md
simonellistonball: Yes, that would break. I'd say that was flat out user error, and not something to guard against, or indeed something it's possible to guard against. I guess the key thing here is that it requires maintaining order in field transformations.

simonellistonball:
Right, that should cover it.

metron-platform/metron-common/src/test/java/org/apache/metron/common/field/transformation/SelectTransformationTest.java
simonellistonball: It's a pretty broad reaching concern when you account for the impact of remove, stellar, and the other field transforms. I'm happy to look at it, but I expect it will just need to be reviewed again as part of a follow on.

metron-platform/metron-parsers/README.md
ottobackwards: It should be documented then in the readme

ottobackwards: We should document that the system fields are protected as well

simonellistonball: agreed, and added.

ottobackwards:
I trying to run this up in full dev.  I have two issues:

1.  expressJS is failing download, obviously outside this PR
2. The usability of this transform.  If a user just wants a couple of fields, then no problem, but If they want anything more than that, then selecting all the fields that they want seems like a bit of a chore.  So if they want to select *most* of the fields, they have to list them all out.   It seems tough.  I wonder if we don't need regex support or something.  

Thoughts?

simonellistonball:
If they want to select most, and remove the ones they don't want, then I would recommend using the remove transformation, or a set null in stellar. Perhaps regex support might be a nice follow on, but it breaks the mental model of people who are used to any other language that handles projection, such as SQL. The goal for this was to allow user to explicitly select only a defined set of fields. To be honest, people have lived with the idea of explicitly choosing fields for decades in SQL and quite liked it, so I suspect adding something that is pattern based might make it less usable than more.

ottobackwards:
Sure, that makes sense.  I am just trying to wrap my head around why I would use this based on the data I would send through.  I don't have to understand it for it to be useful ;)

Do you have a use case or scenario that would help me?

simonellistonball:
A typical case might be something like the CEF parser. You could potentially kick out a lot of fields you really don't care about, which at scale can produce huge amounts of ES and HDFS storage (in addition to the original_string representation. The goal for this is to focus on just outputting fields which match active use cases in the rest of the flow to control data storage costs and data clarity.

This also allows you to map an explicit data model and provide some governability to the data model.

simonellistonball:
Actually, to follow up on that... I have a proxy feed, and some proxy use cases (enrichment, profile, etc). I want to keep my data clean and be explicit about which fields I pass on, so I have a 'library' field set that means "proxy like stuff", these are the fields I push into the output here.

ottobackwards:
I ran the following test:
Modified the default snort parser configuration such that it was :

```json
{
  "parserClassName":"org.apache.metron.parsers.snort.BasicSnortParser",
  "sensorTopic":"snort",
  "parserConfig": {},
  "fieldTransformations" : [
    {
      "output" : ["msg" ],
      "transformation" : "SELECT"
    }
  ]
}

```

And the default snort enrichment configuration such that it was :

```json

{
  "enrichment" : {
  },
  "threatIntel" : {
    }
  }
}

```

I got the following:

```
2.168.138.158,49189,62.75.195.236,80,00:00:00:00:00:00,00:00:00:00:00:00,0x3C,***A****,0x9DFB1927,0xF1BD72CC,,0xFAF0,128,0,2360,40,40960,,,,","enrichmentsplitterbolt.splitter.end.ts":"1512763453749","enrichmentsplitterbolt.splitter.begin.ts":"1512763453749","guid":"08a84757-bf05-431b-9d81-5fa95fb99938","timestamp":1512763452000}
	at org.apache.metron.enrichment.bolt.EnrichmentJoinBolt.getStreamIds(EnrichmentJoinBolt.java:53) ~[stormjar.jar:?]
	at org.apache.metron.enrichment.bolt.EnrichmentJoinBolt.getStreamIds(EnrichmentJoinBolt.java:33) ~[stormjar.jar:?]
	at org.apache.metron.enrichment.bolt.JoinBolt.execute(JoinBolt.java:138) [stormjar.jar:?]
	at org.apache.storm.daemon.executor$fn__6573$tuple_action_fn__6575.invoke(executor.clj:734) [storm-core-1.0.1.2.5.3.0-37.jar:1.0.1.2.5.3.0-37]
	at org.apache.storm.daemon.executor$mk_task_receiver$fn__6494.invoke(executor.clj:466) [storm-core-1.0.1.2.5.3.0-37.jar:1.0.1.2.5.3.0-37]
	at org.apache.storm.disruptor$clojure_handler$reify__6007.onEvent(disruptor.clj:40) [storm-core-1.0.1.2.5.3.0-37.jar:1.0.1.2.5.3.0-37]
	at org.apache.storm.utils.DisruptorQueue.consumeBatchToCursor(DisruptorQueue.java:451) [storm-core-1.0.1.2.5.3.0-37.jar:1.0.1.2.5.3.0-37]
	at org.apache.storm.utils.DisruptorQueue.consumeBatchWhenAvailable(DisruptorQueue.java:430) [storm-core-1.0.1.2.5.3.0-37.jar:1.0.1.2.5.3.0-37]
	at org.apache.storm.disruptor$consume_batch_when_available.invoke(disruptor.clj:73) [storm-core-1.0.1.2.5.3.0-37.jar:1.0.1.2.5.3.0-37]
	at org.apache.storm.daemon.executor$fn__6573$fn__6586$fn__6639.invoke(executor.clj:853) [storm-core-1.0.1.2.5.3.0-37.jar:1.0.1.2.5.3.0-37]
	at org.apache.storm.util$async_loop$fn__554.invoke(util.clj:484) [storm-core-1.0.1.2.5.3.0-37.jar:1.0.1.2.5.3.0-37]
	at clojure.lang.AFn.run(AFn.java:22) [clojure-1.7.0.jar:?]
	at java.lang.Thread.run(Thread.java:745) [?:1.8.0_77]
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
2017-12-08 20:04:17.171 o.a.m.e.b.EnrichmentSplitterBolt [ERROR] Trying to retrieve a field map with sensor type of null
```

So it looks like there are more fields to protect.

simonellistonball:
Good catch, my test included the source.type in the select list, and generalised badly in the instructions. We should include the source.type in the protected system fields, let me update the test and implementation.

ottobackwards:
OK,
Re-ran my scenario after the latest changes, and verified in kibana that only the msg field was indexed ( except for metron system or added fields ).

+1

cestella:
Yeah, this is good.  +1 by inspection.

ottobackwards:
Thank you for the contribution @simonellistonball, please be sure to assign and close METRON-1341 in jira.  Cheers! #ynwa

PR #863, opened 2017-12-11
METRON-1347: Indexing Topology should fail tuples without a source.type

cestella:
## Contributor Comments
If you are sending data into metron indexing without a source.type, which would only happen if you're bypassing our previous topologies, we cannot configure how we write to the indices, so the message should be explicitly failed and reported.

To test this, you can send a message into the indexing topic via kafka console producer that looks like:
```
{
  "value" : "foo"
}
```
Ensure that the tuple is failed and an error message indicating that a `source.type` is required is written to the storm UI.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-indexing/README.md
+++ b/metron-platform/metron-indexing/README.md
@@ -17,7 +17,7 @@ These configs are variable by sensor type.
 
 ## Minimal Assumptions for Message Structure
 
-At minimum, a message should have a `sensor.type` field.
+At minimum, a message should have a `source.type` field.
 Without this field, the message tuple will be failed and not written
 with an appropriate error indicated in the Storm UI and logs.
 
--- a/metron-platform/metron-indexing/README.md
+++ b/metron-platform/metron-indexing/README.md
@@ -15,6 +15,12 @@ Indices are written in batch and the batch size and batch timeout are specified 
 [Sensor Indexing Configuration](#sensor-indexing-configuration) via the `batchSize` and `batchTimeout` parameters.
 These configs are variable by sensor type.
 
+## Minimal Assumptions for Message Structure
+
+At minimum, a message should have a `sensor.type` field.
+Without this field, the message tuple will be failed and not written
+with an appropriate error indicated in the Storm UI and logs.
+
 ## Indexing Architecture
 
 ![Architecture](indexing_arch.png)
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
@@ -17,6 +17,7 @@
  */
 package org.apache.metron.writer.bolt;
 
+import com.google.common.collect.ImmutableList;
 import org.apache.metron.common.Constants;
 import org.apache.metron.common.bolt.ConfiguredIndexingBolt;
 import org.apache.metron.common.configuration.writer.IndexingWriterConfiguration;
@@ -125,6 +126,13 @@ public class BulkMessageWriterBolt extends ConfiguredIndexingBolt {
     return defaultBatchTimeout;
   }
 
+  public BulkWriterComponent<JSONObject> getWriterComponent() {
+    return writerComponent;
+  }
+
+  public void setWriterComponent(BulkWriterComponent<JSONObject> component) {
+    writerComponent = component;
+  }
   /**
    * This method is called by TopologyBuilder.createTopology() to obtain topology and
    * bolt specific configuration parameters.  We use it primarily to configure how often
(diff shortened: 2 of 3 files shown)


metron-platform/metron-indexing/README.md
```diff
@@ -15,6 +15,12 @@ Indices are written in batch and the batch size and batch timeout are specified
 [Sensor Indexing Configuration](#sensor-indexing-configuration) via the `batchSize` and `batchTimeout` parameters.
 These configs are variable by sensor type.
```
ottobackwards: Do we care, here in particular, that we have different required fields listed?  Should this not be cumulative and include the required fields coming out of the parsers?

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
```diff
@@ -229,17 +239,30 @@ public void execute(Tuple tuple) {
       LOG.trace("Writing enrichment message: {}", message);
       WriterConfiguration writerConfiguration = configurationTransformation.apply(
               new IndexingWriterConfiguration(bulkMessageWriter.getName(), getConfigurations()));
-      if(writerConfiguration.isDefault(sensorType)) {
-        //want to warn, but not fail the tuple
-        collector.reportError(new Exception("WARNING: Default and (likely) unoptimized writer config used for " + bulkMessageWriter.getName() + " writer and sensor " + sensorType));
+      if(sensorType == null) {
```
ottobackwards: Maybe we should validate all the required fields?

merrimanr: @ottobackwards which fields should we validate here?

merrimanr:
I would like to hear feedback from @ottobackwards on other required fields but this looks good to me otherwise.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
cestella: Sure thing.  Really the only two required are `timestamp` and `source.type`.  Did I miss any?

metron-platform/metron-indexing/README.md
cestella: So, strictly speaking messages really only require `source.type` (which I typo'd) and `timestamp` (which I should add).  I'll fix that, but did I miss anything?

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
simonellistonball: Strictly speaking that's true, but by convention original_string should be required. There is a broader topic about what should be required, but that certainly doesn't belong in a comment on a PR.

ottobackwards:
The minimum required fields, as far as I can see right now are source.type, original_string and timestamp.  Given the use case for this is something that has skipped the parser topology, we should validate those.

If we think the same can be done for indexing, then we should use the same classes/technique there.

Again, this is based on the presented use case

cestella:
Actually, I don't think `original_string` is required past the parser topology.  For instance, profiler messages into enrichment do not have `original_string`.

metron-platform/metron-indexing/README.md
```diff
@@ -15,6 +15,12 @@ Indices are written in batch and the batch size and batch timeout are specified
...
 
+## Minimal Assumptions for Message Structure
+
+At minimum, a message should have a `sensor.type` field.
```
mmiklavc: sensor.type - did you mean source.type?

cestella: whoops, yes, typo.  corrected.

mmiklavc:
Are all questions on required fields resolved per @cestella's comments/clarification?

mmiklavc:
My +1 still stands on this and this PR has been up for quite some time. I will merge by EOD unless there are any additional concerns.

justinleet:
@mmiklavc @cestella @ottobackwards  Is this good to be merged in?

ottobackwards:
The parser topology PRODUCES the original_string.  By that very action, and it's requirement, it can only be used and required by downstream usages either from rest ( post indexing ) or during enrichment.  It seems inconsistent to validate without including that, if what you want to do is validate the entire set of required fields.  If you only want to handle what is required at a given step, but not what is required 'overall' then this is OK, but something doesn't seem right here conceptually.

ottobackwards:
What I think would be better would be to have a readme about the required fields and refer to that document from these other documents

mmiklavc:
It seems this PR is meant to address the global case, since the writer is general purpose. Per @cestella's comments, I don't know that we can include original_string. I do think there's value in minimally ensuring the timestamp and source.type fields are validated in the near term. It seems that addressing a broader validation strategy might be in order, but maybe we could handle that separately as a compromise? Thoughts @ottobackwards?

ottobackwards:
I think this is a mess that we will need to sort out, but I won't hold this up for that.
I'm +0 I guess.

mmiklavc:
@ottobackwards created this to track it - https://issues.apache.org/jira/browse/METRON-1524

PR #867, opened 2017-12-13
METRON-1350: Add reservoir sampling functions to Stellar

cestella:
## Contributor Comments
Sampling capabilities would fit very well with the profiler and enable algorithms that do not necessarily support our existing probabilistic sketches. We should add a reservoir sampler and utilities to merge and resample.

You can play with `SAMPLE_INIT`, `SAMPLE_ADD`, `SAMPLE_MERGE` and `SAMPLE_GET` via the REPL (via `mvn exec:java -Dexec.mainClass="org.apache.metron.stellar.common.shell.StellarShell" -pl metron-analytics/metron-statistics`):
```
[Stellar]>>> ?SAMPLE_ADD
SAMPLE_ADD
Description: Add to a sample

Arguments:
	sampler - Sampler to use.  If null, then a default Uniform sampler is created
	o - The value to add.  If o is an Iterable, then each item is added.

Returns:
[Stellar]>>> s_10 := SAMPLE_INIT(10)
[Stellar]>>> sample := REDUCE( [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 ], (s, x) -> SAMPLE_ADD(s, x), SAMPLE_INIT(5))
[Stellar]>>> SAMPLE_GET(sample)
[6, 8, 11, 4, 5]
[Stellar]>>> SAMPLE_ADD(s_10, [5, 2, 5, 7, 10 ])
org.apache.metron.statistics.sampling.UniformSampler@3d8d06c0
[Stellar]>>> SAMPLE_GET(SAMPLE_ADD(s_10, [5, 2, 5, 7, 10 ]))
[5, 2, 5, 7, 10, 5, 2, 5, 7, 10]
```
## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-analytics/metron-statistics/README.md
+++ b/metron-analytics/metron-statistics/README.md
@@ -69,7 +69,7 @@ functions can be used from everywhere where Stellar is used.
 * Returns: The resulting sample.
 
 #### `SAMPLE_INIT`
-* Description: Create a uniform reservoir sampler of a specific size or, if unspecified, size 1024.
+* Description: Create a [reservoir sampler](https://en.wikipedia.org/wiki/Reservoir_sampling) of a specific size or, if unspecified, size 1024.  Elements sampled by the reservoir sampler will be included in the final sample with equal probability.
 * Input:
   * size? - The size of the reservoir sampler.  If unspecified, the size is 1024.
 * Returns: The sampler object.
--- a/metron-analytics/metron-statistics/README.md
+++ b/metron-analytics/metron-statistics/README.md
@@ -53,6 +53,32 @@ functions can be used from everywhere where Stellar is used.
   * bounds - A list of value bounds (excluding min and max) in sorted order.
 * Returns: Which bin N the value falls in such that bound(N-1) < value <= bound(N).  No min and max bounds are provided, so values smaller than the 0'th bound go in the 0'th bin, and values greater than the last bound go in the M'th bin.
 
+### Sampling Functions
+
+#### `SAMPLE_ADD`
+* Description: Add a value or collection of values to a sampler.
+* Input:
+  * sampler - Sampler to use.  If null, then a default Uniform sampler is created.
+  * o - The value to add.  If o is an Iterable, then each item is added.
+* Returns: The sampler.
+
+#### `SAMPLE_GET`
+* Description: Return the sample.
+* Input:
+  * sampler - Sampler to use.
+* Returns: The resulting sample.
+
+#### `SAMPLE_INIT`
+* Description: Create a uniform reservoir sampler of a specific size or, if unspecified, size 1024.
+* Input:
+  * size? - The size of the reservoir sampler.  If unspecified, the size is 1024.
+* Returns: The sampler object.
+
+#### `SAMPLE_MERGE`
+* Description: Merge and resample a collection of samples.
+* Input:
+  * samplers - A list of samplers to merge.
+* Returns: A sampler which represents the resampled merger of the samplers.
 
 ### Distributional Statistics
 
--- a/metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/sampling/SamplingInitFunctions.java
+++ b/metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/sampling/SamplingInitFunctions.java
@@ -33,7 +33,7 @@ public class SamplingInitFunctions {
 
   @Stellar(namespace="SAMPLE"
           ,name="INIT"
-          ,description="Create a uniform reservoir sampler of a specific size or, if unspecified, size " + Sampler.DEFAULT_SIZE
+          ,description="Create a [reservoir sampler](https://en.wikipedia.org/wiki/Reservoir_sampling) of a specific size or, if unspecified, size " + Sampler.DEFAULT_SIZE + ".  Elements sampled by the reservoir sampler will be included in the final sample with equal probability."
           ,params = {
             "size? - The size of the reservoir sampler.  If unspecified, the size is " + Sampler.DEFAULT_SIZE
           }
@@ -78,12 +78,7 @@ public class SamplingInitFunctions {
                                        + ", but was " + obj
                                        );
       }
-      else {
-        return Optional.empty();
-      }
     }
-    else {
-      return Optional.ofNullable(ret);
-    }
+    return Optional.ofNullable(ret);
   }
 }
--- a/metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/sampling/SamplingInitFunctions.java
+++ b/metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/sampling/SamplingInitFunctions.java
@@ -1,0 +1,89 @@
+/*
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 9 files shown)


simonellistonball:
Should the size limit on the sample really be a cut off? In a likely usage scenario a users would sample over a window in a profile. Limiting the size is likely to skew to time at the beginning of the window rather than being genuinely uniform. Would a random replacement strategy make more sense when over the limit? This could be a lot heavier in terms of performance, but may be more mathematically sound.

metron-analytics/metron-statistics/README.md
```diff
@@ -53,6 +53,32 @@ functions can be used from everywhere where Stellar is used.
   * bounds - A list of value bounds (excluding min and max) in sorted order.
 * Returns: Which bin N the value falls in such that bound(N-1) < value <= bound(N).  No min and max bounds are provided, so values smaller than the 0'th bound go in the 0'th bin, and values greater than the last bound go in the M'th bin.
 
+### Sampling Functions
+
+#### `SAMPLE_ADD`
+* Description: Add a value or collection of values to a sampler.
+* Input:
```
ottobackwards: This makes it seem like Uniform sampler is a 'known' thing.  But it is not, either by explanation or reference to where it is explained ( as we have done referring to algorithms before ).
Is there another type of sampler?

Somewhere ( I'm not sure where ) we should say:
"A sampler is a xxxxx that is | does | acts as  xxxxx for the sample functions. The default has these properties, but you can override that in init"

Why even mention the Universal?

cestella:
Sorry, I am not sure I understand, this is random replacement when after the size limit.  Am I mistaking your question?

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/sampling/SamplingInitFunctions.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public boolean isInitialized() {
+      return true;
+    }
+  }
+
+
+  public static <T> Optional<T> get(List<Object> args, int offset, String argName, Class<T> expectedClazz) {
+    Object obj = args.get(offset);
+    T ret = ConversionUtils.convert(obj, expectedClazz);
```
ottobackwards: Couldn't this be simplified to :

```java 
if(ret == null ) {
      if(obj != null) {
         throw new IllegalStateException(argName + "argument(" + obj
                                        + " is expected to be an " + expectedClazz.getName()
                                        + ", but was " + obj
                                        );
       }
    }
    return Optional.ofNullable(ret);
 }
```

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/sampling/UniformSampler.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param o
+   */
+  public void add(Object o) {
+    if(o == null) {
+      return;
+    }
+    if (reservoir.size() < size) {
+      reservoir.add(o);
+    } else {
+      int rIndex = rng.nextInt(seen + 1);
```
cestella: Just so I'm clear, up to the reservoir size, we add to the reservoir.  When we're past the reservoir, we do a random replacement as per https://en.wikipedia.org/wiki/Reservoir_sampling

simonellistonball: you are 100% right, that's what I get for skim reading.

metron-analytics/metron-statistics/README.md
cestella: Sorry, `uniform` here is intended to mean that there's each element has equal probability of being in the sample (e.g. the probability is pulled from a [uniform probability distribution](https://en.wikipedia.org/wiki/Uniform_distribution_(continuous))).  I can probably do a better job documenting.

ottobackwards: Couldn't this be simplified to

```java
 if(ret == null ) {
      if(obj != null) {
         throw new IllegalStateException(argName + "argument(" + obj
                                        + " is expected to be an " + expectedClazz.getName()
                                        + ", but was " + obj
                                        );
       }
     }
return Optional.ofNullable(ret);


```

ottobackwards: Ok, It seemed like the Uniform implementation was leaking

cestella: There are definitely other types of reservoir samplers which we will probably want.  Most specifically a sampler that is biased toward recency (so non-uniform in that case).

simonellistonball: Recency would surely be more relevant for merged resampling in a profile context?

cestella: They're both needed.  Some use-cases would be fine without bias and some would be better with bias.  As a follow-on, I was planning on adding a biased sampler, but this is a big enough PR without it.  It'd look something like:
```
samples := SAMPLE_MERGE(PROFILE_GET('samples', ...))
biased_sample := SAMPLE_GET_BIASED(samples, 0.015)
```

ottobackwards: Then we'll have a get sample types method, like we do with other things like this right?

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/sampling/UniformSampler.java
ottobackwards: Shouldn't we reference Reservoir Sampling in the documentation?  Then the use of Universal and other terms would be more in context.

ottobackwards: This makes me think that we need "namespace" scoped documentation

metron-analytics/metron-statistics/README.md
cestella: Actually, more than likely it'd be a separate init since each type are going to have different types of parameters depending on the algorithm.  So a biased sampler would be `SAMPLE_INIT_BIASED(size, ...)`

metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/sampling/UniformSampler.java
cestella: I modified the docs to have a link to the Reservoir sampling wikipedia article and made things a bit more clear regarding Uniform.

ottobackwards:
I'm +1 by inspection, this looks great!

cestella:
I'll keep that +1 in my pocket...I want to spin it up in full-dev and try it in conjunction with a PR that I'm not quite done with yet.  Let's hold off merging until I can validate that I haven't botched the API in some horrible way for sampling ;)

ottobackwards:
@cestella  comments?
https://issues.apache.org/jira/browse/METRON-1361

cestella:
I commented on the ticket, but will repeat a portion of it here at the risk of this being quite off-topic.  @justinleet has some great ideas/prototypes about automatic documentation generation from stellar.  I'd support a new annotation for documenting namespaces.

ottobackwards:
any chance I can see @justinleet 's ideas/prototypes?

justinleet:
@ottobackwards Left a comment on the ticket you made: https://issues.apache.org/jira/browse/METRON-1361?focusedCommentId=16291159&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16291159

PR #879, opened 2017-12-21
METRON-1378: Create a summarizer

cestella:
## Contributor Comments
We have a nice and generalized infrastructure for loading data into HBase and interacting with it via `flatfile_loader.sh` and `ENRICHMENT_GET()`.  It is also useful to summarize a set of data into a static data structure, store it on HDFS and interact with it via stellar.  To this end, to complement the `flatfile_loader.sh`, we should have a `flatfile_summarizer.sh` that, using the same extractor config, will process a flat file and output a serialized object.

The usecase for this is as follows:
Let's say that I have a static list of domains in the second column of a CSV, domains.csv, and I want to generate a bloom filter with those domains in them sans TLD.

I should be able to create a file called `bloom.ser` with the serialized bloom filter given the extractor config:
```
{
  "config" : {
    "columns" : {
       "rank" : 0,
       "domain" : 1
    },
    "value_transform" : {
       "domain" : "DOMAIN_REMOVE_TLD(domain)"
    },
    "value_filter" : "LENGTH(domain) > 0",
    "state_init" : "BLOOM_INIT()",
    "state_update" : {
       "state" : "BLOOM_ADD(state, domain)"
                     },
    "state_merge" : "BLOOM_MERGE(states)",
    "separator" : ","
  },
  "extractor" : "CSV"
}
```

Note, the associated stellar function `OBJECT_GET` is available in #880.

# Testing Plan

We should run the test plan for #445 to ensure no regressions since 80% of this PR is just refactoring existing abstractions to reuse.

## Write out a String Locally
We are going to take the top 10k alexa domains (saved as part of #445 's test plan to `~/top-10k.csv`)
* Keep a running sample of 20 samples per thread
* At the end, merge the samples and get a random domain from the merged samples
* Write out the sample

### Test
* Create a file `~/extractor_sample.json` with the following contents:
```
{
  "config" : {
    "columns" : {
       "rank" : 0,
       "domain" : 1
    },
    "value_transform" : {
       "domain" : "DOMAIN_REMOVE_TLD(domain)"
    },
    "value_filter" : "LENGTH(domain) > 0",
    "state_init" : "SAMPLE_INIT(20)",
    "state_update" : {
       "state" : "SAMPLE_ADD(state, domain)"
                     },
    "state_merge" : "GET_FIRST(SAMPLE_GET(SAMPLE_MERGE(states, SAMPLE_INIT(1))))",
    "separator" : ","
  },
  "extractor" : "CSV"
}
```
* Summarize via `$METRON_HOME//bin/flatfile_summarizer.sh -i ~/top-10k.csv -o ~/sample.ser -e ./extractor_sample.json -p 5 -b 128`
* Execute `hexdump -C ./sample.ser` and ensure that there is a string in there.  It may end or start with some non-ascii bytes at the beginning and end.
e.g.
```
[root@node1 ~]# hexdump -C ./sample.ser
00000000  03 01 37 63 66 6d 6e e6                           |..7cfmn.|
00000008
[root@node1 ~]# cat top-10k.csv | grep 7cfmn
4696,7cfmnf.top
```
### Typosquatting Use-case Testing

You can also follow the testing plan for #882 as this code is merged into that PR and it shows how this feature can be used in a real use-case.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


nickwallen:
IMHO, this seems really complex.  That JSON configuration looks really scary to me and sadly I don't think I could bring myself to use this.

Fundamentally the need you are addressing is that you want to manipulate a text file and leverage Stellar as part of that pipeline, right?

In a perfect world, I'd be able to use Unix-y tools like Grep, Sed, Awk, etc and introduce Stellar expressions as part of the pipeline.  I know it is already possible to some degree with the REPL.  

I am wondering if with a little brainstorming we might be able to arrive at a simpler solution that allows us to leverage tried-and-true UNIX pipelines so that people can use the Unix-y tools they are familiar with.  I feel we'll get better user traction with a simpler, more familiar approach.

ottobackwards:
I think the need is to 'pre-create' complex objects, and re-use them across multiple stellar rule executions, with the bloom-filter being the example.

Is that close?

cestella:
@nickwallen  I definitely hear you, the JSON configs are more complex than I'd like.  I'd like another more composable solution available using lambda functions available in the REPL.  

The intent of this PR, however, was a minimal extension of our existing extractor config to enable this use-case.  This PR isn't making a complex bit of JSON per se, but rather it's reusing our existing complex bit of JSON to enable the use-case.  I would love to create a discuss thread to see about a more REPL-friendly approach to this, but I think that belongs as a follow-on.  Thoughts?

cestella:
@ottobackwards Yes, that's spot on.  It's to enable creation of summarization objects in a method similar (and reusing the configs and infrastructure of) the flat file loader.  The idea is that this is just one way to create these objects, we might (likely will) have more ways to create them in the future.

cestella:
Also, a wizard-like UI could simplify this dramatically.  That was one of the thoughts around extending and reusing the existing infrastructure in the first pass of this rather than creating a new way of iterating over flat files.

cestella:
After more consideration and more egg nog, I decided that I'd create a DISCUSS thread about this entire use-case.  We can move the discussion there.

cestella:
So, the discuss thread has been going for some time now and the discussion is mostly around forward-thinking extensions to this.  Are we at the point to agree that this is a viable first step and this can be reviewed?

metron-platform/metron-data-management/src/main/java/org/apache/metron/dataloads/nonbulk/flatfile/writer/Writer.java
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.metron.dataloads.nonbulk.flatfile.writer;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.metron.common.utils.SerDeUtils;
+
+import java.io.IOException;
+import java.util.Optional;
+
+public interface Writer {
+  void validate(Optional<String> output, Configuration hadoopConfig);
```
justinleet: Why doesn't this either return something or throw a checked exception? Right now the validate method is just something that you call and it doesn't do anything except throw runtime exception in the impls, rather than giving the caller an explicit ability to do anything about the validation failure.

cestella: good catch.

metron-platform/metron-data-management/README.md
```diff
@@ -354,3 +357,91 @@ The parameters for the utility are as follows:
...
+| -e         | --extractor_config  | Yes          | JSON Document describing the extractor for this input data source                                                                                                                   |
+| -m         | --import_mode       | No           | The Import mode to use: LOCAL, MR.  Default: LOCAL                                                                                                                                  |
+| -om        | --output_mode       | No           | The Output mode to use: LOCAL, HDFS.  Default: LOCAL                                                                                                                                  |
+| -i         | --input             | Yes          | The input data location on local disk.  If this is a file, then that file will be loaded.  If this is a directory, then the files will be loaded recursively under that directory.  |
+| -i         | --output            | Yes          | The output data location.    |
```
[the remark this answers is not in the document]
cestella: good catch

justinleet:
I spun this up in the context of the combined PR, and everything worked as advertised, barring the UI because of ES5 issues.  I was able to validate that data flowed through as expected by querying against the indices directly.

I'm good with this being a viable first step, and am +1.

mmiklavc:
@justinleet 

> barring the UI because of ES5 issues

What sort of issues?

justinleet:
@mmiklavc Check out https://github.com/apache/metron/pull/882#issuecomment-356109443.  Looks like the squid mapping @cestella uses doesn't line up (which isn't terribly surprising because it was ES2 until a couple hours ago).

cestella:
Just following up, I have migrated the mapping of existing data to a template in the instructions and the type mismatch for `ip_dst_addr` is no longer an issue.  The PR as it currently stands has the correct instructions and is ready for review to continue.

justinleet:
I'm still +1 on this, thanks again.

cestella:
Ok, @justinleet has given a +1, do we have any existing reservations after the discussion thread and the review here on this work?  If not, then I'm going to commit on Monday.

ottobackwards:
+0

mraliagha:
@cestella Is there any document or description regarding this feature? How would the performance be comparable with normal HBase enrichment?

nickwallen:
+0  I'm sure what's here is solid, but I have not reviewed it myself.  I just want to clear the way for this to get merged.

I don't necessarily like the usability of this approach, but I think it is a good first step.  And merging this doesn't preclude providing alternative approaches or enhancing this approach for better usability later.  

I do love the use case that was added that drove the need for this functionality.  So good stuff @cestella!

PR #899, opened 2018-01-13
METRON-1405: Add Boyer-Moore majority vote algorithm to Stellar

mmiklavc:
## Contributor Comments

Addresses https://issues.apache.org/jira/browse/METRON-1405

A proper majority requires 2 passes, however you can get a sense of plurality by leveraging a single pass model.

Test plan soon to follow.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
@@ -19,16 +19,15 @@
  */
 package org.apache.metron.stellar.dsl.functions;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.apache.metron.stellar.common.utils.ConversionUtils;
 import org.apache.metron.stellar.common.utils.math.MathOperations;
 import org.apache.metron.stellar.common.utils.math.StellarMathFunction;
 import org.apache.metron.stellar.dsl.BaseStellarFunction;
-import org.apache.metron.stellar.dsl.Context;
 import org.apache.metron.stellar.dsl.ParseException;
 import org.apache.metron.stellar.dsl.Stellar;
-import org.apache.metron.stellar.dsl.StellarFunction;
-
-import java.util.List;
-import java.util.function.Function;
 
 public class MathFunctions {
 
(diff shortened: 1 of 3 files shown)


metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
```diff
@@ -220,4 +219,156 @@ public Object apply(List<Object> args) {
...
+      return counter;
+    }
+  }
+
+  @Stellar(namespace = "BOYERMOORE"
+      , name = "PLURALITY"
+      , description = "Calculates the item with current plurality in a Boyer-Moore list. [Boyer-Moore](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_majority_vote_algorithm)"
+      , params = {
+      "state - state holder for list of values. If null, add will initialize a new state value."
+  }
```
ottobackwards: This is confusing.  It is the same description as ADD() but I think that ADD is correct and this should be "Returns the object that has plurality" ?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
```diff
@@ -220,4 +219,156 @@ public Object apply(List<Object> args) {
...
+        }
+      }
+      return m;
+    }
+
+    public Object add(Object item, Long counter) {
+      if (item != null) {
+        if (this.counter == 0) {
+          m = item;
+          this.counter = counter;
```
ottobackwards: So, Stellar will parse and create Numbers, I'm wondering if there is a possibility of strange errors or rounding here.  Like there are two not equal numbers but they end up equal through the conversion process.  Does that make sense?

ottobackwards:
This tests fails, I think something is wrong.

```java
 @Test
  public void boyerMoore_calculates_plurality_from_list_of_mixed_objects_o() throws Exception {
    List<Object> items = Arrays.asList(1,1,1,1, "orange", "orange", "jello", "jello", "jello");
    BoyerMooreState state = (BoyerMooreState) new BoyerMooreAdd().apply(Arrays.asList(null, items));
    Assert.assertThat(new BoyerMoorePlurality().apply(ImmutableList.of(state)), CoreMatchers.equalTo(1));
  }
```

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
mmiklavc: Good catch

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
mmiklavc: It's true, I make no effort to massage data types here - if you want to interpret a 1 as a string "1", then they will be considered distinct. I didn't have a suitable mechanism for determining types. I thought it was reasonable for a user to be able to make that type cast as part of their Stellar expression building, but I'm certainly open to suggestions.

mmiklavc: AFA rounding errors - the counter you're seeing there are internal state. In the case of merging multiple states together, it will effectively merge the counts to nominate a new plurality leader. If we could do this in 2 passes, you'd have a guaranteed majority, but I thought it was still useful to see what item has the most without necessarily being a majority. I'll add some use case documentation this week as I think this might be useful in the profiler.

mmiklavc:
I'll look into the test as it's passing locally for me in my IDE.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
ottobackwards: no, i'm talking about  object.equals(object2)  when the type of the objects is determined by antlr etc

ottobackwards:
Just in case:  Note that that test is modified from your test ( _o )

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
mmiklavc: I know we do this in a number of places where the Objects provided via the inbound List are handed off directly to a java.util.List without any casting or special parsing, e.g. 
* BloomFilter -> https://github.com/apache/metron/blob/master/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/DataStructureFunctions.java#L48
* Data structure functions such as List -> https://github.com/apache/metron/blob/master/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/DataStructureFunctions.java#L189
* HLLP -> https://github.com/apache/metron/blob/master/metron-analytics/metron-statistics/src/main/java/org/apache/metron/statistics/approximation/HyperLogLogPlusFunctions.java#L49

Is that what you're asking?

mmiklavc:
Yes - good catch - this not actually going to work without some modification. The count zeroes out when a majority is held by non-m values and gets reset to a new non-plurality, non-majority object.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MathFunctions.java
ottobackwards: No, more of something that is prob. an edge case, such as fields from input that are not normalized, so they come in as 1 or 1.0 or something and will be unequal.

ottobackwards:
@Test
  public void boyerMoore_calculates_plurality_from_list_of_values_o() throws Exception {
    List<Integer> items = Arrays.asList(1, 1, 1,1, 2,2,3,3,4,4,4);
    BoyerMooreState state = null;
    for (Integer item : items) {
      state = (BoyerMooreState) new BoyerMooreAdd().apply(Arrays.asList(state, item));
    }
    Assert.assertThat(new BoyerMoorePlurality().apply(ImmutableList.of(state)), CoreMatchers.equalTo(1));
  }

as well

mmiklavc:
Pulling this PR until I can resolve and underlying issue with the counting semantics, however I believe this may by theoretically impossible without a 2-pass model.

PR #901, opened 2018-01-19
METRON-1410 [MPACK] Check for existing HBASE tables before creating

ottobackwards:
After upgrade, indexing will not restart because it fails to create HBase tables.  This is because they already exist from the previous version.

This PR changes the base command used to check for the existence of the table, and only create it if it does not already exist.

## Testing
- Full Dev should deploy as normal
- [tbd] some upgrade procedure

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/enrichment_commands.py
```diff
@@ -190,7 +190,7 @@ def create_hbase_tables(self):
                   self.__params.hbase_principal_name,
                   execute_user=self.__params.hbase_user)
 
-        cmd = "echo \"create '{0}','{1}'\" | hbase shell -n"
+        cmd = "if [[ $(echo \"exists '{0}'\" | hbase shell | grep 'not exist') ]]; then echo \"create '{0}','{1}'\" | hbase shell -n; fi"
```
nickwallen: Would it be possible to just add a `create_hbase_table` function in `metron_service` that is used by Enrichment, Indexing, and the Profiler?  Then we don't have to duplicate this in three places?

ottobackwards: I thought of that, the issue would be that that function would have the same parameters as this Execute command.  It wouldn't really make it better.

We could create a shared get_create_hbase_table_fmt() function that returns the command.

ottobackwards: That statement is based on us not always possibly wanting to use the same users etc.

cestella: Still seems like not having to cut and paste that bit of shell script inside a python variable is worth generalizing to a `create_hbase_table` function.

ottobackwards: ok, i'll do that

ottobackwards: done

nickwallen:
+1 Looks good @ottobackwards.  Thanks for fixing this!

I have not tested this myself, but it looks solid.  Let me know if you'd prefer me to spin this up to get a second test run in.  Otherwise, I assume that what you've done is sufficient.

ottobackwards:
I'm going to wait for @lvets a chance to try his scenario

mmiklavc:
Thanks for fixing this @ottobackwards. +1 via inspection from me as well.

PR #902, opened 2018-01-19
METRON-1413 Add Metron Commit Tool

nickwallen:
This PR contributes the `prepare-merge` tool that many (some?) contributors use.  Up until now, it has been managed in a separate repo.

I didn't have a logical place to put this tool, so I had to reorganize a bit.  Since our tooling has been growing, a reorganization shouldn't be too unexpected.

* Creates top level directory called `dev-utilities`.
* Moves existing `build_utils` to `dev-utilities/build-utils`
* Moves existing `build_utils/release-utils` to `dev-utilities/release-utils`
* Creates `dev-utilities/committer-utils`
* Adds the `prepare-commit` script to `committer-utils`.

It is a bit easier to see by just looking at it.
```
dev-utilities/
├── build-utils
│   ├── README.md
│   ├── create_bundled_licenses.sh
│   ├── generate_license.py
│   ├── list_dependencies.sh
│   ├── verify_license.py
│   └── verify_licenses.sh
├── committer-utils
│   ├── README.md
│   └── prepare-commit
└── release-utils
    ├── metron-rc-check
    └── validate-jira-for-release

3 directories, 10 files
```

## Pull Request Checklist


metron-dev-support/committer-utils/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+This project contains tools to assist Apache Metron project committers.
+
+## Prepare Commit
+
+This script automates the process of merging a pull request into `apache/master`.  The script will prompt for the pull request number.  Most of the remaining information is automatically extracted from Github or the Apache JIRA.
+
+When prompted the `[value in brackets]` is used by default.  To accept the default, simply press `enter`.  If you would like to change the default, type it in and hit `enter` when done.
+
+In the following example, I enter the pull request number (`80`) when prompted.   Using the pull request number, the script can extract most of the remaining required information.
```
JonZeolla: `s/80/897/`

mattf-horton:
@nickwallen , excellent to get this into the main repo, and totally agree with the reorganization.

My only suggestion is that the main directory, currently named "metron-dev-support", shouldn't have the word "metron" in it, because it isn't a package that gets built into the Metron product.  "dev-support" or "dev-utilities" would be better.

+1 whether you accept the suggestion or not.

nickwallen:
Thanks @mattf-horton.  Either of those suggestions works for me; 'dev-support' or 'dev-utilities'.  I'll let others chime in with their preference and then update it accordingly.

mattf-horton:
I have a mild preference for dev-utilities, since "support" has many meanings.

ottobackwards:
I think checkout-pr is equally as valuable and should be included

nickwallen:
I don't see the value in `checkout-pr` any longer.  You can just use `prepare-commit` which gives you the exact view of what the code would look-like when its merged.

ottobackwards:
Maybe we can have reviewer tools and put checkout-pr in it?   For those without an apache id?

nickwallen:
The Apache ID isn't really necessary to use `prepare-commit`.  What if I just changed the docs to note that it is not required?  I could document how you could use the script to review a PR.

I would like to try and fill the need that you've identified @ottobackwards, I just really want to keep the amount of code that we have to maintain and support as minimal as possible.

nickwallen:
Travis had a hickup... Doing the "close/reopen" dance to trigger the CI build.

nickwallen:
Also, I've never liked the name `prepare-commit`.  If you guys can think of something better, please let me know.

nickwallen:
> @JonZeolla: ... the curl command HTML encodes the JIRA description but I don't know of a clean bash-native way to decode it.

Yes, I have noticed, especially with apostrophes.  When that happens, I usually just manually override it as a work around.  

My goal was just to get the tool into Apache as-is (warts and all).  We can try and fix annoyances like that on subsequent PRs (IMHO).  

I know you have some enhancements that you would like to make also.  Looking forward to those. :)

justinleet:
Re: the single quoting issue.

Avoiding the XML blob and using the main rest endpoint seems to kick back actual quotation marks. Now you have to read JSON instead of XML, but it might be more convenient to do that.
```
{8:57}~ ➭ curl https://issues.apache.org/jira/rest/api/latest/issue/METRON-1298\?fields\=summary
{"expand":"renderedFields,names,schema,operations,editmeta,changelog,versionedRepresentations","id":"13116358","self":"https://issues.apache.org/jira/rest/api/latest/issue/13116358","key":"METRON-1298","fields":{"summary":"TimeRange Picker doesn't work on Safari"}}%
```

I didn't dig into why things are different, I was mostly more trying to learn more about the API choice in there, and happened to notice it.

Sidenote, that XML blob can have the specific fields you want appended to avoid pulling back everything
```
https://issues.apache.org/jira/si/jira.issueviews:issue-xml/METRON-1298/METRON-1298.xml?field=summary
```

JonZeolla:
I'm game to merge the `prepare-commit` and `checkout-pr` logic into one script; I'm even fine with pushing that to be a follow-on.

Regarding the HTML-encoded JIRA description issue, I'm also fine with this being a follow-on.  I will be submitting a PR to support both of the Metron repos in this script after it's merged, and can take a stab at the JIRA description issue if it's still a problem.

ottobackwards:
I think of there being two users for these scripts.

1. Committers, that will benefit by having scripted commit actions
2. Reviewers who would benefit by having scripted PR downloads for build and review.  These may not be committers.

If we merge the scripts, we should maintain the reviewer only case ( no apache id or email etc )

justinleet:
I'm definitely fine with leaving HTML encoding to a follow-on.

nickwallen:
> @ottobackwards: I think of there being two users for these scripts...

That makes perfect sense to me.  Can we tackle that in a follow-on?

ottobackwards:
I can almost always be bought off with a jira number

ottobackwards:
I have another script I want to add for tracking master in feature branches after this as well

nickwallen:
I merged with master, so I expect Travis to be happy now.  Just need +1s and I'll get this in to allow for any follow-ons.

PR #907, opened 2018-01-24
METRON-1427: Add support for storm 1.1 and hdp 2.6

cestella:
## Contributor Comments
Right now our ambari mpack won't run cleanly on HDP 2.6 and Storm 1.1 because of some classpath issues and not supporting the stack.  We should migrate fulldev to run 2.6 (while still working with 2.5) and validate that we work with storm 1.1.   The only code changes in here is relocating and shading jackson (kerberized storm 1.1 requires a conflicting version of jackson from ES and us, so we shade and relocate it in parsers, but didn't in profiler and elasticsearch uber jars.  This PR corrects that).  

You can validate this by:
* Running up full-dev and ensuring data flows through
* Kerberizing full-dev and ensure things still work

I have done the above, but you can feel free to do the same.  
## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel). 
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  mvn -q clean integration-test install && build_utils/verify_licenses.sh 
  ```

- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [x] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

#### Note:
[...]

PR #910, opened 2018-01-25
METRON-1430: Isolate jackson from being used as arguments or returns from JSONUtils

cestella:
## Contributor Comments
Currently jackson is used as part of our internal API to JSONUtils.  The problem here is when we shade and relocate jackson.  Suddenly we can't use JSONUtils.  We should avoid operating on jackson primitives and rather provide a convenient wrapper around the places where we use jackson as part of the API.

This is required for #907 because this is where we are forced to relocate jackson and there is a competing version on the classpath.

Validate by running up vagrant and ensuring data flows through all topologies, data shows up in the alerts UI.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/IndexDao.java
+++ b/metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/IndexDao.java
@@ -17,7 +17,6 @@
  */
 package org.apache.metron.indexing.dao;
 
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import java.io.IOException;
 import java.util.List;
@@ -143,7 +142,7 @@ public interface IndexDao {
     JsonNode originalNode = JSONUtils.INSTANCE.convert(latest, JsonNode.class);
     JsonNode patched = JSONUtils.INSTANCE.applyPatch(request.getPatch(), originalNode);
     Map<String, Object> updated = JSONUtils.INSTANCE.getMapper()
-        .convertValue(patched, new TypeReference<Map<String, Object>>() {});
+        .convertValue(patched, JSONUtils.MAP_SUPPLIER);
     return new Document( updated
         , request.getGuid()
         , request.getSensorType()
--- a/metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/IndexDao.java
+++ b/metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/IndexDao.java
@@ -17,11 +17,16 @@
  */
 package org.apache.metron.indexing.dao;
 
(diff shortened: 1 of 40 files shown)


ottobackwards:
This looks awesome at first look.  I'm a big fan of doing this regardless of the shading issue.  The only thought that comes to mind is that there is a tipping point where a *Utils class isn't the right thing, that is it outgrown.  I feel like we might be there for JSONUtils.

We can do that in another pr however I guess.

mmiklavc:
@ottobackwards I was envisioning a metron-deps project of some sort. Jackson and Guava are perennial offenders in the Hadoop stack, and a facade of some sort would be useful to insulate us from the ground shifting beneath us.

ottobackwards:
@mmiklavc I think that is a great idea for the project.  And the appropriately factored facade classes as well.

cestella:
Yeah, this is the first step to moving it out, I'd say.  This was useful independent of #907 because it made some things a lot cleaner (namely fewer `new TypeReference<Map<String, Object>>() {}` and more `JSONUtils.MAP_SUPPLIER`).

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/IndexDao.java
```diff
@@ -140,9 +147,9 @@ default Document getPatchedDocument(PatchRequest request
         throw new OriginalNotFoundException("Unable to patch an document that doesn't exist and isn't specified.");
       }
     }
-    JsonNode originalNode = JSONUtils.INSTANCE.convert(latest, JsonNode.class);
-    JsonNode patched = JSONUtils.INSTANCE.applyPatch(request.getPatch(), originalNode);
-    Map<String, Object> updated = JSONUtils.INSTANCE.getMapper()
+    JsonNode originalNode = _mapper.get().convertValue(latest, JsonNode.class);
+    JsonNode patched = JsonPatch.apply(request.getPatch(), originalNode);
+    Map<String, Object> updated = _mapper.get()
         .convertValue(patched, new TypeReference<Map<String, Object>>() {});
```
mmiklavc: JSONUtils.MAP_SUPPLIER here?

cestella: no, I don't think I want to pass in a JsonNode to `JSONUtils.INSTANCE.convert()`.  I'm not sure WHAT happens there, honestly in a situation where the JsonNode we pass in isn't the jackson version that JSONUtils is using in metron-common.  Also, I think I'd just like to keep JSONUtils out of this class since it's using Jackson.

mmiklavc:
Ah, ok I didn't pick that up from the context, but that makes sense. +1 stands.

PR #914, opened 2018-01-29
METRON-1397  Support for JSON Path and complex documents in JSONMapParser

ottobackwards:
It would be useful for implementors to have the ability to ingest more complex documents without having to stand up a NiFi node/cluster.

This PR adds support for splitting multiple messages from a single message by way of using JSON Path statements to select the List element in the document.

Since the parse() interface already returns a List<JSONObject> the scope of this change is in the parser.

Example:

```json
    {
    "foo" :
    [
    { "name" : "foo1", "value" : "bar", "number" : 1.0 },
    { "name" : "foo2", "value" : "baz", "number" : 2.0 }
    ]
    }
 ```

with JSON Path statement "$.foo" will result in two messages emitted from the parser

```json
 { "name" : "foo1", "value" : "bar", "number" : 1.0 }
```

```json
{ "name" : "foo2", "value" : "baz", "number" : 2.0 }
```
Support in the configuration for a new configuration parameter: jsonpQuery has been added.
An example of the new configuration:

```json
{
  "sensorTopic":"jsonMapQuery",
  "parserConfig": {"jsonpQuery":"$.foo"}
}
```

In order to integration test both with and without the query present, a new jsonMapQuery parser configuration has been added, with test data.

##Testing
- tests should run
- deployment should work

> NOTE.  There are three commits here to start
1. The basic functionality
2. checkstyle fixes
3. small fix found later

### For all changes:

### For code changes:

### DIFF
--- a/dependencies_with_url.csv
+++ b/dependencies_with_url.csv
@@ -22,6 +22,9 @@ com.flipkart.zjsonpatch:zjsonpatch:jar:0.3.4:compile,Apache v2, https://github.c
 com.google.protobuf:protobuf-java:jar:2.5.0:compile,New BSD license,http://code.google.com/p/protobuf
 com.google.protobuf:protobuf-java:jar:2.6.1:compile,New BSD license,http://code.google.com/p/protobuf
 com.jcraft:jsch:jar:0.1.42:compile,BSD,http://www.jcraft.com/jsch/
+com.jayway.jsonpath:json-path:jar:2.3.0:compile,Apache v2,https://github.com/json-path/JsonPath
+net.minidev:accessors-smart:jar:1.2:compile,Apache v2,https://github.com/netplex/json-smart-v2
+net.minidev:json-smart:jar:2.3:compile,Apache v2,https://github.com/netplex/json-smart-v2
 com.maxmind.db:maxmind-db:jar:1.2.1:compile,CC-BY-SA 3.0,https://github.com/maxmind/MaxMind-DB
 com.maxmind.geoip2:geoip2:jar:2.8.0:compile,Apache v2,https://github.com/maxmind/GeoIP2-java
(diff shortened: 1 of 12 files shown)


mmiklavc:
This seems cool. Can you provide some guidance via some examples or just a link to what's possible syntax/query-wise with JSON path queries in the docs for this?

ottobackwards:
http://goessner.net/articles/JsonPath/
https://github.com/json-path/JsonPath
http://jsonpath.herokuapp.com

https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-testing.html

ottobackwards:
yeah, you said in the docs, will do

ottobackwards:
Just a note on parser docs, part of the mythical 777 is support for per parser readme, that would be nice here.

simonellistonball:
@ottobackwards Really want that myth to be a reality... as soon as we can get the config overwrite problem solved... but this really isn't the ticket to discuss that on :)

I'm kinda keen on this parser looking at things like CloudTrail logs and such like which appear with merged JSON right now. The other thing that I think this would link up to is the idea that's been floating around about chaining parsers. This parser could well be a split parser, which then routes to other parsers with different transformations for example depending on the message, but again, that's deep future and should probably be on a discuss thread about parser chaining.

simonellistonball:
And in more topical an relevant points... we're introducing a net new dependency here... Do we need to also update the NOTICES file (and the dependencies_with_url.csv list)?

ottobackwards:
Um, it is updated.

simonellistonball:
My bad: the dependencies file is updated, but not the NOTICES.

simonellistonball:
And in fairness to you, we don't do the NOTICES properly for anything else either I see. That's probably a separate scope.

cestella:
I love this, sorry I didn't review it sooner @ottobackwards +1 by inspection.  This is great.

PR #920, opened 2018-02-01
METRON-1438 Move SHELL functions from metron-management to stellar-common

ottobackwards:
Part of making stellar more modular and stand alone is consolidation of stellar functionality scattered through metron into stellar common.   That should be done if the functionality is generally useful and not metron specific. 

Where the functionality *is* metron specific, we should look to factor such functionality as to extract generally useful functionality.

In the case of the SHELL- namespace, it is clearly not metron specific.

This pr moves this functionality from the metron-management module to stellar common.

## Testing
- build and tests should run as normal
- shell functions such as those in the metron-management readme examples should run
- functions should work as before

### For all changes:

### For code changes:
  ```
  ```

- [na] Have you written or updated unit tests and or integration tests to verify your changes?
- [na] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/cli/PausableInput.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/cli/PausableInput.java
@@ -21,6 +21,7 @@ package org.apache.metron.stellar.common.shell.cli;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * An input stream which mirrors System.in, but allows you to 'pause' and 'unpause' it.
@@ -36,8 +37,8 @@ import java.io.InputStream;
  *
  */
 public class PausableInput extends InputStream {
-  InputStream in = System.in;
-  boolean paused = false;
+  private InputStream in = System.in;
+  private AtomicBoolean paused = new AtomicBoolean(false);
   private PausableInput() {
     super();
   }
@@ -46,7 +47,7 @@ public class PausableInput extends InputStream {
    * Stop mirroring stdin
    */
   public void pause() {
-    paused = true;
+    paused.set(true);
   }
 
   /**
@@ -54,8 +55,7 @@ public class PausableInput extends InputStream {
    * @throws IOException
    */
   public void unpause() throws IOException {
-    in.read(new byte[in.available()]);
-    paused = false;
+    paused.set(false);
   }
 
   public final static PausableInput INSTANCE = new PausableInput();
@@ -76,7 +76,14 @@ public class PausableInput extends InputStream {
    */
   @Override
   public int read() throws IOException {
-
+    if(paused.get()) {
+      try {
+        Thread.sleep(1000);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
+      }
+      return 0;
+    }
     return in.read();
   }
 
@@ -116,7 +123,7 @@ public class PausableInput extends InputStream {
   @Override
   public int read(byte[] b) throws IOException {
 
-    if(paused) {
+    if(paused.get()) {
       try {
         Thread.sleep(1000);
       } catch (InterruptedException e) {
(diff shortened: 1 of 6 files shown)


metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/ShellFunctionsTest.java
```diff
@@ -40,8 +45,8 @@
   );
 
   Context context = new Context.Builder()
-            .with(Context.Capabilities.SHELL_VARIABLES , () -> variables)
-            .build();
+            .with(Context.Capabilities.SHELL_VARIABLES , () -> variables).build();
```
nickwallen: I am actually interested in what direction as a project we should be taking with these types of fluent, chained statements.  I run across this all the time and I want to know the 'right' way that I should be doing it for the project.

IMHO, the way it was (separated by a line break) is more readable.  Meaning, a long set of chained statements should be separated by line breaks.  For example...
```
      result = new ProfileMeasurement()
              .withProfileName(profileName)
              .withEntity(entity)
              .withGroups(groups)
              .withPeriod(period)
              .withProfileValue(profileValue)
              .withTriageValues(triageValues)
              .withDefinition(definition);
```

But, of course, in terms of code style my opinion doesn't matter.  It is all about our style guidelines. 
 What does the Google code style guidelines say?   

Doesn't [this](https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break) support what I have said above about line breaks in this case?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/cli/PausableInput.java
```diff
@@ -36,8 +37,8 @@
  *
  */
 public class PausableInput extends InputStream {
-  InputStream in = System.in;
-  boolean paused = false;
+  private InputStream in = System.in;
+  private AtomicBoolean paused = new AtomicBoolean(false);
```
nickwallen: What problem were you solving here @ottobackwards ?  Is this bit access by multiple threads?

ottobackwards: I found in travis and locally, that the PausableInput was hanging, my builds where not completing.      PausableInput.INSTANCE.unpause(); was never returning.

I couldn't figure out why it was locked.  Reading the class documentation and the trying to understand the threading in the shell, I looked at the class and saw some inconsistencies with how and when we checked if we where paused in the different read() calls, and also that the flag field was not volatile or atomic.  I refactored this things to what I *think* would be the correct and consistent approaches and .... it resolved my issue.

nickwallen: Good find!  I'm sure that was frustrating to dig into.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/ShellFunctionsTest.java
ottobackwards: I didn't reformat that like that on purpose.  I had a period where I was setting the CONSOLE capability.  When I removed it, it just worked out like this.

Even if I select and format in intellij it doesn't change the .build() to the next line.

I think what you have above is fine.  I would like it to 'just' happen if I format code though, since it is easy for things to slip through.

It is tough right now, because so much of the codebase isn't formatted to check style, and I don't think we want every pr to include a lot of formatting changes.

ottobackwards: How do you have your formatting preferences set to get the above?

nickwallen: Right now, I never let my IDE reformat for me.  Like you said, if we get the code base matching check style and I can load that style into my IDE, then I'd gladly let it do most of the work for me.

Maybe I'll open a discuss thread.  I don't know how to handle this kind of thing and it happens all the time.

But for this specific scenario in your PR, it really doesn't matter either way.  I think you're good to go either way.

nickwallen:
+1 Thanks @ottobackwards

PR #924, opened 2018-02-02
METRON-1299 In MetronError tests, don't test for HostName if getHostName wouldn't work

ottobackwards:
MetronError ignores exceptions from InetAddress.getLocalHost().getHostName() and leaves the field unset.

The unit test however assumes it would be set, and someone has logged a jira on this, since it makes the build fail.

Changed the test so that it only verifies the hostName if it would have worked.

### Testing
- Code review
- Tests Pass

> no non-test changes in pr

```java
 private void addHostname(JSONObject errorMessage) {
    try {
      errorMessage.put(ErrorFields.HOSTNAME.getName(), InetAddress.getLocalHost().getHostName());
    } catch (UnknownHostException ex) {
      // Leave the hostname field off if it cannot be found
    }
  }
```

### For all changes:

### DIFF
--- a/metron-platform/metron-common/src/test/java/org/apache/metron/common/error/MetronErrorTest.java
+++ b/metron-platform/metron-common/src/test/java/org/apache/metron/common/error/MetronErrorTest.java
@@ -18,15 +18,14 @@
 package org.apache.metron.common.error;
 
 import com.google.common.collect.Sets;
-import com.google.common.primitives.Bytes;
 import org.apache.metron.common.Constants;
 import org.json.simple.JSONObject;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.net.InetAddress;
 import java.util.Arrays;
 
-import static org.apache.metron.common.Constants.ErrorFields.RAW_MESSAGE;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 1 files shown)


metron-platform/metron-common/src/test/java/org/apache/metron/common/error/MetronErrorTest.java
```diff
@@ -53,7 +52,14 @@ public void getJSONObjectShouldReturnBasicInformation() {
     assertEquals(Constants.ErrorType.PARSER_ERROR.getType(), errorJSON.get(Constants.ErrorFields.ERROR_TYPE.getName()));
     assertEquals("error", errorJSON.get(Constants.SENSOR_TYPE));
     assertEquals("sensorType", errorJSON.get(Constants.ErrorFields.FAILED_SENSOR_TYPE.getName()));
-    assertTrue(((String) errorJSON.get(Constants.ErrorFields.HOSTNAME.getName())).length() > 0);
+
+    try {
+      String hostName = InetAddress.getLocalHost().getHostName();
```
cestella: So, in the instance where this happens, is hostName is null or is it that there's an Exception thrown?  If it's null, can't we do a:
```
if(hostName == null) {
  LOG.warn("Unable to resolve local hostname, skipping validation."
}
else {
  assertTrue(((String) errorJSON.get(Constants.ErrorFields.HOSTNAME.getName())).length() > 0);
  assertEquals(hostName, (String) errorJSON.get(Constants.ErrorFields.HOSTNAME.getName()));
}
```

I ask because I worry about extraneous *other* exceptions cropping up and us ignoring them.  It's a very small nit, let me know what you think. :)

merrimanr:
+1 pending @cestella's approval.  Thanks @ottobackwards.

justinleet:
@cestella Bump

cestella:
+1, sorry!

PR #903, opened 2018-01-23
METRON-1370 Create Full Dev Equivalent for Ubuntu

nickwallen:
This PR adds a single node, VM based deployment mechanism based on Ubuntu Trusty.  This has feature parity with the existing CentOS single node VM.  This will help us continue to improve and support the Ubuntu-based DEB packages that were added in #868.

### Changes

* The Metron, Elasticsearch, and Kibana MPacks were updated to support installation on Ubuntu.  This included making the following changes.

    * Adding repo definitions for Ubuntu
    * Changing the default location of the `/etc/sysconfig/elasticsearch` configuration to work across platforms
    * Adding an additional configuration in the MPack to allow Elasticsearch to perform a memory lock on Ubuntu
    * Support configuration of Elasticsearch on a platform running Systemd; like Ubuntu 

* The Elasticsearch MPack was enhanced to correctly perform a service check.  Previously the hostname and port were hard coded and the service check was not functional in most environments.

* The Ubuntu DEBs were enhanced to allow for the addition of "maintainer scripts".  This allows the Management UI and Alerts UI service scripts to be installed with the DEBs.  This was needed to reach feature parity with the RPMs.

* The name `full-dev-platform` no longer makes sense IMHO.  We do not have a `quick-dev-platform` any longer to distinguish it from.  Now the main distinguisher is the underlying operating system.  I renamed our "Full Dev" development environment to `metron-on-centos` and `metron-on-ubuntu`.  I am completely open to community suggestions on what these should be named.

    ```
    metron-deployment/development/
    ├── README.md
    ├── fastcapa
    ├── centos6
    └── ubuntu14
    ```
    
* The organization of `metron-deployment` had to change so that assets could be reused across both the CentOS and Ubuntu deployments.  All shared Ansible assets were moved to `metron-deployment/ansible`.  This includes the Ansible modules, playbooks and roles that are all used across multiple environments.

    ```
    metron-deployment/ansible/
    ├── extra_modules
    ├── playbooks
    └── roles
    ```

* The following Ansible roles had to be updated to support both CentOS and Ubuntu based on small differences in package names and conventions.

    * `ambari-common` 
    * `ambari-config`
    * `ambari-master`
    * `ambari-slave`
    * `libselinux-python`
    * `ntp`

* A new role was added to enable swap space; `enable-swap`.  This is required for the Ubuntu deployment as the underlying Ubuntu Trusty image does not have swap space enabled by default.

[...]

PR #925, opened 2018-02-02
METRON-1443 Missing Critical MPack Install Instruction for Ubuntu

nickwallen:
When installing Elasticsearch with the MPack on Ubuntu, you must manually install the Elasticsearch repositories.  The Mpack itself does not do this, like it does on CentOS. 

When the development environment on Ubuntu is spun-up this step is performed within Ansible as a prerequisite to the Mpack install.  Until this can be fixed so that it matches what happens on CentOS, this needs to be at least documented.

I should have documented this in #903 , but did not do so.  Oops.


cestella:
+1 by inspection

nickwallen:
https://issues.apache.org/jira/browse/METRON-1444

PR #931, opened 2018-02-07
METRON-1449 Set Zookeeper URL for Stellar Running in Zeppelin Notebook

nickwallen:
This is the next step forward to run Stellar in Zeppelin.  The end goal being to define executable use cases in a sharable, Zeppelin Notebook.  To get there we need to be able to connect to Zookeeper to push/pull configuration changes to Metron.  This PR allows us to interact with Zookeeper within the Stellar 'ZInterpreter'.

## Changes

* This enhances the Stellar interpreter for Zeppelin to allow the user to set a `zookeeper.url` property.  

    * If the property is defined, a Zk client connection is accessible by the Stellar functions executed within Zeppelin.  If no `zookeeper.url` is defined, then the behavior remains the same as before.

* Multiple lines of Stellar can now be executed within the same block within a Zeppelin notebook.

* Updates to the README simplify installation of the interpreter in Zeppelin.

screen shot 2018-02-07 at 4 48 41 pm

## Testing

1. Follow the new README instructions (included in this PR) to build the interpreter and install it in Zeppelin.

1. Launch a Zookeeper instance.

1. Load up some basic configuration.  Create the standard Metron settings using `zkCli`.  It would look something like the following.
    ```
    create /metron ""
    create /metron/topology ""
    create /metron/topology/global "{}"
    create /metron/topology/parsers "{}"
    create /metron/topology/enrichments "{}"
    create /metron/topology/indexing "{}"
    ```

1. Set the zookeeper URL.  Go to the Interpreters > Stellar page and define a `zookeeper.url` property.

    ```
    zookeeper.url = localhost:2181
    ```

1. Add the following dependencies to the interpreter so that we can access the metron-management functions.
    
    Yes, this is ugly.  We need to fix some of our dependencies in the metron-management project (and others.)

    | artifact                                  | exclude                          |
    |-------------------------------------------|----------------------------------|
    | org.apache.metron:metron-management:0.4.3 |                                  |
    | org.apache.metron:metron-common:0.4.3     |                                  |
    | io.thekraken:grok:0.1.0                   | org.apache.commons:commons-lang3 |
    | org.apache.commons:commons-lang3:3.2      |                                  |
   
    screen shot 2018-02-07 at 4 36 53 pm

1. Save the interpreter changes, then open a notebook and execute the following. 

    ```
    CONFIG_GET("GLOBAL")
    ```

    An empty set of globals should be returned from Zookeeper.

1.  Run multiple expressions in a single Zeppelin code block.

    ```
    x := 2
    y := 3
    z := x + y
    ```

## Known Problems

1. There is a problem that occurs with the Zk cache for some of the metron-management functions.  Calling `CONFIG_GET` will only ever return the first value that it gets from Zk.  If you change the globals via a `CONFIG_PUT`, the new values will NOT be reflected by calling `CONFIG_GET` until you restart the interpreter.  

    This behavior differs from the CLI REPL.  I have been unable to determine why this is. I do not see anything in stellar-zeppelin that would cause this and am a little suspect of an issue in metron-management.  

   Have an idea what the problem might be?  Let me know!


nickwallen:
The last two commits are refactorings I performed trying to track down what I describe in the PR description as a "Known Problem".  I need to pull each of those out as separate PRs.

ottobackwards:
@nickwallen is this set for review? Or are you working on pulling the commits out?

nickwallen:
This is now ready @ottobackwards .  I reverted the two additional commits and submitted those as #982 and #983 .

ottobackwards:
Ran the steps with local zeppelin to full_dev node.  Works as described. I have found that I needed to reinstall zeppelin, since I had stellar installed already, and just removing it from the configs and /interpreters was not enough to clear it out. 

+1 .  Very nice

nickwallen:
Thanks for the review @ottobackwards !

ottobackwards:
@nickwallen would you like me to commit this? Or would you rather?

PR #934, opened 2018-02-09
METRON-1423: Ambari work to handle Solr configuration

merrimanr:
## Contributor Comments
This PR allows support for Metron-related Solr configuration and management through Ambari.  That does NOT include managing the actual Solr service.  For now this is done outside of Ambari.

High level changes include:

- Solr is now installed automatically in full dev, although not in a running state.  This implementation was inspired by the `install_solr.sh` script.
- Additional scripts for managing the Solr installation provided in full dev and installed with `install_solr.sh`.
- Exposes a single Zookeeper Solr configuration (for now) parameter through Ambari.
- Replaced the random access index writer class Ambari parameter with a random access search engine parameter that is a drop down of "Elasticsearch" and "Solr".  
- Replaced "Elasticsearch" with "Random Access" in all the Ambari parameter descriptions.
- Added the Solr schema creation to Ambari on Indexing start.  This matches the ES template loading implementation with the exception of exposing it through Service actions.

Testing instructions are as follows:
1. Spin up full dev
2. Shutdown ES and Kibana
3. Start Solr from the command line with $METRON_HOME/bin/start_solr.sh
4. Navigate to the Indexing tab in Ambari > Metron > Configs and change the "Random Access Search Engine" setting from "Elasticsearch" to "Solr"
5. Now you can start ingesting data into Solr in 2 ways.  After you make the change in Ambari it will prompt you to restart affected services.  A restart does not automatically create collections so you will need to do this outside of Ambari for bro, snort and error with $METRON_HOME/bin/create_collection.sh (first argument is sensor name) before you restart.
6. Or you can simply stop Metron Indexing and then start it.  This will trigger Ambari to automatically load the collections.
7. Navigate to http://node1:8983/solr/#/snort/query or http://node1:8983/solr/#/bro/query and hit the "Execute Search" button.  You should see data.

There is a lot to digest here and still more questions to answer but this should get us started.  Outstanding items include:

- What Solr parameters do we expose in Ambari?  All of them?
- Do we want to add Solr collection functions to Service actions?  I don't see a way to make this conditional on the selected search engine so both options would always be shown
- Need to review and add documentation
- Support for REST configuration (probably a separate PR)

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
Ran through test, everything worked fine. +1

On the questions:

- What Solr parameters do we expose in Ambari? All of them?

People are going to want to tune indexing.  If ambari is managing the configuration where those tuning parameters happen, then they need to be exposed

- Do we want to add Solr collection functions to Service actions? 

Does that mean that we can stop a certain collection?  We may want to have a custom view for this?

ottobackwards:
Can I ask a question : why aren't we using https://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.6.4/bk_solr-search-installation/content/ch_hdp-search-install-ambari.html?
Is there a reason?  If so we should document that reason.

simonellistonball:
Given that we're happy to accept limiting ourselves to a Hadoop distribution, it doesn't seem unfair to limit ourselves to a Solr distribution. While sticking pure raw Apache has some appeal, it may well make sense to stick to Ambari mpack based installs in the distro @ottobackwards suggest.

justinleet:
I wanted to chime in and make a distinction here for anyone looking in from the outside.  We're limiting ourselves to a Hadoop distribution for our dev environment, not in general.  Things primarily get tested on HDP for this reason, but we don't limit ourselves in general.

The problem with that mpack is that it uses Solr 5.5.2, and it would be nice to use a more modern version, based on the discuss thread. Having said that, I'm not at all opposed to using an mpack if it fits our needs. There are definitely benefits to having an install approach that fits the general mpack install we already do.

ottobackwards:
I would just like it documented as to why we are not using the HDP solr mpack.
Although, I would think we would be using that mpack as the example for ours?

merrimanr:
I agree with @justinleet.  I would prefer we use Solr 6+.  At some point HDP Search will move to 6+ and we can easily switch to the Mpack since installing Solr is a manual step now.  Happy to document that once we are all in agreement.

As for the parameters we expose in Ambari, I'm referring to the parameters stored in Zookeeper that the SolrWriter reads.  They are:

- solr.zookeeper
- solr.commitPerBatch
- solr.commit.waitSearcher
- solr.commit.waitFlush
- solr.commit.soft
- solr.collection
- solr.http.config

Currently Ambari only exposes the "solr.zookeeper" parameter which is necessary to get the indexing topology writing to Solr.  I think a reasonable solution would be to include all the others except "solr.http.config" since it's an advanced config.  Storm tuning parameters are independent of the writer implementation and already exposed in Ambari.  

For the collection service actions, it would mirror what we expose for ES:  creating and deleting templates (schemas in Solr terms).

metron-platform/metron-solr/src/main/scripts/create_collection.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+METRON_VERSION=${project.version}
+METRON_HOME=/usr/metron/$METRON_VERSION
+SOLR_VERSION=6.6.2
```
cestella: I notice that we're depending on maven artifacts of 6.6.0 and we set the solr version to 6.6.2 here, is there a reason for that?  If we can align them, then we could replace this line with SOLR_VERSION=`${global_solr_version}`

This question applies for the rest of the shell scripts too

cestella:
Chiming in late.  I agree that we should not have an explicit dependency on an indexing Mpack, even one not of our own construction.  I think people will have a lot of different ways to install solr and Metron's mpack should just be configured to point to an existing solr instance.

I would generally be in favor of adding support for:
* solr.commitPerBatch
* solr.commit.waitSearcher
* solr.commit.waitFlush
* solr.commit.soft
* solr.collection
* solr.http.config

but sensible defaults are chosen there and people can adjust them in the global config, so I think they can wait for a follow-on.  Some of them may be problematic to encode in a UI (solr.http.config is a map, for instance), but most of them would be pretty trivial.  Frankly, I'm a bit hesitant to give people the ability to screw with transactions details easily.

metron-platform/metron-solr/src/main/scripts/create_collection.sh
merrimanr: Good catch and I agree they should match.  I think the mistake is actually the maven version.  We initially explored using HDP Search (which is version 6.6.2) but it's not quite ready yet so we went ahead with a manual approach in the meantime.  The thinking was this would make it easy to switch to HDP Search in the future.

merrimanr: Latest commit should address this.

cestella:
+1 by inspection

PR #942, opened 2018-02-23
METRON-1461: Modify the MIN, MAX Stellar methods to take a stats or list object and return min and max

MohanDV:
## Contributor Comments
Presently the MIN and MAX stellar function accepts only the list  as input and the list may only contain objects that are mutually comparable / ordinal. Modify the method to take a stats or list object and return min/max.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/Ordinal.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/Ordinal.java
@@ -1,0 +1,24 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.metron.stellar.dsl.functions;
+
+public interface Ordinal {
+
+    double getMin();
+    double getMax();
+}
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/Ordinal.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/Ordinal.java
@@ -17,8 +17,20 @@
  */
 package org.apache.metron.stellar.dsl.functions;
 
+/**
+ * Interface that provides the statistical function get max and min from the implementing object.
+ */
 public interface Ordinal {
 
+    /**
+     * get the min value
+     * @return min value
+     */
     double getMin();
+
+    /**
+     * get the max value
+     * @return max value
+     */
     double getMax();
 }
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/OrdinalFunctions.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/OrdinalFunctions.java
@@ -22,12 +22,10 @@ import com.google.common.collect.Iterables;
 import org.apache.metron.stellar.common.utils.ConversionUtils;
 import org.apache.metron.stellar.dsl.BaseStellarFunction;
 import org.apache.metron.stellar.dsl.Stellar;
-
-import java.util.Collections;
 import java.util.List;
 import java.util.function.BiFunction;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
+import static org.apache.metron.stellar.common.utils.ConversionUtils.convert;
+
 
 public class OrdinalFunctions {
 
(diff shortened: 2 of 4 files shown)


metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/OrdinalFunctions.java
```diff
@@ -37,17 +35,23 @@
...
+          ,"list - List of arguments. The list may only contain objects that are mutually comparable / ordinal (implement java.lang.Comparable interface)" +
           " Multi type numeric comparisons are supported: MAX([10,15L,15.3]) would return 15.3, but MAX(['23',25]) will fail and return null as strings and numbers can't be compared."}
-          , returns = "The maximum value in the list, or null if the list is empty or the input values were not comparable.")
+          , returns = "The maximum value in the list or from stats, or null if the list is empty or the input values were not comparable.")
   public static class Max extends BaseStellarFunction {
 
     @Override
     public Object apply(List<Object> args) {
       if (args.size() < 1 || args.get(0) == null) {
         throw new IllegalStateException("MAX function requires at least a Stellar list of values");
```
nickwallen: With your changes, this error message is now incorrect.  Can you update this? 

This only checks the number of args, so the error message should probably just say that we expect one argument or something to that effect.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/OrdinalFunctions.java
```diff
@@ -37,17 +35,23 @@
...
       }
+      Object firstArg = args.get(0);
+      if(firstArg instanceof Ordinal) {
+        Ordinal stats = convert(firstArg, Ordinal.class);
+        return stats.getMax();
+      }
       Iterable list = (Iterable<Object>) args.get(0);
```
nickwallen: It would make sense to wrap the existing "iterable" handling code in an "else if".  And also handle the possibility that the argument is not an Iterable nor an Ordinal. Perhaps like so...

```
Object firstArg = args.get(0);
if(firstArg instanceof Ordinal) {
  <your new code>

} else if(firstArg instanceof Iterable) {
  <existing code>

} else {
   throw new IllegalStateException("MAX function expects either .... ");

}
```

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/Ordinal.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.metron.stellar.dsl.functions;
+
+public interface Ordinal {
```
nickwallen: Can you add javadocs for the class and each method?

MohanDV:
@nickwallen Thanks for the review, I have incorporated your comments.

nickwallen:
@MohanDV - It looks like some care was taken previously so that getting the max of a list of mixed elements will just work.  For example `MAX([1, 2d, 3f]) == 3f`.  

Did you consider an implementation such that this would continue to work with STATS objects?  I imagine that something like the following could be made to work; `MAX[1, 2d, 3f, stats]`?  

What do you think?

MohanDV:
@nickwallen I dint consider implementation where STATS object can be passed in list of mixed object . That requires to change the STATS object to be 'Comparable' type . Do you think it should be addressed in this pull request ?

nickwallen:
I would like to address the issue somehow or at least garner more community feedback on this change.

As it stands, usage of the function is not very consistent.  For example, I can pass a list of numbers, but I can't pass a list of Stat objects.  I can have a list of mixed numeric types, but I can't have a Stats object in a mixed list.  That is inconsistent IMO.

nickwallen:
The first point here is around semantics.  I am assuming the semantics of this would be a "max of maxes".  So if I have a list of stats objects, I compare the max of each one.  Whichever max is the greatest, that stats object gets returned.  The MIN function would just be a "min of mins".

Does that match your use case?  Is this the right approach?

Given those semantics, I think the 'Comparable' approach could work, but with a twist.  You can't just make the Stats objects Comparable.  Because how do you compare them?  By the average, median, min or max?  There is not one way to do it that is broadly applicable.

The means of comparison for the MAX function should use the max of a stats object.  The means of comparison for the MIN function should use the min of the stats object. 

One way is to create a class that wraps a Stats object and implements the Comparable interface.  In the case of MAX, the wrapper will compare using the max of the underlying stats object.  In the case of MIN function, the wrapper will compare using the min of the underlying stats object.

ottobackwards:
Maybe making this function so generic is going to necessitate it being so complicated that it is harder to maintain etc.

simonellistonball:
I would say performance trumps complexity of functionality here.

cestella:
So, I would suggest rather than accepting a list of stats objects, `MAX` and `MIN` accept one of the following:
* A StatisticsProvider object
* A list of comparables

Essentially, these two would both function:
* `MAX(STATS_ADD(null, 1, 2, 3))`
* `MAX([1, 2, 3])`

simonellistonball:
Do we think we've reached consensus on this? It seems like avoiding the mixed scenarios is a good thing from a performance perspective, which is key to something like this which will be used a lot in baselining use cases. Any one else have any open thoughts on this?

cestella:
I think so.

MohanDV:
If we've reached consensus can I get +1 for this ?

cestella:
+1 by inspection, lgtm

MohanDV:
Thanks @cestella . Can you please merge this pull request.

PR #947, opened 2018-03-02
METRON-1467: Replace guava caches in places where the keyspace might be large

cestella:
## Contributor Comments
Based on the performance tuning exercise as part of METRON-1460, guava has difficulties with cache sizes over 10k.  We, unfortunately, are quite demanding of guava in this regard so we should transition a few uses of guava to Caffeine:

* Stellar processor cache
* The JoinBolt cache
* The Enrichment Bolt Cache

Test plan:
First, spin up full-dev and ensure things continue to work.

For myself, I created a new type called `dummy`, added some stellar enrichments, stellar threat triage enrichments and a triage rule to ensure things continued to work with stellar (exercising the stellar processor cache and the enrichment bolt cache).

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


cestella:
I ran this up with vagrant and ensured:
* Normal stellar works still in field transformations as well as enrichments
* swapped in and out new enrichments live
* swapped in and out new threat intel live

metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/JoinBolt.java
```diff
@@ -89,29 +91,25 @@ public void prepare(Map map, TopologyContext topologyContext, OutputCollector ou
...
-        return new HashMap<>();
-      }
-    };
-    cache = CacheBuilder.newBuilder().maximumSize(maxCacheSize)
-            .expireAfterWrite(maxTimeRetain, TimeUnit.MINUTES).removalListener(new JoinRemoveListener())
-            .build(loader);
+    loader = s -> new HashMap<>();
+    cache = Caffeine.newBuilder().maximumSize(maxCacheSize)
+                         .expireAfterWrite(maxTimeRetain, TimeUnit.MINUTES)
+                         .removalListener(new JoinRemoveListener())
```
nickwallen: It seems like we only want notified of a full cache when ERROR logging is set. Is that the case? In the `JoinRemoveListener` we end up doing some work that we probably don't need to do unless ERROR logging is set.  One easy fix would be to only add the "remove listener" if `LOG.isDebugEnabled()`.

cestella: So, I believe this was intentionally done before this PR (I migrated this to the new caching strategy) and the idea is that if a removal is happening from the join cache under specific circumstances, we want to know about it because a message could be being dropped because the cache is being overwhelmed.  @merrimanr Can you chime in here on the rationale?

nickwallen: Yes, it is pre-existing.  We can address at a later time.

I remember now, maxing out this cache causes the Split/Join to fail, which is a major problem for the Split/Join topology.  And this cache here is only for the Split/Join, not the Unified topology.

We should probably look at adding similar logging (only when ERROR enabled) for the other places where we use the cache.  Or just some mechanism to periodically log cache stats.  Anywho, down the road.

nickwallen:
+1 LGTM

PR #952, opened 2018-03-07
METRON-1480 Add yarn as default build tool for the frontend

xyztdanid4:
Updated the build tool to yarn for the frontend project to avoid npm instability during dependency adding.

### For all changes:
- [X] Is there a JIRA ticket associated with this PR? If not one needs to be created at [https://issues.apache.org/jira/browse/METRON-1480](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel).

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


simonellistonball:
Have the dependencies, License and Notices files been updated?

xyztdanid4:
@simonellistonball I am not sure. Probably no.

metron-interface/metron-alerts/package-lock.json
```diff
@@ -1,6427 +0,0 @@
-{
```
cestella: I suspect the build failure is due to a lack of a license on this file.  Since JSON cannot take comments, you'll need to adjust the top level pom to add an exclusion for `**/package-lock.json` [here](https://github.com/apache/metron/blob/master/pom.xml#L304)  Just make a quick comment above it and note that this is yarn collateral.

xyztdanid4: I think that was the problem!
Thanks. @cestella

metron-interface/metron-config/package.json
```diff
@@ -33,6 +28,11 @@
     "@angular/platform-browser": "2.0.0",
     "@angular/platform-browser-dynamic": "2.0.0",
     "@angular/router": "3.0.0",
+    "@types/ace": "0.0.32",
```
merrimanr: Were there any real changes made to this file?  It looks like some dependencies were simply rearranged.  I do see the addition of `"main": "index.js"`.  What is the purpose of that?

metron-interface/metron-alerts/package.json
```diff
@@ -27,8 +27,8 @@
     "core-js": "^2.4.1",
     "font-awesome": "^4.7.0",
     "moment": "^2.18.1",
-    "pikaday-time": "^1.6.1",
     "ng2-dragula": "^1.5.0",
+    "pikaday-time": "^1.6.1",
```
merrimanr: See comment for metron-config/package.json.

merrimanr:
I'm having trouble understanding what this PR actually does.  I see that package-lock.json was removed and yarn.lock was added.  Is that all that's needed to switch to yarn?  Do we need to also update our build commands to match those [here](https://yarnpkg.com/en/docs/usage)?

What considerations are there for development?  Are we supposed to be using yarn commands now instead of npm commands?  Do we need to install yarn locally?  We probably need to review our documentation and rpm build processes too.  We use npm all over the place.

metron-interface/metron-config/package.json
xyztdanid4: No there were no changes other than rearrange of the dependencies.
About the main attribute, in our case it will do nothing. It defines the entry point of the application if we would SystemJS or requireJS. It is the default setting for yarn.

xyztdanid4:
@merrimanr

metron-interface/metron-config/package.json
merrimanr: What was the reason for rearranging the dependencies?  It's not a big deal but it's distracting and makes reviewing harder.

xyztdanid4:
@merrimanr 
Yes this is the right way to switch to yarn.
Yes we have install yarn locally.
I did not know that we have so much dependency on npm. But for me it was hard to add new dependencies using npm, cause it was constantly failing. And yes I would recommend to use yarn, at least for the front end builds. What yarn does is that reduces the possibility of version mismatch  of the dependencies by importing the dependencies subdependency in the right version. So yes we have to use yarn commands if we merge this PR.

metron-interface/metron-config/package.json
xyztdanid4: Yes. I was not enough carefully. Yarn did the rearrangement, but I can move the dependencies back to their original place if you would like to.

xyztdanid4:
@merrimanr 
although it should work with npm as well, cause then it will use the package.json to install the dependencies.

metron-interface/metron-config/package.json
merrimanr: No I was just curious.  If Yarn did this automatically we might as well leave it that way.  No need to revert.

merrimanr:
I think we are on board with switching to yarn so no concerns there.  There was a discussion on it but if anyone now feels differently, speak up.

Sounds like we still have a lot of work to do.  I would suggest reviewing the metron-config and metron-alerts READMEs and update them appropriately.  In fact, I would just search the whole project for npm commands and switch to yarn as needed.  You will definitely need to run this up in full dev and verify everything works.

Our maven front end plugin will also need to be reviewed and updated:  https://github.com/eirslett/frontend-maven-plugin.  We use that in metron-config and metron-alerts.

xyztdanid4:
@merrimanr 
As we agreed yesterday I did the required changes in the pom xml-s.
I had to update the node dependency of the projects, otherwise the build was failed, cause one of the dependencies was dependent on node8.
There is no equivalent command to npm --prefix in yarn, so I updated the readme to just cd into that directory where the install step is required.
When we build the metron-alert project (angular4), there was an error, mainly because angular4 uses AOT as default builder strategy. I fixed this by addig env=prod to the package json.

sardell:
> What yarn does is that reduces the possibility of version mismatch of the dependencies by importing the dependencies subdependency in the right version

I believe npm also does this with the addition of [package-lock.json](https://docs.npmjs.com/files/package-lock.json).

nickwallen:
@sardell Thanks.

@xyztdanid4 Please let us know if you feel this PR is still relevant.  If not, please close the PR.  Otherwise, the PR may be closed as inactive according to the Metron Development Guidelines.

nickwallen:
-1 as this has become an abandoned PR.  I will open a request to close this.

PR #969, opened 2018-03-20
METRON-1496: ChainLink Parser to reuse parser code at parserConfig level

kevin91nl:
## Contributor Comments
[Please place any comments here.  A description of the problem/enhancement, how to reproduce the issue, your testing methodology, etc.]

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


kevin91nl:
We would like to share our configurable parser architecture for review and discussion. Our parsers are currently found in `metron-platform/metron-parsers-contrib` which might be changed.

cestella:
Thanks for the contribution!  First thing, though, the package naming should be migrated to org.apache.metron, please.

cestella:
Also, just briefly looking at this, can you speak a bit about how the chain links are different from field transformations?  It seems like they're both configurable mechanisms to execute pluggable code in an ordered manner over the message.  I mostly want to understand how they differ and if they differ enough to make sense as a separate thing or if it may be worth-while to migrate them to field transformations.

ottobackwards:
This seems redundant to the metron stellar system.  We should have a discuss thread maybe where Chainlinks are proposed and you can explain why they may be better.

That aside, you probably want to get rid of the references to your company name in the namespaces and classes.  Also, please revert and remove the stye only modifications to the antlr generated classes.

kevin91nl:
The package naming is adjusted in the latest commit.

**Chainparsers versus Stellar**
Here I will present some of the issues we encountered when writing the parsers. I am interested in your thoughts :-).

_Keeping memory_
In one of our to-be-written parsers, we need to read an external file, parse the file and keep the parsed file in memory. As far as I can see, Stellar is stateless and not capable of keeping data into memory. It might be possible by using environment variables, but this approach might get messy.

_Performance_
A ChainParser does not need to interpret a language. We found that interpreting a language was a performance bottleneck in the parsers. In fact, we implemented a template engine in one of our links, but that was breaking the throughput rate of the parsers. Therefore, we removed this mechanism.

_Tests_
There is no clear way how to write end-to-end tests for parsers involving Stellar. Using the ChainParser approach, the data files (input and expected output) and parser config are kept in one place and it is easy to add new tests for a parser.

_Sharing configuration_
When parsers are created using Stellar, there is no clear way how the parser configuration files are shared.

_Post processing_
Note that Stellar can still be used for post-processing the parser output. Therefore, parsers can fully rely on all the Stellar functionality.

nickwallen:
@kevin91nl Thanks for the submission! I really do appreciate the clarity of your code and explanations.

> Performance A ChainParser does not need to interpret a language the way Stellar does. We found that interpreting a language was a performance bottleneck in the parsers. In fact, we implemented a lightweight template engine in one of our links, but that was breaking the throughput rate of the parsers. Therefore, we removed this mechanism.

Can you elaborate on the performance bottleneck that you experienced?  

I have tested this myself and found little to no performance difference between the execution of something written in raw Java versus Stellar.  I compared doing Geo enrichments in the legacy Java adapter versus doing them with `GEO_GET` in Stellar.  I also compared doing enrichments with the legacy HBase Java adapter versus `ENRICHMENT_GET`.

Maybe the root cause of the performance bottleneck that you experienced was something else, a tuning issue or an unexpected problem with your template engine?  There also certainly could be a performance bug somewhere in Stellar that I am not aware of, but I think it is reasonable to expect similar performance between raw Java and Stellar.   If we do have a performance issue with Stellar, then I'd like to isolate and address it.

nickwallen:
> Stateful parsing: In one of our to-be-written parsers, we need to read an external file, parse the file and keep the parsed file in memory. As far as I can see, Stellar is stateless and not capable of keeping data in memory. It might be possible by using environment variables, but this approach might get messy.

With the heavy caching we use I wouldn't expect this to be a problem.  There is also a function like `OBJECT_GET` that allows you to specifically customize a backing cache.  This seems to fit the use case that you are describing, but I could be wrong.

https://github.com/apache/metron/tree/master/metron-stellar/stellar-common#object_get

nickwallen:
@kevin91nl I do think that you've introduced a really useful concept though; chained parsers.  I did not want to leave that unsaid.

Your points about testing and sharing configuration are completely valid and should be addressed.  Maybe there are also usability improvements we could make here based on your experience.

Ultimately, I think doing a little work to merge your chained parsers into the existing architecture and Stellar would be really useful.

kevin91nl:
@nickwallen We used Jinjava for rendering Jinja templates, but that drastically decreased the EPS. Now we use simple replaces of variables, but it is less flexible than the template render engine. I think it is not a fair comparison to the language interpreter of Stellar, since the language interpreter of Stellar would be focused on speed as you showed with the comparison.

I was not aware of the OBJECT_GET method of Stellar, but it might get messy when the object is preprocessed in Stellar.

Next week I will discuss the findings and update the pull request such that it fits into the existing architecture.

kevin91nl:
What exact changes are needed such that it fits in the existing architecture?

PR #975, opened 2018-03-28
METRON-1503: Alerts are not getting populated in alerts UI when search engine is Solr

merrimanr:
## Contributor Comments
Solr throws an error when an index included in a query doesn't exist where Elasticsearch simply ignores that error.  This PR makes the Solr implementation equivalent to the Elasticsearch implementation by removing an indices that don't exist from the query before it is executed.

This has been tested on full dev.  After spinning it up and switching the Indexing search engine to Solr, alerts should appear in the Alerts UI.  I also slightly modified a test case to catch this in the future.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrSearchDao.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrSearchDao.java
@@ -51,6 +51,8 @@ import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.SolrQuery;
 import org.apache.solr.client.solrj.SolrQuery.ORDER;
 import org.apache.solr.client.solrj.SolrServerException;
+import org.apache.solr.client.solrj.request.CollectionAdminRequest;
+import org.apache.solr.client.solrj.response.CollectionAdminResponse;
 import org.apache.solr.client.solrj.response.FacetField;
 import org.apache.solr.client.solrj.response.FacetField.Count;
 import org.apache.solr.client.solrj.response.PivotField;
@@ -84,8 +86,8 @@ public class SolrSearchDao implements SearchDao {
       throw new InvalidSearchException(
           "Search result size must be less than " + accessConfig.getMaxSearchResults());
     }
-    SolrQuery query = buildSearchRequest(searchRequest);
     try {
+      SolrQuery query = buildSearchRequest(searchRequest);
       QueryResponse response = client.query(query);
       return buildSearchResponse(searchRequest, response);
     } catch (IOException | SolrServerException e) {
@@ -153,7 +155,7 @@ public class SolrSearchDao implements SearchDao {
   }
 
   private SolrQuery buildSearchRequest(
-      SearchRequest searchRequest) {
+      SearchRequest searchRequest) throws IOException, SolrServerException {
     SolrQuery query = new SolrQuery()
         .setStart(searchRequest.getFrom())
         .setRows(searchRequest.getSize())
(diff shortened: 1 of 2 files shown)


nickwallen:
Can this fix be tested in a unit or integration test?

merrimanr:
I modified a case in SearchIntegrationTest.java so that the test would have failed before this change but passes now.  

There is no unit test right now for the SolrSearchDao.  I think the original reason was because that class is very well covered by the integration test.  Do you think we need one?

nickwallen:
Yes, I would argue that integration tests are no replacement for unit tests. Unless there is some reason we cannot unit test this thing more directly.  And, if that is the case, then I'd argue we need to refactor.

Think of the dread you get when an integration test fails.  'Oh crap, now what happened?'  And you get that feeling because it is very difficult to pinpoint root cause when an integration test fails.  We really should focus on more direct, targeted testing, unless there is some technical hurdle here.

nickwallen:
I agree.  Doing it right is going to save us in the long run.  I'd be happy to review it in this PR.  If you need me to help with any of this, just let me know.

merrimanr:
The latest commit adds unit tests for all the Solr Dao classes except SolrMetaAlertDao.  Mocking some of the Solr client classes (and client libraries in general) was challenging so I opted to move them to their own methods within the Daos and spy on those methods.  I feel like this achieves the same goal with less complexity.  

I also didn't test for cases where exceptions from Solr client classes and just caught and thrown.  If you think there is value in doing this it probably won't require that much effort to add them.

I ran into this bug while adding these tests:  https://github.com/powermock/powermock/issues/731.  Incrementing the powermock version we use solved it.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/SearchResponse.java
```diff
@@ -61,4 +61,20 @@ public void setResults(List<SearchResult> results) {
   public void setFacetCounts(Map<String, Map<String, Long>> facetCounts) {
     this.facetCounts = facetCounts;
   }
+
+  @Override
+  public boolean equals(Object o) {
```
nickwallen: Would it make sense to throw in a matching `hashCode(..)` just to avoid any potential future headaches?

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
```diff
@@ -72,12 +72,12 @@ public SolrDao() {
   @Override
   public void init(AccessConfig config) {
     if (config.getKerberosEnabled()) {
-      HttpClientUtil.addConfigurer(new Krb5HttpClientConfigurer());
+      enableKerberos();
```
nickwallen: Thank you for the refactorings like this.  Even though small, they help with readability a bunch.  Bravo.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrSearchDao.java
```diff
@@ -176,19 +179,23 @@ private SolrQuery buildSearchRequest(
       facetFields.get().forEach(query::addFacetField);
     }
 
-    String collections = searchRequest.getIndices().stream().collect(Collectors.joining(","));
-    query.set("collection", collections);
+    query.set("collection", getCollections(searchRequest.getIndices()));
```
nickwallen: This is the key fix for the bug that you found.  When I comment this out and run the tests, the integration test does fail, which is great.  

Unfortunately, I would have expected the `SolrSearchDaoTest` to fail too, but it did not.  If you think it is easy enough to fail a unit test on this condition, it would be a good add.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/search/SearchResponse.java
merrimanr: Done

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrSearchDao.java
merrimanr: Done

nickwallen:
+1 Good stuff

PR #984, opened 2018-04-05
METRON-1499 Enable Configuration of Unified Enrichment Topology via Ambari

nickwallen:
We recently introduced an alternative enrichment topology; the Unified enrichment topology.  This is an alternative to the existing Split Join topology.  This work enhances the MPack so that a user can configure, stop, and start either the Unified or the Split Join enrichment topology.  

* The Split Join topology remains the default Enrichment topology.

* A combo box allows the user to choose which Enrichment topology should be run.

    screen shot 2018-04-05 at 5 41 52 pm

* Each topology has a separate configuration section under the Metron > Enrichment settings in Ambari.

    screen shot 2018-04-05 at 5 42 19 pm

* The `bin/start_enrichment_topology.sh` script was enhanced so that you can pass in the `--filter` and `--remote` args directly to launch an alternative topology.  Running the script with no parameters will launch the default topology; Split Join.

* The Unified topology had previously reused some of the same settings from the Split Join topology.  Where it made sense, the Unified topology was updated to use its own logically named properties.

*  There are now separate properties file for each Enrichment topology. 
    `$METRON_HOME/config/enrichment-splitjoin.properties`
    `$METRON_HOME/config/enrichment-unified.properties`

* There are separate flux files for each Enrichment topology. 
    `$METRON_HOME/flux/enrichment/remote-splitjoin.yaml` 
    `$METRON_HOME/flux/enrichment/remote-unified.yaml`

### Manual Testing

To test this change, launch a development environment.  The Split Join Enrichment topology will be running by default.  Switch to the Unified topology.  Change some settings and make sure they are reflected in the topology.

## Pull Request Checklist


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-enrichment-env.xml
```diff
@@ -129,70 +145,156 @@
   <property>
     <name>enrichment_topology_max_spout_pending</name>
     <description>Spout Max Pending Tuples for the Enrichment Topology</description>
-    <value/>
-    <display-name>Enrichment Max Pending</display-name>
+    <value>500</value>
```
cestella: This seems super low..shouldn't we leave this unspecified and default to the storm default?

nickwallen: This is about tuning the user's day 1 experience.  Specifically what I have seen is on day 1 users push some telemetry into a topic, start up the topology, and the topology just blows-up.  They have no idea why because they are on day 1.   That is a bad experience.

The topology blows-up because on day 1 the user is running with a topology that has NOT been tuned.  The topology blows-up because when this value is undefined there is no back pressure.  We need some back pressure here to avoid the day 1 blow-up.

I'd prefer a day 1 user be able to start up the topology and see it actually work, albeit slowly, under most conditions.  A user running against heavier load is more likely to be experienced and should have more knowledge to know that this value needs increased.

nickwallen: But this is definitely not necessary for the task at hand; METRON-1499.  

If this is contentious, I can certainly pull this change out of this PR.  I could then submit a separate PR where we can debate this point.  I don't want to hang up this PR for something unrelated.  

I'm good with either way.

ottobackwards: +1 for a sane default

cestella: I can get behind a sane default.  I retract my comment.

nickwallen:
Let me know if there is anything else we need on this one guys.  Thanks for the looks so far.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-enrichment-env.xml
nickwallen: Ok, that works.  It certainly deserved an explanation from me, so glad you called that out.

cestella:
This looks good to me; +1 by inspection.

mraliagha:
Great Idea to keep both for a while. Thanks.

PR #985, opened 2018-04-09
METRON-1515: Errors loading stellar functions currently bomb the entire topology, they should be recoverable

cestella:
## Contributor Comments
If a bad stellar function gets on the classpath, the entire enrichment topology (or shell) bombs.  We should just log an error and skip the function.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/ClasspathFunctionResolver.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/ClasspathFunctionResolver.java
@@ -34,6 +34,7 @@ import org.apache.metron.stellar.dsl.Context;
 import org.apache.metron.stellar.dsl.Stellar;
 import org.apache.metron.stellar.dsl.StellarFunction;
 
+import org.atteo.classindex.ClassFilter;
 import org.atteo.classindex.ClassIndex;
 import org.reflections.util.FilterBuilder;
 
@@ -219,6 +220,17 @@ public class ClasspathFunctionResolver extends BaseFunctionResolver {
     }
   }
 
+  protected Iterable<Class<?>> getStellarClasses(ClassLoader cl) {
+    return ClassIndex.getAnnotated(Stellar.class, cl);
+  }
+
+  protected boolean includeClass(Class<?> c, FilterBuilder filterBuilder)
+  {
+    boolean isAssignable = StellarFunction.class.isAssignableFrom(c);
+    boolean isFiltered = filterBuilder.apply(c.getCanonicalName());
+    return isAssignable && isFiltered;
+  }
+
   /**
    * Returns a set of classes that should undergo further interrogation for resolution
    * (aka discovery) of Stellar functions.
@@ -254,12 +266,10 @@ public class ClasspathFunctionResolver extends BaseFunctionResolver {
     Set<String> classes = new HashSet<>();
     Set<Class<? extends StellarFunction>> ret = new HashSet<>();
     for(ClassLoader cl : cls) {
-      for(Class<?> c : ClassIndex.getAnnotated(Stellar.class, cl)) {
+      for(Class<?> c : getStellarClasses(cl)) {
         try {
           LOG.debug("{}: Found class: {}", cl.getClass().getCanonicalName(), c.getCanonicalName());
-          boolean isAssignable = StellarFunction.class.isAssignableFrom(c);
-          boolean isFiltered = filterBuilder.apply(c.getCanonicalName());
-          if (isAssignable && isFiltered) {
+          if (includeClass(c, filterBuilder)) {
             String className = c.getName();
             if (!classes.contains(className)) {
               LOG.debug("{}: Added class: {}", cl.getClass().getCanonicalName(), className);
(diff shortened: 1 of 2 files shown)


ottobackwards:
Is there a way to test this in the unit tests?

cestella:
@ottobackwards I'm still contemplating that one ;)

cestella:
Ok, I added a unit test for this as requested.

ottobackwards:
+1 by inspection.  Thanks for the contribution!

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/ClasspathFunctionResolver.java
```diff
@@ -254,18 +266,24 @@ public void initialize(Context context) {
...
+              LOG.debug("{}: Added class: {}", cl.getClass().getCanonicalName(), className);
+              ret.add((Class<? extends StellarFunction>) c);
+              classes.add(className);
+            }
           }
         }
+        catch(Error le) {
+          //we have had some error loading a stellar function.  This could mean that
+          //the classpath is unstable (e.g. old copies of jars are on the classpath).
+          LOG.error("Skipping class: " + le.getMessage()
```
nickwallen: The exception message might not always contain the class name.  We should explicitly log the class name, like we do in the debug statement on line 271.

cestella: Yeah, this is tricky, getting the classname may actually cause the exception (this is precisely the scenario that motivated this, sadly).  That's why I forewent that approach.

cestella: For posterity, this is the stack trace that motivated this:
```
org.apache.metron.stellar.dsl.functions.MathFunctions and org.apache.metron.stellar.dsl.functions.MathFunctions$Sqrt disagree on InnerClasses attribute
at java.lang.Class.getDeclaringClass0(Native Method)
at java.lang.Class.getDeclaringClass(Class.java:1235)
at java.lang.Class.getEnclosingClass(Class.java:1277)
at java.lang.Class.getCanonicalName(Class.java:1392)
at org.apache.metron.stellar.dsl.functions.resolver.ClasspathFunctionResolver.resolvables(ClasspathFunctionResolver.java:250)
```
The problem is in the `getCanonicalName`.  I suppose I can do a `getName`, but I'm a bit wary of even that, frankly.

cestella: On second thought, I think I might've squared the circle there in the latest commit.  How do you like what I came up with?

nickwallen: Commons has a method that handles this situation well.  I tend to always use that in these situations;  `ClassUtils.getCanonicalName(Class<?> cls, String valueIfNull)`.  If I'm not mistaken, it would make your logic simpler; only one error log statement.

cestella: I can certainly use `ClassUtils.getCanonicalName` rather than `Class.getName()`, but I would still feel more comfortable with the 2nd try/catch as I cannot guarantee that ClassUtils won't throw an Error in the case of a malformed class on the classpath.  Am I missing any guarantees offered by ClassUtils here?

ottobackwards:
```java
  /**
     * <p>Gets the canonical class name for a {@code Class}.</p>
     *
     * @param cls the class for which to get the canonical class name; may be null
     * @return the canonical name of the class, or the empty String
     * @since 3.7
     * @see Class#getCanonicalName()
     */
    public static String getCanonicalName(final Class<?> cls) {
        return getCanonicalName(cls, StringUtils.EMPTY);
    }
```

This is in latest commons-lang.

getCanonicalName returns null or the name.  In either  the function or the class function.

cestella: We use commons-lang <= [2.6](http://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/ClassUtils.html), which does not have that method.  I'd prefer to not update as part of this PR as it seems to me to be an overreach from a risk perspective.

ottobackwards: The risk of regression with commons-lang is surely much less than with other components is it not? But, if you are thinking that this would mean updating for the whole project.. then I can understand the scope issues.

nickwallen:
+1 LGTM

PR #986, opened 2018-04-10
METRON-1518 Build Failure When Using Profile HDP-2.5.0.0

nickwallen:
The build fails when using the 'HDP-2.5.0.0' profile.
```
/Users/ottofowler/src/apache/forks/metron/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java:45:
error: cannot find symbol
import org.apache.storm.StormTimer;
                       ^
  symbol:   class StormTimer
  location: package org.apache.storm
/Users/ottofowler/src/apache/forks/metron/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java:167:
error: cannot find symbol
  private StormTimer expiredFlushTimer;
          ^
  symbol:   class StormTimer
  location: class ProfileBuilderBolt
/Users/ottofowler/src/apache/forks/metron/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java:450:
error: cannot find symbol
  private StormTimer createTimer(String name) {
          ^
  symbol:   class StormTimer
  location: class ProfileBuilderBolt
```

This was caused by my PR; #977.  That code relied on a class that is only available in Storm 1.0.3+.  When using the Maven profile `HDP-2.5.0.0`, the code is built against Storm 1.0.1, where that class is not defined.

### Changes

* This completely removes the use of a`StormTimer`.  There is no need for it.  I had thought it would help me drive testing, but it has not ended up being useful at all.  

* An executor is used to drive a separate thread which flushes expired profiles on a regular basis.

### Manual Testing

* Run a development environment and ensure that alerts are visible in the Alerts UI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
   - npm config set prefix $HOME/.npm-prefix --global
 
 install:
-  - time mvn -q -T 2C -DskipTests -PHDP-2.5.0.0 clean install
+  - time mvn -q -T 2C -DskipTests clean install
 
 script:
   - time mvn -q -T 2C surefire:test@unit-tests && time mvn -q surefire:test@integration-tests && time mvn -q test --projects metron-interface/metron-config && time dev-utilities/build-utils/verify_licenses.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
   - npm config set prefix $HOME/.npm-prefix --global
 
 install:
-  - time mvn -q -T 2C -DskipTests clean install
+  - time mvn -q -T 2C -DskipTests -PHDP-2.5.0.0 clean install
 
 script:
   - time mvn -q -T 2C surefire:test@unit-tests && time mvn -q surefire:test@integration-tests && time mvn -q test --projects metron-interface/metron-config && time dev-utilities/build-utils/verify_licenses.sh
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -33,7 +33,7 @@ before_install:
   - npm config set prefix $HOME/.npm-prefix --global
 
 install:
-  - time mvn -q -T 2C -DskipTests clean install
+  - time mvn -q -T 2C -DskipTests -PHDP-2.5.0.0 clean install
```
cestella: I believe the rationale for having travis run without the 2.5.0.0 profile and full-dev build with it is to ensure that we build with both per PR (since both are required for a +1).  I would be more comfortable with a follow-on task to make a proper travis matrix where both are run in travis per build.

cestella: Just so I'm completely clear, by "build matrix" I mean [this](https://docs.travis-ci.com/user/customizing-the-build#Build-Matrix)

nickwallen: That works.

cestella:
+1 by inspection, thanks!

ottobackwards:
+1 ship it

nickwallen:
Thanks for the reviews.  Just to be sure, I ran up the CentOS dev environment, validated that alerts are being created, created a profile and made sure that values were written.

PR #990, opened 2018-04-11
METRON-1520: Add caching for stellar field transformations

cestella:
## Contributor Comments
Given how important caching is in the enrichment topology, we should have caching for stellar field transformations in the parsers as well.

Beyond a smoketest ensuring data flows into the indices, you may test this by creating a new parser with a field transformation and ensuring stellar field transformations function as expected.

* Create a new parser by editing `$METRON_HOME/config/zookeeper/parsers/dummy.json` to have the following content:
```
{
  "sensorTopic":"dummy",
  "fieldTransformations" : [
         { "transformation" : "STELLAR"
         ,"output" : [ "new_field"]
         ,"config" : {
           "new_field" : "JOIN( [ TO_UPPER(source.type), field ], ',' )"
                     }
         }
  ]
}
```
* Create the dummy kafka topic via `/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper node1:2181 --create --topic dummy --partitions 1 --replication-factor 1`
* Push the zookeeper configs via `$METRON_HOME/bin/zk_load_configs.sh --mode PUSH -i $METRON_HOME/config/zookeeper -z node1:2181`
* Start the parser via `$METRON_HOME/bin/start_parser_topology.sh -k node1:6667 -z node1:2181 -s dummy`
* Create some dummy data by creating a file at `~/dummy.dat` with the following:
```
{ "field" : "f1", "source.type" : "dummy" }
{ "field" : "f2", "source.type" : "dummy" }
{ "field" : "f3", "source.type" : "dummy" }
{ "field" : "f4", "source.type" : "dummy" }
{ "field" : "f5", "source.type" : "dummy" }
{ "field" : "f6", "source.type" : "dummy" }
{ "field" : "f7", "source.type" : "dummy" }
```
* Send the dummy data in via `cat ~/dummy.dat | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic dummy`
* Ensure the data written to the indices has a field `new_field` that looks like `DUMMY,${field}`

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/json/JSONMapParser.java
```diff
@@ -99,6 +98,7 @@ public void configure(Map<String, Object> config) {
     String strategyStr = (String) config.getOrDefault(MAP_STRATEGY_CONFIG, MapStrategy.DROP.name());
     mapStrategy = MapStrategy.valueOf(strategyStr);
     if (config.containsKey(JSONP_QUERY)) {
```
ottobackwards: Why was this necessary?

cestella:
This PR provides an expression cache for stellar when using the stellar field transformations in parsers.  Specifically, we found that without caching we were substantially stymied when performance tuning enrichments.  Thus, other places where stellar is in the critical path might derive benefit from similar caching in Enrichments.
A couple of points about the implementation:
* We use the same underlying caching strategy as in enrichments (caffeine-based least frequently used cache).
* This caching is implemented in stellar core as an alternative to the StellarProcessor.  This way it is reusable elsewhere that stellar is used eventually.
* This caching is slightly different than the caching in enrichments in that:
  * Enrichment caching is done regardless of adapter (stellar enrichments uses the same cache as the hbase enrichments)
  * Enrichment caching for stellar is at the group level (e.g. the key is the group of stellar expressions and the input variables used in the group adn the value is the set of results for that group given the input).  This implementation is at the individual expression level.

ottobackwards:
Does this assume, that if you have multiple transformations that all of them should be cached?
Does it assume that every stellar function returns the same value with the same input?

cestella:
Yes, the caching semantics are the same as the enrichment topology; namely that we cache all stellar expressions.  

You can, however, turn caching off in the parsers in the case that you have a nondeterministic stellar expression.  I'd like to be able to augment stellar function annotations to note expressions which are non-deterministic to avoid caching in those cases.

Do you think we should turn it off by default?

ottobackwards:
I think that the assumptions that it makes about the ability to cache the output, and apply to every transformation or every function may be a bit optimistic.

I would have it off by default but put it in the tuning guide, with something written up about the circumstance where it will work and the circumstances where it will not.

A setting the marks the return as cacheable  in the configuration would be the better option, on a per transformation basis.

cestella:
@ottobackwards Sounds good; I'll do that.

cestella:
Ok, turned off by default and added some more detail to the performance tuning guide.

metron-platform/Performance-tuning-guide.md
```diff
@@ -43,6 +43,16 @@ parallelism will leave you with idle consumers since Kafka limits the max number
 important because Kafka has certain ordering guarantees for message delivery per partition that would not be possible if more than
 one consumer in a given consumer group were able to read from that partition.
```
ottobackwards: s/Parser/Sensor/ 
?

metron-platform/Performance-tuning-guide.md
```diff
@@ -43,6 +43,16 @@ parallelism will leave you with idle consumers since Kafka limits the max number
...
+## Parser Tuning Suggestions
+
+If you are using stellar field transformations in your parsers, by default, stellar expressions
+are not cached.  Turning on caching via setting the `cacheConfig` [property](metron-parsers#parser_configuration)
+in your parser configuration can have performance impact if your stellar expressions are
+complex (e.g. `ENRICHMENT_GET` calls or other high latency calls).  The tradeoff, though, is
+that non-deterministic stellar expressions will yield cached results which may be wrong,
+for the period of time in which the data exists in the cache (the max time in the cache is
+configurable).
+
```
ottobackwards: I think something like this is more straightforward:

"Sensors that use stellar field transformations by see a performance boost by turning on caching via setting the `cacheConfig` [property](metron-parsers#parser_configuration).
This is beneficial if your transformations:
- Are complex (e.g. `ENRICHEMENT_GET` calls or other high latency calls)
- All Yield the same results for the same inputs ( caching is either off or applied to *all* transformations)

If *any* of your transformations are non-deterministic, caching should not be used as it will result in the likelihood of incorrect results being returned."

cestella: Thanks for the feedback.  I've accepted your rewordings.

metron-platform/Performance-tuning-guide.md
cestella: Sure

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/CachingStellarProcessor.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * The Caching Stellar Processor is a stellar processor that optionally fronts stellar with an expression-by-expression
+ * LFU cache.
+ */
+public class CachingStellarProcessor extends StellarProcessor {
```
nickwallen: Where are the unit tests for the `CachingStellarProcessor`?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/Context.java
```diff
@@ -36,6 +36,7 @@
     , STELLAR_CONFIG
     , CONSOLE
     , SHELL_VARIABLES
+    , CACHE
```
nickwallen: If you're so inclined, it would be useful to add some javadoc for `CACHE`.  I know we haven't done it on the others, but we should.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/CachingStellarProcessor.java
nickwallen: I'd like to see tests that cover both a "with cache" and a "no cache" use case.

metron-platform/metron-parsers/README.md
```diff
@@ -174,6 +174,19 @@ then it is assumed to be a regex and will match any topic matching the pattern (
 * `spoutConfig` : A map representing a custom spout config (this is a map). This can be overridden on the command line.
 * `securityProtocol` : The security protocol to use for reading from kafka (this is a string).  This can be overridden on the command line and also specified in the spout config via the `security.protocol` key.  If both are specified, then they are merged and the CLI will take precedence.
 * `stormConfig` : The storm config to use (this is a map).  This can be overridden on the command line.  If both are specified, they are merged with CLI properties taking precedence.
+* `cacheConfig` : Cache config for stellar field transformations.   This configures a least frequently used cache.  This is a map with the following keys.  If unconfigured, then no cache will be used.
```
nickwallen: Is "unconfigured" == "default" behavior? I'd like to see mentioned specifically that the default behavior is no cache (at least that's what I am assuming).

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/Context.java
nickwallen: I can remember from experience trying to track down what some of these are used for in the past.  A little doc here would have been very helpful.

cestella:
Ok, @nickwallen I think I reacted to your comments.  I'm going to spin this up one more time since it's churned a bit since I originally submitted.  Let me know if I missed anything.

cestella:
I reran this in full-dev from scratch and it checks out.  Let me know if there are any more comments @nickwallen and @ottobackwards

nickwallen:
+1 Thanks.  Looks good.

PR #995, opened 2018-04-17
METRON-1526: Location field types cause DocValuesField appear more than once error

merrimanr:
## Contributor Comments
Solr automatically expands location type fields internally when a document is indexed.  This causes a problem in our DAO layer because we don't do partial updates (we reindex the whole document) and these expanded fields are included in the updated document.  This PR strips these expanded fields out in the SolrUpdateDao before updating.

I also added a unit test to cover this case.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
+++ b/metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
@@ -46,6 +46,7 @@ public abstract class UpdateIntegrationTest {
   private static String index;
   private static MockHTable table;
   private static IndexDao hbaseDao;
+  private static AccessConfig accessConfig;
 
   protected static MultiIndexDao dao;
   protected static InMemoryComponent indexComponent;
@@ -62,7 +63,7 @@ public abstract class UpdateIntegrationTest {
       table = (MockHTable)tableProvider.getTable(config, TABLE_NAME);
 
       hbaseDao = new HBaseDao();
-      AccessConfig accessConfig = new AccessConfig();
+      accessConfig = new AccessConfig();
       accessConfig.setTableProvider(tableProvider);
       Map<String, Object> globalConfig = createGlobalConfig();
       globalConfig.put(HBaseDao.HBASE_TABLE, TABLE_NAME);
@@ -179,6 +180,28 @@ public abstract class UpdateIntegrationTest {
         Assert.assertNotEquals("Elasticsearch is not updated!", cnt, 0);
       }
     }
+  }
+
+  @Test
+  public void suppress_expanded_fields() throws Exception {
+    dao = new MultiIndexDao(createDao());
+    dao.init(accessConfig);
+
+    Map<String, Object> fields = new HashMap<>();
+    fields.put("guid", "bro_1");
+    fields.put("source.type", SENSOR_NAME);
+    fields.put("ip_src_port", 8010);
+    fields.put("long_field", 10000);
+    fields.put("latitude", 48.5839);
+    fields.put("score", 10.0);
+    fields.put("is_alert", true);
+    fields.put("field.location_point", "48.5839,7.7455");
+
+    Document document = new Document(fields, "bro_1", SENSOR_NAME, 0L);
+    dao.update(document, Optional.of(SENSOR_NAME));
+
+    Document indexedDocument = dao.getLatest("bro_1", SENSOR_NAME);
+    Assert.assertEquals(8, indexedDocument.getDocument().size());
   }
 
   @AfterClass
--- a/metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
+++ b/metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
@@ -36,6 +36,8 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
+import static org.junit.Assert.assertEquals;
+
 public abstract class UpdateIntegrationTest {
 
   private static final int MAX_RETRIES = 10;
@@ -201,6 +203,8 @@ public abstract class UpdateIntegrationTest {
     dao.update(document, Optional.of(SENSOR_NAME));
 
     Document indexedDocument = dao.getLatest("bro_1", SENSOR_NAME);
+
+    // assert no extra expanded fields are included
     Assert.assertEquals(8, indexedDocument.getDocument().size());
   }
 
--- a/metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
+++ b/metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
@@ -49,7 +49,7 @@ public abstract class UpdateIntegrationTest {
   private static String index;
   private static MockHTable table;
   private static IndexDao hbaseDao;
-  protected static AccessConfig accessConfig;
+  private static AccessConfig accessConfig;
 
   protected static MultiIndexDao dao;
   protected static InMemoryComponent indexComponent;
@@ -75,10 +75,14 @@ public abstract class UpdateIntegrationTest {
     }
   }
 
+  protected AccessConfig getAccessConfig() {
+    return accessConfig;
+  }
+
   @Test
   public void test() throws Exception {
     dao = new MultiIndexDao(hbaseDao, createDao());
-    dao.init(accessConfig);
+    dao.init(getAccessConfig());
 
     List<Map<String, Object>> inputData = new ArrayList<>();
     for(int i = 0; i < 10;++i) {
--- a/metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
+++ b/metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
@@ -31,6 +31,7 @@ import org.apache.metron.hbase.mock.MockHTable;
 import org.apache.metron.indexing.dao.update.Document;
 import org.apache.metron.indexing.dao.update.ReplaceRequest;
 import org.apache.metron.integration.InMemoryComponent;
+import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.Before;
@@ -48,7 +49,7 @@ public abstract class UpdateIntegrationTest {
   private static String index;
   private static MockHTable table;
   private static IndexDao hbaseDao;
-  private static AccessConfig accessConfig;
+  protected static AccessConfig accessConfig;
 
   protected static MultiIndexDao dao;
   protected static InMemoryComponent indexComponent;
@@ -71,14 +72,14 @@ public abstract class UpdateIntegrationTest {
       globalConfig.put(HBaseDao.HBASE_TABLE, TABLE_NAME);
       globalConfig.put(HBaseDao.HBASE_CF, CF);
       accessConfig.setGlobalConfigSupplier(() -> globalConfig);
-
-      dao = new MultiIndexDao(hbaseDao, createDao());
-      dao.init(accessConfig);
     }
   }
 
   @Test
   public void test() throws Exception {
+    dao = new MultiIndexDao(hbaseDao, createDao());
+    dao.init(accessConfig);
+
     List<Map<String, Object>> inputData = new ArrayList<>();
     for(int i = 0; i < 10;++i) {
       final String name = "message" + i;
@@ -184,28 +185,9 @@ public abstract class UpdateIntegrationTest {
     }
   }
 
-  @Test
-  public void suppress_expanded_fields() throws Exception {
-    dao = new MultiIndexDao(createDao());
-    dao.init(accessConfig);
-
-    Map<String, Object> fields = new HashMap<>();
-    fields.put("guid", "bro_1");
-    fields.put("source.type", SENSOR_NAME);
-    fields.put("ip_src_port", 8010);
-    fields.put("long_field", 10000);
-    fields.put("latitude", 48.5839);
-    fields.put("score", 10.0);
-    fields.put("is_alert", true);
-    fields.put("field.location_point", "48.5839,7.7455");
-
-    Document document = new Document(fields, "bro_1", SENSOR_NAME, 0L);
-    dao.update(document, Optional.of(SENSOR_NAME));
-
-    Document indexedDocument = dao.getLatest("bro_1", SENSOR_NAME);
-
-    // assert no extra expanded fields are included
-    Assert.assertEquals(8, indexedDocument.getDocument().size());
+  @After
+  public void reset() throws Exception {
+    indexComponent.reset();
   }
 
   @AfterClass
--- a/metron-platform/metron-solr/README.md
+++ b/metron-platform/metron-solr/README.md
@@ -97,8 +97,15 @@ you should refer to the Solr documentation [https://lucene.apache.org/solr/guide
 and [here](https://lucene.apache.org/solr/guide/6_6/documents-fields-and-schema-design.html) if you'd like to know more about schemas in Solr.
 
 Indexing LatLon and PointType field types stores data in internal fields that should not be returned in search results.  For these fields a dynamic field type matching the suffix needs to be added to store the data points.
-For example, a LatLong type should be defined as:
+For example, a LatLongType field should be defined as:
 ```
 <dynamicField name="*.location_point" type="location" multiValued="false" docValues="false"/>
 <dynamicField name="*_coordinate" type="pdouble" indexed="true" stored="false" docValues="false"/>
+<fieldType name="location" class="solr.LatLonType" subFieldSuffix="_coordinate"/>
+```
+A PointType field should be defined as:
+```
+<dynamicField name="*._point" type="location" multiValued="false" docValues="false"/>
+<dynamicField name="*_point" type="pdouble" indexed="true" stored="false" docValues="false"/>
+<fieldType name="point" class="solr.PointType" subFieldSuffix="_point"/>
 ```
--- a/metron-platform/metron-solr/README.md
+++ b/metron-platform/metron-solr/README.md
@@ -96,6 +96,7 @@ Ambari will eventually install these, but at the moment it's manual and
 you should refer to the Solr documentation [https://lucene.apache.org/solr/guide/6_6](here) in general
 and [here](https://lucene.apache.org/solr/guide/6_6/documents-fields-and-schema-design.html) if you'd like to know more about schemas in Solr.
 
+In Metron's Solr DAO implementation, document updates involve reading a document, applying the update and reindexing the whole document.  
 Indexing LatLon and PointType field types stores data in internal fields that should not be returned in search results.  For these fields a dynamic field type matching the suffix needs to be added to store the data points.
 For example, a LatLongType field should be defined as:
 ```
@@ -105,7 +106,8 @@ For example, a LatLongType field should be defined as:
 ```
 A PointType field should be defined as:
 ```
-<dynamicField name="*._point" type="location" multiValued="false" docValues="false"/>
+<dynamicField name="*.point" type="point" multiValued="false" docValues="false"/>
 <dynamicField name="*_point" type="pdouble" indexed="true" stored="false" docValues="false"/>
 <fieldType name="point" class="solr.PointType" subFieldSuffix="_point"/>
 ```
+If any copy fields are defined, stored and docValues should be set to false.
--- a/metron-platform/metron-solr/README.md
+++ b/metron-platform/metron-solr/README.md
@@ -96,9 +96,9 @@ Ambari will eventually install these, but at the moment it's manual and
 you should refer to the Solr documentation [https://lucene.apache.org/solr/guide/6_6](here) in general
 and [here](https://lucene.apache.org/solr/guide/6_6/documents-fields-and-schema-design.html) if you'd like to know more about schemas in Solr.
 
-In Metron's Solr DAO implementation, document updates involve reading a document, applying the update and reindexing the whole document.  
-Indexing LatLon and PointType field types stores data in internal fields that should not be returned in search results.  For these fields a dynamic field type matching the suffix needs to be added to store the data points.
-For example, a LatLongType field should be defined as:
+In Metron's Solr DAO implementation, document updates involve reading a document, applying the update and replacing the original by reindexing the whole document.  
+Indexing LatLonType and PointType field types stores data in internal fields that should not be returned in search results.  For these fields a dynamic field type matching the suffix needs to be added to store the data points.
+Solr 6+ comes with a new LatLonPointSpatialField field type that should be used instead of LatLonType if possible.  Otherwise, a LatLongType field should be defined as:
 ```
 <dynamicField name="*.location_point" type="location" multiValued="false" docValues="false"/>
 <dynamicField name="*_coordinate" type="pdouble" indexed="true" stored="false" docValues="false"/>
--- a/metron-platform/metron-solr/README.md
+++ b/metron-platform/metron-solr/README.md
@@ -95,3 +95,10 @@ As of now, we have mapped out the Schemas in `src/main/config/schema`.
 Ambari will eventually install these, but at the moment it's manual and
 you should refer to the Solr documentation [https://lucene.apache.org/solr/guide/6_6](here) in general
 and [here](https://lucene.apache.org/solr/guide/6_6/documents-fields-and-schema-design.html) if you'd like to know more about schemas in Solr.
+
+Indexing LatLon and PointType field types stores data in internal fields that should not be returned in search results.  For these fields a dynamic field type matching the suffix needs to be added to store the data points.
+For example, a LatLong type should be defined as:
+```
+<dynamicField name="*.location_point" type="location" multiValued="false" docValues="false"/>
+<dynamicField name="*_coordinate" type="pdouble" indexed="true" stored="false" docValues="false"/>
+```
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrUpdateDao.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrUpdateDao.java
@@ -36,6 +36,7 @@ import org.slf4j.LoggerFactory;
 public class SolrUpdateDao implements UpdateDao {
 
   private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private final static String LOCATION_SUB_FIELD_SUFFIX = "_coordinate";
 
   private transient SolrClient client;
 
@@ -92,9 +93,17 @@ public class SolrUpdateDao implements UpdateDao {
     }
   }
 
-  private SolrInputDocument toSolrInputDocument(Document document) {
+  protected SolrInputDocument toSolrInputDocument(Document document) {
     SolrInputDocument solrInputDocument = new SolrInputDocument();
-    document.getDocument().forEach(solrInputDocument::addField);
+    document.getDocument().forEach((field, value) -> {
+      if (!isLocationSubField(field)) {
+        solrInputDocument.addField(field, value);
+      }
+    });
     return solrInputDocument;
+  }
+
+  private boolean isLocationSubField(String field) {
+    return field.endsWith(LOCATION_SUB_FIELD_SUFFIX);
   }
 }
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrUpdateDao.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrUpdateDao.java
@@ -36,7 +36,6 @@ import org.slf4j.LoggerFactory;
 public class SolrUpdateDao implements UpdateDao {
 
   private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  private final static String LOCATION_SUB_FIELD_SUFFIX = "_coordinate";
 
   private transient SolrClient client;
 
@@ -93,17 +92,9 @@ public class SolrUpdateDao implements UpdateDao {
     }
   }
 
-  protected SolrInputDocument toSolrInputDocument(Document document) {
+  private SolrInputDocument toSolrInputDocument(Document document) {
     SolrInputDocument solrInputDocument = new SolrInputDocument();
-    document.getDocument().forEach((field, value) -> {
-      if (!isLocationSubField(field)) {
-        solrInputDocument.addField(field, value);
-      }
-    });
+    document.getDocument().forEach(solrInputDocument::addField);
     return solrInputDocument;
-  }
-
-  private boolean isLocationSubField(String field) {
-    return field.endsWith(LOCATION_SUB_FIELD_SUFFIX);
   }
 }
--- a/metron-platform/metron-solr/src/test/java/org/apache/metron/solr/integration/SolrUpdateIntegrationTest.java
+++ b/metron-platform/metron-solr/src/test/java/org/apache/metron/solr/integration/SolrUpdateIntegrationTest.java
@@ -20,15 +20,21 @@ package org.apache.metron.solr.integration;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+
 import org.apache.metron.indexing.dao.IndexDao;
+import org.apache.metron.indexing.dao.MultiIndexDao;
 import org.apache.metron.indexing.dao.UpdateIntegrationTest;
+import org.apache.metron.indexing.dao.update.Document;
 import org.apache.metron.integration.InMemoryComponent;
 import org.apache.metron.solr.dao.SolrDao;
 import org.apache.metron.solr.integration.components.SolrComponent;
+import org.junit.Assert;
+import org.junit.Test;
 
 public class SolrUpdateIntegrationTest extends UpdateIntegrationTest {
 
-  SolrComponent solrComponent;
+  protected static SolrComponent solrComponent;
 
   @Override
   protected String getIndexName() {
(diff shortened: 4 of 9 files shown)


metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrUpdateDao.java
```diff
@@ -92,9 +93,17 @@ public void batchUpdate(Map<Document, Optional<String>> updates) throws IOExcept
...
+    document.getDocument().forEach((field, value) -> {
+      if (!isLocationSubField(field)) {
+        solrInputDocument.addField(field, value);
+      }
+    });
     return solrInputDocument;
   }
+
+  private boolean isLocationSubField(String field) {
+    return field.endsWith(LOCATION_SUB_FIELD_SUFFIX);
```
cestella: Is there really no better way to determine if a field is a location field than this?  Can we not get the type somewhere?

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrUpdateDao.java
```diff
@@ -92,9 +93,17 @@ public void batchUpdate(Map<Document, Optional<String>> updates) throws IOExcept
     }
   }
 
-  private SolrInputDocument toSolrInputDocument(Document document) {
+  protected SolrInputDocument toSolrInputDocument(Document document) {
     SolrInputDocument solrInputDocument = new SolrInputDocument();
-    document.getDocument().forEach(solrInputDocument::addField);
+    document.getDocument().forEach((field, value) -> {
+      if (!isLocationSubField(field)) {
```
cestella: Are we certain that location fields are the only type that have this issue?

merrimanr: I don't see any other fields like this in the sensors we have but that doesn't mean there are not other field types like this.  I wasn't able to find any documentation around which fields Solr expands either.

cestella:
It seems as though we should be checking the type for `isPolyField()` (see [here](http://lucene.apache.org/solr/6_3_0/solr-core/org/apache/solr/schema/PointType.html#isPolyField--).  

The problem, as I see it, is that we don't ahve access to the type directly.  So, in order to have the type, we may need a type cache.  I'd suggest a LFU caffeine cache that will retrieve the schema for the sensor from solr.  Thoughts?

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrUpdateDao.java
merrimanr: I think there is a better way.  Within our schemas there is a "location" field type that is configured as such:
```
<fieldType name="location" class="solr.LatLonType" subFieldSuffix="_coordinate"/>
```
The constant in this class matches the "subFieldSuffix" attribute.  We could get column metadata, trace fields back to this type, and remove them based on the value of the "subFieldSuffix" attribute instead of hardcoding it.

ottobackwards:
Is it possible for parser writers to create other fields that have this issue?

mmiklavc:
> This causes a problem in our DAO layer because we don't do partial updates (we reindex the whole document) and these expanded fields are included in the updated document.

@merrimanr Can you elaborate on this a bit? I'm not sure I follow the full scope of the problem. Per @ottobackwards comment, would new fields introduced by parsers also cause issues? What are the parameters around when a user would "step in it?" What prophylaxis do we need to ensure, or at least reduce the risk, that this can happen for other fields?

This makes me think of another topic that's been discussed recently about message envelopes (e.g. syslog and parser chaining) that potentially wrap multiple other log types that need to be parsed. If the wrapped messages all routed to the same index, is that a problem?

cestella:
@ottobackwards At the very least, if this is indeed a problem with all [PointType](http://lucene.apache.org/solr/6_3_0/solr-core/org/apache/solr/schema/PointType.html) fields and not just lat/long fields, then this may impact a common data science workflow:
1. Generate a model which outputs a vector representation for some input
2. Enrich data with that model (e.g. generate a vector per data point)
3. Index the data as a Point in solr
4. Find similar data by doing nearest neighbor searches

ottobackwards:
I am thinking of the metron as a framework POV here.  Right now, someone can create a parser, an enrichment a (in this PR's world) solr/es schema for their input.

This PR makes assumptions about a single use case, but the platform allows the capability for this to happen in many other ways that are beyond the control or knowledge of the 'core' platform through the platform's extensibility.

That makes any remediation of this issue a 'one' off it assumes knowledge of the data, which the platform cannot have beyond the parsers that ship.  And even then that knowledge ( based on the known configuration ) is not reliable because they are examples and can be modified.

Does that make sense or am I missing something?

merrimanr:
There are a finite number of field types that ship with Solr so the fact that someone can create an infinite number of fields really isn't an issue.  Any field that is created and stored in Solr has to be one of the finite field types.  I agree that we can't assume knowledge of data hence handling all field types that could be potentially problematic.

Are there other solutions to this problem that we're not thinking of?

ottobackwards:
@cestella I think that must be the case, given that we cannot assume the fields present beyond a certain set.

cestella:
I'm sure you did this, @merrimanr, but I googled your error and ran upon [this](https://stackoverflow.com/questions/44375034/solr-docvaluesfield-appears-more-than-once-in-this-document-solr-6-5).  Are we certain that we're not in this situation where a simple migration to `solr.LatLonPointSpatialField` isn't the solution?

merrimanr:
I did see that but that field type doesn't in Solr 5.5.

cestella:
Yes, but according to that thread, this became a problem in 6.5.  In 5.5, presumably, this worked.  Should we, perhaps, have different schema for 5.x vs 6.x?

cestella:
One thing that I would question, looking at that answer, is whether it's just a LatLonType problem or if it really is a problem with all poly fields.  I would suggest tryign to replicate with a currency type, which is also a polyfield.  If the same error does not happen, then we may be safe with just handling this for point types.

cestella:
Another possibility, which these guys [did](https://github.com/ukwa/webarchive-discovery/issues/105) is disabling docValues for the spatial field.  I'm not clear on the consequences of that.

justinleet:
An interesting consequence of not using docValues is https://lucene.apache.org/solr/guide/6_6/docvalues.html#DocValues-RetrievingDocValuesDuringSearch

>When useDocValuesAsStored="false", non-stored DocValues fields can still be explicitly requested by name in the fl param, but will not match glob patterns ("*").

This sounds more like what we want, right?  From my selfish metaalerts perspective.  Why should be be returning the subfields of polyfields to the user?  My concerns are more around if this causes perf problems around sorting those fields.

merrimanr:
After doing more research on this, managing fields included when updating a document looks to be more complex than we thought.  The `isPolyField()` method is part of the internal API and we don't have access to it with SolrJ.  Also, I found another field that is stored in multiple fields internally (date).  I suspect we will find more special cases like this and if there are cases we don't find, it could break things later if a user adds a field with that type.

Should we investigate adding partial update support to the SolrDao?  There is a `patch` method in the IndexDao so the interfaces would not need to change.  The downside is this will introduce some constraints on schema design:  https://lucene.apache.org/solr/guide/6_6/updating-parts-of-documents.html#UpdatingPartsofDocuments-FieldStorage.

cestella:
Before going down the path of partial document update:
* Can you confirm that this isn't an issue when you set docValues to false, like in my previous comment?
* Are date fields polyfields?
* Is this a problem in 5.x or just 6.x?
  * Is there a workaround for dates like there is for latlon in 6.x (e.g. another type)?  If so, wouldn't differentiating schema per version solve the issue?

If you did want to go with partial updates:
* how do you propose to pull the JSON Patch apart into just the selected fields?
* Since we will possibly expose these generated fields (perhaps not for latlon, but possibly for other polytypes), aren't we in the same spot of having a way to update both a generated field and a derived field if the users can update derived fields?
  * If the answer to that is to whitelist updates on derived fields, aren't we back in the same spot as we are currently in?

justinleet:
It might also be possible to do this the other way, rather than getting a list of all fields then figuring out which are polyFields, we might be able to get just all valid fields.  Substantially more validation would have to be done on what I've tested here though to make sure it's valid and I didn't skip over something important.

I threw this together in one of the metaalerts tests:
```
    MetronSolrClient client = solr.getSolrClient();
    client.setDefaultCollection("test");
    LukeRequest lukeRequest = new LukeRequest();
    lukeRequest.setShowSchema(true);
    LukeResponse result = lukeRequest
        .process(client);
    Map<String, FieldInfo> fieldInfo = result.getFieldInfo();
    for(Entry<String, FieldInfo> entry : fieldInfo.entrySet()) {
      System.out.println("KEY: " + entry.getKey());
      System.out.println("VALUE NAME: " + entry.getValue().getName());
      System.out.println("FLAGS: " + entry.getValue().getFlags());
      System.out.println();
    }
```

Output is
>KEY: new-field
>VALUE NAME: new-field
>FLAGS: [INDEXED, STORED, OMIT_NORMS, OMIT_TF, SORT_MISSING_LAST]
>
>KEY: score
>VALUE NAME: score
>FLAGS: [INDEXED, STORED, DOC_VALUES, OMIT_NORMS, OMIT_TF]
>
>KEY: _version_
>VALUE NAME: _version_
>FLAGS: [DOC_VALUES]
>
>KEY: threat:triage:score
>VALUE NAME: threat:triage:score
>FLAGS: [INDEXED, STORED, DOC_VALUES, OMIT_NORMS, OMIT_TF]
>
>KEY: _root_
>VALUE NAME: _root_
>FLAGS: [INDEXED, OMIT_NORMS, OMIT_TF, SORT_MISSING_LAST]
>
>KEY: location_point
>VALUE NAME: location_point
>FLAGS: [INDEXED, STORED, OMIT_TF]
>
>KEY: name
>VALUE NAME: name
>FLAGS: [INDEXED, STORED, OMIT_NORMS, OMIT_TF, SORT_MISSING_LAST]
>
>KEY: guid
>VALUE NAME: guid
>FLAGS: [INDEXED, STORED, OMIT_NORMS, OMIT_TF, SORT_MISSING_LAST]
>
>KEY: metaalerts
>VALUE NAME: metaalerts
>FLAGS: [INDEXED, STORED, MULTI_VALUED, OMIT_NORMS, OMIT_TF, SORT_MISSING_LAST]
>
>KEY: source.type
>VALUE NAME: source.type
>FLAGS: [INDEXED, STORED, OMIT_NORMS, OMIT_TF, SORT_MISSING_LAST]
>
>KEY: timestamp
>VALUE NAME: timestamp
>FLAGS: [INDEXED, STORED, DOC_VALUES, OMIT_NORMS, OMIT_TF]

Notice there's no "_coordinate" fields. Having "location_point" defined in my schema with:
```
  <field name="location_point" type="location" indexed="true" stored="true"/>
  <dynamicField name="*_coordinate" type="pdouble" indexed="true" stored="false"/>
  ...
  <fieldType name="location" class="solr.LatLonType" subFieldSuffix="_coordinate"/>
```
Resulted in the "_coordinate" fields being present in this query:
```
    client = solr.getSolrClient();
    client.setDefaultCollection("test");
    SolrQuery query = new SolrQuery("*:*");
    query.set("fl", "*");
    QueryResponse response = client.query(query);
    System.out.println(response.getResults());
```
Like so (with all the extra stripped out so it's easier to read here):
```
{
  ...,
  docs=[
    SolrDocument{threat:triage:score=0.0, location_point=10,20, guid=message_0, metaalerts=[meta_active], source.type=test, timestamp=1524157278555, ip_src_port=8010, location_point_1_coordinate=20.0, _version_=1598194742551642112, location_point_0_coordinate=10.0, ip_src_addr=192.168.1.1},
    ...
  ]}
```

There is a catch: the schema has to be updated and I'm not sure it's compatible with both 5.x and 6.x
`<requestHandler name="/admin/luke" class="org.apache.solr.handler.admin.LukeRequestHandler" />`

cestella:
> I'm not following. A patch would only include the field to be added, changed, or removed.

JSONPatch, the input that we accept for the `patch` command, offers more than add, change and remove operations.  There is also move and add supports adding into an array.  My point is that the format is more complex than just passing in a list of fields to add, change or remove.

> If for some reason a user is trying to directly update a generated field that is incorrect and a user error.
 Consider this from a user perspective for a moment.  How would a user know if a field is generated or not? If we could determine which fields were generated to ensure the users aren't given the ability to change them, then we could whitelist them generally and solve this problem.  

I'm not opposed to partial updates, but I don't see that as a solution to the problem from this PR.  It feels like kicking the can down the road and making it the user's problem, which I don't agree with.

cestella:
@justinleet Looks like [LukeRequestHandler](http://lucene.apache.org/solr/5_4_1/solr-core/org/apache/solr/handler/admin/LukeRequestHandler.html) has been around since Solr 1.2.

justinleet:
The schema declaration itself might have changed between 5.x and 6.x.  I saw something about it, but didn't double check since I hadn't looked in to the request at all yet. The request itself should be fine though

mmiklavc:
@justinleet Your example suggests that by using a LukeRequest and calling getFieldInfo() on the result, we should be able to naturally obtain the filtered list of fields that can be used as a whitelist when constructing the index update? And it's available since Solr 1.2? This sounds pretty clean to me and it seems to solve the problem in a general way without exposing Solr implementation details to the end user (whether by API or documentation).

justinleet:
I agree, I'm not seeing how partial updates solves the general problem of "We need to avoid updates against derived fields".  End users aren't going to necessarily know what a generated field is.  Say a data scientist writes a machine learning model, and an analyst wants to update some output of the model.  They may not know, or even be able to easily discover, what fields are poly types and what aren't. They'll just get an error.  I'm not even convinced it'll be an easily understandable error to anyone not familiar with Solr.

It seems like there one real, solution to this problem that doesn't compromise usability: Don't pass sub fields back in queries.

This can be something in the schema itself (e.g. does docValues change work).  It could also be something determined while running (e.g. LukeRequest if it turns out to be viable or some other request / query).

merrimanr:
@justinleet will this return dynamic fields?  Why did you change the schema?

justinleet:
@mmiklavc The main worry is I'm not honestly sure if that's exactly how it works.  I didn't dig into the caveats or any issues that we might hit with it.  I also don't know what constraints get placed on general polyfields in terms of whether they get returned by that or not.  Assuming it works, it would be fairly clean.   But that'll require a bit of research.

justinleet:
@merrimanr Not sure if it returns dynamic fields.  I didn't see them, but there might be a different way to grab them.  Again, it's a really cursory idea that needs more research if other things fall through (e.g. docValues)

I changed the schema in two ways
1. To add the ability to make the call
2. To add the dynamic "*_coordinate" field.  This is (or at least was at some point) required.  Check out https://wiki.apache.org/solr/SpatialSearch#QuickStart.  Assuming this is true, we likely need to add this to all our schemas anyway.

justinleet:
To add onto my last comment, the subfields probably get created right now, but get caught by the "*" field that's type "ignored", which seems wrong.

mmiklavc:
@justinleet 

> They may not know, or even be able to easily discover, what fields are poly types and what aren't. 

@cestella, @merrimanr 

>>Are date fields polyfields?

>I don't know. At this point I am not sure how to determine if a field is a polyfield or not. The isPolyField method is internal and I don't have access to it in Solrj.

I think the proof of @justinleet 's point is made abundantly clear by our very own discussion on the subject.

justinleet:
@merrimanr Re: dynamic fields.  Looks like you can get them in 6.6, but possibly not 5.5

https://lucene.apache.org/solr/6_6_0/solr-solrj/org/apache/solr/client/solrj/response/LukeResponse.html#getDynamicFieldInfo--

http://lucene.apache.org/solr/5_5_2/solr-solrj/org/apache/solr/client/solrj/response/LukeResponse.html#getFieldInfo()

Not sure what it returns, whether it's the actual fields or just what the schema defines though.

merrimanr:
@justinleet I tested that and it looks like it's just what the schema defines.  The only way we could use those is if we did a regex max on the actual fields in a document.

cestella:
Can we get this problem exhibited in the integration tests?  It would help us try potential solutions out easily.

merrimanr:
If you add the "-h node1" option to the start command in install_solr.sh you can develop locally against Solr in full dev.  The line ends up looking like:
```
su $SOLR_USER -c "bin/solr -e cloud -noprompt -h node1"
```

cestella:
@merrimanr Ok, that's fine, we can do it on full-dev.  Can you please submit a step-by-step manual testing script to exhibit this issue?  I was asking for an integration test because then it'll be run every build and easy to adjust the code, but this will work too.

merrimanr:
I've been changing/uploading schemas in full dev with the Solr CLI and using the Solr Admin console to test.  I can write up with a list of commands that mimic what I'm doing in the console or create a simple integration test.  Which one do you prefer?

merrimanr:
Just added an integration test that demonstrates this behavior.  I had to switch from the test schema to the schema that is actually deployed and remove the HBaseDao from this particular test to get it to work.  We will want to expand this test to all types once we agree on a good solution to this problem.  It also might be a good idea to get rid of these test schemas and just use the real ones at some point.

merrimanr:
I did some additional testing and here are my notes.  I've attempted to compile a list of all the proposed solutions and their outcomes.

> checking the type for isPolyField()

This method is not available in the Solrj client library and would require adding a dependency on solr-core which I don't think we want to do.  Compiling a list of field type to polyfield mappings would need to be something we do outside of runtime and maintain across Solr upgrades.  I did this and found that out of the field types include with Solr 6.6, these are polyfields:
- CurrencyField
- DateRangeField
- LatLonPointSpatialField
- LatLonType
- PointType
- SpatialRecursivePrefixTreeFieldType

I also added any types not currently defined in our schemas and indexed sample data.  Only LatLonType and PointType resulted in expanded fields.

> migrate to solr.LatLonPointSpatialField and maintain separate schemas

The solr.LatLonPointSpatialField type worked without issue in Solr 6.6 as expected.  I installed a separate 5.5 instance and indexed the same data.  There is no error thrown but the data ends up looking like this in the index:
```
{
  "enrichments.geo.ip_dst_addr.location_point_0_coordinate": [
    "34.0438",
    "34.0438"
  ],
  "enrichments.geo.ip_dst_addr.location_point_1_coordinate": [
    "-118.2512",
    "-118.2512"
  ],
  "enrichments.geo.ip_dst_addr.location_point": "34.0438,-118.2512"
}
```
That does not look correct to me.  If this were to be acceptable what would our approach be for testing and maintaining separate schemas for different versions?  Our code does not compile with version 5.5.

> disabling docValues for the spatial field

The docValues setting is already disabled in our schemas.

> using LukeRequest to get field type information

This does not handle the case of dynamic fields so we will need to come up with a strategy to map dynamic fields to the field types we care about.  I think we may need to do this regardless of which approach we decide on.  It will be necessary to match dynamic field types to fields that are actually present in a document.

Am I missing any solutions that have been proposed?  Let me know I will investigate.

I also looked into the date field issue I noticed.  Turns out I was wrong and this was actually a result of a copyField being defined.  This is another case we will need to handle.  If a field is a copyField you cannot reindex that field or you will get an error.

merrimanr:
Another potential solution I thought of is to only return the intersection of the fields defined in the schema with fields that are present in a document.  This could potentially eliminate the need to keep track of polyfields and make things simpler.  We would still need to handle dynamic fields and copyFields but I think we'll need to do that anyways.  Still thinking through edge cases but let me know if we think this might be an option worth exploring.

Given all this information, where do we go from here?

mmiklavc:
```
{
  "enrichments.geo.ip_dst_addr.location_point_0_coordinate": [
    "34.0438",
    "34.0438"
  ],
  "enrichments.geo.ip_dst_addr.location_point_1_coordinate": [
    "-118.2512",
    "-118.2512"
  ],
  "enrichments.geo.ip_dst_addr.location_point": "34.0438,-118.2512"
}
```
> That does not look correct to me.

@merrimanr - can you be more specific? I gather it should not be an array, but simple string value for the derived fields, e.g.
```
{
  "enrichments.geo.ip_dst_addr.location_point_0_coordinate":"34.0438",
  "enrichments.geo.ip_dst_addr.location_point_1_coordinate":"-118.2512",
  "enrichments.geo.ip_dst_addr.location_point": "34.0438,-118.2512"
}
```

What was your input doc when doing this test on 5.5? Did you included the expanded fields as part of the insert? To me your results make it look like Solr treated the values as an additive operation vs overwrite, but it's hard for me to say without more detail.

merrimanr:
Yes it should be simple string values or, in other words, the doc should look the same minus the changes.  I should have stated this in my previous comment but ideally those *_coordinate fields wouldn't come back at all.

The test case is:
- read back a doc that is already indexed
- change or add field (not location type)
- reindex the doc
- read back the doc and verify it is unchanged minus the change in step 2

To answer your question, yes the expanded fields are included in the insert because we are reindexing the whole document.  I agree with you, it looks like Solr treats it as an additive operation.

justinleet:
For the integration test, adding the specific dynamic fields for the underlying LatLon seems to take care of the problem for that type.

e.g. this makes the test you added pass:
```
  <!-- Catch all, if we don't know about it, it gets dropped. -->
  <dynamicField name="*_coordinate" type="pdouble" indexed="true" stored="false" docValues="false"/>
  <dynamicField name="*" type="ignored" multiValued="false" docValues="true"/>
```

It appears docValues is not disabled for the subfields, because they get caught by the "*" dynamic fields, so they end up being `docValues=true`.

My concern would be making sure we document what has to happen for LatLon, and making sure that the other types you laid out can be managed in a similar way (and document appropriately),

merrimanr:
Adding this dynamic field worked for me as well.  I tested in a Solr installation as well as the integration test.  I believe this will solve our issue with the older LatLon type.  The only other field type that resulted in 

How should we document this?  Comments in our schemas?  We don't have a dedicated document for creating Solr schemas and I don't think we should maintain detailed documentation on the topic.  Maybe we add a section to the Solr README with tips for Solr schema creation as it relates to Metron (would include this issue)?

justinleet:
If we're going to the `docValues=false` + documentation route,  we should have the documentation be more generic than just the LatLon and Point (e.g. custom types that may get used in more advanced use cases).

I'm also a little surprised that LatLon and Point are the only two types that result in additional fields, given that there were more poly types listed above. Do we know why Currency and other types don't result in additional fields?  Also, is there anything different under the hood about Points vs LatLon?

General questions for people interested
* Are we okay with `docValues=false` and documenting (I personally am, assuming we can provide relatively general guidance, not just something that only works for a couple types).
* If LatLon and Point are the only two mainline fields that produce subfields, should we also provide an example of suppressing the subfields of Point, even though it's not in use right now?

merrimanr:
Just to clarify, both `docValues` and `stored` needs to be set to false.  If one or the other are true then it fails.

At this point I question whether a field being a polyfield is causing this issue.  The testing I've done does not support that.  Javadocs for FieldType.isPolyField() state:

> A "polyField" is a FieldType that can produce more than one IndexableField instance for a single value

The issue here is with stored fields because that's what is returned in a search query.  Indexed fields are not.

merrimanr:
To give an idea of how I arrived at the conclusion that other fields are not a problem, this is how I tested it.  I will use currency as an example.  First I added this to our bro schema:
```
<dynamicField name="*.c"   type="currency" indexed="true"  stored="true"/>
<fieldType name="currency" class="solr.CurrencyField" precisionStep="8"
             defaultCurrency="USD" currencyConfig="currency.xml" />
```
I also added the currency.xml file from the example configs that ship with Solr.  Then I added that field to the integration test added in this PR (SolrUpdateIntegrationTest.suppress_expanded_fields).  It ends up looking like:
```
@Test
  public void suppress_expanded_fields() throws Exception {
    dao = new MultiIndexDao(createDao());
    dao.init(accessConfig);

    Map<String, Object> fields = new HashMap<>();
    fields.put("guid", "bro_1");
    fields.put("source.type", SENSOR_NAME);
    fields.put("ip_src_port", 8010);
    fields.put("long_field", 10000);
    fields.put("latitude", 48.5839);
    fields.put("score", 10.0);
    fields.put("is_alert", true);
    fields.put("field.location_point", "48.5839,7.7455");
    fields.put("field.c", "1000,EUR");

    Document document = new Document(fields, "bro_1", SENSOR_NAME, 0L);
    dao.update(document, Optional.of(SENSOR_NAME));

    Document indexedDocument = dao.getLatest("bro_1", SENSOR_NAME);

    // assert no extra expanded fields are included
    Assert.assertEquals(9, indexedDocument.getDocument().size());
  }
```
If we get the same number of fields back that we indexed originally, we know no extra fields are being returned.

merrimanr:
A note about the added integration test.  I moved the test to the SolrUpdateIntegrationTest because this issue is specific to Solr.  Adding an additional test exposed a problem in our Solr in memory component where reset wasn't working correctly.  I had to fix that to get all tests to pass.

If verifying this in full dev, the "*_coordinate" fields should no longer be present when searching or looking up documents either through our REST endpoint or the Solr admin console.

merrimanr:
Thanks for the thorough reply @justinleet.  Everything you have said makes sense to me.

In my testing I also found copy fields to be an issue if they were set to stored.  Same problem as LatLon types.  Is it worth adding those to our README in this PR as well?

mmiklavc:
@justinleet - Thanks for that write up - that helps considerably.

Just to round out @merrimanr 's comment about polyfields:

>At this point I question whether a field being a polyfield is causing this issue. The testing I've done does not support that. Javadocs for FieldType.isPolyField() state:

>>A "polyField" is a FieldType that can produce more than one IndexableField instance for a single value

So, from the last few examples discussed it suggests to me that being a polyfield *is* actually a problem, but it's only part of the total reason for the problem. Under the hood they are setting stored=false and docValues=false for currency and some of the other polyfields, whereas they aren't doing this for LatLonType and Point. Does that sound about right? I saw mention of Date also, and some comments about LatLonType being the only problem data type, so it would be good to summarize again what field types are specifically a problem.

One bit I'm not completely clear on after the rounds of discussion is when it's desirable to return the dynamic/subfields generated by virtue of being a polyfield, and further how the doc update code is managing that differently from a normal search. For a standard doc query in context of performing an update, we don't want dynamically generated fields returned because that will bork the re-index. For a user glob query, we *do* want then returned because they may be useful to some users? Is that hovering around accurate?

justinleet:
@mmiklavc That sounds right, yes. For returning dynamic fields (at least on the known problematic fields), I don't think it's pretty much ever advantageous. At most it avoids a split on the LatLon result to pull out just latitude or longitude. I think the newer version of geospatial, LatLonPointSpatialField, is in the list not problematic list, so presumably this realization was made and it's handled similar to others.

As an aside, looking through the javadocs, there's this remark on PointFields
>FieldCache is not supported for PointFields, so sorting, faceting, etc on these fields require the use of docValues="true" in the schema.
Even though we don't use PointField in our schemas, we may still want to make that note in our docs.  Seems like a fairly important caveat.

We could also potentially, based on https://lucene.apache.org/solr/guide/6_6/docvalues.html#DocValues-RetrievingDocValuesDuringSearch, set `useDocValuesAsStored="false"`, which might sidestep this problem entirely.  The catch is I think there's a whole set of considerations regarding performance and querying that I do not have the expertise to assess.

@merrimanr Are there any circumstances where we'd want copy fields?  It seems like we could just direct users to make a copy of the field in Metron and just index the copy fields directly. The main concern would be if we're able to manage globs in there, but even if we don't it doesn't seem like super important functionality to support.  At least at a glance; I may definitely be missing use cases.

merrimanr:
> So, from the last few examples discussed it suggests to me that being a polyfield is actually a problem, but it's only part of the total reason for the problem. Under the hood they are setting stored=false and docValues=false for currency and some of the other polyfields, whereas they aren't doing this for LatLonType and Point. Does that sound about right? I saw mention of Date also, and some comments about LatLonType being the only problem data type, so it would be good to summarize again what field types are specifically a problem.

Yes that sounds right to me.  The issue with Date was actually a result of a copy field being defined with stored set to true.  The only types that returned extra fields in my testing were LatLonType and PointType.

> One bit I'm not completely clear on after the rounds of discussion is when it's desirable to return the dynamic/subfields generated by virtue of being a polyfield, and further how the doc update code is managing that differently from a normal search. For a standard doc query in context of performing an update, we don't want dynamically generated fields returned because that will bork the re-index. For a user glob query, we do want then returned because they may be useful to some users? Is that hovering around accurate?

The root of the problem is that updates in our DAOs involve reindexing the whole document.  If internal fields are returned in the lookup it will bork the re-index.  With a couple exceptions, document lookups are done with the SolrClient.getById method which I believe returns all fields.  I don't know that there is a case where we want to return these dynamically generated fields.  In fact, I'm now convinced the LatLonType field was just not configured correctly and is an isolated issue.

> @merrimanr Are there any circumstances where we'd want copy fields? It seems like we could just direct users to make a copy of the field in Metron and just index the copy fields directly. The main concern would be if we're able to manage globs in there, but even if we don't it doesn't seem like super important functionality to support. At least at a glance; I may definitely be missing use cases.

I believe copy fields is a commonly used feature in Solr.  It can be useful for indexing a value in different ways, with a different chain of analyzers.

justinleet:
@merrimanr Can the copyFields destination be `docValues=false`?  I'm inclined to just document that requirement (assuming it works), and not worry too much about it until someone actually needs something that can't be done via the main Metron processes?

metron-platform/metron-solr/README.md
```diff
@@ -95,3 +95,17 @@ As of now, we have mapped out the Schemas in `src/main/config/schema`.
...
+Indexing LatLon and PointType field types stores data in internal fields that should not be returned in search results.  For these fields a dynamic field type matching the suffix needs to be added to store the data points.
+For example, a LatLongType field should be defined as:
+```
+<dynamicField name="*.location_point" type="location" multiValued="false" docValues="false"/>
+<dynamicField name="*_coordinate" type="pdouble" indexed="true" stored="false" docValues="false"/>
+<fieldType name="location" class="solr.LatLonType" subFieldSuffix="_coordinate"/>
+```
+A PointType field should be defined as:
+```
+<dynamicField name="*._point" type="location" multiValued="false" docValues="false"/>
```
justinleet: This looks like holdover copy paste from the location point. The line below looks like what we want.

justinleet:
I think, based on my understanding, that we document around the copy fields limitation.

@cestella @mmiklavc are you both good with the approach of documenting the schema definitions for LatLon and Point along with the restrictions on copy fields?  Are there any outstanding questions or issues you have that need to be addressed? Or new ones to be added?

metron-platform/metron-solr/README.md
merrimanr: Fixed

merrimanr:
Latest commit fixes the copy/paste error and adds copyFields documentation to the README.

cestella:
@justinleet @merrimanr Nope, I'm good.  I'll give my +1 (pending acceptance by the rest of the people who have commented here).

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
```diff
@@ -46,6 +49,7 @@
   private static String index;
   private static MockHTable table;
   private static IndexDao hbaseDao;
+  protected static AccessConfig accessConfig;
```
mmiklavc: As a general access pattern, I think exposing fields for use via inheritance opens up the possibility for hiding fields and non-explicit side effects. Can you either expose this via a `getAccessConfig()` method or invoke it in the subclass via `super.accessConfig`?

https://docs.oracle.com/javase/tutorial/java/IandI/hidevariables.html

metron-platform/metron-solr/README.md
```diff
@@ -95,3 +95,19 @@ As of now, we have mapped out the Schemas in `src/main/config/schema`.
 Ambari will eventually install these, but at the moment it's manual and
 you should refer to the Solr documentation [https://lucene.apache.org/solr/guide/6_6](here) in general
 and [here](https://lucene.apache.org/solr/guide/6_6/documents-fields-and-schema-design.html) if you'd like to know more about schemas in Solr.
+
+In Metron's Solr DAO implementation, document updates involve reading a document, applying the update and reindexing the whole document.
```
mmiklavc: This is a useful description. Can you take it just a bit further and explicitly spell out the final result? Will an update result in 1 doc in the index or 2? ie. does it overwrite the existing indexed doc or insert a new versioned doc?

metron-platform/metron-solr/README.md
```diff
@@ -95,3 +95,19 @@ As of now, we have mapped out the Schemas in `src/main/config/schema`.
 Ambari will eventually install these, but at the moment it's manual and
 you should refer to the Solr documentation [https://lucene.apache.org/solr/guide/6_6](here) in general
 and [here](https://lucene.apache.org/solr/guide/6_6/documents-fields-and-schema-design.html) if you'd like to know more about schemas in Solr.
+
+In Metron's Solr DAO implementation, document updates involve reading a document, applying the update and reindexing the whole document.  
+Indexing LatLon and PointType field types stores data in internal fields that should not be returned in search results.  For these fields a dynamic field type matching the suffix needs to be added to store the data points.
```
mmiklavc: We went through a lot of discussion on this PR to figure out the exact fields and versions of Solr that the LatLon and Point types are problematic for. Can you be a bit more explicit about that here? It's not just "for example" it's literally "these types are a problem bc \<link to problem\> in Solr version X.X, but have been corrected in Solr version X.Y"

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/integration/SolrUpdateIntegrationTest.java
```diff
@@ -20,15 +20,21 @@
...
 import org.apache.metron.integration.InMemoryComponent;
 import org.apache.metron.solr.dao.SolrDao;
 import org.apache.metron.solr.integration.components.SolrComponent;
+import org.junit.Assert;
+import org.junit.Test;
 
 public class SolrUpdateIntegrationTest extends UpdateIntegrationTest {
 
-  SolrComponent solrComponent;
+  protected static SolrComponent solrComponent;
```
mmiklavc: Why `protected static`?

merrimanr: When multiple tests exist in both SolrUpdateIntegrationTest and UpdateIntegrationTest, solrComponent isn't populated across tests because of how we're attempting to reuse it.  Making it static fixes the issue and matches the pattern in UpdateIntegrationTest.

metron-platform/metron-solr/README.md
merrimanr: This problem exists for these types in 5.5 and 6.6.  I added a comment that recommends using a different spatial field type if on version 6+.

metron-platform/metron-solr/README.md
merrimanr: Done

merrimanr:
@mmiklavc let me know if the recent commit satisfies your feedback.

merrimanr:
Thanks @cestella and @mmiklavc.  Just waiting on a +1 from @justinleet and we'll call it done.

justinleet:
+1 by inspection. Thanks a lot for the work on this.

PR #988, opened 2018-04-11
METRON-1445: Update performance tuning guide with more explicit parameter instructions

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1445

The current performance tuning guide can be challenging to follow for those not familiar with the system. This PR updates the guide to provide details about how and where to find tuning parameters along with details as to how the properties are materialized in a running topology.

Additionally, a supplement to the tuning guide has been provided to pull Metron cluster debugging info into a tarball. As pertains to the perf guide, it captures the state of Ambari configuration, flux files, properties files, and zk config. This helps users confirm system state as they make tuning changes.

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel).
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
n/a

### For documentation related changes:
- [ ] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via `site-book/target/site/index.html`:

  ```
  cd site-book
  mvn site
  ```

PR #996, opened 2018-04-17
METRON-1528: Fix missing file in metron.spec

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1528

I missed the spec file addition for cluster_info.py in https://github.com/apache/metron/pull/988 - I must have run full dev up *before* adding the file and testing it in full dev, otherwise the rpm process would have failed.

### For all changes:

### For code changes:
  ```
  ```

Doing this now
UPDATE: done

n/a

#### Note:

### DIFF
--- a/metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
+++ b/metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
@@ -123,6 +123,7 @@ This package installs the Metron common files %{metron_home}
 %dir %{metron_home}/lib
 %{metron_home}/bin/zk_load_configs.sh
 %{metron_home}/bin/stellar
+%{metron_home}/bin/cluster_info.py
 %{metron_home}/config/zookeeper/global.json
 %attr(0644,root,root) %{metron_home}/lib/metron-common-%{full_version}.jar
 


justinleet:
Can you update the changelog?

mmiklavc:
@justinleet Do we still want/need to do this? The last log change was Tue Sep 25 2017 by @merrimanr in METRON-1207. However, there have been 6 changes to the spec since then that have not made it to the change log. I believe there was a reason we started doing this (in duplication of source control), but I don't recall specifically. Do remember why that was?

justinleet:
I believe, and my memory is pretty fuzzy, is that it's best practice to maintain that changelog because it's useful for auditing and tracking purposes given that it's available on the rpm itself.

There's probably a couple questions here
1. Are we going to maintain it going forward? If not, we should just dump it entirely.
2. If we choose to do so, do we want/need to update the changelog for the missing commits (and probably to use the dev list as authors, rather than individuals)?

Might be worth opening a discuss on it.  I could be persuaded either way in terms of whether we update it for this PR or not, but I have a slight preference on adding it until there's agreement we aren't doing it.

nickwallen:
+1 by inspection.  I would suggest we get this fix in ASAP so we can build.  We can resolve the change log issue later.

justinleet:
I'm +1 on this, to get the build working, with the understanding that there's follow up on resolving our best practices around the spec files.

PR #997, opened 2018-04-18
METRON-1529 CONFIG_GET Fails to Retrieve Latest Config When Run in Zeppelin REPL

nickwallen:
The cache used to back the `CONFIG_GET` function does not update when changes are made in Zookeeper.  This problem only occurs when the REPL is run in Zeppelin.  It does not occur when the REPL is run in the CLI.  [METRON-1529](https://issues.apache.org/jira/browse/METRON-1529) includes step to replicate the issue.

**NOTE:** This change is dependent on #982 .  It should only be merged after #982.

## Changes

The root cause is not clear, but must be due to the way in which Zeppelin launches the interpreters and keeps them isolated from the core of Zeppelin itself.  

That being said, there really is no need for a cache to back `CONFIG_GET`.  The function should always go to Zookeeper to get the latest configuration values.  This simplifies the implementation and solves the problem at-hand.

The changes relevant to this PR include the following.

* [`metron-platform/metron-management/src/main/java/org/apache/metron/management/ConfigurationFunctions.java`](https://github.com/apache/metron/compare/master...nickwallen:METRON-1529?expand=1#diff-00b9178e2e80b694bac96e387f434ec9)

* [`metron-platform/metron-management/src/test/java/org/apache/metron/management/ConfigurationFunctionsTest.java`](https://github.com/apache/metron/compare/master...nickwallen:METRON-1529?expand=1#diff-187b5eaaf0be7dc001aaa47388444036)

* [`metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigurationsUtils.java`](https://github.com/apache/metron/compare/master...nickwallen:METRON-1529?expand=1#diff-b8d2f6514e881628a94f49e342bc7be)

## Testing

* The unit tests were enhanced to validate this change.  

* I have validate this change in the CentOS dev environment

* I have tested this in the Zeppelin REPL by following the steps outlined in [METRON-1529](https://issues.apache.org/jira/browse/METRON-1529).

    screen shot 2018-04-18 at 11 51 06 am

* I have tested this in the CLI REPL by launching the REPL using the following command and then running the same set of Stellar commands.

        java -cp "metron-stellar/stellar-common/target/stellar-common-0.4.3.jar:metron-platform/metron-management/target/metron-management-0.4.3.jar:metron-platform/metron-parsers/target/metron-parsers-0.4.3-uber.jar:~/.m2/repository/org/slf4j/slf4j-simple/1.7.9/slf4j-simple-1.7.25.jar"  org.apache.metron.stellar.common.shell.cli.StellarShell -z localhost:2181

    screen shot 2018-04-18 at 11 54 21 am

## Pull Request Checklist


metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/StellarProcessorUtils.java
```diff
@@ -1,34 +1,31 @@
 /**
- * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
- * regarding copyright ownership.  The ASF licenses this file
- * to you under the Apache License, Version 2.0 (the
- * "License"); you may not use this file except in compliance
- * with the License.  You may obtain a copy of the License at
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
```
mmiklavc: Probably don't want to change the license header format on this - can you revert to the normal formatting? e.g. https://github.com/nickwallen/metron/blob/5e08e28b4f2c67679199fe0bf4af3b66c99630e6/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/shell/DefaultStellarShellExecutor.java

nickwallen: Fixed. Thanks

PR #1000, opened 2018-04-20
METRON-1533 Create KAFKA_FIND Stellar Function

nickwallen:
I created a `KAFKA_FIND` function that allows you to provide a filter expression so that only messages satisfying a condition are returned.   For example...

- Find a message that has been enriched with geolocation data.
    ```
    KAFKA_FIND('indexing', m -> MAP_EXISTS('geo.city', m))
    ```

- Find a Bro message.
    ```
    KAFKA_FIND('indexing', m -> MAP_GET('source.type', m) == 'bro')
    ```

The message is presented to the filter lambda expression as a map of field values.  This makes creating the filter expression a bit simpler.

Like the other `KAFKA_*` functions, this is not intended to be highly performant.  This is only intended to make the process of creating and modifying enrichments simpler in the REPL.  See the **Use Case** section for more details on how I see this being used.

### Future

If we were in the future to provide map literals in Stellar, this would become a fair bit simpler.
```
KAFKA_FIND('indexing', m -> m['source.type'] == 'bro')
```

### Use Case

When creating enrichments, I often find that I want to validate that the enrichment I just created was successful on the live, incoming stream of telemetry. My workflow looks something like this.

1. Create and test the enrichment that I want to create.
    ```
    [Stellar]>>> ip_src_addr := "72.34.49.86"
    72.34.49.86

    [Stellar]>>> geo := GEO_GET(ip_src_addr)
    {country=US, dmaCode=803, city=Los Angeles, postalCode=90014, latitude=34.0438, location_point=34.0438,-118.2512, locID=5368361, longitude=-118.2512}
    ```

2. That looks good to me. Now let's add that to my Bro telemetry.
    ```
    {
      "enrichment" : {
        "fieldMap": {
          "stellar": {
            "config": [
               "geo := GEO_GET(ip_src_addr)"
            ]
          }
        }
      },
      "threatIntel": {
      }
    }

    [Stellar]>>> CONFIG_PUT("ENRICHMENTS", e, "bro")
     ```

3.  It looks like that worked, but did that really work?

    At this point, I would run `KAFKA_GET` as many times as it takes to retrieve a Bro message. You would just have to get lucky and hope that the enrichment worked and secondly that you would pull down a Bro message (as opposed to a different sensor).

    I would rather have a function that lets me only pull back the messages that I care about. In this case I could either retrieve only Bro messages.
    ```
    KAFKA_FIND('indexing', m -> MAP_GET('source.type', m) == 'bro')
    ```

    Or I could look for messages that contain geolocation data.
    ```
    KAFKA_FIND('indexing', m -> MAP_EXISTS('geo.city', m))
    ```

### Changes

* Created the `KAFKA_FIND` function along with unit tests.

* Defined the global property `bootstrap.servers` by default during the MPack install.  This allows all of the `KAKFA_*` functions to work out-of-the-box.  Previously, a user would have to manually define this value before using any of the `KAFKA_*` functions.

###  Pull Request Checklist

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_service.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/metron_service.py
@@ -104,6 +104,11 @@ def build_global_config_patch(params, patch_file):
         "op": "add",
         "path": "/user.settings.hbase.cf",
         "value": "{{user_settings_hbase_cf}}"
+    },
+    {
+        "op": "add",
+        "path": "/bootstrap.servers",
+        "value": "{{kafka_brokers}}"    
     }
   ]
   """
--- a/metron-platform/metron-management/src/main/java/org/apache/metron/management/KafkaFunctions.java
+++ b/metron-platform/metron-management/src/main/java/org/apache/metron/management/KafkaFunctions.java
@@ -18,31 +18,51 @@
 
 package org.apache.metron.management;
 
+import org.apache.commons.lang.ClassUtils;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
+import org.apache.metron.profiler.client.stellar.Util;
+import org.apache.metron.stellar.common.LambdaExpression;
+import org.apache.metron.stellar.common.utils.ConversionUtils;
+import org.apache.metron.stellar.common.utils.JSONUtils;
 import org.apache.metron.stellar.dsl.Context;
 import org.apache.metron.stellar.dsl.ParseException;
 import org.apache.metron.stellar.dsl.Stellar;
 import org.apache.metron.stellar.dsl.StellarFunction;
-import org.apache.metron.stellar.common.utils.ConversionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.*;
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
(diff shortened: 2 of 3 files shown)


merrimanr:
I tested this in full dev and the results were somewhat inconsistent.  I listened on the enrichments topic with the kafka-console-consumer tool in one window:
```
/usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh -z node1:2181 --topic enrichments
```
While repeatedly running this command in another:
```
KAFKA_FIND('enrichments', m -> MAP_GET('source.type', m) == 'snort')
```
About 25-50% of the time the Stellar shell returned `[]` and the other times it would return a snort message as expected.

How long will this command listen until it times out (or is it based on number of messages read)?  Sometimes it returned an empty array immediately.  Is this configurable?

merrimanr:
I should add that I have both bro and snort parser topologies running.

nickwallen:
Thanks for taking it for a test drive.  I think all your observations are explainable, but they all point out usability issues that I think I can improve on.

#### 1. Offsets

`KAFKA_FIND` 'sticks' on its consumer offset.  It operates more like `KAFKA_GET` than `KAFKA_TAIL`.  This is how I described it in the docs.

> Finds messages that satisfy a given filter expression. Subsequent calls will continue retrieving messages sequentially from the original offset.

When you first run `KAFKA_FIND`, its consumer offset will not be set.  It will pick-up from the end of the topic.  When you run it again in the same session, it will continue filtering from those same offsets, rather than going to the end of the topic.  

The `kafka-console-consumer` tool always seeks to the end when it is run.  In your test its likely that `kafka-console-consumer` and `KAFKA_FIND` are at completely different offsets as you try to compare the two.

I had actually already been working on a version of this that always seeks to the end and so behaves more like `KAFKA_TAIL` and `kafka-console-consumer`.

Per the use case I described in the PR, I think 'seek to end' makes more sense.  You make a change on a live stream and want to see the immediate results.  If `KAFKA_TAIL` 'sticks' on an earlier offset, you're not going to see the most recent messages, which can be confusing for the user.

#### 2. Timeouts

> How long will this command listen until it times out (or is it based on number of messages read)? ...  Is this configurable?

The command will poll for up to 5 seconds, by default.  This can be adjusted by defining a global property `stellar.kafka.max.wait`.

> Sometimes it returned an empty array immediately. 

In this case, it probably pulled in messages from the topic, none of those messages matched your filter, and so returned an empty array to you.

I probably need to look at the timeout logic under these conditions.  It should probably 'try harder' to find matching messages and not return immediately.  I'll take a look at this and see if it can be improved.

nickwallen:
I made a bunch of enhancements based on the feedback I outlined above.  I am in the process of breaking that work out into multiple PRs so that it can be reviewed more easily.

PR #1009, opened 2018-05-08
METRON-1549: Add empty object test to WriterBoltIntegrationTest implementation

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1549

Adding a new integration test case to the writer bolt integration test to verify an edge case with how our parserbolt, parsers, and writers handle empty objects returned from custom parsers.

## Pull Request Checklist

### For all changes:

### For code changes:
(pending travis)
  ```
  ```

n/a

### DIFF
--- a/metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/integration/WriterBoltIntegrationTest.java
+++ b/metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/integration/WriterBoltIntegrationTest.java
@@ -17,32 +17,44 @@
  */
 package org.apache.metron.writers.integration;
 
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
 import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import javax.annotation.Nullable;
 import org.adrianwalker.multilinestring.Multiline;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.metron.common.Constants;
 import org.apache.metron.common.configuration.SensorParserConfig;
-import org.apache.metron.stellar.dsl.Context;
 import org.apache.metron.common.field.validation.FieldValidation;
 import org.apache.metron.common.utils.JSONUtils;
 import org.apache.metron.enrichment.integration.components.ConfigUploadComponent;
-import org.apache.metron.integration.*;
+import org.apache.metron.integration.BaseIntegrationTest;
+import org.apache.metron.integration.ComponentRunner;
+import org.apache.metron.integration.Processor;
+import org.apache.metron.integration.ProcessorResult;
+import org.apache.metron.integration.ReadinessState;
 import org.apache.metron.integration.components.KafkaComponent;
-import org.apache.metron.integration.processors.KafkaMessageSet;
 import org.apache.metron.integration.components.ZKServerComponent;
+import org.apache.metron.integration.processors.KafkaMessageSet;
 import org.apache.metron.integration.processors.KafkaProcessor;
-import org.apache.metron.parsers.csv.CSVParser;
 import org.apache.metron.parsers.integration.components.ParserTopologyComponent;
-import org.apache.metron.test.utils.UnitTestHelper;
+import org.apache.metron.parsers.interfaces.MessageParser;
+import org.apache.metron.stellar.dsl.Context;
 import org.json.simple.JSONObject;
-import org.json.simple.parser.ParseException;
 import org.junit.Assert;
 import org.junit.Test;
-
-import javax.annotation.Nullable;
-import java.io.IOException;
-import java.util.*;
 
 public class WriterBoltIntegrationTest extends BaseIntegrationTest {
 
@@ -60,6 +72,14 @@ public class WriterBoltIntegrationTest extends BaseIntegrationTest {
     public void initialize(Map<String, Object> validationConfig, Map<String, Object> globalConfig) {
     }
   }
+
+
+  /**
+   * { }
+   */
+  @Multiline
+  public static String globalConfigEmpty;
+
   /**
    * {
    *   "fieldValidations" : [
@@ -68,7 +88,7 @@ public class WriterBoltIntegrationTest extends BaseIntegrationTest {
    * }
    */
   @Multiline
-  public static String globalConfig;
+  public static String globalConfigWithValidation;
 
   /**
    * {
@@ -88,9 +108,7 @@ public class WriterBoltIntegrationTest extends BaseIntegrationTest {
   public static String parserConfigJSON;
 
   @Test
-  public void test() throws UnableToStartException, IOException, ParseException {
-
-    UnitTestHelper.setLog4jLevel(CSVParser.class, org.apache.log4j.Level.FATAL);
+  public void parser_with_global_validations_writes_bad_records_to_error_topic() throws Exception {
     final String sensorType = "dummy";
 
     SensorParserConfig parserConfig = JSONUtils.INSTANCE.load(parserConfigJSON, SensorParserConfig.class);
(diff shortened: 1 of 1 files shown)


metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/integration/WriterBoltIntegrationTest.java
```diff
@@ -166,6 +185,139 @@ public void test() throws UnableToStartException, IOException, ParseException {
...
+
+    // the input messages to parser
+    final List<byte[]> inputMessages = new ArrayList<byte[]>() {{
+      add(Bytes.toBytes("foo"));
+      add(Bytes.toBytes("bar"));
+      add(Bytes.toBytes("baz"));
+    }};
+
+    // setup external components; zookeeper, kafka
+    final Properties topologyProperties = new Properties();
```
nickwallen: Both test cases **seem** to do the same or similar component setup for Zk, Kafka, etc.  Can we extract any duplicate logic into a shared method?  Maybe just creating a "setup(@Before)" and "teardown(@After)" methods could handle that logic.

Pulling that logic out will also make it easier for me to understand the specifics of what each test case is intended to test.

metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/integration/WriterBoltIntegrationTest.java
```diff
@@ -166,6 +185,139 @@ public void test() throws UnableToStartException, IOException, ParseException {
...
+   *    "sensorTopic":"emptyobjectparser",
+   *    "outputTopic": "enrichments",
+   *    "errorTopic": "parser_error"
+   * }
+   */
+  @Multiline
+  public static String offsetParserConfigJSON;
+
+  @Test
+  public void commits_kafka_offsets_for_emtpy_objects() throws Exception {
```
nickwallen: Are `methods_named_like_this` acceptable according to our style guidelines?  If the Google Java guidelines say yes, then I am ok with it.  Otherwise, can we camel hump these names?

metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/integration/WriterBoltIntegrationTest.java
```diff
@@ -88,9 +108,7 @@ public void initialize(Map<String, Object> validationConfig, Map<String, Object>
   public static String parserConfigJSON;
 
   @Test
-  public void test() throws UnableToStartException, IOException, ParseException {
-
-    UnitTestHelper.setLog4jLevel(CSVParser.class, org.apache.log4j.Level.FATAL);
+  public void parser_with_global_validations_writes_bad_records_to_error_topic() throws Exception {
```
nickwallen: Maybe just a brief javadoc would help here.  I have to read a lot of code to understand what each of these tests are trying to do. One sentence of javadoc would save me some time.

metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/integration/WriterBoltIntegrationTest.java
mmiklavc: I'll look at pulling common logic.

metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/integration/WriterBoltIntegrationTest.java
mmiklavc: I think the quantity of code is mainly due to the existing setup code. I'm happy to refactor that. I really don't like adding or reading javadoc where code can communicate clearly. I think it's redundant to add it where method and class names can appropriately capture the intent. With extracting out the common setup code and the Processor logic, does the test name you've referenced on line 111 along with the test assertions make it clear enough?

nickwallen: > I really don't like adding or reading javadoc where code can communicate clearly. I think it's redundant to add it where method and class names can appropriately capture the intent. 

Yes, I can see your point, but the definition of what is clear can be very different between the original author and others maintaining that code.  Or even the original author and their future self, months later. Redundancy is a small price to pay for clarity. This is especially true in an open source project which we hope long outlives ourselves.  But we can disagree on this point.

I offered the refactor/rename comments only as suggestions.  Do whatever you think is sufficient to make it more clear.  I'm sure any little effort to improve clarity will win me over.  I'll be exceptionally pragmatic here.

metron-platform/metron-parsers/src/test/java/org/apache/metron/writers/integration/WriterBoltIntegrationTest.java
mmiklavc: My thoughts here probably belong in a discuss thread, but the example serves well in the moment. Just for disclosure, I almost always format my code using the IDE formatting tool that Justin shared some time ago with checkstyle. I'm inclined to want to break convention when the outcome can be of greater benefit than adhering strictly to the rules. The underscore pattern happens to be a test method naming convention that a group of 30 engineers and myself landed on 7 years ago while re-writing a million+ line application that had no tests from the ground up to include unit, integration, and acceptance tests.  After a lot of trial and error, it was the most clear for everyone involved.

Which do you prefer to read? Which reads most naturally? Which provides context about the test without needing extraneous javadoc?

```
commits_kafka_offsets_for_emtpy_objects()
CommitsKafkaOffsetsForEmtpyObjects()
testCommits()
test()
```

**TL;DR**

Per the Google style guide, this naming convention is allowable:
> Underscores may appear in JUnit test method names to separate logical components of the name, with each component written in lowerCamelCase. One typical pattern is <methodUnderTest>_<state>, for example pop_emptyStack. **There is no One Correct Way to name test methods.**

https://google.github.io/styleguide/javaguide.html#s5.2-specific-identifier-names
https://stackoverflow.com/questions/96297/what-are-some-popular-naming-conventions-for-unit-tests
https://dzone.com/articles/7-popular-unit-test-naming

justinleet: I vote for one that doesn't contain "emtpy" in it.

nickwallen: No doubt I've seen it done in other code bases.  Its just not common in the Metron code base, thus I called it out.  

Since it doesn't violate the code style guidelines, then I have no basis to complain.  I wasn't aware that was acceptable.  Thanks for clarifying.

nickwallen: > @justinleet I vote for one that doesn't contain "emtpy" in it.

And "emtpy" is actually the Olde English spelling, which I am fine with.

mmiklavc: I agree @nickwallen - it isn't the prevailing idiom that we've used. And regardless of whatever the current standards are, I'm a big fan of open dialogue and comments like the ones you made here.

mmiklavc:
@nickwallen and @ottobackwards - I extracted the common setup components into one parameterized method. I opted for setting private test class fields with the components rather than invoking getComponent() because I thought it was a bit cleaner. I also looked at doing this in @Before and @After setup methods, but setup for each of the tests changes on a per-test basis. 

We're currently employing a useful builder pattern to configure the test components/infrastructure which generally means you provide invariants and then build the object from them. It might also be worthwhile to look at a way to use the builder pattern, but also change certain aspects of the setup within the individual test cases, e.g. upload a new global config after spinning up the components. This could make it more efficient to run multiple tests by using @Before and @After to spin things up to run a battery of similar integration tests. Do you guys think we need it for this test change?

ottobackwards:
I think this is good.  My work on the builder/Processor stuff was a decent attempt to reduce duplicate code in the integration tests without a total rethink.  I think what we see across the integration tests is so much commonality that it is apparent that some re-think could be useful to have more standardize test classes and use cases structured such that there is less overhead of duplicated code required.

As I have done some work on other projects and seen other approaches this has only become more clear.  commons-vfs for example runs the same tests against every provider, with each provider providing specialization of it's suites. 

It would seem that we should be able to have a more generic suite or testing harness for *n* topologies/topics working together, as a next step refactoring of my prior `Process` work.

I don't think this is the PR for that though

nickwallen:
+1 This looks good.  Thanks!

ottobackwards:
+1

PR #1010, opened 2018-05-11
METRON-1548: Remove hardcoded source:type from Alerts UI

sardell:
## Contributor Comments
Instead of 'source:type' being hard coded as a string in the Alerts UI, we now use a global-config service to retrieve whether a user's setup is using "source:type" or "source.type" syntax (e.g. elastic search or solr). The app default's to using "source:type" if no configuration is specified. If a user wants, they can update their configuration to use source.type by adding `"sourceType": "source.type"` to their configuration.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
@@ -179,9 +179,12 @@ export class AlertsListComponent implements OnInit, OnDestroy {
   ngOnInit() {
     this.globalConfigService.get().subscribe((config: {}) => {
       this.globalConfig = config;
-      if (this.globalConfig['source.type.field'] === 'source.type' && !this.alertsColumns['source.type']) {
-        this.alertsColumns = this.alertsColumns.filter(colName => colName.name !== 'source:type');
-        this.alertsColumns.splice(2, 0, new ColumnMetadata(config['source.type.field'], 'string'));
+      if (this.globalConfig['source.type.field']) {
+        let filteredAlertsColumns = this.alertsColumns.filter(colName => colName.name !== 'source:type');
+        if (filteredAlertsColumns.length < this.alertsColumns.length) {
+          this.alertsColumns = filteredAlertsColumns;
+          this.alertsColumns.splice(2, 0, new ColumnMetadata(this.globalConfig['source.type.field'], 'string'));
+        }
       }
     });
     this.getAlertColumnNames(true);
--- a/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
@@ -177,13 +177,17 @@ export class AlertsListComponent implements OnInit, OnDestroy {
   }
 
   ngOnInit() {
+    this.globalConfigService.get().subscribe((config: {}) => {
+      this.globalConfig = config;
+      if (this.globalConfig['source.type.field'] === 'source.type' && !this.alertsColumns['source.type']) {
+        this.alertsColumns = this.alertsColumns.filter(colName => colName.name !== 'source:type');
+        this.alertsColumns.splice(2, 0, new ColumnMetadata(config['source.type.field'], 'string'));
+      }
+    });
     this.getAlertColumnNames(true);
     this.addAlertColChangedListner();
     this.addLoadSavedSearchListner();
     this.addAlertChangedListner();
-    this.globalConfigService.get().subscribe((config: {}) => {
-      this.globalConfig = config;
-    });
   }
 
   onClear() {
--- a/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
@@ -42,7 +42,7 @@ import {Pagination} from '../../model/pagination';
 import {META_ALERTS_SENSOR_TYPE, META_ALERTS_INDEX} from '../../utils/constants';
 import {MetaAlertService} from '../../service/meta-alert.service';
 import {Facets} from '../../model/facets';
-import { environment } from 'environments/environment';
+import { GlobalConfigService } from '../../service/global-config.service';
 
 @Component({
   selector: 'app-alerts-list',
@@ -76,6 +76,7 @@ export class AlertsListComponent implements OnInit, OnDestroy {
   pagination: Pagination = new Pagination();
   alertChangedSubscription: Subscription;
   groupFacets: Facets;
+  globalConfig: {} = {};
 
   constructor(private router: Router,
               private searchService: SearchService,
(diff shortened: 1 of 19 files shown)


justinleet:
Thanks for the contribution, getting this cleaned up is really helpful!

Could you update https://github.com/apache/metron/tree/master/metron-platform/metron-common#global-configuration with the new config property?

I'd also like to see the property be something like `source.type.field` to be a bit more explanatory and consistent with the other non-Object type fields.

metron-interface/metron-alerts/src/app/service/global-config.service.ts
```diff
@@ -0,0 +1,88 @@
...
+import {HttpUtil} from '../utils/httpUtil';
+
+@Injectable()
+export class GlobalConfigService {
+  url = 'api/v1/global/config';
+  defaultHeaders = {'Content-Type': 'application/json', 'X-Requested-With': 'XMLHttpRequest'};
+
+  private globalConfig = {};
+
+  constructor(private http: Http) {
```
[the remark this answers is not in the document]
sardell: Thanks for pointing this out, @merrimanr. Most of the file was carried over from metron-config, and after taking another look I agree that the constructor should be empty. I'm also going to remove the post and delete methods, as they are not used in metron-alerts currently.

sardell:
@justinleet I've changed the property to be `source.type.field` and added the property to the documentation at https://github.com/apache/metron/tree/master/metron-platform/metron-common#global-configuration. I was slightly confused about the subset field, but I took an educated guess and entered `UI` as the value. Let me know if this is not correct and I'll update it.

justinleet:
@sardell Looks like Travis failed for unrelated reasons (which I'm curious about on it's own).  Could you just close and reopen the PR to kick Travis off again?

justinleet:
Also, UI for the subsytem should be fine. It's purely informational.  If you wanted to be more explicit, you could put "Alerts UI" or something similar, but I think just "UI" is fine.

justinleet:
I spun this up, changed the config, restarted the UI and had a couple comments.

1. Every time you log in, it keeps adding new source.type fields in the table. Unselecting these in the options doesn't work properly (only the top one toggles no matter which I click on).  Possibly doubles every time?
2. The filters on the left still say "source:type", even after the config is changed.
3. The group by still says "source:type", even after the config is changed. This grouping still works fine (which I would expect if it's still grouping on source:type).

I would expect both the filter and group by to act on the configured `source.type.field` by default.  Otherwise, the user has to know what's going on and change it manually.

I'd also like to add a bit in the documentation that updating the `source.type.field` requires a UI restart.  This is pretty typical of the UI, but most global config things get picked up automagically (e.g. topology changes and such), so a quick blurb would be nice.

metron-interface/metron-alerts/src/app/alerts/alerts-list/table-view/table-view.component.ts
```diff
@@ -77,12 +80,24 @@ export class TableViewComponent implements OnChanges {
...
     this.metronDialogBox = metronDialogBox;
     this.updateService = updateService;
     this.metaAlertService = metaAlertService;
+    this.globalConfigService = globalConfigService;
+  }
+
+  ngOnInit() {
+    this.globalConfigService.get().subscribe((config: {}) => {
+      this.globalConfig = config;
+      if (this.globalConfig['source.type.field'] === 'source.type' && !this.alertsColumnsToDisplay['source.type']) {
```
merrimanr: I don't think this logic is correct.  What happens if you have "source.type" selected as a table column?  I went in a updated by local storage directly and ended up with 2 "source.type" columns.

metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
```diff
@@ -174,6 +177,13 @@ export class AlertsListComponent implements OnInit, OnDestroy {
   }
 
   ngOnInit() {
+    this.globalConfigService.get().subscribe((config: {}) => {
+      this.globalConfig = config;
+      if (this.globalConfig['source.type.field'] === 'source.type' && !this.alertsColumns['source.type']) {
```
merrimanr: Same as https://github.com/apache/metron/pull/1010/files#r189060226

merrimanr:
Most of this worked, I could only find one issue.  We store a user's selected columns in local storage right now.  These columns are selected in the Configure Table view and the list of available columns comes from a REST call that returns ES (or Solr) metadata.  The available column list will always be correct since that information is coming straight from the index templates or schemas but the column name in local storage will become incorrect since it's still the old source type.

There is an attempt to just transform the name in the selected column list but I think this is flawed (I added an inline comment on that).  I am not sure what the best way to handle this is.  I feel like if we find ourselves hardcoding either "source:type" or "source.type" we're doing something wrong.  

One idea is to limit the source.type.field property to be either "source:type" or "source.type".  This would at least allow us to perform those hardcoded checks to see if we need to substitute the new value.  Otherwise we'll need to evolve our local storage model by using boolean fields like "includeSourceType" instead of the actual source.type.field value.  This would make upgrading harder though.

justinleet:
@merrimanr After thinking about this a bit, is that actually a serious problem?  This setting is likely to only be done on initial install (e.g. when selecting Solr or ES).  Switching midlife is unlikely to be a problem for the vast majority of users.  Most users should never see the problem of it being switched because they'd never see `source:type` instead of `source.type` or vice versa, because you'd have to reindex all your old data into the new format anyway.  We'd need to improve documentation in the Solr feature branch to ensure that users initially set it up appropriately.

Another way of going about this might be to have the app itself get config updates on the fly similar to how our topologies do, given that this particular one is from ZK.  At that point, you could do the mapping from old source type to new source type and update the storage appropriately.  I'm sure there's way more to it than just doing that though (when do you update UI for users, what are the potential snags in updating that when users are querying, etc.).

merrimanr:
@justinleet I agree with you that it's not a serious problem.  I think it can be solved more easily with documentation and some manual steps from the user.  There is some logic we need to fix, specifically around substituting the source.type.field value for "source:type" (I added comments for those).  Other than that, I'm good with documenting how to switch to Solr in the Solr feature branch.

There is also the case of upgrading ES and using "source.type" since that name is allowed in later versions.  Is that something we need to think about and document?

justinleet:
I think the general ES migration to source.type is outside the scope of this PR.  That would have to be part of a larger migration anyway since you'd have to reindex data on an existing metron cluster to do that.  I'm sure we'll want to think about it at some point, though.

metron-interface/metron-alerts/src/app/alerts/alerts-list/table-view/table-view.component.ts
sardell: @merrimanr Good point. What if we check if a config property of ['source.type.field'] exists and change it to whatever value it has. If no config exists, default to `source:type` (to catch existing users without a config property set)?

```
if (this.globalConfig['source.type.field']) {
        this.alertsColumnsToDisplay.splice(2, 0, new ColumnMetadata(config['source.type.field'], 'string'));
      } 
else {
        this.alertsColumnsToDisplay.splice(2, 0, new ColumnMetadata(config['source:type'], 'string'));
      }
```

merrimanr:
+1 pending travis

PR #1014, opened 2018-05-14
METRON-1563 : Base Stellar assign for feature branch

ottobackwards:
repackage: 
https://github.com/apache/metron/pull/687

Please sanity check and see that PR


ottobackwards:
@cestella I rebased this on the new feature branch ( after rebasing on the same master ) and I get all of these other commits.  I don't know how to get rid of them?

ottobackwards:
ok, @cestella I have fixed up the feature branch and this pr so it is clean

nickwallen:
> Should assignment be = or :=?

Ultimately, I'd like to see us move to supporting `=` as that is just way more intuitive for most people.  Maybe that means supporting both `=` and `:=` for some period of time?

It looks like `StellarAssignment` was not altered here, so maybe both are supported in this PR?  Is that what you are thinking @ottobackwards ?

I am not sure how this PR impacts all the places where Stellar is used and assignment might already occur; field transforms, enrichment,  REPL, profiler, etc.  There is some other way in which assignment is handled in at least some of these places.  I imagine that we would need to update these to use the new mechanism here?  Or maybe that is left to a future PR?

nickwallen:
> I am not sure how this PR impacts all the places where Stellar is used and assignment .. Or maybe that is left to a future PR?

Sorry, its a little hard to digest that old PR, but I think from [this comment](https://github.com/apache/metron/pull/687#issuecomment-387840955), @ottobackwards is saying that all the other touch points would be updated in a future PR on this feature branch.

cestella:
>Ultimately, I'd like to see us move to supporting = as that is just way more intuitive for most people. Maybe that means supporting both = and := for some period of time?

In my opinion we should support both for at least a release.

cestella:
Do we want to expand assignment to `:=` as part of this PR or make it a follow-on in this feature branch?  I'm ok with either way.

ottobackwards:
Can I ask how we do ++, --, += etc with the := notation?

JonZeolla:
I wonder if it's good enough just to keep `:=` and `=` around, but not have that affect the format of `++`, `+=`, etc.

ottobackwards:
I'll expand the assignment to include := in literal stellar, maybe that will bump some review

ottobackwards:
ok, I updated the feature base to track master, and pulled that into this.
I also added support for `:=` @cestella

cestella:
Ok, this looks good.  +1 to go to the feature branch

ottobackwards:
This is in, not sure why it isn't closed

justinleet:
@ottobackwards If it's a feature branch, the PR on Github doesn't automatically close. It only happens for merges into a master.  It's a bit annoying, but I'm not sure there's a whole lot to do about it

PR #1015, opened 2018-05-15
METRON-1544 Flaky test: org.apache.metron.stellar.common.CachingStell…

nickwallen:
This fixes the `CachingStellarProcessorTest` which is failing intermittently.

```
Running org.apache.metron.stellar.common.CachingStellarProcessorTest
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.054 sec <<< FAILURE! - in org.apache.metron.stellar.common.CachingStellarProcessorTest
testCaching(org.apache.metron.stellar.common.CachingStellarProcessorTest) Time elapsed: 0.053 sec <<< FAILURE!
java.lang.AssertionError: expected:<6> but was:<4>
 at org.junit.Assert.fail(Assert.java:88)
 at org.junit.Assert.failNotEquals(Assert.java:834)
 at org.junit.Assert.assertEquals(Assert.java:645)
 at org.junit.Assert.assertEquals(Assert.java:631)
 at org.apache.metron.stellar.common.CachingStellarProcessorTest.testCaching(CachingStellarProcessorTest.java:73)
 at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
 at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
 at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
 at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
 at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
 at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
 at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
 at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
 at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
 at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:283)
 at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:173)
 at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:153)
 at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:128)
 at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:203)
 at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:155)
 at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:103)
```

### Changes

* Added a property `stellar.cache.record.stats` that allows recording of cache usage stats. 

* Updated the tests to use the stats produced by Caffeine.

## Pull Request Checklist

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/CachingStellarProcessor.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/CachingStellarProcessor.java
@@ -161,14 +161,18 @@ public class CachingStellarProcessor extends StellarProcessor {
   }
 
   /**
-   * Create a cache key.
+   * Create a cache key using the expression and all variables used by that expression.
    *
    * @param expression The Stellar expression.
    * @param resolver The variable resolver.
    * @return A key with which to do a cache lookup.
    */
   private Key toKey(String expression, VariableResolver resolver) {
+
+    // fetch only the variables used in the expression
     Set<String> variablesUsed = variableCache.get().computeIfAbsent(expression, this::variablesUsed);
+
+    // resolve each of the variables used by the expression
     Map<String, Object> input = new HashMap<>();
     for(String v : variablesUsed) {
       input.computeIfAbsent(v, resolver::resolve);
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/CachingStellarProcessor.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/CachingStellarProcessor.java
@@ -19,11 +19,16 @@ package org.apache.metron.stellar.common;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.metron.stellar.common.utils.ConversionUtils;
 import org.apache.metron.stellar.dsl.Context;
 import org.apache.metron.stellar.dsl.VariableResolver;
 import org.apache.metron.stellar.dsl.functions.resolver.FunctionResolver;
+import org.slf4j.LoggerFactory;
 
+import java.lang.invoke.MethodHandles;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


mmiklavc:
Can you call out the lines relevant to the fix for reviewers?

nickwallen:
> @mmiklavc: Can you call out the lines relevant to the fix for reviewers?

I provided some more color in the description this time.  Also, copied below.

> The core problem is that the test makes assumptions about the cache state which are not always true. Specifically, even though the cache max size is 2, the cache can exceed that size for short periods of time until cache maintenance is performed.

> Since the cache can be larger than expected for brief periods of time, the expectations of what is a hit or miss were incorrect. This is why the test would fail intermittently.

> To address this, I rewrote the test so that the cache stats are validated after each expression is run and I also perform cache maintenance between each execution and validation of cache state.

> Without the additional logging, the ability to grab more detailed cache stats, and breaking down that test case, I wouldn't have been able to uncover the bug.

nickwallen:
Ok, now its really working.  

The second problem was that there is some variability in which items get expired from the cache according to the LRU algorithm.  In about 1/250 runs the algorithm would not choose the expected item to expire.  I imagine this is only a problem when running at such small scale with a cache size of 2.

I updated the test case so that it looks into the cache to determine whether a hit should occur or not.  The test passes, but I am not sure if the test is really useful.

ben-manes:
Cache eviction is performed asynchronously by default using the FJP common pool. For testing you can set `executor(Runnable::run)` and it will be on the calling thread. The penalty is small so it’s sometimes used in practice anyway.

The eviction is not LRU and has a random seed to guard against HashDoS attacks. This means that which entry is not predictable without forcing the seed, which is not exposed. Typically only the cache’s test should do that. You’re code shouldn’t make assumptions on what is evicted as we will try to maximize the hit rate using various techniques.

ottobackwards:
Is there not some relevant Caffeine test we can ape for this?

nickwallen:
> @ben-manes You’re code shouldn’t make assumptions on what is evicted as we will try to maximize the hit rate using various techniques.

That makes a ton of sense.  Thanks for the tips.

nickwallen:
In the latest, I removed the problematic test case.  There is no need to test the eviction policy of the underlying cache, which is effectively what is happening.  If anyone can find gaps in test coverage, please let me know.

nickwallen:
@mmiklavc bump. Did I answer your question?  How is this looking?

PR #1019, opened 2018-05-16
METRON-1555: Update REST to run YARN and MR jobs

merrimanr:
## Contributor Comments
This PR sets us up to run YARN and MR jobs inside our REST application.  Changes include:
- addition of maven dependencies
- addition of -Dhdp.version parameter to the REST startup script
- MPack now supplies the hdp.version parameter
- MPack now sets up a "metron" service user HDFS directory needed for running MR jobs
- MPack now sets up a pcap HDFS directory 
- addition of a Pcap controller with a single Fixed Pcap Query endpoint and service to demonstrate running MR jobs in REST

The fixed pcap query endpoint submitted here should match the functionality in the metron-api module with a few minor differences:  

- the default input and output paths are spring properties instead of hardcoded in classes (this will make it easier to expose them in Ambari if we choose to)
- query results are not cleaned up automatically since that work is captured in a separate Jira
- num reducers is defaulted to 1 instead of 10

Unit and integration tests are included and this has been tested in full dev.  I tested this by generating sample pcap data with the PcapTopologyIntegrationTest.  You can do this by either:

- running the test in your IDE and pausing it after the topology has generated data
- commenting out `clearOutDir(outDir);` and running the test

Pcap data should be present in `/metron/metron-platform/metron-pcap-backend/target/pcap/data_dir`.  Upload the `pcap*` files to the `/apps/metron/pcap` directory in HDFS.  You should be able to perform the tests in PcapTopologyIntegrationTest using REST and get the same results.  

For example:
```
  "endTime": 1458240269424,
  "startTime": 1458240269419
}' 'http://node1:8082/api/v1/pcap/fixed'
```
should return 2 pcap results.

```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{}' 'http://node1:8082/api/v1/pcap/fixed'
```
should return 20 pcap results.

```
  "ipDstAddr":"207.28.210.1"
}' 'http://node1:8082/api/v1/pcap/fixed'
```
should return no pcap results.

```
  "ipDstPort": 22
}' 'http://node1:8082/api/v1/pcap/fixed'
```
should return 10 pcap results.

Related discussion can be found here:

http://mail-archives.apache.org/mod_mbox/metron-dev/201805.mbox/%3CCAEVkqPbxzJNU_WGRbFWnZ-MVqnKb7mthEdvEQ9PLYhwfiT7F0g@mail.gmail.com%3E

## Pull Request Checklist

### For all changes:

### For code changes:
- [] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


JonZeolla:
Based on the temporary/example state of some of this PR would it make sense to explicitly keep this one out of the upcoming release, regardless of the review cycle?

ottobackwards:
This pr is against the feature branch @JonZeolla so, it is not in play

JonZeolla:
Gotcha, GitHub mobile seems to collapse that information.  👍

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/PcapQueryService.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.metron.rest.service;
+
+import org.apache.metron.rest.model.PcapsResponse;
+
+public interface PcapQueryService {
+
+  PcapsResponse query();
```
mmiklavc: Noting this as an incomplete implementation.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/PcapQueryServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.List;
+
+@Service
+public class PcapQueryServiceImpl implements PcapQueryService {
+
+  @Override
+  public PcapsResponse query() {
```
mmiklavc: Noting as incomplete API definition implementation.

merrimanr:
I added a link to the discuss thread to the description.  There is also a link in the top-level Jira:  https://issues.apache.org/jira/browse/METRON-1554.

The latest commit removes the Service classes so that only a sample controller remains.  I added a comment to that controller stating it should be replaced.  There was a Jira created for service development and API design:  https://issues.apache.org/jira/browse/METRON-1559.  Hopefully that makes it clearer that API design is out of scope for this PR.  I can also add a comment to that Jira once this PR gets accepted.

I feel like spinning up and testing this in full dev should be part of the review.  Most of the changes here are Ambari and Maven related.  To test this properly there will need to be some way to submit a MR job (any job) from REST, hence the throwaway sample controller.  If you want to create your own sample controller and test that way I can remove the one in this PR.  Is there an easier option I'm not thinking of?

Maintaining a needs addressed list sounds like a good idea.  I'm not clear on how this would work in practice though.  Where would we keep it?  Would Jira tasks serve this purpose?  This seems like it could be applicable to all tasks in this feature branch (all feature branches really) so I don't think this is the appropriate place to discuss it.  Can we move it back to a discuss thread so you can clarify and expand on your idea?  I think it needs some more thought if we're to adopt it as a process.

mmiklavc:
@merrimanr Thanks for adding the discuss thread links.

It sounds like this doesn't do much aside from a pcap service. Why not just handle them together? That way we're not dealing with TODO's and extra lists, and together they would form a concise unit of work.

merrimanr:
This was my attempt to split it up into a more manageable review.  I think this will end up being a large PR but that's fine with me as long as everyone else is ok with it.  Happy to keep going.

mmiklavc:
I guess I'm saying combine this with a basic PcapService that has an api somewhat flushed out - not necessarily final, but close enough to start filling in implementation details when new features/PR's with those features are added. It's effectively what you have already with just a bit more of the service implemented/defined.

merrimanr:
Sure.  In that case we need to break up https://issues.apache.org/jira/browse/METRON-1559 into smaller tasks which we need to do that anyways.  Which endpoints would you include in a basic PcapService?  Maybe just these 2:

- POST /api/v1/pcap/fixed
- POST /api/v1/pcap/query

We could manually test in full dev by inspecting results in HDFS.  I still think a basic PcapService with a couple endpoints is going to require a large PR.  Just want to make sure you're ok with that.

mmiklavc:
How about starting with 1 of those, fixed, and maybe making it synchronous in the first pass so you can integration test? I'd consider many of those additional api endpoints as features to be built out in this feature branch.

merrimanr:
The latest commit adds the fixed endpoint.  I updated the PR description with steps to test it in full dev.

justinleet:
I'm good with this right now with the follow-on changes and tests, assuming @mmiklavc is good.  I do have the same concerns Mike had, but I think being watchful going forward is a good compromise, but that involves a bit of vigilance here.

justinleet:
@merrimanr Can you kick Travis?

mmiklavc:
Thanks for the adjustments @merrimanr. Looks good, +1.

PR #1021, opened 2018-05-21
METRON-1568: Stellar should have a _ special variable which returns the message in map form

cestella:
## Contributor Comments
In order to support functions which operate on the whole message, we should have a special variable (_, keeping with the vaguely scala theme) which can return the entire underlying message.  This map should be immutable.

Manual testing can be done by evaluating `MAP_GET('source.type', _)`:
* via the REPL (presuming `source.type` is defined in the REPL)
* adding a field transformation to a parser
* a stellar enrichment

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/enrichment/handler/StellarConfig.java
```diff
@@ -142,8 +143,14 @@ else if(kv.getValue() instanceof List) {
   {
```
ottobackwards: Why would this be an either or situation?
What if I have a 'match' expression that uses both _ and other vars?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/ConcatMap.java
```diff
@@ -0,0 +1,202 @@
...
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Sets;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
```
ottobackwards: Can throw some java doc on this about what it is supposed to provide?  And the assumptions that it makes ( like the uniqueness of the keys in the different maps and what happens etc )

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/enrichment/handler/StellarConfig.java
cestella: If you are using a `_`, then you cannot make a judgement about which fields you're using, so you have to include all fields.  If you aren't, then you can make some decision and only want to send the fields required to the adapter.  That's the basis for the either/or situation.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/ConcatMap.java
cestella: Yep, absolutely.

cestella:
@ottobackwards I don't see it really affecting assignment.  `_` should behave just like any other variable to my mind and return a map.  For instance, if you say `new_field := _`, then I'd expect `new_field` to be a map with all of the existing fields for a message.  We should prevent `_` from being on the left-hand-side, though, I think.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/enrichment/handler/StellarConfig.java
cestella: Well, I added it because I came across it multiple times in data science use-cases.  Specifically, the way that we do model application via MaaS (and other model applications) presumes that the user will filter just the fields into the model, rather than having the model select it.  This puts an undue burden on the user as the model knows which fields it wants and the user may not, so being able to pass the whole message lets the model do the decision.  I think that `_` is a natural way to do that.  To be fair, also, this is a capability inherent to the VariableResolver, not the grammar itself, which I thought was reasonable.  SOme variable resolvers won't support it and some will.

Thoughts?

ottobackwards: I don't doubt the need or applicability, I guess I'm saying why not just have the _ in the configuration side, and just have the scripts reference the vars by name and not have to MAP_GET()?

That way, the vars are the vars are the vars in Stellar, and the external configuration can have the option for putting all the message in the var namespace.

cestella:
[Multimaps](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Multimap.html#entries()) don't have unique keys.  Making the unique keys would necessitate copying the maps, which seems excessive to me.

metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/enrichment/handler/StellarConfig.java
ottobackwards: I'm trying to wrap my head around why you need the special handling in the variable resolver.
If one of the vars points to a map, that is it.  You don't need to do any special dance.

You can have the variable context 'builder' ( config whatever ) allow naming the message as a var with a certain name and that is it.

I don't understand why all the extra stuff is required.  IE.  why the variable resolver needs to know it is a map.  It is just a var.

cestella: I thought about about the variable context builder idea over lunch.  Is what you are proposing to change this so that `MapVariableResolver`, instead of taking a list of maps, we take a Map of maps.  e.g. `MapVariableResolver(Map<String, Map<String, Object>> variables)` and the resolution would check that map first?  So, for instance:
```
MapVariableResolver resolver = new MapVariableResolver(ImmutableMap.of("_", message, "global", globalConfig));
```

ottobackwards: Something like that, in leu of literal scopes ( which would have a hierarchy ) might work out.  

I'm not trying to shoot down your idea, I just want to understand it, and we approach things differently.

I think in the end we want to get to
interface Scope {
Scope getParent()
Resolve()
Get()
Put()
}

before calling a function/lambda

Scope scope = new Scope(stack.peek())
stack.push(scope)
call function(scope)
stack.pop(scope)

cestella: So, what you're trying to solve, I think, is scoped variable resolution.  I think this is a bit of a different use-case here and one that doesn't really have an analogue in programming languages at large.  This is, rather, getting a list of the variables and their values within the current scope.  For our purposes, though, we only have one scope, global scope.  I'd consider non-global scope resolution to be out of scope for this.

Ultimately, what I'm arguing for is the map variable resolver to have a special variable that returns the full variable scope. 

Is this becoming clearer or am I spreading more mud than I'm clearing?

ottobackwards: My impression is that you are creating an implicit scope with the message, maybe that is where we are crossing

cestella: The PR as it stands does not make any scoping designation beyond global.  Indeed, it is pulling in ALL of the available variables.  My adjustment above adds namespacing (e.g. _ implies the message variables, `global` imples just the global config variables).

ottobackwards:
@cestella re:assignment.  If all assignment is in the language, and _ is either or, then how can we assign to a message field when _ is active?

cestella:
@ottobackwards As long as the underlying map is updated in `MapVariableResolver` as per the `assign` function in that variable resolver, it should be supported without any change as the approach is to return the ConcatMap, which is backed by the real maps, which are updated.  ConcatMap is immutable, but the underlying maps that backs it aren't and assign would update those maps directly.

cestella:
Ok, just ran this up in full-dev and it works as advertised.  If this needs to change the fundamental approach, let me know, but as it is submitted it functions as advertised.

mmiklavc:
+1 via inspection.

PR #1022, opened 2018-05-22
METRON-1569 Allow user to change field name conversion when indexing …

nickwallen:
The `ElasticsearchWriter` has a mechanism to transform the field names of a message before it is written to Elasticsearch.  Right now this mechanism is hard-coded to replace all '.' dots with ':' colons.

This mechanism was needed for Elasticsearch 2.x which did not allow dots in field names.  Now that Metron supports Elasticsearch 5.x this is no longer a problem. A user should be able to configure the field name transformation when writing to Elasticsearch, as needed.  

While it might have been simpler to just remove the de-dotting mechanism, this would break backwards compatibility.  Taking this approach provides users with an upgrade path.

## Changes

This change allows the user to configure the field name converter as part of the index writer configuration.  

Acceptable values include the following.
* `DEDOT`: Replaces all '.' with ':' which is the default, backwards compatible behavior.
* `NOOP`: No field name change.

If no "fieldNameConverter" is defined, it defaults to using `DEDOT` which maintains backwards compatibility.

A cache of `FieldNameConverter`s is maintained since the index writer configuration can be changed at run-time and each sensor has its own index writer configuration.

An example configuration looks-like the following.
```
    {
      "hdfs" : {
        "enabled" : false
      },
      "elasticsearch" : {
        "index" : "bro",
        "batchSize" : 5,
        "enabled" : true,
        "fieldNameConverter": "NOOP"
      },
      "solr" : {
        "enabled" : false
      }
    } 
```

## Code Changes

* Added the `fieldNameConverter` parameter to the Index writer configuration.

* Moved the `FieldNameConverter` implementations to a dedicated package in `metron-common`.

* Renamed `ElasticsearchFieldNameConverter` to `DeDotFieldNameConverter`.

* Implemented the `NoopFieldNameConverter` which does not modify the field name.

* Created `FieldNameConverters` class that allows a user to specify either `DEDOT` or `NOOP` to choose the appropriate implementation.

* Implemented a `CachedFieldNameConverterFactory` that encapsulates all the logic for choosing and instantiating the appropriate `FieldNameConverter`.  

* Updated `ElasticsearchWriter` to use the `CachedFieldNameConverterFactory`.

* Updated the README to document the new configuration parameter.

## Manual Testing

1. Launch a development environment and login.

    ```
    vagrant ssh
    sudo su -
    source /etc/default/metron
    ```

1. Validate the environment by ensuring alerts are visible in the Alerts UI and that the Ambari Service Check completes successfully.  This ensures that the change is backwards compatible.

1. Login to the Storm UI and enable DEBUG logging for `org.apache.metron.common` and `org.apache.metron.elasticsearch`.

1. The Storm worker logs in `/var/log/storm/worker-artifacts/random_access_indexing*/worker.log` should contain the following log statements, if you have enabled DEBUG logging correctly.  This shows that the default `DEDOT` converter is in-use.

    ```
    2018-05-22 14:38:... [DEBUG] Renamed dotted field; original=source.type, new=source:type
    2018-05-22 14:38:... [DEBUG] Renamed dotted field; original=adapter.geoadapter.end.ts, new=adapter:geoadapter:end:ts
    2018-05-22 14:38:... [DEBUG] Renamed dotted field; original=threatintelsplitterbolt.splitter.end.ts, new=threatintelsplitterbolt:splitter:end:ts
    2018-05-22 14:38:... [DEBUG] Renamed dotted field; original=adapter.threatinteladapter.begin.ts, new=adapter:threatinteladapter:begin:ts
    2018-05-22 14:38:... [DEBUG] Renamed dotted field; original=enrichments.geo.ip_dst_addr.location_point, new=enrichments:geo:ip_dst_addr:location_point
    2018-05-22 14:38:... [DEBUG] Renamed dotted field; original=adapter.threatinteladapter.end.ts, new=adapter:threatinteladapter:end:ts
    2018-05-22 14:38:... [DEBUG] Renamed dotted field; original=enrichmentsplitterbolt.splitter.end.ts, new=enrichmentsplitterbolt:splitter:end:ts
    ```

1. Launch the REPL.

    ```
    ./bin/stellar -z $ZOOKEEPER
    ```

1. Change the field name converter to NOOP.

    ```
    {
      "hdfs" : {
        "enabled" : false
      },
      "elasticsearch" : {
        "index" : "bro",
        "batchSize" : 5,
        "enabled" : true,
        "fieldNameConverter": "NOOP"
      },
      "solr" : {
        "enabled" : false
      }
    }

    [Stellar]>>> CONFIG_PUT("INDEXING", conf, "bro")
    ```

1. It can take up to 5 minutes for the topology to pick-up this change.  The old `FieldNameConverter` needs to expire from the cache first.

1. Go back to the Storm worker logs.  When the change takes effect, we should see a log like the following indicating that the `NoopFieldNameConverter` was created.

    ```
    2018-05-22 16:... [DEBUG] Created field name converter; sensorType=bro, configuredName=NOOP, class=NoopFieldNameConverter
    ```

1. In the same logs, we will start to see tuples fail to be indexed.  Elasticsearch complains because the templates have been created to expect `source:type`, but that field no longer exists because the `FieldNameConverter` was changed.

    ```
    2018-05-22 16:0...[ERROR] Failing 1 tuples
    org.elasticsearch.index.mapper.MapperParsingException: Could not dynamically add mapping for field [source.type]. Existing mapping for [source] must be of type object but found [keyword].
    	at org.elasticsearch.index.mapper.DocumentParser.getDynamicParentMapper(DocumentParser.java:876) ~[stormjar.jar:?]
    	at org.elasticsearch.index.mapper.DocumentParser.parseValue(DocumentParser.java:596) ~[stormjar.jar:?]
    	at org.elasticsearch.index.mapper.DocumentParser.innerParseObject(DocumentParser.java:396) ~[stormjar.jar:?]
    	at org.elasticsearch.index.mapper.DocumentParser.parseObjectOrNested(DocumentParser.java:373) ~[stormjar.jar:?]
    	at org.elasticsearch.index.mapper.DocumentParser.internalParseDocument(DocumentParser.java:93) ~[stormjar.jar:?]
    	at org.elasticsearch.index.mapper.DocumentParser.parseDocument(DocumentParser.java:66) ~[stormjar.jar:?]

    ```

## Pull Request Checklist

### DIFF
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/FieldNameConverters.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/field/FieldNameConverters.java
@@ -15,10 +15,34 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.metron.common.interfaces;
 
-public interface FieldNameConverter {
+package org.apache.metron.common.field;
 
-    String convert(String originalField);
+/**
+ * Allows the field name converter to be specified using a short-hand
+ * name, rather than the entire fully-qualified class name.
+ */
+public enum FieldNameConverters {
 
+  /**
+   * A {@link FieldNameConverter} that does not rename any fields.  All field
+   * names remain unchanged.
+   */
+  NOOP(new NoopFieldNameConverter()),
+
+  /**
+   * A {@link FieldNameConverter} that replaces all field names containing dots
+   * with colons.
+   */
+  DEDOT(new DeDotFieldNameConverter());
+
+  private FieldNameConverter converter;
+
+  FieldNameConverters(FieldNameConverter converter) {
+    this.converter = converter;
+  }
+
+  public FieldNameConverter get() {
+    return converter;
+  }
 }
--- a/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
+++ b/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
@@ -1,0 +1,155 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.metron.elasticsearch.writer;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import org.apache.commons.lang.ClassUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.apache.metron.common.configuration.writer.WriterConfiguration;
+import org.apache.metron.common.field.DeDotFieldNameConverter;
+import org.apache.metron.common.field.FieldNameConverter;
+import org.apache.metron.common.field.FieldNameConverters;
+import org.apache.metron.common.field.NoopFieldNameConverter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.invoke.MethodHandles;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A {@link FieldNameConverterFactory} that is backed by a cache.
+ *
+ * <p>Each sensor type can use a different {@link FieldNameConverter} implementation.
+ *
+ * <p>The {@link WriterConfiguration} allows a user to define the {@link FieldNameConverter}
+ * that should be used for a given sensor type.
+ *
+ * <p>The {@link FieldNameConverter}s are maintained in a cache for a fixed period of time
+ * after they are created.  Once they expire, the {@link WriterConfiguration} is used to
+ * reload the {@link FieldNameConverter}.
+ *
+ * <p>The user can change the {@link FieldNameConverter} in use at runtime. A change
+ * to this configuration is recognized once the old {@link FieldNameConverter} expires
+ * from the cache.
+ *
+ * <p>Defining a shorter expiration interval allows config changes to be recognized more
+ * quickly, but also can impact performance negatively.
+ */
+public class CachedFieldNameConverterFactory implements FieldNameConverterFactory {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  /**
+   * A cache that contains a {@link FieldNameConverter} for each sensor type.
+   *
+   * A user can alter the {@link FieldNameConverter} for a given sensor at any time
+   * by altering the Indexing configuration.  The actual {@link FieldNameConverter}
+   * in use for a given sensor will only change once the original converter has
+   * expired from the cache.
+   */
+  private Cache<String, FieldNameConverter> fieldNameConverters;
+
+  /**
+   * Creates a {@link CachedFieldNameConverterFactory}.
+   *
+   * @param expires The duration before {@link FieldNameConverter}s are expired.
+   * @param expiresUnits The units before {@link FieldNameConverter}s are expired.
+   */
+  public CachedFieldNameConverterFactory(int expires, TimeUnit expiresUnits) {
+
+    fieldNameConverters = createFieldNameConverterCache(expires, expiresUnits);
+  }
+
+  /**
+   * Creates a {@link CachedFieldNameConverterFactory} where the cache expires after 5 minutes.
+   */
+  public CachedFieldNameConverterFactory() {
+
+    this(5, TimeUnit.MINUTES);
+  }
+
+  /**
+   * Creates a {@link CachedFieldNameConverterFactory} using the given cache.  This should only
+   * be used for testing.
+   *
+   * @param fieldNameConverters A {@link Cache} containing {@link FieldNameConverter}s.
+   */
+  public CachedFieldNameConverterFactory(Cache<String, FieldNameConverter> fieldNameConverters) {
+
+    this.fieldNameConverters = fieldNameConverters;
+  }
+
+  /**
+   * Creates a cache of {@link FieldNameConverter}s, one for each source type.
+   *
+   * @return A cache of {@link FieldNameConverter}s.
+   */
+  private Cache<String, FieldNameConverter> createFieldNameConverterCache(int expire, TimeUnit expireUnits) {
+
+    return Caffeine
+            .newBuilder()
+            .expireAfterWrite(expire, expireUnits)
+            .build();
+  }
+
+  /**
+   * Create a new {@link FieldNameConverter}.
+   *
+   * @param sensorType The type of sensor.
+   * @param config The writer configuration.
+   * @return
+   */
+  @Override
+  public FieldNameConverter create(String sensorType, WriterConfiguration config) {
+
+    return fieldNameConverters.get(sensorType, (s) -> createInstance(sensorType, config));
+  }
+
+  /**
+   * Create a new {@link FieldNameConverter}.
+   *
+   * @param sensorType The type of sensor.
+   * @param config The writer configuration.
+   * @return
+   */
+  private FieldNameConverter createInstance(String sensorType, WriterConfiguration config) {
+
+    // default to the 'DEDOT' field name converter to maintain backwards compatibility
+    FieldNameConverter result = new DeDotFieldNameConverter();
+
+    // which field name converter should be used?
+    String converterName = config.getFieldNameConverter(sensorType);
+    if(StringUtils.isNotBlank(converterName)) {
+
+      try {
+        result = FieldNameConverters.valueOf(converterName).get();
+
+      } catch(IllegalArgumentException e) {
+        LOG.error("Unable to create field name converter, using default; value={}, error={}",
+                converterName, ExceptionUtils.getRootCauseMessage(e));
+      }
+    }
+
+    LOG.debug("Created field name converter; sensorType={}, configuredName={}, class={}",
+            sensorType, converterName, ClassUtils.getShortClassName(result, "null"));
+
+    return result;
+  }
+}
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/IndexingConfigurations.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/IndexingConfigurations.java
@@ -31,6 +31,7 @@ public class IndexingConfigurations extends Configurations {
   public static final String ENABLED_CONF = "enabled";
   public static final String INDEX_CONF = "index";
   public static final String OUTPUT_PATH_FUNCTION_CONF = "outputPathFunction";
+  public static final String FIELD_NAME_CONVERTER_CONF = "fieldNameConverter";
 
   public Map<String, Object> getSensorIndexingConfig(String sensorType, boolean emptyMapOnNonExistent) {
     Map<String, Object> ret = (Map<String, Object>) getConfigurations().get(getKey(sensorType));
@@ -61,7 +62,8 @@ public class IndexingConfigurations extends Configurations {
   }
 
   public Map<String, Object> getSensorIndexingConfig(String sensorType, String writerName) {
-    Map<String, Object> ret = (Map<String, Object>) getConfigurations().get(getKey(sensorType));
+    String key = getKey(sensorType);
+    Map<String, Object> ret = (Map<String, Object>) getConfigurations().get(key);
     if(ret == null) {
       return new HashMap();
     }
@@ -147,6 +149,10 @@ public class IndexingConfigurations extends Configurations {
     return getOutputPathFunction(getSensorIndexingConfig(sensorName, writerName), sensorName);
   }
 
+  public String getFieldNameConverter(String sensorName, String writerName) {
+    return getFieldNameConverter(getSensorIndexingConfig(sensorName, writerName), sensorName);
+  }
+
   public static boolean isEnabled(Map<String, Object> conf) {
     return getAs( ENABLED_CONF
                  ,conf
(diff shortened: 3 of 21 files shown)


metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
```diff
@@ -0,0 +1,155 @@
...
+   * Create a new {@link FieldNameConverter}.
+   *
+   * @param sensorType The type of sensor.
+   * @param config The writer configuration.
+   * @return
+   */
+  private FieldNameConverter createInstance(String sensorType, WriterConfiguration config) {
+
+    // default to the 'DEDOT' field name converter to maintain backwards compatibility
+    FieldNameConverter result = new DeDotFieldNameConverter();
```
mmiklavc: Might we want to call the enum when we create these objects?

```
FieldNameConverters.DEDOT.get();
```

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
```diff
@@ -0,0 +1,155 @@
...
+ * <p>Each sensor type can use a different {@link FieldNameConverter} implementation.
+ *
+ * <p>The {@link WriterConfiguration} allows a user to define the {@link FieldNameConverter}
+ * that should be used for a given sensor type.
+ *
+ * <p>The {@link FieldNameConverter}s are maintained in a cache for a fixed period of time
+ * after they are created.  Once they expire, the {@link WriterConfiguration} is used to
+ * reload the {@link FieldNameConverter}.
+ *
+ * <p>The user can change the {@link FieldNameConverter} in use at runtime. A change
```
mmiklavc: I want to say that this is handled through the ConfigurationsUpdater classes already already. I believe the writer passes in an updated config on each call.

https://github.com/apache/metron/blob/master/metron-platform/metron-common/src/main/java/org/apache/metron/common/zookeeper/configurations/ConfigurationsUpdater.java

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
nickwallen: Sure, that makes sense.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
nickwallen: Maybe I am misunderstanding you, but the ConfigurationsUpdater ensures that our configuration classes stay in-sync with Zk.  That's now what this class does. 

This maintains a `FieldNameConverter` instance for each source type that is then used by the `ElasticsearchWriter` to do the field name conversion.  Are you proposing a different way to do this?

nickwallen: To help clarify, I could have implemented a different `FieldNameConverterFactory`; maybe one called `SimpleFieldNameConverterFactory`.  

This factory wouldn't even have to use a cache.  Every time `ElasticsearchWriter` calls `create(String sensorType, WriterConfiguration config)`, it could just instantiate a new instance of the right `FieldNameConverter`.  

But of course, I didn't do that because this occurs for every field in every message that is written to Elasticsearch.  The cache is there for performance.

I offer this only as an example to help clarify the purpose of this class.

mmiklavc: I think I'm just not fully grokking why we need additional caching here rather than `FieldNameConverters.DEDOT.get();`. It seems like the enum handles caching by virtue of the constructors in the enum itself? `NOOP(new NoopFieldNameConverter())`

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * Create a new {@link FieldNameConverter}.
+   *
+   * @param sensorType The type of sensor.
+   * @param config The writer configuration.
+   * @return
+   */
+  private FieldNameConverter createInstance(String sensorType, WriterConfiguration config) {
+
+    // default to the 'DEDOT' field name converter to maintain backwards compatibility
```
cestella: This looks interesting, but one bit of functionality has changed as part of doing this PR.  Currently, we specify the field converter at the writer implementation level, so:
* ES uses dedot
* HDFS uses noop
* solr uses noop

By doing this, we're actually not maintaining backwards compatibility, we're changing the behavior for the HDFS writer and Solr to dedot.  What I'd suggest doing is adding a method to the `BulkMessageWriter` interface like so:
```
default FieldNameConverter defaultFieldNameConverter() {
  return FieldNameConverters.NOOP.get();
}
```
and in ElasticsearchWriter specify DEDOT as the default.  Also, here, you probably want to pass in the default field name converter if unspecified as a 3rd argument.

This would allow us to maintain backwards compatibility and enable users to override.

nickwallen: Remember that only the `ElasticsearchWriter` does field name conversion.  HDFS and Solr do not do field name conversion and so are backwards compatible.  

In the README, I think I specified that this field is only applicable for 'elasticsearch'.

In the future, we could certainly role this functionality out to the other Writers, but right now its not there. AFAIK

cestella: OH! yes, you're right.  Sorry for jumping to conclusions, I assumed you generalized this to the bulk message writer level, rather than keeping it at the ES writer level.  My bad, I retract the objection.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
nickwallen: So you're saying, hey why not just...
1. Take out the caching and just have something like a `SimpleFieldNameConverterFactory`. This would translate the (sourceType, WriterConfiguration) to the right `FieldNameConverter`.
2. The caching would be handled by `FieldNameConverters` in the sense that they are effectively singletons.

I think that would work if all `FieldNameConverter` implementations are stateless.  And they all are stateless today, but I tried not to make that a constraint.  I actually tried to make this work whether they are stateless or not.  

But alas there is a bug in what I did because I need each call to FieldNameConverters.get() to return a new instance.

I'd be open to removing the cache, if its not needed and we think `FieldNameConverter`s would always remain stateless.

mmiklavc: I definitely appreciate the thinking around state though I think we can safely leave that out for now. This isn't client-facing, so refactoring later would have less consequence than something we expose via config or the UI.

I think you could probably use the enum as a factory directly and use FieldNameConverter converter = FieldNameConverters.valueOf(converterName).get() in ElasticsearchWriter line 80 directly. We do this a number of places as a combination strategy/factory pattern. 

https://github.com/apache/metron/blob/master/metron-platform/metron-common/src/main/java/org/apache/metron/common/utils/CompressionStrategies.java

https://github.com/apache/metron/blob/master/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/UnifiedEnrichmentBolt.java#L194

I definitely get wanting to provide a factory, and I think you effectively get that with the enum you already created.

nickwallen: I still need a place to house the (sourceType, WriterConfiguration) -> FieldNameConverter logic.  I will remove the cache mechanism though.

nickwallen:
FYI - I also confirmed that the configuration value can be changed using the Advanced mode in the Management UI.

screen shot 2018-05-25 at 11 31 17 am

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
mmiklavc: I actually liked what you had with your enum pattern. I was imaging a very minor change to your code, like below. The other change would be having the default converter value provided by the config class, or making it an Optional<String> if you wanted the es writer to own setting the default:

```
public interface FieldNameConverter {
    String convert(String originalField);
}

public enum FieldNameConverters implements FieldNameConverter {

  NOOP(new NoopFieldNameConverter()),
  DEDOT(new DeDotFieldNameConverter());

  private FieldNameConverter converter;

  FieldNameConverters(FieldNameConverter converter) {
    this.converter = converter;
  }

  @Override
  public String convert(String originalField) {
    return converter.convert(originalField);
  }
}

public class ElasticsearchWriter implements BulkMessageWriter<JSONObject>, Serializable {
    write( ... ) {
        String converterType = config.getFieldNameConverter(sensorType);
        // or this -> String converterType = config.getFieldNameConverter(sensorType).orElse("DEDOT");
        // fetch the field name converter for this sensor type
        FieldNameConverter fieldNameConverter = FieldNameConverters.valueOf(converterType);
        final String indexPostfix = dateFormat.format(new Date());
        BulkRequestBuilder bulkRequest = client.prepareBulk();
        for(JSONObject message: messages) {
          JSONObject esDoc = new JSONObject();
          for(Object k : message.keySet()){
            copyField(k.toString(), message, esDoc, fieldNameConverter);
          }
          ...
        }
        ...
    }

    private void copyField(... FieldNameConverter fieldNameConverter) {
         String destinationFieldName = fieldNameConverter.convert(sourceFieldName);
        ...
    }
...
}

```

nickwallen: The primary reason the logic was extracted into a separate class is because the `ElasticsearchWriter` cannot be easily unit tested as it stands today.  I ran into tons of library conflicts when trying to do that.    

Keeping it separate also will make it easier to push out similar logic to other writers in the future; HDFS, Solr, should that be needed.

nickwallen: The prior commit 'b18c3bf' was just me fixing my original caching implementation, so at least I had a version of that with the bug fixed.

The latest commit 'cee3994' removes the cache and relies on the `FactoryNameConverters` class to hand out shared instances.  This is what we have discussed.

mmiklavc: Yes, agreed on the possible need for handling other writers. Is there a reason we wouldn't put that logic in the enum factory rather than creating another new class for it? It wouldn't even need to depend on the WriterConfiguration, actually.

```
public enum FieldNameConverters implements FieldNameConverter {

  NOOP(new NoopFieldNameConverter()),
  DEDOT(new DeDotFieldNameConverter());

  private FieldNameConverter converter;

  FieldNameConverters(FieldNameConverter converter) {
    this.converter = converter;
  }

  public FieldNameConverter create(String sensorType, Optional<String> converterName) {
    // default to the 'DEDOT' field name converter to maintain backwards compatibility
   return FieldNameConverters.valueOf(converterName.orElse("DEDOT"));
  }

  @Override
  public String convert(String originalField) {
    return converter.convert(originalField);
  }
}
```

I'm not sure I follow about the testing - do you mean integration testing? We have an ElasticsearchWriterTest class that @justinleet wrote - https://github.com/apache/metron/blob/master/metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriterTest.java

nickwallen: > I'm not sure I follow about the testing

If you look closely at those tests, all that class tests is the response using `buildWriteReponse`.  It does not actually test the class using the `write` method.  When you try to do that, it blows up because of dependency issues.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/SharedFieldNameConverterFactory.java
```diff
@@ -0,0 +1,72 @@
...
+import java.lang.invoke.MethodHandles;
+
+/**
+ * A {@link FieldNameConverterFactory} that returns instances of {@link FieldNameConverter}
+ * that are shared and reused.
+ *
+ * <p>The instances are created and managed by the {@link FieldNameConverters} class.
+ */
+public class SharedFieldNameConverterFactory implements FieldNameConverterFactory {
+
```
cestella: This looks like it's moving in the right direction, but as @mmiklavc gave an example for, why would we not replace this class with a `static create(String sensorType, WriterConfig config)` method on the enum?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/CachedFieldNameConverterFactory.java
mmiklavc: Ok, I see what you mean. Agreed, it would be good to have some tests around that `write` method at some point, but I think it's ok to keep that out of scope for this particular PR. We have a WriterBoltIntegrationTest, but it's only covering parsers right now. At a later time we could look at expanding it to cover multiple writers and topologies and provide hooks on those builders to enable one-time setup and teardown across a suite of tests.

nickwallen: I can add the `create` to the `FieldNameConverters`.  That makes sense here.  Its going to be slightly different than what you have above, but is what you were going for.

nickwallen:
All review feedback should have been addressed.  Let me know if there is anything else @mmiklavc @cestella

metron-platform/metron-common/src/main/java/org/apache/metron/common/field/FieldNameConverters.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Enumerates a set of {@link FieldNameConverter} implementations.
+ *
+ * <p>Provides shared instances of each {@link FieldNameConverter}.
+ *
+ * <p>Allows the field name converter to be specified using a short-hand
+ * name, rather than the entire fully-qualified class name.
+ */
+public enum FieldNameConverters {
```
cestella: Minor nit, can we have this implement `FieldNameConverter` so the enum can be used directly without calling `get()`?

nickwallen: Ok.  Check 'er out.

cestella:
Ok, I'm comfortable with the change now.  +1 by inspection, pending @mmiklavc

PR #1036, opened 2018-05-29
METRON-1584 Indexing Topology Crashes with Invalid Message

nickwallen:
If the indexing topology receives a message containing invalid JSON, the topology will crash.  The topology needs to handle these types of errors and continue processing.

## Changes

* Updated the `BulkWriterComponent` to provide a means for erring out tuples that do not contain valid JSON messages.

* Updated `BulkMessageWriterBolt` to handle the case where a tuple does not contain a valid JSON message.

* Updated `BulkMessageWriterBoltTest` to validate that error'd messages are being ack'd and emitted on the correct error stream.  Previously a mock was being used that hid a bug when erring out tuples with the `BulkWriterComponent` that do not contain a valid JSON message.

* Added a unit test for the case where an invalid message is sent as input to the indexing topology.

## Testing

1. Spin-up the development environment.

1. Ensure that messages are successfully being indexing by running the Service Check and/or validating the Alerts UI.

1. Change the Indexing error topic to "indexing_errors" and restart the Indexing topology.

    screen shot 2018-05-29 at 3 11 33 pm

1. Write an invalid message to the "indexing" topic.

   ```
   export PATH=$PATH:/usr/hdp/current/kafka-broker/bin/
   source /etc/default/metron
   echo "this is a bad message" | kafka-console-producer.sh --broker-list $BROKERLIST --topic indexing
   ```

1. Wait until the invalid message gets picked-up by the Indexing topology, then ensure the invalid message is sent to the "error" topic.

   ```
   kafka-console-consumer.sh --zookeeper node1:2181 --topic indexing_errors --from-beginning
   ```

1. Ensure that an error is reported in the Storm UI.

    screen shot 2018-05-29 at 3 24 33 pm

1. Ensure the indexing topology did not crash and messages continue to be indexed.

## Pull Request Checklist

### DIFF
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
@@ -42,6 +42,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
+import static java.lang.String.format;
+
 /**
  * This component implements message batching, with both flush on queue size, and flush on queue timeout.
  * There is a queue for each sensorType.
@@ -116,17 +118,13 @@ public class BulkWriterComponent<MESSAGE_T> {
   }
 
   public void error(String sensorType, Throwable e, Iterable<Tuple> tuples, MessageGetStrategy messageGetStrategy) {
-    if(!Iterables.isEmpty(tuples)) {
-      LOG.error(String.format("Failing tuples; count=%d, error=%s",
-              Iterables.size(tuples), ExceptionUtils.getRootCauseMessage(e)), e);
-    }
-    tuples.forEach(t -> collector.ack(t));
+    LOG.error(format("Failing %d tuple(s); sensorType=%s", Iterables.size(tuples), sensorType), e);
     MetronError error = new MetronError()
             .withSensorType(sensorType)
             .withErrorType(Constants.ErrorType.INDEXING_ERROR)
             .withThrowable(e);
     tuples.forEach(t -> error.addRawMessage(messageGetStrategy.get(t)));
-    ErrorUtils.handleError(collector, error);
+    handleError(tuples, error);
   }
 
   /**
@@ -139,14 +137,21 @@ public class BulkWriterComponent<MESSAGE_T> {
    * @param tuples The tuples to error that may not contain valid messages.
    */
   public void error(Throwable e, Iterable<Tuple> tuples) {
-    if(!Iterables.isEmpty(tuples)) {
-      LOG.error(String.format("Failing tuples; count=%d, error=%s",
-              Iterables.size(tuples), ExceptionUtils.getRootCauseMessage(e)), e);
-    }
-    tuples.forEach(t -> collector.ack(t));
+    LOG.error(format("Failing %d tuple(s)", Iterables.size(tuples)), e);
     MetronError error = new MetronError()
             .withErrorType(Constants.ErrorType.INDEXING_ERROR)
             .withThrowable(e);
+    handleError(tuples, error);
+  }
+
+  /**
+   * Errors a set of tuples.
+   *
+   * @param tuples The tuples to error.
+   * @param error
+   */
+  private void handleError(Iterable<Tuple> tuples, MetronError error) {
+    tuples.forEach(t -> collector.ack(t));
     ErrorUtils.handleError(collector, error);
   }
 
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
@@ -19,6 +19,7 @@
 package org.apache.metron.writer;
 
 import com.google.common.collect.Iterables;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.metron.common.Constants;
 import org.apache.metron.common.configuration.writer.WriterConfiguration;
 import org.apache.metron.common.error.MetronError;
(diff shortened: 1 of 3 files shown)


metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
```diff
@@ -115,15 +116,37 @@ public void commit(BulkWriterResponse response) {
...
+   * <p>Without a valid message, the JSON message cannot be added to the error.
+   *
+   * @param e The exception that occurred.
+   * @param tuples The tuples to error that may not contain valid messages.
+   */
+  public void error(Throwable e, Iterable<Tuple> tuples) {
+
     if(!Iterables.isEmpty(tuples)) {
-      LOG.error("Failing {} tuples", Iterables.size(tuples), e);
+      LOG.error("Failing tuples; count={}, error={}", Iterables.size(tuples), ExceptionUtils.getRootCauseMessage(e));
```
cestella: Does `ExceptionUtils.getRootCauseMessage()` return the whole stack trace or just the root message?  I think we'd like to keep the stack trace here.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
```diff
@@ -115,15 +116,37 @@ public void commit(BulkWriterResponse response) {
...
     }
-    tuples.forEach(t -> error.addRawMessage(messageGetStrategy.get(t)));
+    tuples.forEach(t -> collector.ack(t));
```
cestella: wow, we did we really not ack here before?  That's a bug. good catch.

mmiklavc: Wait, @cestella isn't it already acked in line 123?

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
```diff
@@ -213,60 +214,144 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
+  }
+
+  /**
+   * Retrieves the JSON message contained in a tuple.
+   *
+   * @param tuple The tuple containing a JSON message.
+   * @return The JSON message contained in the tuple. If none, returns null.
+   */
+  private JSONObject getMessage(Tuple tuple) {
+
```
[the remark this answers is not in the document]
nickwallen: What newlines?  Can you refer to line #s?

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
nickwallen: We were not logging the the exception message or stack trace at all before; just "Failing 8 tuples".  

The error method was not being used as intended AFAIK.  Having the exception as the last parameter was just doing a `{}` substitution for a `{}` that did not exist.

I can change this to log the stack trace, if you like.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
cestella: AH yes, it is, but this is a new error method.  It feels that both `error` method should be calling another `error`method to do the basics like `ack()`.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/bolt/BulkMessageWriterBolt.java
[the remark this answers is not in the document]
cestella: Lines 284, 288, 292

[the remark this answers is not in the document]
nickwallen: I'll remove the leading whitespace in each method if that helps like 284, 305, etc.  

But 288 and 292 are intentional.  They make it much easier to read IMO.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
cestella: Ugh, yeah, so I think we refactored at some point to move to parameterized logging statements and ran into [this](https://www.slf4j.org/faq.html#paramException).

Perhaps we can use `LOG.error(String.format("Failing tuples; count=%d, error=%s", ...), e);` and get the best of both worlds?

nickwallen: Sure will do.

mmiklavc:
I want to take a step back a moment here. I think there are some useful, clarifying refactorings in this PR. I'm generally in favor of refactoring improvements e.g. "handleTick(tuple)" that sum up the code nicely. I also like adding sensible clarifying comments and making tests more clear. However, I'm not sure how I feel about mixing refactoring like this with the bug fix itself. It seems to me that the actual bug fix is roughly 5-10 lines of code. It makes it more difficult to determine code that's specific to the change in mind. I noticed this while reviewing https://github.com/apache/metron/pull/1015 as well. Might it be better to do separate refactoring PR's? I am certainly not suggesting we be pedantic about it - I'm not even clear how I would even suggest a reasonable rule to follow, to be honest. Part of me thinks "ok, broken window theory and boyscout rules - this is great stuff." But in general, I think that splitting the refactoring will allow reviewers to voice opinions specific to coding style and refactoring changes as distinct from a feature or bug fix itself. It's clear that we all have varying opinions on what reads best in our code, and I think splitting those discussions off will make the review process more expedient.

cestella:
Yeah, I have mixed reactions.  I think refactoring opportunistically is damned useful, but if it becomes excessive it can make it hard to review and make it hard for those with open branches in the same codebase.

This seems like a broader discussion is merited, though.  Can we maybe move this to a mailing list DISCUSS thread?  This feels like one of those things we should hash out and should end up in the dev guidelines.

nickwallen:
Undid refactors.  This should include only what is needed.

merrimanr:
I ran this up in full dev and everything worked as expected.  +1 pending approval from other reviewers.

nickwallen:
@mmiklavc Are you good with this?

mmiklavc:
+1

PR #1037, opened 2018-05-30
METRON-1547: Solr Comment Fields

justinleet:
## Contributor Comments
Adding support for comments on Alerts in Solr.  Nested objects are absurdly painful, so this follows the ES expectation of these not needing to be queryable.

Solr doesn't have a great way to store these as-is, so they're stored as raw JSON, which is converted appropriately to ensure everything sees the same data as in the ES case.

The REST API has been adjusted to have add/remove comment calls, rather than just patching (since that won't work in Solr directly). However, index is required to avoid an error for Solr right now as a result of METRON-1585.  Once that is fixed, index will be optional.

There are currently a couple bugs that I would love some help with.

-  ~In ES, when you add a comment, it doesn't immediately show up on the comments list.  It looks like a findOne is fired off, but I assume ES isn't updated yet.  This doesn't appear to happen in Solr.~ Preexisting, related to UI's handling of new comments.
- ~When I hit delete on a comment in the UI (at least in Solr, unsure about ES), the comment is visually removed, but no remove request is fired off. I assume I have something misaligned, but I'm not sure what.~  Appears to be a browser issue, as it hasn't been reproducible and there's no code path difference between ES and Solr in this case.
- ~I patterned the optional index off of findOne, because it handles it a bit specially.  Delightfully, now I have an extra indexString field showing up in both findOne and the add remove comment.  There's some JsonGetter annotation thing I'm probably not using correctly.  This also happens in findOne currently, so it looks like it's probably a matter of getting the right fix and updating this to match.~ Index removed entirely, so this is moot.

To test,
This can be tested both via REST and the UI.  As noted above, at least right now the UI has more instability associated with it.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


cestella:
Did you specifically run this up in the UI backed by both solr and ES?

justinleet:
@cestella I have, yes, but it'll need to happen again because there's an issue in ES, and issue in Solr, and a general issue.

The easiest way to go about it is generally to just spin up full dev, make sure it works in ES, swap over to Solr, make sure it works there.

justinleet:
Having spun up #1040, it looks like the rest API indexString weirdness might be outside the scope of this PR, but I'll want to dig in a bit more before promising that.

justinleet:
The ES not updating appears to be a preexisting issue, and I will file another bug.

justinleet:
Comment not appearing created as https://issues.apache.org/jira/browse/METRON-1596

justinleet:
I believe the Travis issues have been resolved by the commit yesterday reverting a log level change in testing.

justinleet:
I haven't been able to reproduce Solr not making the delete call, and given that the code is the same between ES and Solr, I'm inclined to call it a browser issue.

I have spun this back up for both ES and Solr (with the caveat that the default "source.type.field" for ES is broken until master is merged all the way down.

At this point both work, after a minor fix that would result in an off by one error in the UI when deleting.  The UI is functional for both, although with the previously mentioned preexisting latency issue in updating, I do recommend closing and reopening the tab or things don't line up correctly.

At this point, the only issue is the indexString.

justinleet:
@cestella For that indexString thing, we could do a couple things

- Leave it as-is for now, and fix both when they need to be.
- Figure out how to fix it and fix both here.
- Drop index from this query entirely. Given that we have source type, it should be sufficient.

I personally lean towards the last one, unless we either know a fix is imminent or it's easy to do.  Index is pretty extraneous given that we use source type everywhere and we can do a lookup on the source type  anyway.

Any preference?

justinleet:
Kick Travis for cache bug

merrimanr:
I agree with you @justinleet, we should drop the optional index from the query.

cestella:
Yes, we should drop index from the query, agreed.

justinleet:
Index removed from rest calls. I spun up full dev and confirmed both ES and Solr still work as expected, and that the calls do not pass along index anymore.  Feature branch merged in to resolve conflicts.

cestella:
This looks good to me.  +1 by inspection

justinleet:
Closing PR now that this is merged into the feature branch.

PR #1040, opened 2018-05-31
METRON-1589: '/api/v1/search/search' fails when 'Solr Zookeeper Urls' has comma separated multiple zookeeper urls

justinleet:
## Contributor Comments
Refactored `SolrDao` to kill `getZkHost` and made it `getZkHosts` and return a `List`. `getSolrClient` now takes a `List<String>`, and passes it down.

I also added a `getSolrClient` method that's just blank, and defers to the `getZkHosts` method.  Our calls to `getSolrClient` are generally `solrDao.getSolrClient(solrDao.getZkHosts)`, which seems unnecessary. I did not however, change everything to use that, but I could definitely be persuaded to as part of this.

Additionally, `SolrColumnMetadataDao` now takes a` SolrClient` instead of a Zookeeper host string, to make it consistent with the other DAOs.

There's one catch however: I haven't integration tested this, because I'd need multiple Zk nodes and we don't really have a clean way to get it (neither the `SolrComponent` or `ZkComponent` really let you do it).  I'm not sure if we want or need those tests, or if alternatively we want to spin up a multinode environment and manually test, or if we're fine just assuming things line up (given that it's really testing if the underlying method works as expected at this point). There are unit tests agains the methods to ensure splitting is handled properly in the methods.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
@@ -17,9 +17,9 @@
  */
 package org.apache.metron.solr.dao;
 
+import com.google.common.base.Splitter;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -37,9 +37,9 @@ import org.apache.metron.indexing.dao.search.SearchResponse;
 import org.apache.metron.indexing.dao.update.Document;
 import org.apache.metron.indexing.dao.update.OriginalNotFoundException;
 import org.apache.metron.indexing.dao.update.PatchRequest;
+import org.apache.metron.solr.SolrConstants;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.impl.CloudSolrClient;
-import org.apache.solr.client.solrj.impl.CloudSolrClient.Builder;
 import org.apache.solr.client.solrj.impl.HttpClientUtil;
 import org.apache.solr.client.solrj.impl.Krb5HttpClientConfigurer;
 import org.slf4j.Logger;
@@ -156,8 +156,8 @@ public class SolrDao implements IndexDao {
    */
   public List<String> getZkHosts() {
     Map<String, Object> globalConfig = accessConfig.getGlobalConfigSupplier().get();
-    String solrZookeeper = (String) globalConfig.get("solr.zookeeper");
-    return Arrays.asList(solrZookeeper.split(","));
+    return Splitter.on(',').trimResults()
+        .splitToList((String) globalConfig.getOrDefault("solr.zookeeper", ""));
   }
 
   void enableKerberos() {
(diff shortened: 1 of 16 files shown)


metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
```diff
@@ -134,13 +134,30 @@ public void patch(RetrieveLatestDao retrieveLatestDao, PatchRequest request,
...
+
+  /**
+   * Get ZK hosts from the global config.
+   * @return List of ZkHosts
+   */
+  public List<String> getZkHosts() {
     Map<String, Object> globalConfig = accessConfig.getGlobalConfigSupplier().get();
-    return (String) globalConfig.get("solr.zookeeper");
+    String solrZookeeper = (String) globalConfig.get("solr.zookeeper");
+    return Arrays.asList(solrZookeeper.split(","));
```
cestella: Have you tried this when people put a space between the nodes?  Perhaps something like `Lists.newArrayList(Iterables.transform(Splitter.on(",").split(globalConfig.getOrDefault("solr.zookeeper", ""), x -> x.trim()));`


Also, `solr.zookeeper` should be a constant.

merrimanr:
I think `solrDao.getSolrClient()` is easier to understand and if I had to choose I would use that.  Very minor though.

For integration testing I think it will take a non-trivial amount of work to test with a cluster of zk components.  I would be willing to compromise and leave this to multinode testing outside of our CI process.  As far as I know we don't test other components in this manner (ES for example).

justinleet:
@cestella Check out the adjustment for the trim and see how you feel about it.

Where should solr.zookeeper live?  I was a going to drop it in SolrConstants, but discovered it currently lives in SolrWriter, which doesn't feel like the right place for it since it's more widely used.  Do we want to drop it into SolrConstants, and adjust SolrWriter to use that instead?

cestella:
That looks good @justinleet Yeah, I think that key deserves to be in SolrConstants.

+1 by inspection, pending moving the constant.

justinleet:
@cestella Before I merge this in, are you okay with this PR doing this for the existing constant?

https://github.com/apache/metron/pull/1040/commits/c04b9462669bbe26de0ee1f6c7d0d39a348e865b#diff-47dd38427181750f7a4d42b813b28f43R57

I wanted to keep the existing stuff consistent and avoid too much digging into something not entirely related.

Having said that, it seems like we could probably clean up some of configuration constants in a more holistic way, but I'd argue that's outside the scope of this PR.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/writer/SolrWriter.java
```diff
@@ -17,47 +17,46 @@
...
-
-import static org.apache.metron.solr.SolrConstants.SOLR_WRITER_NAME;
-
 public class SolrWriter implements BulkMessageWriter<JSONObject>, Serializable {
 
   public static final String JAVA_SECURITY_CONFIG_PROPERTY = "java.security.auth.login.config";
 
   public enum SolrProperties {
-    ZOOKEEPER_QUORUM("solr.zookeeper"),
+    ZOOKEEPER_QUORUM(SolrConstants.SOLR_ZOOKEEPER),
```
cestella: Yep, this is fine.  +1 by inspection.

justinleet:
Closing PR that's in the feature branch

PR #1043, opened 2018-05-31
METRON-1587: Make collection utility work for HDP search

merrimanr:
## Contributor Comments
This PR improves the metron-solr create_collection.sh and delete_collection.sh scripts by switching to the Solr REST api instead of the Solr CLI tools.  This removes the requirement that Metron is colocated with a Solr client and makes the scripts easier to use with different versions and distributions.  Now only a zookeeper client is required for retrieving a Solr node address.

I refactored the Ambari code a little bit to match the rest of our MPack.  The solr_schema_install and 
solr_schema_delete functions were moved from indexing_master.py to indexing_commands.py because specific, low-level functions are generally in the *_commands.py scripts.  If this is confusing, doesn't make sense or makes reviewing this harder I'm happy to move them back.

I also added Kerberos support to these scripts.  The way I've done it creates a dependency on the HDP Search 3 Mpack but makes configuring Kerberos easy with no configuration required.  I would like to hear what the community thinks about this approach.  If we do decide to make this a dependency I can also remove the Solr zookeeper configuration setting in Ambari.  At that point using Solr is as easy as switching the "Random Access Search Engine" setting.

This has been tested in full dev with Kerberos enabled.  To test this PR, first install HDP Search 3 in full dev (https://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.6.4/bk_solr-search-installation/content/hdp-search30-install-mpack.html).  Kerberos can also be enabled now if desired.  Then in Ambari set "Solr Zookeeper Urls" in the "Index Settings" tab to `node1:2181/solr` and change the "Random Access Search Engine" setting in the Indexing tab to `Solr`.  Now stop the Metron Indexing component, then start it.  Note that restart does not trigger automatic Solr collection creation, matching how ES template loading works.  If we think adding collection creation on restart is desirable I can make that change.  Collections (bro, snort, metaalert, yaf, error) should now be present in Solr at node1:8983.

The scripts can also be tested independently from Ambari.  Just have to set a few environment variables first:
```
export ZOOKEEPER=node1:2181/solr
export SECURITY_ENABLED=true  # if Kerberos is enabled
```
Then the scripts should work when executed directly.  For example, to create the bro collection:
```
$METRON_HOME/bin/create_collection.sh bro
```
and to delete the bro collection:
```
$METRON_HOME/bin/delete_collection.sh bro
```

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_commands.py
@@ -25,6 +25,7 @@ from resource_management.core.exceptions import Fail
 from resource_management.core.logger import Logger
 from resource_management.core.resources.system import Execute, File
 from resource_management.libraries.functions import format as ambari_format
+from resource_management.libraries.functions.format import format
 
 import metron_service
 import metron_security
@@ -85,14 +86,13 @@ class IndexingCommands:
         :return: Dict where key is the name of a collection and the
           value is a path to file containing the schema definition.
         """
-        from params import params
-        return {
-            "bro": params.bro_schema_path,
-            "yaf": params.yaf_schema_path,
-            "snort": params.snort_schema_path,
-            "error": params.error_schema_path,
-            "metaalert": params.meta_schema_path
-        }
+        return [
+            "bro",
+            "yaf",
+            "snort",
+            "error",
+            "metaalert"
+        ]
 
     def is_configured(self):
         return self.__configured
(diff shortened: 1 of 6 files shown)


merrimanr:
As I update the documentation in the Solr README it occurs to me that we should consider installing HDP Search 3 in full dev with our Ansible scripts.  I have a feeling this will end up being the recommendation for using Solr with Metron.  The documentation in the README is a little outdated so I think it's worth figuring this out before I update it so it's accurate and complete.  Do you agree with that?  

Would you be ok doing it in this PR or would you prefer a separate PR?  If separate, it might be easier to update the documentation in the next PR because we'll have a clear installation path at that point (and the README will have to be rewritten anyways).  My preference would be to just do it here but I understand if you want to review them separately.  I guess my point is updating the documentation once makes more sense to me.

justinleet:
@merrimanr There was a bit of discussion around this in #934, and it looks like the conclusion was that since HDP Search didn't support Solr 6 at the time, that we would hold off on it.  Given that HDP Search 3.0 does support Solr 6, it might be worth revisiting that discussion.

Personally, I'm okay with doing full dev off HDP Search (and having Ambari be able to setup config as needed automagically), as long as indexing can still be configured to work with a non HDP Solr instance.

Re: this ticket in particular, I'd do that as a separate PR.  Would we still want to use these scripts as part of that?  If we still want the scripts, I'd want them documented as part of this ticket (even if it's not in the perfect place).

There was also a hesitance on relying on any mpack (internal or external) explicitly in that discussion.  I'm unsure if that objection extended to a full dev or not.  @cestella Can you expand on your POV?

merrimanr:
Ok I'll hold off on that for now and will also hold off on documenting anything Ambari related since it's going to depend on what we decide to do with HDP Search.  For this PR I will document standalone usage of the collection scripts.  Will that work?

merrimanr:
The latest commit adds documentation on script usage to the Solr README.

cestella:
I'm ok with full-dev off of HDP search, but that's a separate ticket than this IMO.

merrimanr:
I created a Jira for adding HDP Search 3 to full dev.

merrimanr:
@justinleet is the documentation I added comprehensive enough for this PR?

PR #1047, opened 2018-06-04
METRON-1599 Allow user to define global property 'source.type.field' in Ambari

nickwallen:
The user can specify the name of the message field containing the source type in the global configuration.  This is used by the Alerts UI.

Currently this necessitates a user using the CLI to add the field. This property should be exposed to the user via Ambari.  The user should be able to define this property directly in Ambari.

## Testing

1. Launch the development environment.  Ensure alerts are visible in the Alerts UI and that the Service Check passes.

1. Open the REPL and validate the current value of the global property 'source.type.field'.  The value here, the default value should be `source:type`.

    ```
    [root@node1 ~]# source /etc/default/metron
    [root@node1 ~]# $METRON_HOME/bin/stellar -z $ZOOKEEPER
    Stellar, Go!
    ...
    [Stellar]>>> globals := CONFIG_GET("GLOBAL")
    {
      ...
      "source.type.field" : "source:type",
      ...
    }
    ```

1. Change the value in Ambari by going to Metron > Configs > REST > Source Type Field Name.

    screen shot 2018-06-18 at 2 33 11 pm

1. After saving the change, Ambari should prompt for the following services to be restarted.

    > 1 Metron Alerts UI, 1 Metron Management UI, 1 Metron REST

    screen shot 2018-06-18 at 2 34 57 pm

1. Restart all affected services.

1. After the services have been restarted, ppen the REPL and validate that the value of the global property 'source.type.field' changed in the global config.

    ```
    [Stellar]>>> globals := CONFIG_GET("GLOBAL")
    {
      ...
      "source.type.field" : "different:source:type:field",
      ...
    }
    ```

## Pull Request Checklist


nickwallen:
FYI - If we standardize on dots, we won't need this PR.  There is an ongoing discussion thread around this.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
```diff
@@ -137,4 +137,10 @@
         <value>yyyy.MM.dd.HH</value>
         <display-name>Elasticsearch Date Format</display-name>
     </property>
+    <property>
+        <name>source_type_field</name>
+        <display-name>Source Type Field</display-name>
+        <description>Name of the message field that contains the source type.</description>
+        <value>source.type</value>
```
merrimanr: Shouldn't the default be "source:type"?

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
```diff
@@ -137,4 +137,10 @@
         <value>yyyy.MM.dd.HH</value>
         <display-name>Elasticsearch Date Format</display-name>
     </property>
+    <property>
```
merrimanr: Do you think It might be more appropriate to have this property in metron-rest-env instead?  Now when I change it I'm prompted to restart all Metron components when really the only one that needs to be restarted is REST and the UIs.  Both UIs depend on REST so they will also be included in the restart list.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/alerts_ui_master.py
```diff
@@ -40,10 +41,9 @@ def install(self, env):
         self.install_packages(env)
 
     def configure(self, env, upgrade_type=None, config_dir=None):
-        print 'configure alerts_ui'
         from params import params
         env.set_params(params)
-
+        metron_service.refresh_configs(params)
```
merrimanr: If we move the property to metron-rest-env then I'm not sure this is necessary.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
cestella: Currently the default in the UI, if unspecified, is `source:type`, so I'd vote keeping it as that.

nickwallen: Ah, I was a little confused by this.  I set it to `source.type` based on the value defined in [`Constants`](https://github.com/apache/metron/blob/32d132b3f8ac7d915180882934d38504263d86ae/metron-platform/metron-common/src/main/java/org/apache/metron/common/Constants.java#L29).  The Enrichment and Indexing topologies use `MessageUtils.getSensorType(msg)` to get the source type of a message.  Right now this will always look for a field called `source.type`.  

But after digging a little more, it seems that this setting is currently only intended to define where the UI looks for the source type in the search indices, not to where Enrichment or Indexing look.  That explains my confusion.

I will update this to use `source:type` as you guys have mentioned and also update the description to clarify that this field is only intended for the UI.

nickwallen: I might also update this so the field only shows up under the "Advanced" settings, since it is not something we expect the user to have to, nor want to change in most cases.  I *think* I can do that easily.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
nickwallen: Yes, assuming this field is applicable only to the REST/UI/DAO, then this makes complete sense.  Will address this.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/alerts_ui_master.py
nickwallen: The call to `refresh_configs` is the only thing that pushes changes to the global config.  I don't see how the setting would be pushed to the global config without it.  Let me know if I am misunderstanding something.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
nickwallen: I addressed this by changing the default to `source:type` and improving the description.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/alerts_ui_master.py
nickwallen: You are correct, of course.  Doh.  I'll drop this.

merrimanr:
I tested this in full dev and it worked as advertised.  +1

PR #1055, opened 2018-06-07
METRON-1608: Add configuration for threat.triage.field name

merrimanr:
## Contributor Comments
This PR adds a configuration to the global config for the `threat.triage.score` field name, similar to what was done in https://github.com/apache/metron/pull/1010 for `source.type`, minus the UI changes.

I also opportunistically fixed a bug where the `source.type` field name wasn't being read from the global config.  Normally this would be a separate PR but the fix overlaps with what was done here so I included it.  I also added a constant for the `source.type.field` property.

## Changes
- Added constants for `source.type.field` and `threat.triage.field` (would `threat.triage.score.field` be better?)
- Added getter/setter for AccessConfig in the ElasticsearchDao (necessary for testing)
- Refactored the default threat triage field name in ElasticsearchMetaAlertDao to match source type field name pattern
- Added a `getField` method that gets the field name from the global config or returns the default field name if not found
- Added unit and integration tests for both `source.type.field` and `threat.triage.field`

## Testing
Testing this in full dev requires changing our indexing topology to use fields with '.'s rather than ':'s.  To do that:

1. Stop the snort sensor with `service sensor-stubs stop snort`
2. Delete the snort ES index with `curl -XDELETE http://node1:9200/snort_index*`
3. Edit the template at `/var/lib/ambari-agent/cache/common-services/METRON/0.5.0/package/files/snort_index.template` by changing `source:type` to `source.type` and `threat:triage:*score` to `threat.triage.*score`
4. Reinstall the template using the "Elasticsearch Template Install" in Ambari
5. Update the global config with the new field names:
```
  "es.clustername": "metron",
  "es.ip": "node1:9300",
  "update.hbase.cf": "t",
  "es.client.settings": {
    "client.transport.ping_timeout": "500s"
  },
  "profiler.client.period.duration": "15",
  "profiler.client.period.duration.units": "MINUTES",
  "user.settings.hbase.table": "user_settings",
  "user.settings.hbase.cf": "cf",
  "bootstrap.servers": "node1:6667",
  "geo.hdfs.file": "/apps/metron/geo/default/GeoLite2-City.mmdb.gz",
"source.type.field":"source.type",
"threat.triage.score.field":"threat.triage.score"
}' 'http://node1:8082/api/v1/global/config'
```
6. Update the snort indexing config to not DEDOT the field names:
```
  "hdfs": {
    "index": "snort",
    "batchSize": 1,
    "enabled": true
  },
  "elasticsearch": {
    "index": "snort",
    "batchSize": 1,
    "enabled": true,
    "fieldNameConverter": "NOOP"
  },
  "solr": {
    "index": "snort",
    "batchSize": 1,
    "enabled": true
  }
}' 'http://node1:8082/api/v1/sensor/indexing/config/snort'
```
7. Start the snort sensor again with `service sensor-stubs start snort`
8.  Wait for data to appear in the snort ES index and then perform a search:
```
"facetFields":[],
  "from": 0,
  "indices": [
    "snort"
  ],
  "query": "*",
  "size": 5
```
9. Grab a guid from one of the snort search results and create a metaalert:
```
  "alerts": [
    {
      "guid": "1c082f1a-ad3a-4e46-ae1e-828b73c9a016",
      "sensorType": "snort"
    }
  ],
  "groups": [
    "string"
  ]
```
10. Now find the metaalert from the guid returned in the previous step:
```
  "guid": "1ad8be85-3164-4be9-a773-379abb044d0a",
  "sensorType": "metaalert"
}' 'http://node1:8082/api/v1/search/findOne'
```

Before this PR there were a couple problems.  First the metaalert had the `source:type` field even though we've switched to "source.type" in the global config.  Second, the `threat.triage.score` field in the metaalert would be 0 because the wrong threat triage field name is used to get scores from alerts.  With this PR the metalaert should correctly have the `source.type` field and the `threat.triage.score` field should be > 0 (assuming the snort alert has a threat triage score):
```
{
  "average": 10,
  "max": 10,
  "threat.triage.score": 10,
  "count": 1,
  "groups": [
    "string"
  ],
  "sum": 10,
  "source.type": "metaalert",
  "min": 10,
  "median": 10,
  "alert": [
    ...
  ]
}
```

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/SearchServiceImpl.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/SearchServiceImpl.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.metron.common.Constants;
+import org.apache.metron.common.configuration.ConfigurationsUtils;
 import org.apache.metron.indexing.dao.IndexDao;
 import org.apache.metron.indexing.dao.search.FieldType;
 import org.apache.metron.indexing.dao.search.GetRequest;
@@ -147,13 +148,9 @@ public class SearchServiceImpl implements SearchService {
     if (!alertUserSettings.isPresent() || alertUserSettings.get().getFacetFields() == null) {
       String facetFieldsProperty = environment
           .getProperty(SEARCH_FACET_FIELDS_SPRING_PROPERTY, String.class, "");
-
-      Map<String, Object> globalConfig = globalConfigService.get();
-      String sourceTypeField = Constants.SENSOR_TYPE.replace('.', ':');
+      String sourceTypeField = ConfigurationsUtils.getFieldName(globalConfigService.get(), SENSOR_TYPE_FIELD_PROPERTY,
+              Constants.SENSOR_TYPE.replace('.', ':'));
       List<String> facetFields = new ArrayList<>();
-      if (globalConfig != null) {
-        sourceTypeField = (String) globalConfig.getOrDefault(SENSOR_TYPE_FIELD_PROPERTY, sourceTypeField);
-      }
       facetFields.add(sourceTypeField);
       if (facetFieldsProperty != null) {
         facetFields.addAll(Arrays.asList(facetFieldsProperty.split(",")));
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/SearchServiceImpl.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/SearchServiceImpl.java
@@ -18,6 +18,7 @@
 package org.apache.metron.rest.service.impl;
 
 import static org.apache.metron.common.Constants.ERROR_TYPE;
+import static org.apache.metron.common.Constants.SENSOR_TYPE_FIELD_PROPERTY;
 import static org.apache.metron.indexing.dao.MetaAlertDao.METAALERT_TYPE;
 import static org.apache.metron.rest.MetronRestConstants.INDEX_WRITER_NAME;
 import static org.apache.metron.rest.MetronRestConstants.SEARCH_FACET_FIELDS_SPRING_PROPERTY;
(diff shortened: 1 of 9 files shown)


metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
```diff
@@ -720,4 +719,12 @@ public int getPageSize() {
   public void setPageSize(int pageSize) {
     this.pageSize = pageSize;
   }
+
+  private String getField(String globalConfigKey, String defaultField) {
```
nickwallen: It seems like we have duplicated this logic here and in `SearchServiceImpl`.  Would it make sense to extract this out?

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/SearchServiceImpl.java
```diff
@@ -151,7 +152,7 @@ public GroupResponse group(GroupRequest groupRequest) throws RestException {
       String sourceTypeField = Constants.SENSOR_TYPE.replace('.', ':');
```
nickwallen: Just another point of duplication.  We define what the default value is here and in `ElasticsearchMetaAlertDao`.  Could be extracted out along with the other bits.

nickwallen:
FYI - There is also another `source.type.field` reference that would be an easy update [here](https://github.com/apache/metron/blob/71a015fabc47b23ea0c5400fa239fee9f5c0d194/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/service/impl/SearchServiceImplTest.java#L186)

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertDao.java
merrimanr: The latest commit extracts this to ConfigurationsUtils class in metron-common.  This class is already really big so I was hesitant to add it there but I'm not sure where else it belongs.

metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/SearchServiceImpl.java
merrimanr: This one is tricky.  We can't just reference ElasticsearchMetaAlertDao.SOURCE_TYPE because metron-elasticsearch is a runtime dependency in metron-rest.  The only other solution I can think of is to define a second constant in Constants but I think that is confusing since it's specific to ES.  I remember @justinleet working through this in another, maybe he can chime in.

merrimanr:
Good feedback @nickwallen.  The latest commit should satisfy your comments except the duplicated SOURCE_TYPE issue.  Not sure what the best approach is so I'm inclined to leave it alone.

nickwallen:
+1 looks good. Thanks

I wouldn't worry about that constant. Not worth it.

PR #1056, opened 2018-06-07
METRON-1609 Elasticsearch settings in Ambari should not be required if Solr is the Indexer

nickwallen:
NOTE: This PR has been submitted against the Solr feature branch; not master.

## Description

  When a user deploys Metron with Solr, the Mpack requires that the user populate the Index Settings > Elasticsearch Hosts field to continue, even when deploying a Metron cluster that indexes to Solr and does not use Elasticsearch.

  * The Elasticsearch related properties under the Ambari 'Index settings' tab should only be required if Elasticsearch is chosen as the indexer.

  * The Solr related properties under the 'Index Settings' tab should only be required if Solr is chosen as the indexer.

## Changes

  * Removed the static validation checks for each of the parameters on the 'Index Settings' tab.  Static validation no longer works because validation depends on which Indexer the user has chosen.

  * Added a function `check_indexing_parameters` that validates the parameters based on the Indexer that has been chosen by the user on the "Indexing" tab.

  * The parameters are checked before any of the following actions:
      * Updating global properties, which occurs when Parsers, Enrichment, Indexing, Profiler, and REST are started/configured in Ambari.
      * Start/Configure indexing
      * Service Check
      * Install Elasticsearch Template
      * Delete Elasticsearch Template
      * Install Kibana Dashboard

  * When updating the global properties, we only add either Solr or Elasticsearch properties, but not both as before. 

  * It seems logical to move the 'Indexing Settings' parameters to the "Indexing"  tab, especially when it on this page where the user selects the Indexer; Elasticsearch or Solr.  I am not sure if there are other implications in doing this, so I did not do that as part of this PR.

## Testing

  1. Start the development environment.

  1. Validate that the Alerts UI is receiving alerts and that the Ambari 'Service Check' passes.

  **Missing Elasticsearch parameters should be caught**

  1. Stop the Metron services in Ambari.

  1. Using Ambari, delete (define a blank value for) the `metron-env/es_binary_port` parameter in Metron > Index Settings.

  1. Attempt to start each of the Metron services; Parsers, Enrichment, Indexing, Profiler, and REST.  These should fail with the following error message.

      ```
      resource_management.core.exceptions.Fail: Missing required indexing parameters(s): indexer=Elasticsearch, missing=['metron-env/es_binary_port']
      ```

  1. Attempt the Elasticsearch Template Install.  This should fail with the same error message.

  1. Attempt the Elasticsearch Template Delete.  This should fail with the same error message.

  1. Attempt the Kibana Dashboard Install.  This should fail with the same error message.

  **Missing Elasticsearch parameters should not impact Solr**

  1. Go to Metron > Indexing and change "Random Access Search Engine" to "Solr".

  1. Attempt to start the Indexing topology in Ambari.  This should succeed even though we are missing Elasticsearch parameters.

  **Ensure Elasticsearch parameters are defined in globals**

   1. Open up the REPL.

   1. Check for the Elasticsearch parameters in the globals.

      ```
      [Stellar]>>> CONFIG_GET("GLOBAL")
      {
        "es.clustername" : "metron",
        "es.ip" : "node1:9300",
        "es.date.format" : "yyyy.MM.dd.HH",
        ...
      }
      ```

  **Missing Solr parameters should be caught**

  1. Stop the Indexing topology in Ambari.

  1. Using Ambari, delete (define a blank value for) the `metron-env/solr_zookeeper_url` parameter in Metron > Index Settings.

  1. Attempt to start the Indexing topology in Ambari.  This should fail with the following error message.

      ```
      resource_management.core.exceptions.Fail: Missing required indexing parameters(s): indexer=Solr, missing=['metron-env/solr_zookeeper_url']
      ```
      
  **Ensure Solr parameters are defined in globals**

   1. Open up the REPL.

   1. Check for the Elasticsearch parameters in the globals.

      ```
      [Stellar]>>> CONFIG_GET("GLOBAL")
      {
        ...
        "solr.zookeeper" : "node1:9983",
        ...
      }
      ```

## Pull Request Checklist


merrimanr:
I spun this up in full dev and the only issue I could find was that starting REST doesn't throw an error when a required parameter is missing.  The action succeeds but the app won't start and Ambari will still report it as down.  Other than that everything else worked.

nickwallen:
Thanks @merrimanr.  Good catch.  I will fix that.

justinleet:
@nickwallen Can you conflict this?

nickwallen:
I deconflicted and pushed some changes to catch any problems when the REST service is started.  I am in the middle of testing those changes now.

nickwallen:
This is ready to review again.  Cheers

justinleet:
I'm +1 by inspection assuming @merrimanr is good.

merrimanr:
I tested this in full dev and went through the testing instructions.  REST now throws an error on restart.  Thanks!  +1

nickwallen:
This has been merged

PR #1060, opened 2018-06-12
METRON-1615 Default endpoint locations based on Ambari service locations and config

simonellistonball:
## Contributor Comments
For certain values, the mpacks will now default values for the following settings:

- location of storm ui
- location of elastic search
- location of zeppelin notebook server

Testing involves spinning up full dev, and ensuring that while the settings es_url, storm_rest_url and zeppelin_url are set blank, the correct values (node1...) are set in the config files on disk. 

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ temp/**
 temp/**/*
 metron-interface/metron-alerts/node/
 repodata/
+
+.pydevproject
--- a/metron-deployment/ansible/roles/ambari_config/vars/single_node_vm.yml
+++ b/metron-deployment/ansible/roles/ambari_config/vars/single_node_vm.yml
@@ -37,11 +37,13 @@ metron_parsers : [METRON_PARSERS]
 metron_rest: [METRON_REST]
 metron_management_ui: [METRON_MANAGEMENT_UI]
 metron_alerts_ui: [METRON_ALERTS_UI]
+zeppelin: [ZEPPELIN_MASTER]
 
 metron_components: >
   {{ hadoop_master | union(zookeeper_master) | union(storm_master) | union(hbase_master) | union(hadoop_slave) | union(zookeeper_slave) |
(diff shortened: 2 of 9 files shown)


mmiklavc:
@simonellistonball - I'm happy to test this out if you can resolve the conflicts.

mmiklavc:
@simonellistonball - Bump

simonellistonball:
Turns out this is (in theory) done by the service advisor at             https://github.com/apache/metron/blob/ff1f9cf52b31ae866eb3de1cf8993af97af1790f/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/service_advisor.py#L137 which doesn't look to have been significantly tested (doesn't for example work on a real cluster, and only works with overrides to blueprint templates to force application of settings (since the required field insists you enter a custom value). That therefore doesn't fire in default Ambari blueprint installs. 

You won't see that on full dev, or a platform that explicitly sets storm_rest_addr in blueprints to a correct value, but on something like a cloudbreak install (where you can't change the ambari configStrategy) it is impossible to get the service advisor to run. Since the fields in question are marked required, the service advisor cannot override them, and the blueprint cannot be submitted without setting them, so you're in a catch 22 if you do not know the host names upfront prior to install (the cloud install case).

In any case, this is probably not a good Apache fix, since from the Apache point of view this already 'works as designed'. Maybe I should file a separate ticket on the removing the required attribute from the settings which are not actually required, or it should be handled by rewriting the automation engine elsewhere, which would allow the service advisor to set them.

mmiklavc:
> In any case, this is probably not a good Apache fix, since from the Apache point of view this already 'works as designed'.

Why would a bug fix for incomplete behavior not be a good fix for Apache?

simonellistonball:
The fix is not necessary in pure Apache Ambari, and is only required in a non-Apache toolchain, hence from a pure Apache perspective 'works as designed' (just not very inclusively designed)

PR #256, opened 2016-09-16
Metron 257 Enable pcap result pagination from the Pcap CLI

mmiklavc:
This builds on efforts from [https://github.com/apache/incubator-metron/pull/217](https://github.com/apache/incubator-metron/pull/217)

This closes [https://issues.apache.org/jira/browse/METRON-257](https://issues.apache.org/jira/browse/METRON-257)

The purpose for this PR is to give the user the ability to specify how many records per file should be written by the PCAP CLI tool. For example, if 1,000 records are returned by a PCAP query and the user specifies 200 records per file, then the user should expect 5 PCAP files to be written to the current working directory.

Note - I tested this on quick-dev

**Testing**
Get PCAP data into Metron: Install and setup pycapa - the instructions below reference/mirror those in [PR-93](https://github.com/apache/incubator-metron/pull/93#issue-151705836)

1. Install the pycapa library & utility ```$ cd /opt/pycapa/pycapa && pip install -r requirements.txt && python setup.py install```
2. (if using singlenode vagrant) Kill the enrichment and sensor topologies via for i in bro enrichment yaf snort;do storm kill $i;done
3. Start the pcap topology via /usr/metron/0.2.0BETA/bin/start_pcap_topology.sh
4. Start the pycapa packet capture producer on eth1 via /usr/bin/pycapa --producer --topic pcap -i eth1 -k node1:6667
5. Watch the topology in the Storm UI and kill the packet capture utility from before, when the number of packets ingested is over 3k.
6. Ensure that at at least 3 files exist on HDFS by running hadoop fs -ls /apps/metron/pcap
7. Choose a file (denoted by $FILE) and dump a few of the contents using the pcap_inspector utility via /usr/metron/0.2.0BETA/bin/pcap_inspector.sh -i $FILE -n 5
8. Choose one of the lines and note the protocol.
9. Note that when you run the commands below, the resulting file will be placed in the execution directory where you kicked off the job from.

### Fixed filter

1. Run a fixed filter query by executing the following command with the values noted above (match your start_time format to the date format provided - default is to use millis since epoch)
2. `/usr/metron/0.2.0BETA/bin/pcap_query.sh fixed -st <start_time> -df "yyyyMMdd" -p <protocol_num> -rpf 500`
3. Verify the MR job finishes successfully. Upon completion, you should see multiple files named with relatively current datestamps in your current directory, e.g. pcap-data-20160617160549737+0000.pcap
[...]

PR #1081, opened 2018-06-27
METRON-1641: Enable Pcap jobs to be submitted asynchronously

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1641

This enables Pcap Jobs to be run asynchronously. The PcapJob class itself is now a Statusable implementation that allows clients to poll for current JobStatus. This implementation exposes the new functionality on the job class but keeps the existing PcapCli functionality intact and unchanged. The tests for this will be in a comment below, taken from https://github.com/apache/metron/pull/256.

This validation should check that the current pcap functionality does not break. Follow on PR's will leverage the new asynchronous capabilities.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/pom.xml
+++ b/metron-platform/pom.xml
@@ -46,6 +46,7 @@
 		<module>metron-enrichment</module>
 		<module>metron-solr</module>
 		<module>metron-parsers</module>
+    <module>metron-job</module>
 		<module>metron-pcap-backend</module>
 		<module>metron-data-management</module>
 		<module>metron-pcap</module>
--- a/metron-platform/metron-common/src/main/java/org/apache/metron/common/utils/HDFSUtils.java
+++ b/metron-platform/metron-common/src/main/java/org/apache/metron/common/utils/HDFSUtils.java
@@ -23,6 +23,7 @@ import java.util.List;
 import org.apache.commons.io.IOUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
@@ -56,4 +57,21 @@ public class HDFSUtils {
     return IOUtils.readLines(inputStream, "UTF-8");
   }
 
+  /**
+   * Write file to HDFS. Writes to local FS if file:// used as protocol.
+   *
+   * @param config filesystem configuration
+   * @param bytes bytes to write
+   * @param path output path
+   * @throws IOException This is the generic Hadoop "everything is an IOException."
+   */
+  public static void write(Configuration config, byte[] bytes, String path) throws IOException {
+    FileSystem fs = FileSystem.newInstance(config);
+    Path outPath = new Path(path);
+    fs.mkdirs(outPath.getParent());
+    try (FSDataOutputStream outputStream = fs.create(outPath)) {
+      outputStream.write(bytes);
+      outputStream.flush();
+    }
+  }
 }
--- a/metron-platform/metron-common/src/test/java/org/apache/metron/common/utils/HDFSUtilsTest.java
+++ b/metron-platform/metron-common/src/test/java/org/apache/metron/common/utils/HDFSUtilsTest.java
@@ -1,0 +1,48 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.metron.common.utils;
+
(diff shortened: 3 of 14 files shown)


metron-platform/pom.xml
```diff
@@ -46,6 +46,7 @@
 		<module>metron-enrichment</module>
 		<module>metron-solr</module>
 		<module>metron-parsers</module>
+    <module>metron-job</module>
```
cestella: can we adjust the indentation here?

mmiklavc: Heh, the pom.xml has tabs instead of spaces. Rather than reformat everything in the file I just changed that line use tabs.

merrimanr:
Looks good!  One thing I'm trying to wrap my head around is how we get status if we only have a job id or unique identifier for a job?  JobStatus doesn't have an id so I'm assuming resultPath is the unique identifier here.  

As far as I can tell an instance of org.apache.hadoop.mapreduce.Job is kept in memory and is responsible for reporting status.  I can think of a couple scenarios where this might be problematic. 
 One is if I ran a query from the CLI but then wanted to get status from REST.  How would that work?  That's probably not a likely use case so maybe not an issue there.  What happens if I submit a query through REST and REST is restarted while jobs are running?  Do we lose job status information?

cestella:
I think for a first cut, it's ok to have the restrictions that:
* the REST API controls only the jobs it creates.  Otherwise, we would need more refactoring in the CLI to drop the output in the same HDFS directory rather than it be user specifiable and output locally.  Ultimately, while they use the same mechanism, the UX is different between the two approaches (e.g. the CLI entirely cleans up after itself and outputs to the local directory whereas the REST approach stores the results in HDFS until manual cleanup).
* If a job is running while the REST API dies, we should consider that job to be runaway and needs to be killed by the admin or left to complete without the result being published.  One thing that we might consider doing is enabling the job naming to have a prefix of `METRON_REST_PCAP` so, upon REST start, it can kill existing jobs.  I think for THIS PR, we should just have REST pcap jobs have that prefix and leave it to a follow-on PR to do the actual killing.

cestella:
One thing I didn't see.  Can we make sure we pass along the yarn queue to the job?

cestella:
One more comment about restartability, I think we could potentially support this with this architecture in the future.  You can recover the Job object from MR via the JobClient
```
Configuration conf = new Configuration();
JobClient jobClient = new JobClient(new JobConf(conf)); // deprecation WARN
JobID jobID = JobID.forName("job_201107011451_0001");   // deprecation WARN
RunningJob runningJob = jobClient.getJob(jobID);
```

We could look for jobs which are completed but not in the HDFS structure and recover them on REST start.  I would suggest doing that as a follow-on though.

merrimanr:
Perfect.  That addresses my concern.  Doing this in a follow on is fine since it's not necessary when using the CLI.

merrimanr:
I spun this up in full dev, ran a fixed query, and got this error:
```
Exception in thread "main" java.lang.IllegalArgumentException: Wrong FS: file://./, expected: hdfs://node1:8020
	at org.apache.hadoop.fs.FileSystem.checkPath(FileSystem.java:666)
	at org.apache.hadoop.hdfs.DistributedFileSystem.getPathName(DistributedFileSystem.java:214)
	at org.apache.hadoop.hdfs.DistributedFileSystem$25.doCall(DistributedFileSystem.java:1181)
	at org.apache.hadoop.hdfs.DistributedFileSystem$25.doCall(DistributedFileSystem.java:1177)
	at org.apache.hadoop.fs.FileSystemLinkResolver.resolve(FileSystemLinkResolver.java:81)
	at org.apache.hadoop.hdfs.DistributedFileSystem.mkdirsInternal(DistributedFileSystem.java:1195)
	at org.apache.hadoop.hdfs.DistributedFileSystem.mkdirs(DistributedFileSystem.java:1169)
	at org.apache.hadoop.fs.FileSystem.mkdirs(FileSystem.java:1925)
	at org.apache.metron.common.utils.HDFSUtils.write(HDFSUtils.java:71)
	at org.apache.metron.pcap.writer.ResultsWriter.write(ResultsWriter.java:38)
	at org.apache.metron.pcap.mr.PcapJob.writeResults(PcapJob.java:270)
	at org.apache.metron.pcap.query.PcapCli.run(PcapCli.java:155)
	at org.apache.metron.pcap.query.PcapCli.main(PcapCli.java:52)
	at org.apache.hadoop.util.RunJar.run(RunJar.java:233)
	at org.apache.hadoop.util.RunJar.main(RunJar.java:148)
```
Looks like the MR jobs succeeded but partitioning the files to the local FS did not work.

mmiklavc:
Good catch on the FS @merrimanr - also finding that via manual testing. I believe I have a workaround that degrades nicely to the configuration default and also allows you to pass in the scheme in the path.

```
FileSystem fs = FileSystem.newInstance(outPath.toUri(), config);
```

mmiklavc:
Also @merrimanr agreed about the points you made regarding restart-ability, etc. in the long run. In the short term, @cestella has done a rather good job of summarizing my thoughts on a v1 pass at this feature set. 

I will:
- add Job ID in the return status (was going to add that when I do the job manager follow-on to this PR, but I'll just do it here)
- add the ability to pass a job name, ie job.setJobName(name) - I'll handle the actual job naming in the pcapservice and pass that as a parameter. I think that's a natural place for that logic.
- Add ability to pass the queue name

How's that sound to you both?

mmiklavc:
Heads up, hadoop config class is where you set queue name iirc. We already pass that in as an arg. This would simply need to be provided via the calling job manager class. `config.setProperty("mapreduce.job.queuename", "somequeue");`

merrimanr:
That sounds good to me.

merrimanr:
I tested this and it's working for me in full dev.  I think it's good enough to go into the feature branch.  +1

merrimanr:
Can we merge this?  Any other items you would like addressed @cestella?

cestella:
+1, lgtm

PR #1082, opened 2018-06-27
METRON-1642: KafkaWriter should be able choose the topic from a field in addition to topology construction time

cestella:
## Contributor Comments
Currently, we choose the kafka topic via the kafka.topic field.  It would be useful to allow people to specify the topic via a field.  This would enable multi-stage (or chain) parsing, among other use-cases.

## Manual Test
Ensure that `ZOOKEEPER`, `BROKERLIST` and `METRON_HOME` are set
* Create a topics for 2 sensors:
```
/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper $ZOOKEEPER --create --topic 1642_pre --partitions 1 --replication-factor 1
/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper $ZOOKEEPER --create --topic 1642_second --partitions 1 --replication-factor 1
```
* Open `$METRON_HOME/config/zookeeper/parsers/1642_pre.json` and save
```
{
  ,"sensorTopic" : "1642_pre"
  , "parserConfig": {
     "columns" : { "name": 0
                 , "first_field": 1
                 , "timestamp" : 2
                 },
     "kafka.topicField" : "route_field"
   }
  ,"fieldTransformations" : [
    {
     "transformation" : "STELLAR"
    ,"input" :  ["name"]
    ,"output" :  ["route_field"]
    ,"config" : {
      "route_field" : "match{ name == 'metron' => '1642_second', default => 'enrichments'}"
                }
    }
                           ]

}
```
* Open `$METRON_HOME/config/zookeeper/parsers/1642_second.json` and save
```
{
  "parserClassName" : "org.apache.metron.parsers.json.JSONMapParser"
  ,"sensorTopic" : "1642_second"
  , "parserConfig": {
   }
  ,"fieldTransformations" : [
    {
     "transformation" : "STELLAR"
    ,"input" :  ["first_field"]
    ,"output" :  ["new_field"]
    ,"config" : {
      "new_field" : "TO_UPPER(first_field)"
                }
    }
                           ]
}
```
* Push configs via `$METRON_HOME/bin/zk_load_configs.sh --mode PUSH -z $ZOOKEEPER -i $METRON_HOME/config/zookeeper/`
* Stop the parsers in ambari
* Start the `1642_pre` and `1642_second` parsers
```
$METRON_HOME/bin/start_parser_topology.sh -k $BROKERLIST -z $ZOOKEEPER -s 1642_pre
$METRON_HOME/bin/start_parser_topology.sh -k $BROKERLIST -z $ZOOKEEPER -s 1642_second
```
* Open `~/data.json` and save:
```
notmetron,f1,1
metron,f2,2
metron,f3,3
metron,f4,4
metron,f5,5
metron,f6,6
metron,f7,7
notmetroneither,f8,8
```
* Send data into the `1642_pre` topic via
```
cat data.json | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list $BROKERLIST --topic 1642_pre
```

From here you should see data flowing into the `1642_pre` and `1642_second` indices depending on the first field.  When the field is `metron`, then it should go into the `1642_second` index and a new field `new_field` should be created.  Otherwise, it will go in as-is in the `1642_pre` index.
## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


cestella:
Ok, the manual test script is ready for this and can be reviewed.

metron-platform/metron-parsers/README.md
```diff
@@ -166,6 +166,15 @@ then it is assumed to be a regex and will match any topic matching the pattern (
   * `batchTimeout` : The timeout after which a batch will be flushed even if batchSize has not been met.  Optional.
     If unspecified, or set to `0`, it defaults to a system-determined duration which is a fraction of the Storm
     parameter `topology.message.timeout.secs`.  Ignored if batchSize is `1`, since this disables batching.
+  * The kafka writer can be configured within the parser config as well.  (This is all configured a priori, but this is convenient for overriding the settings)  :
```
nickwallen: The ability to route messages based on a message field applies for all topologies that use a `KafkaWriter`, not just the Parsers, right?  That would include Enrichment and Profiler?

This is documented under the Parsers.  Would it be worth mentioning that the same settings would work in any topology that uses a `KafkaWriter` although it may not be advised.

It might make sense to put these docs in `metron-writer` (with the `KafkaWriter` class) and then link to those docs from the Parser docs here.  Then in `matron-writer` it would make sense to mention that this potentially could be used by any topology that uses a `KafkaWriter`.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/kafka/KafkaWriter.java
```diff
@@ -197,6 +209,15 @@ public void init(Map stormConf, TopologyContext topologyContext, WriterConfigura
     return producerConfig;
   }
 
+  public Optional<String> getKafkaTopic(JSONObject message) {
```
nickwallen: I am assuming this logic applies to any topology that uses a `KafkaWriter`.  Would it be easy enough for a user to run into an infinite loop scenario if they have any two sequential topologies both using a `KafkaWriter`?

Parser -> Stage1 -> Stage2 -> ...

* In the Parser, it ingests a message where the "outputTopic" = "stage1".  
* This sends the message to my Stage1 processing
* If the Stage1 logic does not change the value of that field for whatever reason, then the message will go right back to Stage1 and be reprocessed.
* Wash, rinse, repeat and you've got a mess on your hands that is difficult to debug.

Maybe I am thinking too hard about this.  There may be nothing we can really do about that.  With power comes responsibility.  If an advanced user wants to customize routing, then they need to own this risk.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/kafka/KafkaWriter.java
```diff
@@ -197,6 +209,15 @@ public void init(Map stormConf, TopologyContext topologyContext, WriterConfigura
...
+    if(kafkaTopicField != null) {
+      return Optional.ofNullable((String)message.get(kafkaTopicField));
+    }
+    else {
+      return Optional.ofNullable(kafkaTopic);
+    }
```
nickwallen: It would make sense to add some debug statements showing which topic was chosen and what the field value is.  This would make debugging a routing issue much simpler for the user.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/kafka/KafkaWriter.java
cestella: They absolutely could create loops in kafka, which wouldn't be ideal.  I don't know how I could prevent it, sadly.  I think this falls in the "with great power comes great responsibility".  Should we call this a spiderman bug (spiderbug?).

metron-platform/metron-parsers/README.md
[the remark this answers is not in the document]
cestella: That's a very good point, I can move this documentation into metron-writer and link to it.

cestella:
Ok, I think I addressed the issues here and I also added an integration test that will exercise this particular scenario.  Let me know what you think, @nickwallen et al

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/kafka/KafkaWriter.java
nickwallen: What if we unset the value of the topic field after the redirect occurs?  This would force the user to set the field again, if they really want another redirect.

It seems like on day 1 of a user attempting to use this functionality, they are going to fall into this trap.  To really use this functionality, you need to set the value in one step and then unset or change it in the next step.  If you don't, its going to loop.

cestella: It's unclear to me that we really want to do that.  It could be that users are going to have logic that depends on that field downstream.  It seems wrong to me to remove a field that they're adding or have in the message.  

I mean, in order to create a loop, one of two things would have to happen:
* the user specifies the input topic as the output topic (aka a simple loop), which removing the kafka topic field wouldn't help because it is likely to be computed (e.g. in the manual test script)
* the user creates a non-simple loop where by sensor A -> B -> ... C -> A, but it's exceedingly unlikely that they're all going to be parsers of the same type, so messages from C will unlikely fail to parse in A.  In the case that they do, it's likely that the kafka topic will be computed in first parser, so it'll be recomputed and removing the field after the first parse won't have helped.

TL;DR
Ultimately, I think removing the field won't appreciably help the situation and puts us into the state of removing data, which makes me uncomfortable.

cestella: I think I see your point.  I think we're ok, though, because we're not defaulting `kafka.topicField`.  If it's not specified AND the `kafka.topic` isn't specified, then we don't send the message anywhere.  I did this specifically so people didn't accidentally forget to unset a field and end up in a loop.  You have to go out of your way (and set the `kafka.topicField`) to make the mistake.

nickwallen: Ok, that makes sense to me.  A user would have to take additional steps to really shoot their self in the foot here.  If 'topic field' was a global setting this might happen, but its not.  

I agree with you.  I don't think there is anything we need to change here.

nickwallen:
+1 Nice bit of functionality.  Thanks!

PR #1089, opened 2018-06-28
METRON-1647 Fix logging level

pravin-raha:
This is quite misleading. logging level should be info rather than error.


cestella:
+1 by inspection, thanks for the contribution!

cestella:
And congrats on your first time contribution!

pravin-raha:
Thanks @cestella

PR #1090, opened 2018-07-02
METRON-1649 Intermittent Test Failure ProfileBuilderBoltTest#testFlushExpiredProfiles

nickwallen:
```
Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.412 sec <<< FAILURE! - in org.apache.metron.profiler.bolt.ProfileBuilderBoltTest
testFlushExpiredProfiles(org.apache.metron.profiler.bolt.ProfileBuilderBoltTest)  Time elapsed: 0.026 sec  <<< FAILURE!
org.mockito.exceptions.verification.TooManyActualInvocations: 
outputCollector.emit(
    "hbase",
    <Capturing argument>
);
Wanted 1 time:
-> at org.apache.metron.profiler.bolt.ProfileBuilderBoltTest.getProfileMeasurements(ProfileBuilderBoltTest.java:265)
But was 2 times. Undesired invocation:
-> at org.apache.metron.profiler.bolt.HBaseEmitter.emit(HBaseEmitter.java:54)

	at org.apache.metron.profiler.bolt.ProfileBuilderBoltTest.getProfileMeasurements(ProfileBuilderBoltTest.java:265)
	at org.apache.metron.profiler.bolt.ProfileBuilderBoltTest.testFlushExpiredProfiles(ProfileBuilderBoltTest.java:205)
   ...
```

The test case ensures that expired profiles are flushed successfully.  The test forces a flush of expired profiles and validates that a value was emitted from the expired profile.  During normal execution, any expired profiles are flushed regularly on a separate thread.

The test case can fail intermittently because profiles are also being flushed in a separate thread at the same time that the test case is forcing a flush.  When the test fails, both the flush triggered by the test and the flush triggered by the background thread have occurred, which results in two values being emitted and the test failing.

Instead of scheduling the background thread so that it executes immediately, execution is delayed by the profile TTL.  There is no need to run the flush immediately as there will never be any expired profiles to flush at least until after a period of time equal to the TTL has passed.  Introducing this initial delay also gives the test plenty of time to execute without interference.

## Pull Request Checklist

### DIFF
--- a/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
+++ b/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
@@ -423,8 +423,10 @@ public class ProfileBuilderBolt extends BaseWindowedBolt implements Reloadable {
    */
   private void startFlushingExpiredProfiles() {
 
+    long initialDelay = profileTimeToLiveMillis;
+    long period = profileTimeToLiveMillis;
     flushExpiredExecutor = Executors.newSingleThreadScheduledExecutor();
-    flushExpiredExecutor.scheduleAtFixedRate(() -> flushExpired(), 0, profileTimeToLiveMillis, TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


cestella:
+1 by inspection, thanks Nick for the quick turnaround

PR #1096, opened 2018-07-05
METRON-1476: Update to Angular 6.1.3

sardell:
## Contributor Comments
This PR updates the Alerts UI to use the latest stable release of Angular (6.1.3). [The original ticket in the ASF JIRA](https://jira.apache.org/jira/browse/METRON-1476) was vague and probably originally intended to update both the Alerts and Management UI. However,[ I created a separate ticket](https://issues.apache.org/jira/browse/METRON-1749) for the Management UI to make the review process easier. 

### Changes included:
* Updated .gitignore to more closely match that generated by Angular CLI and include common files, such as system and IDE specific files.
* Added an .nvmrc to metron-alerts to help ensure developers are using the correct node version.
* Updated documentation to recommend using a node version manager.
* Renamed angular-cli to angular.json and updated the contents to work with the updated Angular CLI.
* Updated prepend_license_header.sh to include css files.
* Updated from using HttpModule and the Http service to HttpClientModule and the HttpClient service.
* Updated from using RequestOptions to HttpHeaders.
* Added an HttpInterceptor to set default headers in requests.
* Added global-shim.ts to prevent polyfill.ts from hoisting and executing before initialization. See here for more details: [https://github.com/swimlane/ngx-dnd/issues/81]( https://github.com/swimlane/ngx-dnd/issues/81).
* Updated Angular CLI flags to use double dashes (--).
* Added aot compilation flag to serve script.
* Removed and updated deprecated RxJS 6 features using [rxjs-tslint auto update rules](https://github.com/ReactiveX/rxjs-tslint).
* Updated auth service to use ReplaySubject instead of EventEmitter.
* Created separate unit test scripts for Travis CI and local usage to prevent launching of two browsers.
* Updated pom.xml to utilize `npm ci` instead of `npm install`.
* Updated installed npm version to 6.2.0 (which allowed us to use `npm ci` instead of `npm install`).
* Updated metron.spec to account for new Angular dist css file naming convention to prevent failures during a build.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


sardell:
Closing and reopening to re-run Travis.

sardell:
To give a brief explanation about my 'add minified dist files to exclude list' commit, I noticed that the new Angular production build files are renamed and stripped of all comments. This caused apache-rat to fail when looking for licenses in those files, which caused Travis to also fail. I added these files to the pom.xml's excludes at the root of the project. I'm making an assumption here that these files are fine to exclude since they aren't source files but rather the result of running `ng build`.

tiborm:
Reviewed +1

pom.xml
```diff
@@ -348,6 +348,16 @@
                              metron-interface/metron-config README and NOTICE files -->
                         <exclude>dist/*.svg</exclude>
                         <exclude>dist/styles.*.bundle.css</exclude>
+                        <!-- Generated assets in the dist folder have licenses present in their source files -->
```
simonellistonball: We should not exclude these from RAT, but should have the licences attached to these 'binaries' using the prepend_licence_header step.

sardell:
Per @simonellistonball's comment, I removed the compiled Angular code from the pom.xml excludes, and instead added them to the prepend_license_header.sh script. In addition, I've added .nvmrc files to both UI projects and updated the documentation accordingly.

justinleet:
@sardell Can you deconflict this?

justinleet:
Could we migrate the `npm ci` change from [this discussion](https://lists.apache.org/thread.html/d0da3647f2955b4257c3eb0d89235779aed64a58097b416a18de6cd9@%3Cdev.metron.apache.org%3E) out to a separate PR?

I personally would like to see that small change go into the upcoming 0.6.0 release so that everything is deterministic, but pulling in a large upgrade like this might be a little riskier (although let me know if you disagree).

sardell:
@justinleet I completely agree. This PR is pretty big, so submitting the `npm ci` fix in a separate PR would mitigate our risk if trying to include it in the next release. I created a separate ASF issue [here](https://issues.apache.org/jira/browse/METRON-1752). I'll try to submit a PR for that today.

metron-interface/metron-alerts/src/app/alerts/alerts-list/table-view/table-view.component.spec.ts
```diff
@@ -42,17 +42,18 @@ describe('TableViewComponent', () => {
...
     TestBed.configureTestingModule({
-      imports: [ HttpModule ],
+      imports: [ HttpClientModule ],
       providers: [
         SearchService,
         UpdateService,
         GlobalConfigService,
         MetaAlertService,
         MetronDialogBox,
+        HttpClient,
```
ruffle1986: Isn't it enough to import the `HttpClientModule` only? Is it necessary to explicitly set `HttpClient` as a provider?
As far as I know, you only have to set the HttpClient when you want to set a mock instead of the original http module.

metron-interface/metron-alerts/.nvmrc
```diff
@@ -0,0 +1 @@
+v9.11.1
```
ruffle1986: This version of node includes npm v5.6.0 which hasn't got `ci` command yet. Should `npm` be updated manually to the latest version? If yes, we should add the update command to the maven config as well.

sardell: Thanks for pointing this out! The maven configuration is fine because we specify the npm version to install in the pom.xml (6.2.0). However, this will effect developers working on Metron. As far as I can tell, there isn't a way for us to include an npm version in our .nvmrc, so I think the next best thing we can do is add a step to the README that instructs developers to install npm 6.2.0.

metron-interface/metron-alerts/src/app/alerts/alerts-list/table-view/table-view.component.spec.ts
sardell: You're right, `HttpClientModule` should be enough here. We should remove this line and the import. Thanks!

ruffle1986:
It has turned out that Angular's changed the way how it renders templates by default in version 6. Earlier, it preserved whitespaces by default but in version 6 the `preservedWhitespace` option is false by default when using the aot (head of time) compiler. Actually, it's a good thing because it reduces the size of the production bundle but during the regression test, we noticed a few differences in the layout before and after the version upgrade. So since we've been relying on whitespaces, we have to set `preserveWhitespace` true.

https://github.com/sardell/metron/pull/10

justinleet:
@sardell To get this running in full dev, I had to change a line in the metron.spec file for the RPMs to build properly.

`%attr(0644,root,root) %{metron_home}/web/alerts-ui/*.bundle.css` to `/usr/metron/0.5.1/web/alerts-ui/styles.*.css`. I assume this is a just a naming change, but I wanted to confirm that with you.

tiborm:
Tested against full dev. With all the fixes you guys made seems good to me.
Huge improvement and great job! Thanks 
+1

nickwallen:
@sardell Seems like we are still adding in changes.  Let us know when this has stabilized and is ready to be reviewed.

sardell:
@justinleet I tested locally and ran into the same thing. Thanks for catching this! The spec file has been updated.

sardell:
@nickwallen Will do.

sardell:
@nickwallen I've added the last of my changes. Almost all were related to review feedback, but I did notice a deprecated class still being imported while applying these feedback updates. My last 3 commits are related to fixing this. This PR is ready for review again.

tiborm:
@justinleet @sardell Just to double check I made a full dev rebuild after fix for metron.spec added. UI builds fine and looks healthy.

sardell:
@nickwallen @tiborm Sorry for the ongoing nature of this PR, but I've made two small but important updates: 
* The first has to do with rxjs import statements that weren't updated automatically via the [rxjs-tslint auto update command](https://github.com/ReactiveX/rxjs-tslint). This command was supposed to resolve all import statements and is recommended for upgrading by the rxjs team, but when I was reviewing a piece of code in the PR today, I noticed an old import statement. As a result, I went through manually and made sure all rxjs imports and operators were updated. 

* The second update is related to [this bug](https://github.com/angular/angular/issues/25424), which led to an @angular/core patch release yesterday. As a result, I ran `ng update @angular/core` and resolved the issue.

metron-interface/metron-alerts/README.md
```diff
@@ -45,7 +45,7 @@ Alerts that are contained in a a meta alert are generally excluded from search r
 * The Management UI should be installed (which includes [Express](https://expressjs.com/))
 * The alerts can be populated using Full Dev or any other setup
 * UI is developed using angular4 and uses angular-cli
-* node.JS >= 7.8.0
+* nvm (or a similar node verison manager) should be installed. The node version required for this project is listed in the [.nvmrc](https://github.com/creationix/nvm#nvmrc) file.
```
nickwallen: Do you have any recommendations for how we should install NVM?  I am assuming a `brew install nvm` will do the trick on a Mac?  Or do you prefer what is mentioned [here](https://github.com/creationix/nvm#install-script)?

Maybe we should embed a link to [this](https://github.com/creationix/nvm#install-script) if you think this is the right way to install.

nickwallen: However we update the instructions here, we should also do the same in `metron-interface/metron-config/README.md`.

nickwallen:
+1 Looks good.  Thanks for all the hard work on this.  I validated full dev by spinning it up, running service check, manually validated the Alerts UI, manually validated the Management UI, and successfully ran the e2e tests.

nickwallen:
I'd prefer this to not be included in the next release; 0.6.0.  This is a rather large change and I'd like it to get more cycles in master before we include it in a release.  

Let's wait for @justinleet to kick-out at least the first release candidate of 0.6.0 before we merge this.  Does that work for you @justinleet ?

metron-interface/metron-alerts/README.md
ruffle1986: We should also update the line above (line 47.): 

> UI is developed using angular4 and uses angular-cli

to: 

> UI is developed using angular 6 and uses angular-cli

sardell: @ruffle1986 Fixed. I also updated `angular-cli` to `Angular CLI` since that seems to be how it's referenced [when mentioned in documentation](https://cli.angular.io/). Thanks for catching that!

PR #1084, opened 2018-06-27
METRON-1644: Support parser chaining

cestella:
## Contributor Comments
Currently we have only one layer of parsing prior to enrichment, but often real data is more complex.  For instance, often data is wrapped in an envelope (e.g. syslog data which contains a field which needs to be parsed).

This Jira allows us to support a DAG of parsers prior to enrichment by allowing us to provide a strategy for interpreting what is data and what is metadata in the parser bolt.  This enables upstream parsers to pass in a JSON Blob which contains the metadata and have the parser bolt choose which field is the data to be parsed, the remaining fields would be considered metadata (and merged into the resulting data or not depending on our existing rules for handling metadata).

To illustrate this better, I've provided a use-case with an example.  Note, this PR depends on METRON-1643 and METRON-1642, so those should be reviewed prior to this.

Manual Test script:
* Go through the use-case doc.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel).
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  mvn -q clean integration-test install && dev-utilities/build-utils/verify_licenses.sh 
  ```

- [x] Have you written or updated unit tests and or integration tests to verify your changes?
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
[...]

PR #1099, opened 2018-07-10
METRON-1657: Parser aggregation in storm

justinleet:
## Contributor Comments
This PR allows for users to specify multiple parsers to be run in one aggregated Storm topology.

Essentially, the ParserBolt (and the supporting infrastructure) has been generalized to take multiple sensors.  This gives us a structure where there's a Storm spout per sensor.  These all lead to a parser bolt that delegates based on the topic metadata and output appropriately.

### Current assumptions/restrictions

- Topic metadata must be enabled in order to be able to delegate properly in the aggregated case (still unneeded for standard case).
- Configs that are shared across parsers generally apply in a "Last-one wins" manner.  Theres's a couple minor exceptions (e.g. if a parser says Kafka security is not PLAINTEXT, that'll win over anyone who says PLAINTEXT).
- All error topics for the aggregated parsers are the same.  This restriction could be lifted if we generalize that infrastructure a bit, but I think it's reasonable to leave as a follow-on if there's enough demand.
- Order matters in how the sensors are specified.  "bro,yaf" is not the same as "yaf,bro".  There are two places this matters, configs and the name of the Storm topology. This could pretty easily be lifted by sorting for the Storm topology (and I might just go ahead and do it anyway).

### Testing
To ensure that single sensor parsers work, just spin up full dev and ensure everything is passing data as expected.

On fulldev, the REST API has been altered to accept a comma separated list in the parser start and parser stop endpoints.  Just kill bro, and start up "bro,yaf".  An aggregated parser should be be launched in Storm with the names of both sensors used, and data should flow through both.   Because no UI is attached to aggregated parsers, I chose not to expand this REST API out for now until we know what's actually needed to properly manage it. The upshot of this is that you can't specify advanced configs directly here.

On the CLI, the start_parser_topology.sh can be used to submit
e.g.
```
${METRON_HOME}/bin/start_parser_topology.sh -z node1:2181 -s bro,yaf
```

If you want to override certain parameters via command line this is still possible:
```
${METRON_HOME}/bin/start_parser_topology.sh -z node1:2181 -s bro,yaf -snt 2,3 -pp 2 -pnt 5
```
This will spin up a topology with a spout number of tasks of 2 for bro, 3 for yaf, along with a parser parallelism of 2 and a parser number of tasks of 5.  This can be seen in the Storm UI by looking at the number of executors and tasks for each Storm component after launch.  As a heads up, because this is full dev and it's only running with one Storm supervisor/worker/etc, it's likely to be unhappy with the config, but it'll still submit the job.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


justinleet:
As a heads up, this will have conflicts with https://github.com/apache/metron/pull/1084

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/topology/ParserTopologyBuilder.java
```diff
@@ -91,14 +101,14 @@ public Config getTopologyConfig() {
    */
   public static ParserTopology build(String zookeeperUrl,
                                       Optional<String> brokerUrl,
-                                      String sensorType,
-                                      ValueSupplier<Integer> spoutParallelismSupplier,
-                                      ValueSupplier<Integer> spoutNumTasksSupplier,
+                                      List<String> sensorTypes,
+                                      ValueSupplier<List> spoutParallelismSupplier,
```
cestella: This is a nit, but is there any reason why we couldn't make these `ValueSupplier<List<Integer>>` and `ValueSupplier<List<Map>>`?

justinleet: I actually forgot to ask about this in the main thing.  It's because of the use of the Class argument in ValueSupplier breaks the typing involved elsewhere. It's the same reason it's ValueSupplier<Map> instead of ValueSupplier<Map<Foo, Bar>> elsewhere.

Having said that, it seems like it's entirely unnecessary to have the class argument at all and we could just actually type everything everywhere and just remove the argument entirely. I tried this out briefly and it seemed fine, but I didn't spin it up through full dev since it seemed optional to do.

cestella: Ah, I see.  man, I miss C++ templates sometimes.  I retract the nit; don't worry about it.

justinleet:
Kicking Travis

cestella:
This looks good to me, I'm +1 by inspection, but I want to make sure enough time has passed so enough people can look at it.  I'll hold my +1 until EOD monday.

ottobackwards:
I have been on vacation, but will be reviewing Monday and Tuesday.  Please do not commit

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -182,40 +185,61 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
+        cacheConfig.putAll(config.getCacheConfig());
       }
     }
+    cache = CachingStellarProcessor.createCache(cacheConfig);
 
-    parser.init();
+    // Need to prep all sensors
+    for (Map.Entry<String, ParserComponents> entry: sensorToComponentMap.entrySet()) {
+      String sensor = entry.getKey();
+      MessageParser<JSONObject> parser = entry.getValue().getMessageParser();
```
ottobackwards: I don't believe this is correct.  We want to initialize stellar PER parser.  Each should have it's own stellar instance and cache.

metron-platform/metron-parsers/README.md
```diff
@@ -82,6 +82,12 @@ topology in kafka.  Errors are collected with the context of the error
 (e.g. stacktrace) and original message causing the error and sent to an
 `error` queue.  Invalid messages as determined by global validation
 functions are also treated as errors and sent to an `error` queue. 
+
+Multiple sensors can be aggregated into a single Storm topology. When this is done, there will be
+multiple Kafka spouts, but only a single parser bolt which will handle delegating to the correct
```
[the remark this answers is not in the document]
justinleet: Correct, it's aggregating existing topics / sensors into a single topology (with multiple spouts).  It's pulling from each individual topic/sensor as a spout, and then passing to a single parser bolt which handles parsing and output as needed.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
justinleet: I left it as a single shared cache on purpose. 

I don't believe that there'd be any incorrect evictions by sharing the cache, and I think evicting based on the overall usage in the aggregated parser is the appropriate place to handle it. Since the cache is (mostly) LRU, I'd prefer to drop the least recently used entry of all parsers rather than dropping for each parser.  LRU of the overall flow seems better than LRU of each of the sensors. Assuming a single cache, you'd bump up the cache configs to account for this, rather than having to optimize each config individually (and potentially as a group afterwards).

Caching is also off by default for the parsers, so this is a case that's only hit if the user explicitly chooses to do so.

Having said that, I do think I need to shore up the documentation around that logic, assuming we choose to go forward with it.  Let me know what you think, and I can adjust appropriately.

justinleet:
@ottobackwards Is there anything we want to do in this PR about the ParserBolt? I agree that it's getting unwieldy, and if there's easy wins it's not a bad opportunity to fix it up.

metron-platform/metron-parsers/README.md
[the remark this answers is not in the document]
ottobackwards: There is another, more likely use case where we have a transport wrapper on another message, and 1 topic split into many parsers as well.  How can we handle that?

Specifically -> Syslog (Many Msg types) -> kafka -> bolt -> Split per message

I expect to add the ability for syslog parsing later, so set that aside.  The issue is we *will* have more than one use case wrt topics.  

I am not going to say this PR needs to address it, but I would want us to understand our path forward and minimize the churn.

It would be best if we did not have to redo this work when accounting for that.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
ottobackwards: We may need a test then.

ottobackwards:
@justinleet the main things I saw that I would think of cutting down, or I though about looking into ( the idea may turn out to be bad ) are places where the bolt 'knows' a lot of weird or complicated initialization logic around the configurations or classes it uses, like what we do initializing Stellar, or in getComponentConfiguration.

I'm ok with a follow on,  I don't think this PR is creating that issue.

metron-platform/metron-parsers/README.md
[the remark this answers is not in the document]
cestella: @ottobackwards I think that's essentially the parser chaining stuff I added earlier, am I misunderstanding?  The [use-case](https://github.com/apache/metron/tree/master/use-cases/parser_chaining) is using the "we get a ton of types of data in syslog" example.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
ottobackwards: I'm still not sure sharing the stellar Context is a good thing. @cestella

metron-platform/metron-parsers/README.md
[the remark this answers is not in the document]
ottobackwards: I think I'm misunderstanding between your diagram and this implementation.  There will be one kafka topic monitored by the bolt,  then routed to the right parser, then output to a different spout per parser?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
cestella: So, the consequences of this decision are as follows:
* You share an expression cache (i.e. the statement -> abstract syntax tree cache; distinct from the expression -> evaluated return cache)
* You share an stellar value cache (expression -> evaluated return)
* You share the state in the Context (e.g. hbase connections, zookeeper connections).

On the whole, anything shared in the context is intended to be shared across users and sensors by virtue of Stellar being used in the enrichment topology (where it's not sensor-by-sensor), so we shoudl be ok there.  The real question is whether users would prefer to have one knob per topology for stellar cache sizing or whether they would prefer to have one knob per sensor.  I'd say that I'm ok with how this PR is doing it, because it's easier to reason about resources, IMO, on a per-topology perspective.

metron-platform/metron-parsers/README.md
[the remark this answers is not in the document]
justinleet: Essentially the flow in the syslog case is:

1. Syslog message comes in through the wrapper's topic.
2. The spout for syslog passes it through to the aggregate bolt.
3. The aggregated bolt does its job unwrapping and outputs the inner messages to the appropriate topics for each.
4. The spouts for the inner messages then pick up from the now populated topics.
5. All spouts are now passing messages as they receive them to the aggregate bolt.
6. All messages get delegated and handled as needed.

This implementation doesn't do anything clever around avoiding output topics if the aggregate parser is already contained here.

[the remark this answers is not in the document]
cestella: This PR gives us the ability to group the parsers into a single topology if we so desire.  You would still write through to kafka.  So, the topology in the example would have 3 kafka spouts:
* One for monitoring `pix_syslog_router` (the syslog parser aka the routing parser)
* One for monitoring `cisco-5-304`
* One for monitoring `cisco-6-302`

There would be one parser bolt, though, which would handle parsing all 3 sensor types.  That is the contribution of this PR, the ability to determine the parser and filter and field transformations from the input kafka topic and use the appropriate one to parse the messages.  There is not, however, any code here that would bypass the intermediate kafka write (e.g. from the router topology to the individual `cisco-5-304` or `cisco-6-302` topics).  That's a current gap.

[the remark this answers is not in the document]
ottobackwards: The bolt is actually executing the parser, not sending it to kafka though.

Let's say that I route all my syslog stuff ( multiple message types ) through 1 kafka topic ( not tied to _ANY_ sensor.  I would want to point this bolt at that one topic, and have it route to multiple parsers. I think then they all just go to enrichment?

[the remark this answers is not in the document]
cestella: Yeah, in order to do that, we'd need to execute the DAG without the intermediate kafka step.  That'd be a follow-on.  This is the first step in that.  We have all the information here (the input -> output mapping of kafka topics), so it shouldn't be so hard to intervene and cut the intermediate kafka writing step out.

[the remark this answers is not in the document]
cestella: You also might want 2 different parsers that source from the same kafka topic (think: 1 parser to send the data to enrichment and 1 parser to send to hbase as a streaming enrichment for authentication data)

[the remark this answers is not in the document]
cestella: @justinleet can you maybe create a data flow diagram or sequence diagram that shows a syslog record from the use-case flowing through this topology and add it to the use-case around parser chaining?

It'd be something like, given a `cisco-6-302` record, it'll go:
* From NiFi to the `pix_syslog_router` kafka topic
* From the `pix_syslog_router` kafka topic to the `pix_syslog_router` spout in the aggregated storm topology
* From the `pix_syslog_router` kafka spout to the parser bolt, which will run the `pix_syslog_router` Grok parser and write out to the `cisco-6-302` kafka topic
* From the `cisco-6-302` kafka topic to the `cisco-6-302` spout in the aggregated storm topology
* From the `cisco-6-302` kafka spout to the `cisco-6-302` Grok parser and write out to the `enrichments` kafka topic, where it's picked up by the enrichment topology.

Eventually, we should consider taking out the writing to the `cisco-6-302` topic (optionally), but even eventually there may be value in those intermediate kafka topics due to how users may want to group sensors (e.g. grouping may be done via velocity or scalability requirements, rather than logical connection).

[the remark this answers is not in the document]
justinleet: Yeah, I'll go ahead and add a diagram to the doc, and flesh out that explanation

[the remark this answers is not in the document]
justinleet: Added a diagram [here](https://github.com/justinleet/metron/blob/995fdd7d4f8ce3b49c7e29aaa7842eb99137a813/use-cases/parser_chaining/README.md#aggregated-parsers-with-parser-chaining), and also linked to it in the parsers readme.

cestella:
@ottobackwards I agree, our bolts tend to do the same magic set of incantations to set up stellar.  It'd be better to either try to infer that initialization where possible or to externalize that logic.  Agreed that it's not this PR that should do that, but I would expect that we should probably consider a separate JIRA to the effect:
* Decouple ParserBolt the parserbolt initialization and execution logic into a separate component from the Storm Bolt infrastructure (this will aid in NiFi or spark integration).

Thoughts?  If we agree, then I'll create the JIRA.

use-cases/parser_chaining/aggregated_parser_chaining_flow.svg
```diff
@@ -0,0 +1,2 @@
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
cestella: You need to add the license header here.

use-cases/parser_chaining/aggregated_parser_chaining_flow.xml
```diff
@@ -0,0 +1 @@
...TrVmbZ8uUetk9yeMfgTzMBRtHDkYPYCCHZU2Oyg7bkR2+6sj7sqMOm5SNYwR2CD+2ix0mH9/yxGGZV7HpAm1gvNet99qNu/gzgCevbU14cuHZWywBJpbUwjf58jqQ2KFwk5XALOaXN1EaliXdniUU2kpsovFdVtul48a9N97+h/D2fFPgXXxbsI2wHuj2XYNhVyzdadieATFvIKevWKNqjn3NOpDnqdsjmLUh4RHBHg10MxHuRaiM8IV3AViEUXXTqdsjPkPE/iGiZ0v4eo4h3h8tRnRMMWIX4j5F3PlE/K2IewFQUrx3xtzTMJct+qMm3m+zQzCTo7DgPfNuRw+3ZZs7S0y8qXVMTKa/Q8ylJ+KGpL2Vs+GCLHGCszBtZXKHRmVeYAjLnP3Z+Ihhmafk0K4an/eNy7SJxgzMgpOnogjxB6HbLmIPSkVwTCrW84gUYabUc3pn/kqOUMcTY1BxdipU7F8/Oiw3PZBoxhpTFXOdSCoa2MOkolN1mx2QZqBHhem8aWaqeFRh5NH82UwpZDpKutLbn6kTzUYkmh7lwaxA0XJF1V9+prW9g3UgdljhZUzB+mhJLdC3JUPRooHxjmqgWl7JDpstmk3l24RKGachTBVEZD0XsES/wnmdK+WMltu1eFcT74blXWuCy+ozDruVhqVwQQxJGGVWk0z9yxo3F+4OtzMAdq6SaE0DT8fOMWCnFjUP+xxAP3SWChLWCpZlmECDNZ4HPJ7yRYBnsCzXgI4aBh6Gjn70IFnW2aPjK2HRO8OjV/Muk6SASUhgTOWiYHQvPN0cp6QHMPLu0b3h9K8pmTbCotoIOfh7drvxIFT3LjvQT/eMe9cwEHYV/6Qjls69a6/V/WYnsVMZK+AGGlaBASq1PHMYVHq0+Jow48ygcv1jQnUyNbNXZJA9vp6Sc0t7zPKtOCmSv6U6ao4J7JlyeOb7B2aZwNemspWpBswz3Y9YN+ss6h6Hjr6JjtNj0tENbJVDakjQ+zwBqFN56neCA9LR6/Fx6MnRsSGdLRfYvlh1+x4WiCqChaJ7zsFG46lr4ik4Kk+VCrBteQeyNFAnUo/PhuSontp8JI4Gr+aowc+OxtHpJ0cH4ejbcrczrzuqJ6DvW3f09AzB9Dukz5SuXSnxHQkxxzFUSixHR6wWvgIy2mx+7FdZZ/OLSnD7Pw==</diagram></mxfile>
```
cestella: You need to add the license header here.

use-cases/parser_chaining/README.md
```diff
@@ -233,3 +233,10 @@ cat ~/data.log | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --b
 ```
 
 You should see indices created for the `cisco-5-304` and `cisco-6-302` data with appropriate fields created for each type.
+
+# Aggregated Parsers with Parser Chaining
+Chained parsers can be run as aggregated parsers. These parsers continue to use the sensor specific Kafka topics, and do not do internal routing to the appropriate sensor.
+
```
ottobackwards: Should we explicitly say that when you use this type of topology so do NOT create a sensor topology for any sensor included in the aggregate?

ottobackwards: yes.  I'm not sure we need to document the UI or somewhere else?   Do we have to document removing the default processors from ambari if you are going to aggregate them etc?  Or how to do it in the management ui?

We may need more practical steps.

ottobackwards:
Ok @justinleet thanks for the diagram.  That really helps.  I did not see in the code how we were sending out to the sensor topic and then into the sensor, I though the bolt was just calling the parser ( which makes the sensor specific topic not make sense ).  I think that this flow is a valid case, one where the data could be coming from either an aggregated flow into the topic *or* straight to the topic.  My question looking at this, is why ( esp when people are asking about reduction of kafka overhead ) can't we have the *option* to eliminate the sensor topic completely?

use-cases/parser_chaining/README.md
justinleet: Right now, as noted in the description, there's no UI attached to this.  Even the REST API's update is pretty minimal (just to take comma separated lists).  I didn't want to build that out, because the management UI requires some decent amount of thought put into it and that'll ripple through REST as needed (e.g. needing/wanting to pass spout num tasks, parallelism, etc.).

Right now I look at this as providing a low level way of being able to get some of the benefits of this type of aggregation, with making it more user friendly being follow-on since it'll require nontrivial effort and design. I can go ahead and create follow-on tickets for that work, if that works for you.

For the default Ambari processors, I'm not particularly inclined to worry about it, although I could be persuaded that we need to.  That feels like something that can be addressed as this is made more user friendly (i.e. I expect people familiar enough with the system to make the determination to aggregate parsers right now to also be familiar enough to stop the topologies).  I could add a warning or something like that in the docs to not run an aggregated parser with sensor X alongside a dedicated topology for sensor X, but I'm not sure that's necessary.

I also went ahead and added the actual command to the chain parsers README, so the practical example is complete.

justinleet:
@ottobackwards I definitely think it's something we should consider adding and that's worthwhile.  I feel it's more of a follow-on task, and the follow-on is to allow for optionally direct routing. i.e. this PR generalizes what's already there in the ParserBolt and a follow-on is to provide that option to do direct routing in the aggregate case.

Do you feel that we need to provide that option in this PR, or is it fine to be left as a follow-on?

use-cases/parser_chaining/README.md
ottobackwards: I think adding the follow on is a great idea.  I wonder if we shouldn't change the default install to aggregate the default sensors?  I don't think that will work though, because the comma separation in ambari is the list, and you won't be able to do it.

My main concern is that the reviewers and committers of this pr are going to be the only ones who can do it, and everyone in user land is going to be lost.  If this is going to be expert only ( until the UI comes ) and not recommend, or a preview thing, maybe mark it as such.

ottobackwards:
@justinleet I am fine with that as a follow on, I would like the task or issue created.

use-cases/parser_chaining/README.md
cestella: I think we should adjust, as a follow-on, ambari to accept proper CSV for parsers.  So, if you want groups you just quote the groups.  E.g. let's say you input the following: `bro,"snort,yaf"`
That would create the following topologies:
* `bro`
* `snort` and `yaf` aggregated

cestella:
Ok, so I want to capture the follow-on tasks:
* Change Ambari to accept quoted parser groups
* Decouple the ParserBolt from the Parse execution logic
* Allow the option for intermediate kafka topics to be removed

Did I miss anything?

justinleet:
@cestella I think we want at least one more to capture making this available in the management UI (I'd argue that should inform any REST API task, so I'm not sure if we want that also created as a ticket).

justinleet:
@ottobackwards Is there anything else we want as follow-on?  Assuming we're good with this where its at (or at least close to), I'll go ahead and create those tickets.

ottobackwards:
A mechanism for the routing process to apply a transform or some such. @cestella may have a better design idea.

What I would like us to do is remove the transport from the message where there are common wrappers.

Example:  All the message types delivered by syslog.   The parsers should not have to all parser syslog AND the message.  I imagine defining a transform/parser in the router that takes every message and transforms it ( parses syslog fields + structured data if 5424 into meta and MSG to the bytes ) and then passes it on to a parser that only needs to know the message.

That way we can have simpler parsers, and even support the same message when transported in different wrappers.

We can talk about if this is a function of parser chaining, such that each specialized parser will be second in a chain with the syslog unwrapped being first, or part of routing, but I think it is part of routing personally.

ottobackwards:
All that being said I am a big +1 on this.  Great work @justinleet, thanks for taking the time to work it through my thick skull.

justinleet:
Happy to help explain it, and it's really helpful to get feedback on it because it led to improved docs, so hopefully it makes sure things are clearer to anyone looking at this later.

I created tickets for:

- [Management UI changes for sensor aggregation](https://issues.apache.org/jira/browse/METRON-1678)
- [REST changes to support parser aggregation](https://issues.apache.org/jira/browse/METRON-1679)
- [Allow the option for intermediate kafka topics to be removed in aggregated sensors](https://issues.apache.org/jira/browse/METRON-1682)
- [Allow Ambari to accept quoted parser groups](https://issues.apache.org/jira/browse/METRON-1680)
- [Decouple the ParserBolt from the Parse execution logic](https://issues.apache.org/jira/browse/METRON-1681)

I held off on creating a ticket for the routing process applying a transform.  @ottobackwards Would you mind creating that, to make sure the use case and examples or assumption, etc. are what you expect?

ottobackwards:
Sure, actually I'll do a discuss thread when this all goes through.  That way I can try again to get @cestella to comment

PR #1111, opened 2018-07-19
METRON-1665 Host UIs on Spring Boot and add SSO authentication

simonellistonball:
This is part of a move towards using KnoxSSO, and LDAP as an authentication method for Metron UIs.

The current UIs are served up by very lightweight nodejs applications, which serve the static bundle files produced by the angular build process, and proxies the rest api. This PR replaces this with Spring Boot applications, using the Spring Security stack. 

It also provides a shared project used as the base for the UI host application, which ensures consistent security across all UIs (metron-interface/metron-ui-security) an abstract web hosting application with zuul embedded for proxying rest to avoid cors problems (metron-interface/metron-ui-host) and two further concrete applications to do the actual hosting, the angular front-ends have been subtley changed to package themselves in jars which are included in the hosting applications, aiding the ease and integrity of the front end code on deployment. 

Some tests in the rest api have been changed slightly, and a fix has been applied to AlertsUiController to allow users access to delete their own settings. The metron-rest project has also been adapted to work with metron-ui-security, and so inherits the same authentication benefits. 

A follow on PR will address changes to the mpack to implement necessary config changes.

Nodejs has been removed from the RPMs, and the deployment process, since it is no longer needed beyond the build stage. 

An additional feature is the filter provided to allow JWT cookies to be passed to a backend through zuul as Authorization Bearer headers, allowing simpler stateless interaction with the rest api for example, which opens the possibility of deploying each layer in a load balanced or HA mode using ribbon config. 

To test at this stage, you will need to manually configure and launch the metron-alerts and metron-config jars, which behave like init.d scripts. This can be done by launching the jars through the bin/ scripts, but would be a lot easier with the follow on work on mpack elements to configure them. For now, integration tests have been included for key function on the security layer and should be reviewed.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-interface/metron-alerts-host/pom.xml
```diff
@@ -0,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
merrimanr: Our xml files generally follow the formatting style of 4 character indents.  Reformatting should be easy and might as well do it now.

metron-interface/metron-alerts-host/src/main/java/org/apache/metron/ui/AlertsApplication.java
```diff
@@ -0,0 +1,31 @@
+/**
```
merrimanr: Our formatting for Java is generally 2 spaces for indents.  I don't feel like it's much effort and might as well make it consistent.

metron-interface/metron-alerts-host/pom.xml
merrimanr: I'm not even sure this is accurate according to our style guide but almost every xml file follows this.

metron-interface/metron-alerts-host/src/main/scripts/metron-alerts.sh
```diff
@@ -0,0 +1,55 @@
...
+    echo "METRON_SYSCONFIG=${METRON_SYSCONFIG}"
+    set -a
+    . "$METRON_SYSCONFIG"
+fi
+
+echo "METRON_SPRING_PROFILES_ACTIVE=${METRON_SPRING_PROFILES_ACTIVE}"
+
+METRON_CONFIG_LOCATION=" --spring.config.location=classpath:/application.yml,$METRON_HOME/config/alerts_ui.yml"
+echo "METRON_CONFIG_LOCATION=${METRON_CONFIG_LOCATION}"
+METRON_SPRING_OPTIONS+=${METRON_CONFIG_LOCATION}
```
merrimanr: I'm not sure if this is intentional but this property is also used by REST (METRON_SPRING_OPTIONS).  I can't see many cases where we would want this to be shared.

metron-interface/metron-rest-client/bin/pom.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
merrimanr: Is adding this file a mistake?

metron-interface/metron-ui-host/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+    <name>Metron Generic UI Host</name>
+    <description>Spring Server to host config ui</description>
+
+    <parent>
+        <groupId>org.apache.metron</groupId>
+        <artifactId>metron-interface</artifactId>
+        <version>0.5.1</version>
+    </parent>
+
+    <properties>
```
merrimanr: Ideally these properties would be defined at the metron-interface module level.  Do you mind changing in this PR?  Should be an easy fix and we should do it eventually.

merrimanr:
I really like this feature and think it's an important improvement.  This PR is a great start and I don't see any major architectural issues.  I've added some comments inline for the really minor issues.  I noticed that application.yml in metron-config-host and metron-alerts-host is different between the 2 but I can't think of any reason why.  I didn't comment directly but it's probably worth reviewing.

There are some bigger issues I want to call out that are more complex.  

I noticed there is a fairly significant gap with respect to groups and roles.  As far as I can tell it's not supported right now which is a regression from the JDBC solution currently in place.  I'm sure this is expected and will be solved in follow on tasks in this feature branch.  I do think we should set ourselves up to be able to do that in future PRs though.  

In this PR, metron-ui-host and metron-rest use 2 different authentication mechanisms for testing.  I feel they should use the same one and we should eliminate JDBC or in memory authentication all together.  To do this we need to move this testing infrastructure (the EmbeddedLdap class and supporting files, Spring properties, etc) to metron-security.  This will allow us to switch REST to a common security model during testing, which is essentially what we're doing with this feature.  I've submitted a PR with an example of how we could do that (switching REST to embedded ldap isn't included yet and should be done later to avoid conflicts with other PRs).  This will allow us to eliminate redundant properties across modules but I think the biggest benefit is that our testing will more accurately match a real installation and allow us to catch bugs sooner.

I think it's fair to say we need some documentation.  I would say at a minimum the metron-rest README should be updated to reflect the new authentication approach.  Eventually it would be nice to have architectural level documentation because there are several new components/frameworks being introduced here.  I have no problem with that being done in future PRs.

I also noticed there are several TODOs.  We should at some point document these as follow on tasks so we have a clear understanding of what is expected to work in each PR.

metron-interface/metron-rest-client/bin/pom.xml
simonellistonball: Yes. Will remove.

metron-interface/metron-alerts-host/src/main/scripts/metron-alerts.sh
simonellistonball: They're actually in different environments, so didn't feel the need to change the convention. These scripts take all their environment from Ambari launchers in a subsequent PR, rather than the /etc/defaults approach, so there is no actual collision. I can see an argument for changing them all for clarity, but could also see the argument for consistency in the context of potentially growing number of services like this.

merrimanr: This environment variable is sourced from /etc/default/metron which is bound to the spring options setting in the REST tab in Ambari.  I tested it in full dev and the option is indeed added to alerts and config UI start commands.

metron-interface/metron-ui-host/pom.xml
merrimanr: Sure if you think certain versions should be configured separately.  At the very least I think we want the spring boot version to be consistent.

merrimanr:
Here is what I think is pending on this PR:
- Address the minor comments I've made inline
- Update the metron-rest and/or other relevant READMEs with the new authentication approach
- Create a follow on Jira to implement a simple LDAP group to role mapping so that we're not introducing a regression
- Create a follow on Jira to implement roles properly (would that be outside of this feature branch?)
- Add architecture documentation either in this PR or create a follow on Jira for it
- Create a follow on Jira to convert metron-rest to use the testing infrastructure in metron-security

metron-interface/metron-alerts-host/src/main/scripts/metron-alerts.sh
simonellistonball: Right, my mistake. We should change it and carry through. Would it make sense to do that as part of the follow on PR for the mpack work, since that will be where the bulk of the changes for this would be?

metron-interface/metron-ui-host/pom.xml
simonellistonball: Makes sense. I guess I'm half thinking about the idea of 3rd party extensions, but we can at least be consistent within the official bits and let people override if they absolutely for some strange reason have to definitely use something different.

simonellistonball: Done, I moved the spring versions up to interface.

metron-interface/metron-alerts-host/src/main/java/org/apache/metron/ui/AlertsApplication.java
simonellistonball: Done

metron-interface/metron-alerts-host/pom.xml
simonellistonball: Done, and agreed, it should be a formal style guide addition, and part of the formatting configs on the contributor wiki really.

metron-interface/metron-alerts-host/src/main/scripts/metron-alerts.sh
merrimanr: Sure that's fine.

merrimanr:
Thanks @simonellistonball.  I think this is good enough to start with.  +1

merrimanr:
This has been merged into the feature branch.  Can you close @simonellistonball?

PR #1112, opened 2018-07-19
METRON-1668 Remove login services and screens from UIs

simonellistonball:
## Contributor Comments

Removing the login screens and modules from the UIs since they are superfluous in the the SSO context, and no-longer have an authentication engine to back them. 

This can be tested by authentication with basic auth against the new hosts in METRON-1665 and then ensuring the UIs still function, and by ensuring the UIs do not redirect to a login. 

The authentication will be handled by the hosts that allow loading of the UIs redirecting the browser to a KnoxSSO endpoint, handled in METRON-1665.

Note, handling jwt cookie expiry is not included here, but will be left as a follow on task with separate ticketing.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
"The authentication will be handled by the hosts that allow loading of the UIs redirecting the browser to a KnoxSSO endpoint, handled in METRON-1665"

How is this going to work in vagrant?
What will be the new minimum required setup for security to use Metron's UI's now?

simonellistonball:
Patience, patience :) https://issues.apache.org/jira/browse/METRON-1664 will set all this up in ambari to work on full-dev with an embedded ldap (instead of the current embedded h2) and Knox doing the SSO.

ottobackwards:
It seems strange to me to ONLY support SSO.  Most things support a local configuration and SSO or 'other' on demand.

simonellistonball:
It would also support ldap authentication, via basic auth.

simonellistonball:
And infact through form based auth if sso is off (see the backend issue METRON-1665). This PR just removes the need to process a spring login form in the Angular app to lighten the load a bit and avoid having multiple front-end login implementations to keep in sync (config and alerts have their own at present). This is more about moving the same functionality out of the angular layer and into a different page.

ottobackwards:
I don't understand, how are you going to do the auth without the login screen?

ottobackwards:
this might be worth a discuss thread @simonellistonball

ottobackwards:
@simonellistonball, thank you.  I didn't get that from the PR description.  Sorry for the noise.

merrimanr:
I noticed that I can no longer start the Alerts UI locally for development purposes.  Since the spring boot hosting application handles SSO with knox there is no way to authenticate running outside of that with something like `ng serve`.   We will need to come up with a strategy for this case.

simonellistonball:
@mmiklavc the knoxsso is enabled with a toggle in the mpack, to follow the pattern for NiFi, Ranger and Atlas etc. There is a kind of failover of auth methods: knoxsso, ldap, nothing in our case. Many of the API calls require auth because they use username to load settings in the rest backend. As such I’m not sure no-auth makes sense, but maybe a default to anon would allow @rmerriman’s use case of dev runs, I’d still say that was a case for a mock backend though, in angular rather than rest.

PR #1132, opened 2018-07-25
METRON-1695: Expose pcap properties through Ambari

anandsubbu:
## Contributor Comments
This change add a new jinja template for the pcap properties and adds a new tab in Ambari Metron config for the PCAP properties. Here are the screenshots:
pcap-config-1
pcap-config-2

**Testing Done**
Built mpack out of these changes, deployed a multi-node cluster and validated that the pcap.properties is being updated during Parser startup scripts per the screenshot below:
parser-startup-screenshot

Also validated that the pcap properties is updated with config changes (ZK quorum, kafka broker etc.)
```
[root@metron-12 config]# cat pcap.properties

##### Storm #####
topology.worker.childopts=
topology.auto-credentials=[]
topology.workers=1

##### Kafka #####
spout.kafka.topic.pcap=pcap
kafka.zk=metron-10:2181,metron-11:2181,metron-1:2181
hdfs.sync.every=1
hdfs.replication.factor=--1
kafka.security.protocol=PLAINTEXT

# One of EARLIEST, LATEST, UNCOMMITTED_EARLIEST, UNCOMMITTED_LATEST
kafka.pcap.start=UNCOMMITTED_EARLIEST

kafka.pcap.numPackets=1000
kafka.pcap.maxTimeMS=300000
kafka.pcap.ts_scheme=FROM_KEY
kafka.pcap.out=/apps/metron/pcap/input

##### Parallelism #####
kafka.pcap.ts_granularity=MICROSECONDS
kafka.spout.parallelism=1
```
**Verification Steps**
* Spin up full dev (or a multi-node cluster)
* Ensure that the PCAP tab in Ambari Metron config allows for values to be entered.
* Deploy cluster and validate that $METRON_HOME/config/pcap.properties file has the proper config values set.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-pcap-env.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-pcap-env.xml
@@ -1,0 +1,141 @@
+<?xml version="1.0"?>
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<configuration supports_final="true">
+    <property>
+        <name>pcap_topology_worker_childopts</name>
+        <description>PCAP Topology JVM Options</description>
+        <value/>
+        <display-name>PCAP Topology childopts</display-name>
+        <value-attributes>
+            <empty-value-valid>true</empty-value-valid>
+        </value-attributes>
+    </property>
+    <property>
+        <name>pcap_topology_workers</name>
+        <description>Number of PCAP Topology Workers</description>
+        <value>1</value>
+        <display-name>Workers for PCAP Topology</display-name>
+    </property>
+    <property>
+        <name>spout_kafka_topic_pcap</name>
+        <description>PCAP Input Topic</description>
+        <value>pcap</value>
+        <display-name>PCAP Input Topic</display-name>
+    </property>
+    <property>
+        <name>hdfs_sync_every</name>
+        <description>HDFS Sync Every</description>
+        <value>1</value>
+        <display-name>HDFS Sync Every</display-name>
+    </property>
+    <property>
+        <name>hdfs_replication_factor</name>
+        <description>HDFS Replication Factor</description>
+        <value>-1</value>
+        <display-name>HDFS Replication Factor</display-name>
+    </property>
+    <property>
+        <name>kafka_pcap_start</name>
+        <description>PCAP Topology Spout Offset</description>
+        <value>UNCOMMITTED_EARLIEST</value>
+        <display-name>PCAP Topology Offset</display-name>
+        <value-attributes>
+            <type>value-list</type>
+            <entries>
+                <entry>
+                    <value>EARLIEST</value>
+                </entry>
+                <entry>
+                    <value>LATEST</value>
+                </entry>
+                <entry>
+                    <value>UNCOMMITTED_EARLIEST</value>
+                </entry>
+                <entry>
+                    <value>UNCOMMITTED_LATEST</value>
+                </entry>
+            </entries>
+            <selection-cardinality>1</selection-cardinality>
+        </value-attributes>
+    </property>
+    <property>
+        <name>kafka_pcap_numpackets</name>
+        <description>Number of Packets</description>
+        <value>1000</value>
+        <display-name>Number of Packets to keep in one file</display-name>
+    </property>
+    <property>
+        <name>kafka_pcap_maxtimems</name>
+        <description>Max Time in milliseconds</description>
+        <value>300000</value>
+        <display-name>Number of packets to keep in terms of duration</display-name>
+    </property>
+    <property>
+        <name>kafka_pcap_tsscheme</name>
+        <description>Kafka PCAP Timestamp Scheme. Use FROM_KEY which works with current tooling</description>
+        <value>FROM_KEY</value>
+        <display-name>Kafka PCAP Timestamp Scheme</display-name>
+        <value-attributes>
+            <type>value-list</type>
+            <entries>
+                <entry>
+                    <value>FROM_KEY</value>
+                </entry>
+                <entry>
+                    <value>FROM_VALUE</value>
+                </entry>
+            </entries>
+            <selection-cardinality>1</selection-cardinality>
+        </value-attributes>
+    </property>
+    <property>
+        <name>kafka_pcap_out</name>
+        <description>HDFS Directory to store PCAPs</description>
+        <value>/apps/metron/pcap/input</value>
+        <display-name>HDFS Directory to store PCAPs</display-name>
+    </property>
+    <property>
+        <name>kafka_pcap_ts_granularity</name>
+        <description>Granulariy of Timing in Timestamps</description>
+        <value>MICROSECONDS</value>
+        <display-name>Granulariy of Timing in Timestamps</display-name>
+        <value-attributes>
+            <type>value-list</type>
+            <entries>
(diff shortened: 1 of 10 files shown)


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/parser_master.py
```diff
@@ -54,6 +57,12 @@ def configure(self, env, upgrade_type=None, config_dir=None):
             commands.init_kafka_acls()
             commands.set_acl_configured()
 
+        File(format("{metron_config_path}/pcap.properties"),
```
anandsubbu: I had earlier place this under the [indexing_master.py ](https://github.com/apache/metron/blob/master/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/indexing_master.py#L62)script, but felt that it is more related to parser than with indexing. I am happy to move this block to any other place that is deemed fit.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-pcap-env.xml
```diff
@@ -0,0 +1,141 @@
...
+        <display-name>Workers for PCAP Topology</display-name>
+    </property>
+    <property>
+        <name>spout_kafka_topic_pcap</name>
+        <description>PCAP Input Topic</description>
+        <value>pcap</value>
+        <display-name>PCAP Input Topic</display-name>
+    </property>
+    <property>
+        <name>hdfs_sync_every</name>
```
anandsubbu: The PCAP [config readme](https://github.com/apache/metron/tree/master/metron-platform/metron-pcap-backend#configuration) does not talk more about the units for this parameter (whether seconds, or minutes or something else). I looked through the code as well, but was not able to find more info. If someone could clarify about this, I can add the units so it is clearer.

mmiklavc:
@anandsubbu Thanks for contributing this!  The only thing I'm unsure of after reviewing this is your comment regarding this belonging in parser_master.py or not. On one hand, you could make a case that this is a parser like any other with the exception that it is spout-only. On the other hand, we don't currently provide a mechanism (iirc) to start/stop/manage the PCAP topology from Ambari. Starting parsers doesn't start PCAP, does it? 

It may be that PCAP deserves its own "master", e.g. pcap_master.py, and lifecycle as a proper first class citizen (topology) within Ambari. Below is the current list of components, and PCAP is not in it afaik, even as a parser (there's a start_pcap_topology.sh script that is NOT the same as start_parser_topology.sh). It seems like if we're going to add this to Ambari we should also provide the ability to manage it. What do you think of that approach?

image

simonellistonball:
+1 to that, PCAP should definitely get its own service, but agreed with @anandsubbu that should probably be a follow on item.

mmiklavc:
I think I'm ok with this as a first step. Please document the lifecycle for how pcap.properties is updated in the README. It should be clear to users that they will need to restart parsers for the changes to take effect after the initial install is performed. Can you confirm, is that the way the properties are/will be deployed with this PR @anandsubbu? 

https://github.com/apache/metron/pull/1134 should be included in the refactoring once we make PCAP a proper component.

mmiklavc:
I'm +1 via inspection pending the requested documentation. Also, please create the follow-on Jira, have it linked to this Jira, and also link it here for reference in the comments on this PR. Both this Jira/PR and https://github.com/apache/metron/pull/1134#issuecomment-409658527 should link up.

anandsubbu:
> Please document the lifecycle for how pcap.properties is updated in the README. It should be clear  to users that they will need to restart parsers for the changes to take effect after the initial install is performed. Can you confirm, is that the way the properties are/will be deployed with this PR @anandsubbu?

Hi @mmiklavc - there is no change in behavior from earlier to now. Nothing changes from a usage perspective. Let me clarify more with an example for a multi-node deployment...

**Earlier**
* User deploys metron
* `pcap.properties` file created with defaults under `$METRON_HOME/config`
* User hand edits the properties file (sets ZK quorum and other parameters as required)
* Starts the topology per the instructions [here](https://github.com/apache/metron/tree/master/metron-platform/metron-pcap-backend#starting-the-topology).

**Now**
* During deploy time, user is presented with a separate config tab in Ambari to configure/reconfigure PCAP properties. If user chooses to leave them untouched, the pcap.properties is initialized with appropriate defaults. ZK quorum is auto-populated as well.
* `pcap.properties` file is created during metron-parsers startup step.
* Starts the topology per the instructions [here](https://github.com/apache/metron/tree/master/metron-platform/metron-pcap-backend#starting-the-topology).

The steps in the [README](https://github.com/apache/metron/tree/master/metron-platform/metron-pcap-backend) still holds. 

Now, we can add a note in the README to indicate that the PCAP properties can be configured via the 'PCAP' tab in Ambari Metron config. But I noticed this was not explicitly mentioned for other components (e.g. [REST](https://github.com/apache/metron/tree/master/metron-interface/metron-rest#configuration), [elasticsearch](https://github.com/apache/metron/tree/master/metron-platform/metron-elasticsearch#properties)). Let me know if you prefer to have this added.

anandsubbu:
Btw, I created [METRON-1719](https://issues.apache.org/jira/browse/METRON-1719) to track PCAP sensor being an independent service entity.

mmiklavc:
@anandsubbu, considering the changes in this PR:

1. Let's say I manually make changes to `pcap.properties`. I restart the parser topologies. Later, I manually restart pcap with the shell command. What happens to my `pcap.properties` changes?
2. Similarly, I make changes via Ambari's new PCAP panel. Now I go start pcap using `$METRON_HOME/bin/start_pcap_topology.sh`. What does pcap.properties look like?

In 1, your changes are overwritten by the parsers' restart. In 2, you've made changes in Ambari but they don't get materialized in the properties file because you haven't restarted the parsers - I may be mistaken, but I'm pretty sure clicking "save" in Ambari does not deploy the new properties, and this is the reason I ask for a doc change. Previously, configuring pcap was all manual steps. Now it will be a combo of manual steps for start/stop, automated steps for the config updates. And more specifically for the config updates, you will *only* see them if you make them in Ambari and restart the parsers. Also, if you choose not to use Ambari for this config handling and instead make your config changes to pcap.properties locally, you will have to remember that restarting the parsers will overwrite your pcap config - something you're only likely to realize the next time you start/restart pcap. This is the behavior that I think we should document if we accept this. Does that make sense?

anandsubbu:
Yup, @mmiklavc I see what you are saying. Make sense to me. Let me know if the latest README update conveys the message. 

I also added a fix to prompt a service restart upon changes to the PCAP config settings. Thanks @MohanDV for the pointer!

mmiklavc:
@anandsubbu thanks for the updates, I'm going merge your PR.

merrimanr:
I don't see any good reason to put off making pcap it's own Ambari component.  Having to restart the parsers component (which is unrelated to pcap) to propagate pcap changes is a deal breaker for me. 

Had this been submitted against the pcap feature branch I would be fine with it as an incremental change but I don't feel like this puts master in a good state and I definitely would not want it in a release.  Why not just do it here?

mmiklavc:
@anandsubbu, I hate to give you whiplash, but @merrimanr may have a point here. There benefit of exposing the pcap properties via Ambari is lessened by now requiring the user to do 3 things instead of 2 to get the properties set for Pcap. 

**Old**

1. Update properties
2. Restart pcap from command line

**New**

1. Update properties
2. Restart parsers to deploy the properties (users may not like this requirement)
3. Restart pcap from command line

The work is by no means throw away in its current form, but if we're going to do this it's probably worth going the whole way.

anandsubbu:
This PR lays the foundation for exposing the properties to begin with. My thought was that it reduces one error prone manual step of hand-editing the `pcap.properties` file. I agree with your point about how users might be concerned about restarting all parsers after modifying PCAP config. 

I did not create this earlier under the feature branch since this was not related specifically to the PCAP query panel, it was a more generic change.

I am fine to move this under the feature branch or I could wait until the fix for METRON-1709 (making the PCAP service its own) is available and then submit a fresh PR. Let me know which one works better.

merrimanr:
If you want to do the work in a separate PR (METRON-1709) then that's fine.  As long as they are tested and committed together that works for me.  If it were me, I would just do the work in this PR and save the trouble of managing two different PRs.  We have several different components already that you can use as a template.  I don't think this is that much work.

Since this was developed against master I wouldn't switch to the feature branch.

anandsubbu:
Sure @merrimanr , sounds good.

@MohanDV has already begun work on METRON-1709. I will wait for him to complete and then submit this pull request so it would be a natural fit.

nickwallen:
@anandsubbu I am conflicted on this one. :) Can you de-conflict?

anandsubbu:
Hey @nickwallen ... Let me close this PR. Will create one afresh based on the latest changes from #1201

PR #1138, opened 2018-08-03
METRON-1722: PcapCLI should print progress to stdout

merrimanr:
## Contributor Comments
This PR updates the Pcap CLI to print the status of a running job.  I took a fairly simple approach as a first pass:  

- Added a `PrintStream` member variable to PcapJob 
- Added a `reportStatus` method that configures that variable
- Added logic to the `PcapJob.get` method to print status if the printStream variable is set

This approach keeps us from having to change the Statusable interface.  For those reviewing, what do you think of the message?  Should we format it differently or include more information?

I am also considering adding a unit test for `PcapJob.get` in this PR since it doesn't exist and the changes in this PR affect that method.

Let me know what you think.  There are several different ways to solve this.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


justinleet:
Given that this is solving a regression, I'd like a quick update on what the previous state was and how this compares.  In particular, I'd like to know how it functions in terms of status and error / exception handling?  I believe the old version printed out some of the underlying MR progression along with underlying errors.  Do we have a sample of the old output at available?

I'm not sure how concerned we need to be with it being exactly the same (and I think there is definitely a fairly strong argument towards being consistent with the UI because the MR job isn't the only time consuming item there), but we do need to make sure it's at least as informative.

merrimanr:
I believe I found the solution.  There is a method called `monitorAndPrintJob` on the `org.apache.hadoop.mapreduce.Job` class that blocks and prints status just as if you had submitted the job synchronously.  I compared the Pcap CLI output in this PR with the output in master and it is exactly the same as far as I can tell.

mmiklavc:
@merrimanr I think this is closer with your recent change. I think it should still use the get() method because you need to wait for the finalizer to complete. Can we add a param for the CLI like `PcapOptions.PRINT_JOB_STATUS.put(commonConfig, true)`?

merrimanr:
@mmiklavc the latest commit includes the changes you suggested.  Let me know what you think.

justinleet:
+1 by inspection, assuming  @mmiklavc is good.

merrimanr:
Are you good @mmiklavc?

merrimanr:
Let me know what you think of the latest commit.

metron-platform/metron-pcap-backend/src/test/java/org/apache/metron/pcap/query/PcapCliTest.java
```diff
@@ -166,12 +166,12 @@ public void runs_fixed_pcap_filter_job_with_full_argument_list_and_default_datef
     PcapOptions.START_TIME_MS.put(config, 500L);
     PcapOptions.END_TIME_MS.put(config, 1000L);
     PcapOptions.NUM_RECORDS_PER_FILE.put(config, 1000);
+    PcapOptions.PRINT_JOB_STATUS.put(config, true);
```
mmiklavc: Probably remove these as they're not used anymore by the test.

merrimanr: This needs to be there since the print status command line argument is set.

mmiklavc: Ah, I see. Missed that with the diff only showing the option line in the latest commit. Full diff shows the extra option in the cli options arg list.

mmiklavc:
+1

justinleet:
I'm still +1

PR #1140, opened 2018-08-06
METRON-1726: Refactor PcapTopologyIntegrationTest

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1726

While working on issues with multi-threading and the pcap job I found it useful  to split up the monolithic pcap integration test into multiple tests with a separate setup and teardown method using @Before/AfterClass. Also fixed a regression/bug where the integration tests were writing out to the module root instead of `target`. Added back the file output path prefix along with defaults for the PcapCLI to write to the exec dir. Verification of this PR is checking that the tests appear correct and that the integration test passes.

### For all changes:

### For code changes:
  ```
  ```

n/a

#### Note:

### DIFF
--- a/metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
+++ b/metron-platform/metron-pcap-backend/src/main/java/org/apache/metron/pcap/query/PcapCli.java
@@ -87,6 +87,7 @@ public class PcapCli {
       try {
         config = fixedParser.parse(otherArgs);
         commonConfig = config;
+        PcapOptions.FINAL_OUTPUT_PATH.put(commonConfig, new Path(execDir));
       } catch (ParseException | java.text.ParseException e) {
         System.err.println(e.getMessage());
         System.err.flush();
@@ -112,6 +113,7 @@ public class PcapCli {
       try {
         config = queryParser.parse(otherArgs);
         commonConfig = config;
+        PcapOptions.FINAL_OUTPUT_PATH.put(commonConfig, new Path(execDir));
       } catch (ParseException | java.text.ParseException e) {
         System.err.println(e.getMessage());
         queryParser.printHelp();
--- a/metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/finalizer/PcapCliFinalizer.java
+++ b/metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/finalizer/PcapCliFinalizer.java
@@ -37,7 +37,7 @@ public class PcapCliFinalizer extends PcapFinalizer {
    * as a formatted timestamp + uuid. A final sample format will look as follows:
    * /base/output/path/pcap-data-201807181911-09855b4ae3204dee8b63760d65198da3+0001.pcap
    */
-  private static final String PCAP_CLI_FILENAME_FORMAT = "pcap-data-%s+%04d.pcap";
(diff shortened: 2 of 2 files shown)


mmiklavc:
I also ran this up in full dev to verify the pcap_query.sh CLI tool still works as expected.

`$METRON_HOME/bin/pcap_query.sh fixed -bp /apps/metron/pcap/input -df "yyyy-MM-dd-HH-mm" -st 2018-07-26-00-00 -et 2018-07-26-00-10 -rpf 500`

This output 1,518 records in 4 files, as expected (1518 / 500 = 3 R 18), in the user's current execution directory.

justinleet:
+1 by inspection. This is great cleanup and makes it a lot easier to understand.

merrimanr:
First off, I think this is definitely an improvement.  Putting the individual pcap queries into separate tests makes it easier to understand when things go wrong.

Would it make sense to move the pcap job/query related code to it's own test?  For me it makes more sense to have these tests in something like PcapJobIntegrationTest in metron-pcap.  It feels awkward for the pcap topology test to be in a setup or `@BeforeClass` method.  If I introduced a bug to the pcap topology code, which test would fail?

What do you think?

mmiklavc:
> If I introduced a bug to the pcap topology code, which test would fail?

@merrimanr They would all fail because the writing to HDFS sequence files is incorrect. Prior to this refactoring, only 1 test would fail if a bug was introduced, and likely only the first test. This test is, and has been, a 2-phase broad e2e/acceptance test since the beginning. I didn't seek to change the fundamental operation of the test in this refactoring, but rather the organization of those units to make it easier to pinpoint specific failures. I definitely see your point about the setup, though this is still a definite improvement imo that moves us in the direction of being able to make additional refactorings should we so choose.

To answer your question more broadly, I've dealt with similar questions when testing DAO layers with databases. The approach I've used in the past that yielded optimal results without duplicating almost verbatim the code under test and test code itself was to use the code under test to read/write to the system and verify the expected values. To your point, however, this test is really 2 phases. 
1. `Raw pcap -> pcap topology processes it -> pcap as sequence files in HDFS`
2. `pcap as sequence files in HDFS -> pcap query filters it -> written to local  FS or HDFS`

Each of those can be tested independently, similar to what I described about DAO's. Even so, I think there's probably value in doing the full e2e as it's currently being done since we also have some spot-check unit tests around the individual components as well. We could spin up a test for the pcaptopology distinct from the querying, but there's definitely more work there. We would need
to:
1. write test code that correctly reads the sequence files from HDFS
2. choose additional test conditions for the topology validation (e.g. validating specific pcap values)
3. create a sequence file (probably just pulled from the output of the existing integration test after the topology creates the files in HDFS) for input to the query test.

All that being said, could I convince you that the improvements made by this PR are a good first step that moves us closer to additional improvements should we choose to make them?

merrimanr:
That's fine with me.  Just wanted to point out that this test is still deficient because the pcap topology part really isn't tested that well and the only coverage we have right now is that it sets up other tests.  It also doesn't match the way we integration test other topologies or DAOs in Metron.  For example, our indexing integration test and search DAO integration tests are completely separate from each other.

Maybe that's something we improve in the future.  I imagine it will make sense to do it when (if) the pcap topology is updated.  +1

mmiklavc:
@merrimanr - agreed. Pcap hasn't gotten nearly as much attention as the rest of the core infrastructure for quite some time. I created a Jira to track this - https://issues.apache.org/jira/browse/METRON-1729

PR #1146, opened 2018-08-07
METRON-1704 Message Timestamp Logic Should be Shared [Feature Branch]

nickwallen:
The Profiler can operate using either processing time or event time.  This is controlled by the user by defining the "timestampField" option in their Profiler configuration.

There is logic that determines the timestamp of a message.  If the Profiler is configured to use processing time, then system time is returned by this logic.  If the Profiler is configured to use event time, then the timestamp is extracted from a message field.

This logic is currently duplicated across both ports of the Profiler; the REPL and Storm.  This should be pulled into `metron-profiler-common/` so that the logic can be shared and also used by the Spark port.

## Testing

## Pull Request Checklist

### DIFF
--- a/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageDistributor.java
+++ b/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageDistributor.java
@@ -146,16 +146,14 @@ public class DefaultMessageDistributor implements MessageDistributor {
   /**
    * Distribute a message along a MessageRoute.
    *
-   * @param message The message that needs distributed.
-   * @param timestamp The timestamp of the message.
    * @param route The message route.
    * @param context The Stellar execution context.
    */
   @Override
-  public void distribute(JSONObject message, long timestamp, MessageRoute route, Context context) {
+  public void distribute(MessageRoute route, Context context) {
     try {
       ProfileBuilder builder = getBuilder(route, context);
-      builder.apply(message, timestamp);
+      builder.apply(route.getMessage(), route.getTimestamp());
 
     } catch(ExecutionException e) {
       LOG.error("Unexpected error", e);
--- a/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageRouter.java
+++ b/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageRouter.java
@@ -22,6 +22,9 @@ package org.apache.metron.profiler;
 
 import org.apache.metron.common.configuration.profiler.ProfileConfig;
 import org.apache.metron.common.configuration.profiler.ProfilerConfig;
+import org.apache.metron.profiler.clock.Clock;
+import org.apache.metron.profiler.clock.ClockFactory;
+import org.apache.metron.profiler.clock.DefaultClockFactory;
 import org.apache.metron.stellar.common.DefaultStellarStatefulExecutor;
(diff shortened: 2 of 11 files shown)


metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageDistributor.java
```diff
@@ -146,16 +146,14 @@ public DefaultMessageDistributor(
...
    * Distribute a message along a MessageRoute.
    *
-   * @param message The message that needs distributed.
-   * @param timestamp The timestamp of the message.
    * @param route The message route.
    * @param context The Stellar execution context.
    */
   @Override
-  public void distribute(JSONObject message, long timestamp, MessageRoute route, Context context) {
+  public void distribute(MessageRoute route, Context context) {
```
nickwallen: The `MessageRoute` abstraction now contains both the message (a JSONObject) and the timestamp (a Long).  Previously this was passed along separately.  It is much simpler for all of the ports (Spark included) to wrap this all into the `MessageRoute`.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageRouter.java
```diff
@@ -53,10 +56,16 @@
    */
   private StellarStatefulExecutor executor;
 
+  /**
+   * Responsible for creating the {@link Clock}.
+   */
+  private ClockFactory clockFactory;
```
nickwallen: The `ClockFactory` helps with the timestamp logic.  This used to live in the `ProfileSplitterBolt` as that is where the timestamp extract logic lived.  Now the `MessageRouter` does this for all Profiler ports.

metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileSplitterBolt.java
```diff
@@ -162,15 +156,9 @@ private void doExecute(Tuple input) throws ParseException, UnsupportedEncodingEx
     // ensure there is a valid profiler configuration
     ProfilerConfig config = getProfilerConfig();
     if(config != null && config.getProfiles().size() > 0) {
+      routeMessage(input, message, config);
 
-      // what time is it?
```
nickwallen: Here I removed the timestamp logic from Storm.  The `MessageRouter` does this for us now.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageRouter.java
```diff
@@ -86,20 +96,24 @@ public DefaultMessageRouter(Context context) {
...
     final Map<String, Object> state = (Map<String, Object>) message;
-
     try {
       // is this message needed by this profile?
       if (executor.execute(profile.getOnlyif(), state, Boolean.class)) {
 
-        // what is the name of the entity in this message?
-        String entity = executor.execute(profile.getForeach(), state, String.class);
-        route = Optional.of(new MessageRoute(profile, entity));
+        // what time is is? could be either system or event time
```
nickwallen: The timestamp logic now lives here.  This allows it to be shared by all of the Profiler ports; Storm, REPL, and Spark.

mmiklavc:
lgtm @nickwallen. +1 via inspection.

nickwallen:
Thanks for the review. Merged.

PR #1147, opened 2018-08-07
METRON-1706: HbaseClient.mutate should return the number of mutations [Feature Branch]

nickwallen:
The `HbaseClient.mutate` method should return the number of mutations.

This is a pull request against the METRON-1699-create-batch-profiler feature branch.

## Testing

## Pull Request Checklist


merrimanr:
Looks good to me. +1

nickwallen:
Thanks for the review

PR #1148, opened 2018-08-07
METRON-1705 Create ProfilePeriod Using Period ID [Feature Branch]

nickwallen:
In some cases you need to be able to create a `ProfilerPeriod` using an epoch timestamp.  In other cases, you need to be able to create one using the period ID.  The period ID  is a monotonically increasing value which uniquely identifies each profile period.

Since both an epoch timestamp and a period ID are `Long`s, I created two separate factory methods to create a `ProfilePeriod` rather than using a public constructor directly.  This makes it much more clear as to the intended purpose.
   * `ProfilePeriod.fromTimestamp`
   * `ProfilePeriod.fromPeriodId`

## Testing

## Pull Request Checklist


justinleet:
+1, I was able to run this up and run through the readme.

nickwallen:
Many thanks... merged!

PR #1145, opened 2018-08-07
METRON-1703 Make Core Profiler Components Serializable [Feature Branch]

nickwallen:
All of the core Profiler components are now tagged as Serializable.  This allows users to use Java serialization if desired, when running the Profiler in Spark.

This is a pull request against the `METRON-1699-create-batch-profiler` feature branch.

## Testing

1. Launch the development environment.
2. Ensure alerts are created in the Alerts UI.
3. Ensure the Service Check in Ambari passes.
4. Follow the instructions in the Profiler README to create a simple profile using the Profiler topology in Storm.

## Pull Request Checklist

- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
- [x] Have you written or updated unit tests and or integration tests to verify your changes?
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [x] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

PR #1150, opened 2018-08-07
METRON-1707 Port Profiler to Spark [Feature Branch]

nickwallen:
This ports the Profiler to run in Spark.

This is dependent on the follow PRs.  By filtering on the last commit, this PR can be reviewed before the others are reviewed and merged.
- [x] #1145 
- [x] #1146 
- [x] #1148 
- [x] #1147

This does not include a mechanism to launch the Batch Profiler yet.  A follow-on PR will provide a means to launch the Batch Profiler from the command line.

### Test Existing Functionality

### Test New Functionality

1. Build and install Metron.
    ```
    mvn clean install -DskipTests -T2C
    ```

1. Edit `metron-analytics/metron-profiler-spark/src/test/resources/log4j.properties` and set `log4j.logger.org.apache.metron.profiler.spark=DEBUG`.

1. Run the `BatchProfilerIntegrationTest` either in your IDE or using Maven like so.
    ```
    cd metron-analytics/metron-profiler-spark/
    mvn -Dtest=BatchProfilerIntegrationTest test    
    ```

1. The output should look something like the following which shows that 2 profile measurements were written by the Batch Profiler.
```
Running org.apache.metron.profiler.spark.BatchProfilerIntegrationTest
2018-08-13 09:09:53 DEBUG BatchProfiler:70 - Building 1 profile(s)
2018-08-13 09:09:53 DEBUG BatchProfiler:102 - Loading telemetry from 'src/test/resources/telemetry.json'
2018-08-13 09:09:56 DEBUG BatchProfiler:109 - Found 13710 telemetry record(s)
ANTLR Tool version 4.5 used for code generation does not match the current runtime version 4.7ANTLR Runtime version 4.5 used for parser compilation does not match the current runtime version 4.7ANTLR Tool version 4.5 used for code generation does not match the current runtime version 4.7ANTLR Runtime version 4.5 used for parser compilation does not match the current runtime version 4.72018-08-13 09:09:58 DEBUG BatchProfiler:127 - Generated 13710 message route(s)
2018-08-13 09:10:00 DEBUG ProfileBuilderFunction:84 - Building a profile for group 'profile1-192.168.66.1-1704628' from 4017 message(s)
2018-08-13 09:10:00 DEBUG ProfileBuilderFunction:98 - Profile measurement created; profile=profile1, entity=192.168.66.1, period=1704628, value=4017
2018-08-13 09:10:00 DEBUG ProfileBuilderFunction:84 - Building a profile for group 'profile1-192.168.138.158-1704628' from 9693 message(s)
2018-08-13 09:10:00 DEBUG ProfileBuilderFunction:98 - Profile measurement created; profile=profile1, entity=192.168.138.158, period=1704628, value=9693
2018-08-13 09:10:00 DEBUG BatchProfiler:167 - Produced 2 profile measurement(s)
2018-08-13 09:10:00 DEBUG HBaseWriterFunction:111 - About to write profile measurement(s) to HBase
2018-08-13 09:10:00 DEBUG HBaseWriterFunction:123 - 0 profile measurement(s) written to HBase
2018-08-13 09:10:00 DEBUG HBaseWriterFunction:111 - About to write profile measurement(s) to HBase
2018-08-13 09:10:00 DEBUG HBaseWriterFunction:123 - 1 profile measurement(s) written to HBase
2018-08-13 09:10:00 DEBUG HBaseWriterFunction:111 - About to write profile measurement(s) to HBase
2018-08-13 09:10:00 DEBUG HBaseWriterFunction:123 - 1 profile measurement(s) written to HBase
2018-08-13 09:10:00 DEBUG HBaseWriterFunction:111 - About to write profile measurement(s) to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:123 - 0 profile measurement(s) written to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:111 - About to write profile measurement(s) to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:123 - 0 profile measurement(s) written to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:111 - About to write profile measurement(s) to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:123 - 0 profile measurement(s) written to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:111 - About to write profile measurement(s) to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:123 - 0 profile measurement(s) written to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:111 - About to write profile measurement(s) to HBase
2018-08-13 09:10:01 DEBUG HBaseWriterFunction:123 - 0 profile measurement(s) written to HBase
2018-08-13 09:10:01 DEBUG BatchProfiler:185 - 2 profile measurement(s) written to HBase
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.967 sec - in org.apache.metron.profiler.spark.BatchProfilerIntegrationTest
```

## Pull Request Checklist


metron-analytics/metron-profiler-spark/src/main/java/org/apache/metron/profiler/spark/BatchProfiler.java
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * Execute the Batch Profiler.
+   *
+   * @param spark The spark session.
+   * @param properties The profiler configuration properties.
+   * @param profiles The profile definitions.
+   * @return The number of profile measurements produced.
+   */
+  public long execute(SparkSession spark,
```
nickwallen: This is the main method that drives the Batch Profiler.  Its a good place to start reviewing.

simonellistonball:
Do we have to use groupByKey in the spark implementation, is it not possible to use reduceByKey to build the profiles, since profilers are by definition reducible. I've seen groupByKey cause performance problems (see https://databricks.gitbooks.io/databricks-spark-knowledge-base/content/best_practices/prefer_reducebykey_over_groupbykey.html for a good discussion on this).

nickwallen:
> @simonellistonball: Do we have to use groupByKey in the spark implementation, is it not possible to use reduceByKey to build the profiles...

I had in the back of my mind that groupByKey might not be the most performance option, but I just didn't focus any energy on that for the first pass.

I will take a look and see if we can't use your advice.  Thanks for the pointer @simonellistonball !

metron-analytics/metron-profiler-spark/src/main/java/org/apache/metron/profiler/spark/function/ProfileBuilderFunction.java
```diff
@@ -0,0 +1,107 @@
...
+    // sort the messages/routes
+    List<MessageRoute> routes = toStream(iterator)
+            .sorted(comparing(rt -> rt.getTimestamp()))
+            .collect(Collectors.toList());
+    LOG.debug("Building a profile for group '{}' from {} message(s)", group, routes.size());
+
+    // apply each message/route to build the profile
+    for(MessageRoute route: routes) {
+      distributor.distribute(route, context);
+    }
```
nickwallen: > @simonellistonball: Do we have to use groupByKey in the spark implementation, is it not possible to use reduceByKey to build the profiles...

What we do now is group by (profile, entity, period) to aggregate all of the messages needed to produce a measurement for any given profile period.  Then those messages are sorted by timestamp and applied to the profile in that order.

I didn't see an easy way to use `reduceByKey` and ensure that the messages are applied to the profile in timestamp order.  Can you think of an alternative that maintains the ordering?

nickwallen: Or maybe we not need to apply the timestamps in order?  There are no strict guarantees of ordering when the Profiler runs in Storm really. Hmm.

nickwallen:
I just refactored `BatchProfiler` hopefully to make it simpler to grok.  The functionality was not changed.  I wanted to refactor prior to additional work on @simonellistonball 's `reduceByKey` suggestion.

metron-analytics/metron-profiler-spark/src/main/java/org/apache/metron/profiler/spark/function/ProfileBuilderFunction.java
nickwallen: So assuming timestamp ordering doesn't matter, I am also trying to understand whether a groupByKey vs reduceByKey is applicable with the Dataset API.  The Catalyst engine is doing some optimization under-the-hood.  

At the very least, I might add the ability to log an explain plan, but leave the effort to optimize this as a follow-on PR.

simonellistonball: I can see how people would want ordered timestamps. If you're using profiles to find sequences of events that indicate state of an attack, you do need order. This does make it hard to use reduceByKey thinking about it, because that assumes distribution. Maybe we do have to use the shuffled method after all. There may be an argument for group and sort in a 'sub-query' and then a grouped aggregate over a deterministic ordering, which I believe you could do with spark sql easily enough, and let catalyst sort out the shuffling and sorting for you, but that's a top of the head theory at the moment.

nickwallen: What we have (which maintains ordering) is probably good enough for a first pass.  Please tell me if you disagree.

--

> @simonellistonball  I can see how people would want ordered timestamps. If you're using profiles to find sequences of events that indicate state of an attack, you do need order. 

My gut tells me not necessarily, but we probably need to dig into specific examples to see who is right. ;) Definitely worthy of digging into though.

Trying to explicitly support ordering complicates matters when streaming; no closed data set.  If you haven't set your time lag sufficiently large enough then an out-of-order messages could break your use case.

Storm's event time processing is simple, so if an out-of-order event arrives beyond your time lag, it gets dropped on the floor.  And increasing your time lag, equally increases latency which is a tough trade-off.  

Other platforms like Spark support recalculation, so we could use a higher time lag without suffering equally increased latency.  Even then, it only supports recalculation so long after the fact and does come at a cost.

All that being said (whether Storm, Spark, etc) a user's time lag setting would impact the accuracy of their results if they depend on strict ordering. If the user doesn't expect strict ordering, then we eliminate most of the problem.

I am also thinking that our goal here is to make the output produced in batch indistinguishable from that produced when streaming.  Its much easier to meet that goal if there is no expectation of ordering.

metron-analytics/metron-profiler-spark/src/main/java/org/apache/metron/profiler/spark/BatchProfiler.java
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Execute the Batch Profiler.
+   *
+   * @param spark The spark session.
+   * @param properties The profiler configuration properties.
+   * @param profiles The profile definitions.
+   * @return The number of profile measurements produced.
+   */
+  public long run(SparkSession spark,
+                  Properties properties,
```
nickwallen: This is a good place to start a review.  This is the main driver for the Spark job.

merrimanr:
I was able to understand everything by following BatchProfilerIntegrationTest.  Looks like a great start to me. +1

nickwallen:
Thanks for the review guys!  I merged this one into the feature branch.  

For those reviewers wanting further punishment, #1161 is the next PR to review for this feature branch.

PR #1151, opened 2018-08-07
METRON-1728: Handle null values in config in Pcap backend more gracefully

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1728

I modified the ConfigOptions interface to provide the ability to handle default values similar to the Map interface. ie getOrDefault. For PcapJob, I only improved/changed the values we expect to be user-supplied, ie start time, end time, etc., so we don't throw NPE's if a value gets lost in translation. I think we should definitely still add some validation to users via both REST and the UI similar to what is done in PcapCLI. 

**Note:** This PR is an improvement to existing functionality, not a regression fix.

### For all changes:

### For code changes:
  ```
  ```

n/a

#### Note:

### DIFF
--- a/metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/mr/PcapJob.java
+++ b/metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/mr/PcapJob.java
@@ -20,6 +20,7 @@ package org.apache.metron.pcap.mr;
 
 import static org.apache.metron.pcap.PcapHelper.greaterThanOrEqualTo;
 import static org.apache.metron.pcap.PcapHelper.lessThanOrEqualTo;
+import static org.apache.metron.pcap.config.PcapGlobalDefaults.NUM_REDUCERS_DEFAULT;
 
 import com.google.common.base.Joiner;
 import java.io.IOException;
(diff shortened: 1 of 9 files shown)


metron-platform/metron-pcap/src/main/java/org/apache/metron/pcap/mr/PcapJob.java
```diff
@@ -216,20 +218,22 @@ public void setCompleteCheckInterval(long interval) {
...
-      startTime = PcapOptions.START_TIME_NS.get(configuration, Long.class);
+      startTime = PcapOptions.START_TIME_NS.getOrDefault(configuration, Long.class, 0L);
     } else {
-      startTime = PcapOptions.START_TIME_MS.get(configuration, Long.class) * 1000000;
+      startTime = PcapOptions.START_TIME_MS.getOrDefault(configuration, Long.class, 0L) * 1000000;
     }
     long endTime;
     if (configuration.containsKey(PcapOptions.END_TIME_NS.getKey())) {
-      endTime = PcapOptions.END_TIME_NS.get(configuration, Long.class);
+      endTime = PcapOptions.END_TIME_NS.getOrDefault(configuration, Long.class, System.nanoTime());
```
justinleet: Nanotime isn't actually current time in nanos or anything objectively meaningful (and can vary by jvm or be negative or whatever).  Check out https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#nanoTime()

Particularly,
> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative). 

I'd just make it System.currentTimeMillis() and adjust, since I don't think there's a clean way to get a real nanos timestamp in Java (at least until 9, maybe)

mmiklavc: @justinleet good catch! Thanks.

justinleet:
+1 by inspection. Thanks for the improvement!

mmiklavc:
Closing as merged

PR #1172, opened 2018-08-22
METRON-1724: Date/time validation missing in PCAP query

tiborm:
## Contributor Comments
This PR contains date range validation. We added validation messages with hints also, to help the user correct a particular input problem.

screen shot 2018-08-15 at 3 38 30 pm

As a part of this changeset, we refactored pcap-filter.component to make it more aligned with some Angular best practices. We introduced ReactiveFormsModule to make form handling similar to other observable based solutions in Angular.
ReactiveFormsModule also helped us to remove validation logic from the HTML template, make it part of component.ts and handle more complex validations like in case of the date range.

For more details please check the following Jira ticket: https://issues.apache.org/jira/browse/METRON-1724

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-alerts/src/app/pcap/model/pcap.request.ts
+++ b/metron-interface/metron-alerts/src/app/pcap/model/pcap.request.ts
@@ -17,13 +17,13 @@
  */
 
 export class PcapRequest {
-  startTimeMs: number = 0;
-  endTimeMs: number = 150000000000000000;
-  ipSrcAddr: string = '';
-  ipSrcPort: number;
-  ipDstAddr: string = '';
-  ipDstPort: number;
-  protocol: string = '';
-  packetFilter: string = '';
-  includeReverse: boolean = false;
+  startTimeMs = 0;
+  endTimeMs = 150000000000000000;
+  ipSrcAddr = '';
+  ipSrcPort = '';
+  ipDstAddr = '';
+  ipDstPort = '';
+  protocol = '';
+  packetFilter = '';
+  includeReverse = false;
 }
--- a/metron-interface/metron-alerts/src/app/pcap/model/pcap.request.ts
+++ b/metron-interface/metron-alerts/src/app/pcap/model/pcap.request.ts
@@ -17,8 +17,8 @@
  */
 
 export class PcapRequest {
-  startTimeMs = 0;
-  endTimeMs = 150000000000000000;
+  startTimeMs: number;
+  endTimeMs: number;
   ipSrcAddr = '';
   ipSrcPort = '';
   ipDstAddr = '';
--- a/metron-interface/metron-alerts/src/app/pcap/pcap-filters/pcap-filters.component.spec.ts
+++ b/metron-interface/metron-alerts/src/app/pcap/pcap-filters/pcap-filters.component.spec.ts
@@ -16,22 +16,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { async, ComponentFixture, TestBed, fakeAsync, tick } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 
-import { PcapFiltersComponent } from './pcap-filters.component';
-import { FormsModule } from '../../../../node_modules/@angular/forms';
 import { Component, Input, Output, EventEmitter, DebugElement, SimpleChange } from '@angular/core';
-import { PcapRequest } from '../model/pcap.request';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
 
-@Component({
-  selector: 'app-date-picker',
-  template: '<input type="text" [(value)]="date">',
-})
-class FakeDatePickerComponent {
-  @Input() date: string;
-  @Output() dateChange = new EventEmitter<string>();
-}
+import { PcapFiltersComponent } from './pcap-filters.component';
+import { DatePickerModule } from '../../shared/date-picker/date-picker.module';
+import { PcapRequest } from '../model/pcap.request';
 
 describe('PcapFiltersComponent', () => {
   let component: PcapFiltersComponent;
@@ -40,10 +33,11 @@ describe('PcapFiltersComponent', () => {
   beforeEach(async(() => {
     TestBed.configureTestingModule({
       imports: [
-        FormsModule
+        FormsModule,
+        ReactiveFormsModule,
+        DatePickerModule,
       ],
       declarations: [
-        FakeDatePickerComponent,
         PcapFiltersComponent,
       ]
     })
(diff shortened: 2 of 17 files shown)


metron-interface/metron-alerts/src/app/pcap/pcap-filters/pcap-filters.component.ts
```diff
@@ -15,63 +15,116 @@
...
-import {Component, Input, Output, EventEmitter, OnInit, OnChanges, SimpleChanges} from '@angular/core';
+import {Component, Input, Output, EventEmitter, OnChanges, SimpleChanges} from '@angular/core';
+import { FormGroup, FormControl, Validators, ValidationErrors } from '@angular/forms';
+
 import * as moment from 'moment/moment';
 import { DEFAULT_TIMESTAMP_FORMAT } from '../../utils/constants';
 
 import { PcapRequest } from '../model/pcap.request';
 
+const DEFAULT_END_TIME = new Date();
```
sardell: I noticed these static values and functions are outside of the PcapFiltersComponent class. What are the benefits to doing things this way instead of keeping them inside the PcapFiltersComponent class? Or if they are intended to be shared, why don't we move them into the utils folder and import as needed?

ruffle1986: For the record, in this case, it would be ok to put these things inside the class definition because they're connected to the pcap filter so they would be hardly reused by other components, but who knows. It's hard to predict.

I'm not sure about how these kind of things work in Java or other OOP heavy programming languages but Javascript (NodeJS) doesn't enforce us to put constants and static functions inside a class. 

The functions are static pure functions. They have nothing to do with the class' instance (this) so they should not be a private or a public class methods. they could be a static functions (a property of the class object itself). From testing perspective, I would keep it as is but export them of course so we wouldn't have to create a new instance of the component to test them. The best would be if they were replaced into a separate file in the appropriate folder based on whether it's shared across other components.

simonellistonball: These names do look like they are likely to collide with other potential areas of functionality though, so if they are genuinely global concerns, then fine, otherwise they should probably belong to some form of scope related to the relevant component.

metron-interface/metron-alerts/src/app/pcap/model/pcap.request.ts
```diff
@@ -17,13 +17,13 @@
...
-  endTimeMs: number = 150000000000000000;
-  ipSrcAddr: string = '';
-  ipSrcPort: number;
-  ipDstAddr: string = '';
-  ipDstPort: number;
-  protocol: string = '';
-  packetFilter: string = '';
-  includeReverse: boolean = false;
+  startTimeMs = 0;
+  endTimeMs = 150000000000000000;
```
ruffle1986: Why the magic number?

We use this value to produce a formatted date string by using the Date constructor. But

```js
new Date(150000000000000000) // -> Invalid Date
```

so we have to do extra checks to avoid showing "Invalid Date" in the date range selector:
https://github.com/tiborm/metron/blob/METRON-1724/metron-interface/metron-alerts/src/app/pcap/pcap-filters/pcap-filters.component.ts#L59

which makes it hard to understand and maintain the code.

Using magic numbers is a bad practice because it's hard to understand their meaning. We should add a hint as a code comment at least. Or use another, more straightforward value. 
My first thought was `Date.now()` but it would introduce side effects so I let it go.

What do you think?

metron-interface/metron-alerts/src/app/pcap/pcap-filters/pcap-filters.component.ts
ruffle1986: They belong to the file's scope.

They could collide with others but we can import them in this way:

```
import { DEFAULT_END_TIME as DEFAULT_PCAP_FILTER_END_TIME } from './pcap-filter.component';
```

But I can hardly imagine that we would use them somewhere else out of this scope.

tiborm: I think the DEFAULT_START_TIME and DEFAULT_END_TIME as all uppercase variables are a bit miss leading. These are not static finals like in java. These are actually private constants. I suggest to rename them to normal camelCase. DEFAULT_END_TIME might be reusable somewhere else but DEFAULT_START_TIME in its current form really belongs to the filtering. I think the same true to the validator and the transform functions too. Validator might be reusable if we turn our other forms reactive too in the future. For me, the question is how much we like to prepare our current code to an imagined future. And how much it supports the understanding of the existing functionality. I would leave it as it is right now but opened to refactor it if we see value in it.

metron-interface/metron-alerts/src/app/pcap/model/pcap.request.ts
tiborm: I agree. These default values should be fixed. Do we needs to assign value to the dates here?

tiborm: I think it would be better to leave them undefined here and let this file to be a simple type definition.

metron-interface/metron-alerts/src/app/pcap/pcap-filters/pcap-filters.component.ts
sardell: Agreed with @tiborm on renaming those constants. In fact, the Angular Style Guide [prefers to use lower camel case (though it also suggests tolerating existing upper snake case) for constants.](https://angular.io/guide/styleguide#constants) I also prefer not to worry about an imagined future, but I do think we should move some of this out to separate files if it doesn't belong in the class to make the file more readable, if anything.

tiborm: @sardell @ruffle1986  I'll refactor out the whole PcapRequest related code from the filter. It was part of the original implementation but I think it does not belongs to here. Filter component should produce a filter output instead of a request object.

metron-interface/metron-alerts/src/app/pcap/model/pcap.request.ts
ruffle1986: It's not necessary to set these values here but I'm afraid the server relies on it as default values. So it might crash if we send it undefined values instead of magic numbers.

tiborm:
@sardell @ruffle1986 I made the following changes based on your feedback:

- default values for the timestamps removed
- DEFAULT_START_TIME and DEFAULT_END_TIME moved to constants.ts (theoretically these values are reusable)
- transform functions moved to utils.ts to make the filter component cleaner

I think this is cleaner and more aligned with the other parts of the code.

metron-interface/metron-alerts/src/app/pcap/pcap-filters/pcap-filters.component.spec.ts
```diff
@@ -179,19 +172,6 @@ describe('PcapFiltersComponent', () => {
     component.onSubmit();
   });
 
-  it('Port fields should be removed from request when set to empty', () => {
```
merrimanr: Why was this test removed?

metron-interface/metron-alerts/src/app/pcap/pcap-packet-line/pcap-packet-line.component.spec.ts
```diff
@@ -143,38 +132,4 @@ describe('PcapPacketLineComponent', () => {
       "protos": null 
     });
   });
-
-  it('should extract UDP ipSrcPort fields', () => {
```
merrimanr: Why were these tests removed?

tiborm: Merge issue. I checked these and the related code change too. Got no idea how could I missed these. Fixing.

metron-interface/metron-alerts/src/app/pcap/pcap-filters/pcap-filters.component.spec.ts
tiborm: It was intentional. We changed back the port type to string so this behavior not needed anymore.

tiborm:
As a result of the followup discussion about the placing of the transform functions, I moved them back to the filter component.

ruffle1986:
It's not just simply about sharing the same validator function. It's more complicated than that. Here's a possible fix for that: https://github.com/tiborm/metron/pull/11

tiborm:
@merrimanr Tha bug you find was very interesting. We tried two different approaches to fix, but I think @ruffle1986 solution is more suitable and elegant here. I merged it. Please take a look and give a +1 if it's good to go.

PR #1174, opened 2018-08-23
METRON-1748 Improve Storm Profiler Integration Test

nickwallen:
Improved the Storm Profiler integration tests based on improvements I made for the Spark Profiler feature branch.

* The output of the Profiler is now validated using the client library; `PROFILE_GET`.  This is better validation that things are working end-to-end.

* Using more telemetry messages to validate event time processing.

* Added logging around cache maintenence.

* Added logging of the tuple windows that are passed from Storm's WindowManager.  This can help debug issues where the time lag and window size are incorrectly set.

## Testing

This only changes the integration tests and adds logging.  Running the integration tests is sufficient to test the changes there.  Running up the development environment is sufficient for the extra caution.

## Pull Request Checklist


nickwallen:
I am seeing sporadic integration test failures from `ProfilerIntegrationTest.testProcessingTimeWithTimeToLiveFlush` with this change.  I need to track that down before this gets merged.

nickwallen:
The CI build hit what seems to be an unrelated, intermittent test failure.
```
ZKConfigurationsCacheIntegrationTest.validateUpdate:230->lambda$validateUpdate$9:230 
expected:<{hdfs={index=yaf, batchSize=1, enabled=true}, elasticsearch={index=yaf, batchSize=25, batchTimeout=7, enabled=false}, solr={index=yaf, batchSize=5, enabled=false}}> 
but was:<{}>
```
Kicking Travis.

nickwallen:
BTW - This should be ready to go.  I fixed the sporadic test failures.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
```diff
@@ -150,27 +159,98 @@ public void testProcessingTime() throws Exception {
...
+   * measurements being flushed.
+   *
+   * (2) If no messages arrive to advance time, then the "time to live" mechanism will flush a profile
+   * after a period of time.
+   *
+   * <p>This test specifically tests the *second* mechanism when a profile is flushed by the
+   * "time to live" mechanism.
+   */
+  @Test
+  public void testProcessingTimeWithTimeToLiveFlush() throws Exception {
```
mmiklavc: Might it be better to have the core of this test's javadoc documentation in the main README and simply reference the core terminology from the test? Actually, your test name already has that info (test flushing via TTL), so even that is probably redundant. My concern is that the info is encoded in all of:
* test code
* test javadoc
* source code
* source javadoc (maybe it's not, but I'm assuming there's also doc there)
* project README's

I think expressing the functionality in test/source code is of course good because it's executable. And README's are great as a set of expectations about how the project works. But my concern would be that if this functionality gets updated by someone other than you, then they'll have to be sure it's updated in 5 places. If a test gets updated it might be easy to miss updating associated javadoc of this detail. It's the kind of thing that Bob Martin, Martin Fowler, and Kent Beck are constantly railing about wrt stale docs.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
```diff
@@ -150,27 +159,98 @@ public void testProcessingTime() throws Exception {
...
+
+    // upload the config to zookeeper
```
mmiklavc: Can we remove the redundant javadoc? Your code speaks for itself here. I'd probably just rename that method precisely as "uploadConfigToZookeeper". Very clear and with 1 less non-executable line.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
```diff
@@ -150,27 +159,98 @@ public void testProcessingTime() throws Exception {
...
 
-    // validate what was flushed
-    List<Integer> actuals = read(
-            profilerTable.getPutLog(),
-            columnFamily,
-            columnBuilder.getColumnQualifier("value"),
-            Integer.class);
-    assertEquals(1, actuals.size());
+    // the profile should count at least 3 messages
+    assertTrue(actuals.size() > 0);
```
mmiklavc: For my own understanding, is the List that's returned as `actuals` a List of profile periods? I looked at the profiler.json in test resources and it looks like it's counting the number of messages that have ip_src_addr. Is that correct? Under what condition(s) would actuals.size() != 1 for the purposes of this test? I think that having a clear understanding of the associated profile for this test and why I would expect a value of 3 (as it's not just bc of the flushing mechanism) would be useful for future maintainers.

Along those lines, as a small recommendation for readability (by no means a firm requirement here), but you might consider using the multiline string annotation that we've leveraged in other areas of code. It allows the config to sit right along side the test code itself. Just easier to read imo when the size of the multiline string is still manageable/small.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
```diff
@@ -186,21 +266,40 @@ public void testEventTime() throws Exception {
...
+    List<String> messages = FileUtils.readLines(new File("src/test/resources/telemetry.json"));
+    kafkaComponent.writeMessages(inputTopic, messages);
+
+    long timestamp = System.currentTimeMillis();
+    LOG.debug("Attempting to close window period by sending message with timestamp = {}", timestamp);
+    kafkaComponent.writeMessages(inputTopic, getMessage("192.168.66.1", timestamp));
+    kafkaComponent.writeMessages(inputTopic, getMessage("192.168.138.158", timestamp));
+
+    // create the 'window' that looks up to 5 hours before the last timestamp contained in the telemetry
+    assign("lastTimestamp", "1530978728982L");
```
mmiklavc: What's the significance of this specific value `1530978728982L`? Is that tied to the last record timestamp in the telemetry input data file?

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
```diff
@@ -150,27 +159,98 @@ public void testProcessingTime() throws Exception {
...
-            Integer.class);
-    assertEquals(1, actuals.size());
+    // the profile should count at least 3 messages
+    assertTrue(actuals.size() > 0);
     assertTrue(actuals.get(0) >= 3);
   }
 
+  /**
+   * The Profiler can generate profiles based on processing time.  With processing time,
+   * the Profiler builds profiles based on when the telemetry is processed.
```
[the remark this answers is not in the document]
nickwallen: The same.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
nickwallen: > Along those lines, as a small recommendation for readability (by no means a firm requirement here), but you might consider using the multiline string annotation that we've leveraged in other areas of code.

I could certainly add that.  This is just how the tests were originally written.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
nickwallen: Ok, fair point. Let me see if I can't move this to the README and clean it up.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
mmiklavc: On looking at that data file more closely, it looks like the records are not in time order. That TS shows up on line 92 in the telemetry file rather than the last line. Is that expected? I think I'm misunderstanding something. Can you elaborate on that a bit more?

nickwallen: Really what I mean is its the latest/most recent timestamp. The records are not expected to be in order.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
nickwallen: I removed the javadocs from the integration test and put it in the bolt's javadoc.  It seemed to fit there nicely. 

Hopefully this satisfies your valid point around DRY.  But let me know if there is something else I can do.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
nickwallen: Done

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
nickwallen: > I could certainly add that. This is just how the tests were originally written.

So I looked at this.  I'd rather not do this now, only because the ConfigUploader component doesnt just let me pass it a String to upload to Zk.  I need to give it a path.  But I would definitely like to come back to this and embed the profile definitions in the test like you suggested.

nickwallen: >  I looked at the profiler.json in test resources and it looks like it's counting the number of messages that have ip_src_addr. Is that correct? Under what condition(s) would actuals.size() != 1 for the purposes of this test?

Hopefully the latest commit clears up your questions around this.

metron-analytics/metron-profiler/src/test/java/org/apache/metron/profiler/integration/ProfilerIntegrationTest.java
```diff
@@ -127,94 +127,53 @@
...
-      // retrieve the profile measurement using PROFILE_GET
-      actuals = execute(profileGetExpression, List.class);
+      // try again to retrieve the profile measurement using PROFILE_GET
+      measurements = execute(profileGetExpression, List.class);
     }
 
-    // the profile should count at least 3 messages
-    assertTrue(actuals.size() > 0);
-    assertTrue(actuals.get(0) >= 3);
+    // expect to see only 1 measurement, but could be more (one for each period) depending on
```
mmiklavc: Ok, that makes sense - that's more what I meant when I said "period" in my original inquiry. It's a measurement that is associated with the k-eth period.

mmiklavc:
This looks great @nickwallen. Thanks for the contribution and for taking the time with my feedback. +1

nickwallen:
Great.  I appreciate the review.  Very helpful feedback.

PR #1012, opened 2018-05-12
METRON-1551 Profiler Should Not Use Java Serialization

nickwallen:
When running the Profiler in a topology where serialization occurs, the following error happens.  This can occur when the number of workers is greater than 1.

The topology should not be using Java serialization for serializing tuple values as this will negatively impact performance. 

```
2018-05-09 10:48:35.136 o.a.s.d.executor [ERROR] 
java.lang.RuntimeException: java.lang.RuntimeException: java.io.NotSerializableException: org.apache.metron.common.configuration.profiler.ProfileResult
 at org.apache.storm.utils.DisruptorQueue.consumeBatchToCursor(DisruptorQueue.java:485) ~[storm-core-1.1.0.2.6.4.0-91.jar:1.1.0.2.6.4.0-91]
 at org.apache.storm.utils.DisruptorQueue.consumeBatchWhenAvailable(DisruptorQueue.java:451) ~[storm-core-1.1.0.2.6.4.0-91.jar:1.1.0.2.6.4.0-91]
 at org.apache.storm.disruptor$consume_batch_when_available.invoke(disruptor.clj:73) ~[storm-core-1.1.0.2.6.4.0-91.jar:1.1.0.2.6.4.0-91]
 at org.apache.storm.disruptor$consume_loop_STAR_$fn__7183.invoke(disruptor.clj:83) ~[storm-core-1.1.0.2.6.4.0-91.jar:1.1.0.2.6.4.0-91]
 at org.apache.storm.util$async_loop$fn__553.invoke(util.clj:484) [storm-core-1.1.0.2.6.4.0-91.jar:1.1.0.2.6.4.0-91]
 at clojure.lang.AFn.run(AFn.java:22) [clojure-1.7.0.jar:?]
 at java.lang.Thread.run(Thread.java:748) [?:1.8.0_162]
Caused by: java.lang.RuntimeException: java.io.NotSerializableException: org.apache.metron.common.configuration.profiler.ProfileResult
 at org.apache.storm.serialization.SerializableSerializer.write(SerializableSerializer.java:41) ~[storm-core-1.1.0.2.6.4.0-91.jar:1.1.0.2.6.4.0-91]
 at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:628) ~[kryo-3.0.3.jar:?]
 at com.esotericsoftware.kryo.serializers.CollectionSerializer.write(CollectionSerializer.java:100) ~[kryo-3.0.3.jar:?]
[...]

PR #1178, opened 2018-08-28
METRON-1757 Storm Profiler Serialization Exception

nickwallen:
### Description

* When using Kryo serialization, specifically in Storm, the class being serialized must have a no-arg constructor.

* The existing unit tests that validated Kryo serialization for the Profiler classes (that I added for #1012 ) use the `SerDeUtils` class to test Kryo serialization.  I found that this class sets up Kryo in a slightly different way from how Storm does internally.  This is one reason why the tests did not catch the missing no-arg constructor.

* The `ProfileMeasurementTest`'s setup method was missing the `@Before` annotation so it was never called.  This caused the test to validate serialization of a null, rather than an actual object.  Doh!

* Removed the 'asm' dependency from 'metron-profiler-common'.  Conflicting version of this library was causing issues with the Kryo serialization tests (once they actually tested serializing something other than nulls).  There should now be no dependency on Storm from `metron-profiler-common`, which is how it should be.

* The unit tests now catch this bug and highlighted a few other classes that needed a no-arg constructor. 

* The only piece left wanting is that I'd like to have seen the integration tests fail if Kryo serialization fails.  That was the original intent of using `topology.testing.always.try.serialize=true` for the integration test.  But that only seems to catch Java serialization problems, not Kryo for some reason.

* This serialization error will never show itself if you are running with a single worker as there is no need for Storm to serialize.  Even running the Profiler in a cluster on multiple workers, it will only show-up if Storm runs the `ProfileBuilderBolt` and the downstream bolts (`HBaseBolt` or `KafkaBolt`) on a separate worker.  This is why you may not always see the problem when running on a cluster (and why it was confusing to track down.)

### Work Around

If you are experiencing this issue, you can work around the problem by changing the following values in your `profiler.properties`.  This will cause Storm to fall-back to Java serialization when Kryo serialization fails in Storm for `org.apache.metron.common.configuration.profiler.ProfileResult`.

```
topology.fall.back.on.java.serialization=true
topology.kryo.register=[ org.apache.metron.profiler.ProfileMeasurement, \
    org.apache.metron.profiler.ProfilePeriod, \
    org.apache.metron.common.configuration.profiler.ProfileResultExpressions, \
    org.apache.metron.common.configuration.profiler.ProfileTriageExpressions, \
    org.apache.metron.common.configuration.profiler.ProfilerConfig, \
    org.apache.metron.common.configuration.profiler.ProfileConfig, \
    org.json.simple.JSONObject, \
    org.json.simple.JSONArray, \
    java.util.LinkedHashMap, \
    org.apache.metron.statistics.OnlineStatisticsProvider ]
```

### Testing

1. Spin-up Full Dev

1. Create  a profile by following the Profiler README as a guide.

1. Spin-up this change on a multi-node cluster.  Run the profiler with multiple workers.  Create the same profile as before and ensure that values can be persisted in HBase.

## Pull Request Checklist

### DIFF
--- a/metron-analytics/metron-profiler-common/pom.xml
+++ b/metron-analytics/metron-profiler-common/pom.xml
@@ -109,26 +109,6 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>org.apache.storm</groupId>
-            <artifactId>storm-core</artifactId>
-            <version>${global_storm_version}</version>
-            <scope>provided</scope>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.apache.logging.log4j</groupId>
-                    <artifactId>log4j-api</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.apache.logging.log4j</groupId>
-                    <artifactId>log4j-core</artifactId>
-                </exclusion>
-                <exclusion>
-                    <artifactId>log4j-slf4j-impl</artifactId>
-                    <groupId>org.apache.logging.log4j</groupId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-all</artifactId>
             <version>${global_mockito_version}</version>
--- a/metron-analytics/metron-profiler-common/pom.xml
+++ b/metron-analytics/metron-profiler-common/pom.xml
@@ -55,6 +55,10 @@
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-log4j12</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>asm</groupId>
+                    <artifactId>asm</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
--- a/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/ProfileMeasurement.java
+++ b/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/ProfileMeasurement.java
@@ -20,6 +20,8 @@
 
 package org.apache.metron.profiler;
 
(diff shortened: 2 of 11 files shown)


JonZeolla:
While you're at it, what do you think about making a tweak to the related instructions?  Updating [this](https://github.com/apache/metron/tree/master/metron-analytics/metron-profiler#deploying-profiles-with-the-stellar-shell) to reference `source /etc/default/metron` and then use `$ZOOKEEPER` instead of `ZK:2181` so the steps more copy/pastable.

nickwallen:
Sure. Done.

JonZeolla:
Looks good.  I wasn't able to finish my testing last night but I am going to take a stab at validating.  Based on your comments above though this was great digging, thanks.

cestella:
This is really good work, @nickwallen.  I'm +1 by inspection here, pending @JonZeolla's +1

JonZeolla:
This looks good to me +1 thanks @nickwallen

nickwallen:
Here is one other detail that I forgot to add to the description.  This may be important if you are trying to validate this fix on a multi-node cluster.

* This serialization error will never show itself if you are running with a single worker as there is no need for Storm to serialize.  

* Even when running the Profiler in a cluster on multiple workers, it will only show-up if Storm runs the `ProfileBuilderBolt` and the downstream bolts (`HBaseBolt` or `KafkaBolt`) on a separate worker.  This is why you may not always see the problem when running on a cluster (and why it was confusing to track down.)

ottobackwards:
The work-around to this issue, and some documentation of it to the extent you feel necessary should go out to the users list.

nickwallen:
FYI - I tested this on a multi-node cluster.  I was experiencing the problem and the topology could not write out any profile measurements. 

I updated the Profiler lib with these changes and restarting the topology.  After the restart, I was able to successfully write out measurements to HBase. I also ensured that the `ProfileBuilderBolt` and `HBaseBolt` were running on different workers during this test.

PR #1181, opened 2018-08-31
METRON-1762: Change default Elasticsearch field rename behavior to NOOP

merrimanr:
## Contributor Comments
This PR changes the default field renaming behavior in the ElasticsearchWriter to NOOP.  Metron uses '.' characters in field names by default so this means the separator in ES field names is changed from ':' to '.'.  While the this specific change was simple, it cascaded to several different parts of the code base.  This is the reason for the rather large PR.

One important implication I would like to point out is that changing field names to '.'s causes ES to interpret them as object field types.  There are various places in this PR where I had to account for that and is something reviewers should keep in mind while testing.  This is a fairly significant change and it makes me a little nervous.

### Changes Included
- Change the FieldNameConverter to default to NOOP instead of DEDOT
- Updated constants (MetaAlertConstants.THREAT_FIELD_DEFAULT, etc) to use '.'s
- Changed the legacy `source:type` field to `metron_sensor_type`.  I was not able to change it to `source.type` because the bro template defines a `source` field.  ES does not allow this (see my object field type comment above):  https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html#_upgrading_fields_with_dots_to_5_x.  The `metron_sensor_type` is similar to other dedicated fields like `metron_alert` so it made sense to me.  There is really no reason we need to have a separate in that field name anyways.
- Change fields in tests and test data to use '.' instead.  I only changed this in tests that were failing so there still may be tests with ':' in field names.
- Changed the field names in the alerts UI code to use '.' instead of ':' and also made the `metron_sensor_type` change.  I tried to replace hard-coded field names with constants where appropriate to make this easier in the future (hopefully we never have to do this again).
- Updated ES templates to use the new field name conventions and updated Solr schemas to use the new `source.type` name.  ES dynamic templates also required a switch from the `match` attribute to `path_match` (see my object field type comment above).
- I had to significantly change the ElasticsearchColumnMetadataDao class to pull field names out of the column metadata response recursively since they are now nested (see my object field type comment above).  For example, without this change `threat.triage.score` would be returned only as `threat` because there are other field names that start with `threat.triage.*`.  I also updated ElasticsearchSearchIntegrationTest to test for this case.
- Updated the global config `source.type.field` and `threat.triage.score.field` settings to match the new field names.  These settings are probably useless now.
- Updated the Solr README by removing some steps that are now unnecessary for enabling Solr in Metron.

### Testing
I am still working my way through testing.  I have run several tests in full dev including:

- Verified data in bro and snort indexes looks correct.
- Verified the basic features in the alerts UI still work.  This includes all fields showing up and being able to filter on the fields that were updated to follow the new naming format.
- Verified the column metadata REST endpoint returns fields and types as expected.  As a bonus, dynamic fields that have been indexed are now returned with the correct type (I don't think this worked before).
- Verified the metaalert feature still works correctly
- Verified switching to Solr works as expected and performing the tests listed above

I'm sure there are test cases I'm missing since this is change cuts across so many parts of Metron.  Still trying to think of more but feel free to add to this list.

### Next Steps
There are a couple outstanding items and unfinished tasks:

- I only updated the Solr README.  There are several places in our READMEs that reference fields with the ':' separator.  Is it worth updating our docs to reflect the changes to our field names?  I am still looking through the READMEs for spots that obviously need updating (related to `metron_sensor_type` specifically).
- The change to the `metron_sensor_type` field name is reflected in the UI.  Is this ok?
- Should we review and update field names in all tests, even unit tests where the actual field name isn't crucial?

Thanks in advance for reviewing this beast.  Hopefully the complexity and size of this PR is worth the benefits it offers.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

- [] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

  ```
  cd site-book
  mvn site
  ```

#### Note:


mmiklavc:
What's the state of this item?

PR #1184, opened 2018-09-04
METRON-1761, allow application of grok statement multiple times

ottobackwards:
This PR adds support for incoming messages to grok parsers that have multiple log lines.

Instead of having to split the logs before sending to metron/kafka, you could just send the logs in batches.

# todo testing

### For all changes:

### For code changes:
- [-] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
  ```
  ```

  ```
  cd site-book
  mvn site
  ```


metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/MultiLineGrokParserTest.java
```diff
@@ -0,0 +1,112 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class MultiLineGrokParserTest extends GrokParserTest {
```
mmiklavc: Comparing this test with the base class and `SampleGrokParserTest`, there doesn't seem to be any value in having this extend the abstract `GrokParserTest` class. The `test()` method is overridden, and the other nugget of value, `compare(...)`, is not used. I think your test would be much cleaner and succinct if it was simply separated from the 1-line-per-message test.

mmiklavc: Also, can you add some tests around parsing failure scenarios in the batch case?

justinleet:
@ottobackwards Any update on this? I think it's valuable to have in.

ottobackwards:
I am sorry, I missed the comments on this.  I will try to have something soon.

ottobackwards:
@mmiklavc wrt failing the whole message or some sort of partial failure scheme.  I don't like failing multiples if we _can_ parse some lines, but I don't see a good way to cleanly handle it.  Or we need to talk it through.

We could track exceptions and lines processed.... and if we got an exception for every line fail the file, else log the execeptions and a message about only parsing x of y lines and pass back a partial message list.
Thoughts?

mmiklavc:
@ottobackwards Yeah, I took a look through the ParserBolt and the MessageParser interface. We process a raw byte array. A clean way of doing that isn't immediately obvious to me either. Perhaps these semantics could be documented for the parser? Minimally, I think we should describe the 2 cases I outlined above.

ottobackwards:
If you think the "every line fails" == fail, some fails = emit and log works, we can do that, but I don't know how or if we want to put things in the error stream.  I need to look at it more closely.  I think that solving this might be beyond this pr, with a documentation note as you state.

ottobackwards:
@mmiklavc Can you take a look at the parser.validate() stuff in the bolt?  Maybe the answer is put a dummy invalid record in there and fail validation for each parse failure?

ottobackwards:
@mmiklavc I looked through the validation stuff more, I think that validation is the way to go here.  The grok parser will add invalid message for each exception, parser failure, and then in the validation call fail those messages.  It will have to be done so that the returned message makes sense when it is sent to the error topic.  

What do you think?   @cestella ?

mmiklavc:
@ottobackwards I see what you're saying. It looks like that could definitely work. Thinking out loud here, but might that conflate the semantics of our validation a bit? Validate currently does things like ensure that a timestamp exists on the message, though I don't see why we couldn't expand it to validations outside of our global Metron context.

One class that might be worth checking out is the unified enrichment topology. This was changed to include a parallel enricher that handles errors and message results in an EnrichmentResult class.

1. https://github.com/apache/metron/blob/master/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/bolt/UnifiedEnrichmentBolt.java#L270
2. https://github.com/apache/metron/blob/master/metron-platform/metron-enrichment/src/main/java/org/apache/metron/enrichment/parallel/ParallelEnricher.java#L63

It looks to me like there might be some possible collaboration opportunity and overlap with the work you're doing here and the work @merrimanr is doing on this PR - https://github.com/apache/metron/pull/1213#pullrequestreview-161248142

I'm just wondering if we might be able to kill 2 birds with one stone. We probably don't want to change the MessageParser interface, but maybe we can manage the bulk processing through a more generalized bridge between the ParserBolt and parser implementations. I haven't dug too deep into implementation feasibility, but it seems worth considering.

ottobackwards:
@mmiklavc  but we don't have messages to split, we have bytes.  If we where going to leave the 'parser's as single object -> single result | single exceception', ie not change the interface and not subvert validate, then we would have to introduce 'splitStrategies' at the bolt.

mmiklavc:
@ottobackwards Yeah, understood about the message vs raw bytes. It seems like there are 2 options here - use validate or modify the message parser interface to return a List of something other than a `List<JSONObject>`. If you were to go the validate route, how would you modify and return the messages that failed to parse so that validate can handle them? I'd expect them to be indexed to the error topic in the same style as the single message parsers.

mmiklavc:
Looking at the code in parserbolt, I think the only thing missing with the validate route is the original parser exception. Here's how errors are handled when validation error occurs:
```
ErrorUtils.handleError(collector, error);
```
here's how they're handled when an exception occurs during parsing:
```
} catch (Throwable ex) {
    handleError(originalMessage, tuple, ex, collector);
}
```

ottobackwards:
So the idea would be that the JSONObject returned for the failed line ( that would be passed to handle error ) would be a new object that had the raw line, the exception type, the exception message in it.
That would all get passed through to the Error stuff.  We could put anything we wanted in there.

mmiklavc:
@ottobackwards That sounds reasonable to me. One thing I would definitely add to the documentation for this is some comment about how it changes processing semantics and might require additional tuning. Since we now have batch sizing specified for the writers to Kafka, there may be a little bit of tweaking there, but I'm just not sure.

ottobackwards:
It think the API should be improved ideally, but that is in the future

ottobackwards:
Let me give it a shot, I'll document the semantics of the failure mode and we can look again

ottobackwards:
If it turns out that this is just such a duck tape job, we can always close the PR and open a jira for the new api

ottobackwards:
actually, I'm just going to close this.  Once I step back from "how could I do this" to look at the big picture, it doesn't seem like a good idea.  This is not a field issue, more of an improvement, so there is no need to hammer something in.

ottobackwards:
Closing this pr. I will create a jira for api improvement

ottobackwards:
re-opening for input

lvets:
So I talked @ottobackwards into initially adding this feature 😄 The reason I asked for this improvement is that have a bunch of log files which contain a relevant entry per line. I currently have to rely on 3rdparty tools to split these logfiles into messages (where 1 log line == 1 message) before sending them to Metron. I was hoping to just ingest the log and Metron would take care of this.
To me, relying on 3rdparty tools for what I assume to be a normal use case for a SIEM seems a bit strange.

merrimanr:
I prefer @mmiklavc's suggestion several comments back:  return a List of something other than a List<JSONObject>.  Could we use a wrapper that can contain results and errors?  The calling class could then determine the best way to handle errors without having to fail the whole batch.

ottobackwards:
@merrimanr @mmiklavc First pass of what we discussed

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/MultiLineGrokParserTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Test that if a byte[] with multiple lines of log is passed in
+   * it will be parsed into the correct number of messages.
+   * @throws IOException if we can't read from disk
+   * @throws ParseException if we can't parse
+   */
+  @Test
+  @SuppressWarnings("unchecked")
+  public void test() throws IOException, ParseException {
```
mmiklavc: Can you provide a descriptive name here?

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/MultiLineWithErrorsGrokParserTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Test that if a byte[] with multiple lines of log is passed in
+   * it will be parsed into the correct number of messages.
+   * @throws IOException if we can't read from disk
+   * @throws ParseException if we can't parse
+   */
+  @Test(expected = RuntimeException.class)
+  @SuppressWarnings("unchecked")
+  public void test() throws IOException, ParseException {
```
mmiklavc: Can you provide a descriptive name here?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -383,7 +408,7 @@ public void execute(Tuple tuple) {
     }
   }
 
-  protected void handleError(byte[] originalMessage, Tuple tuple, Throwable ex, OutputCollector collector) {
+  protected void handleError(Object originalMessage, Tuple tuple, Throwable ex, OutputCollector collector) {
```
mmiklavc: Does this need to be relaxed to Object? It looks to me like byte[] is still used everywhere for original message.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/GrokParser.java
```diff
@@ -134,26 +144,102 @@ public void init() {
...
+          message.remove(patternLabel);
+          postParse(message);
+          messages.add(message);
+          LOG.debug("Grok parser parsed message: {}", message);
+        } catch (Exception e) {
+          LOG.error(e.getMessage(), e);
+          errors.put(originalMessage, e);
+        }
+      }
+    } catch (IOException e) {
```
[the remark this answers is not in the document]
ottobackwards: That is right

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
ottobackwards: Sorry, that was from an earlier rev. where I was using that.

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/MultiLineWithErrorsGrokParserTest.java
ottobackwards: done

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/MultiLineGrokParserTest.java
ottobackwards: done

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
```diff
@@ -31,23 +35,41 @@
...
   }
 
+  /**
+   * Take raw data and convert it to messages.  Each raw message may produce multiple messages and therefore
+   * multiple errors.  A {@link MessageParserResult} is returned, which will have both the messages produced
+   * and the errors.
+   * @param parseMessage the raw bytes of the message
+   * @return Optional of {@link MessageParserResult}
+   */
+  default Optional<MessageParserResult<T>> parseOptionalResult(byte[] parseMessage) {
```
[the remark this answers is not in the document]
ottobackwards: Right, the Optional interface is the newer, so I 'built' on that approach.

mmiklavc:
This looks good to me @ottobackwards. I think the only thing left is some docs telling users how to add both existing and custom parsers to take advantage of this feature.

ottobackwards:
@mmiklavc please see latest commit

ottobackwards:
@merrimanr any comment?

merrimanr:
+1 from me too.  I will merge this into https://github.com/apache/metron/pull/1213 once it's in master and we can continue moving forward there.  Parsing is getting an upgrade!

PR #1194, opened 2018-09-12
METRON-1775 Transient exception could prevent expired profiles from being flushed

nickwallen:
In the Storm Profiler, the ProfileBuilderBolt uses a separate thread to flush expired profiles.  Expired profiles are those that have not received a message for an extended period of time.  That period of time is called the profile time-to-live (profiler.ttl).  This is done to avoid potential memory leaks.

This separate thread is scheduled using `java.util.concurrent.ScheduledExecutorService#scheduleAtFixedRate`.  The documentation for `scheduleAtFixedRate` says that if an exception is thrown, subsequent executions will be suppressed.  We do not want this to happen.  If an exception occurs, we need to log it and continue, ensuring that expired profiles will continue to flush.

I have no direct proof that this has caused a problem, but it certainly is a potential issue that should be addressed.

## Testing

There is not an easy way to manually trigger this in a development environment.  I have added a unit test to exercise the problem.

## Pull Request Checklist

### DIFF
--- a/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
+++ b/metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
@@ -20,6 +20,7 @@
 
 package org.apache.metron.profiler.bolt;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.curator.RetryPolicy;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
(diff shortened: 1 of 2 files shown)


metron-analytics/metron-profiler/src/main/java/org/apache/metron/profiler/bolt/ProfileBuilderBolt.java
```diff
@@ -372,15 +372,21 @@ protected void flushActive() {
...
-      measurements = messageDistributor.flushExpired();
-      emitMeasurements(measurements);
+    } catch(Throwable t) {
+      // need to catch the exception, otherwise subsequent executions would be suppressed.
+      // see java.util.concurrent.ScheduledExecutorService#scheduleAtFixedRate
+      LOG.error("Failed to flush expired profiles", t);
     }
 
-    LOG.debug("Flushed expired profiles and found {} measurement(s).", measurements.size());
+    LOG.debug("Flushed expired profiles and found {} measurement(s).", CollectionUtils.size(measurements));
```
mmiklavc: Just a note - CollectionUtils.size() is null-safe. I'm assuming that's the reason for the tweak in implementation here.

https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#size-java.lang.Object-

nickwallen: Yep, exactly.

PR #1207, opened 2018-09-21
METRON-1695: Expose pcap properties through Ambari

anandsubbu:
## Contributor Comments
This change makes the PCAP config properties available via the Metron service -> Configs tab in Ambari. It also takes care of updating the `pcap.properties` file with relevant values depending on the deploy environment (single node vs. multi node vs. kerberized cluster). 

Here's a screenshot of the PCAP Config:
pcap-config-values

### Testing Steps
* Launch full dev
* Metron PCAP service should come up successfully.
* Go to Ambari -> Metron services -> Configs -> PCAP tab; change any config parameter; 
* Restart the required services and the config change take effect.

### Testing Done
I ran the following different tests to validate the fix.

#### A. Full Dev
* Verified Metron PCAP service comes up successfully 
* Changed some of the PCAP config values from the Ambari UI, validated that this is reflected in the `pcap.properties` file upon service restart.
- **Note**: I ran into [METRON-1786](https://issues.apache.org/jira/browse/METRON-1786) during testing. I was able to resolve the issue by applying the fix in PR #1206 

#### B. Multi-node Cluster
* Validated that the `kafka.zk` property in `pcap.properties` is updated with the correct Broker details
* Changed a PCAP service-related property (E.g. `kafka_pcap_numpackets`) and validated that the `pcap.properties` file is updated.
* Changed a PCAP REST-related property (E.g. `pcap_page_size`) and validated that the `/etc/default/metron` file is updated.

#### C. Kerberized Cluster
* Kerberized the multi-node cluster and setup pycapa
* Validated that `topology.worker.childopts` and `topology.auto-credentials` are 
* Ingested PCAP into the cluster
* Ran queries through the PCAP panel and validated that the modified config settings are in effect (E.g. `pcap_page_size` set to 50). 

## Pull Request Checklist

### For all changes:

### For code changes:
- [NA] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
```diff
@@ -387,18 +388,32 @@
 bolt_hdfs_rotation_policy_units = config['configurations']['metron-indexing-env']['bolt_hdfs_rotation_policy_units']
 bolt_hdfs_rotation_policy_count = config['configurations']['metron-indexing-env']['bolt_hdfs_rotation_policy_count']
 
-# Pcap
+# PCAP
 metron_pcap_topology = 'pcap'
 pcap_input_topic = 'pcap'
```
MohanDV: pcap input topic is also exposed , so shouldn't we be getting the value from config tab ?

MohanDV: pcap input topic is also hardcoded in [status_param.py ](https://github.com/MohanDV/metron/blob/a46fda45b46a9f317e2ef7c3755ce6f072ed893b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/status_params.py#L123)  we should get the value from config tab here too.

nickwallen:
This is causing one problem on Full Dev. We have 6 slots in Full Dev, but since adding the Pcap Topology, we now require 7 slots for all of the topologies to run.  

What ends up happening is that the 'random_access_indexing' often gets starved.  When spinning up Full Dev, no alerts are visible in the Alerts UI.  We want alerts to be immediately visible as this is a standard validation step.  

We need to figure out a solution around this, which I am not sure what that looks-like right now.  Add a slot?  Do not immediately start 'pcap' topology?  Other?

ottobackwards:
We have an overall deficiency here I believe.  The Slot management is a oft forgotten and manual task, for adding new parsers from the ui etc, you ( I believe ) still have to remember to have already added a slot before starting your parser topology for example.  We may want to consider an overall underlying approach to this that is usable where we fall short.

nickwallen:
> @ottobackwards The Slot management is a oft forgotten and manual task, for adding new parsers from the ...

Sure, worthy of a discuss.  

@anandsubbu I would just drop "snort" as a default for the parser configuration in Ambari.  Then we save a slot and get around this issue for now.  As a follow-on, we could explore other alternatives.

mmiklavc:
Might be worth looking at how we set what spins up by default. For instance, we might not want Pcap to start  by default in *any* installation, let alone full dev. Is there an option in the mpack config for one, or two, is there a way to disable it via our blueprints? This way we don't alter the existing set of topologies. One reason for this is that it's at least mildly useful to have 2+ sensors worth of data writing out to ES that can be viewed in Kibana. And while pcap doesn't go to indexing anyhow, we don't have anything generating data into the pcap topology currently at all. It would just be taking up a slot and doing nothing.

nickwallen:
@anandsubbu I am also running into a separate issue.  After some period of time Ambari will start reporting the status of the pcap topology as stopped, even though it remains running in Storm.  Are you able to replicate this?  I am digging through the logs trying to find some sort of stack trace, but have not found anything yet.

ottobackwards:
@mmiklavc I agree with that, but I also think that we should separate the default sensors from the regular metron install, and have them be their own optional thing.  IE> you would choose to install the snort , yaf, bro things we install automatically now.

anandsubbu:
Latest commit modifies full dev to start only the bro parser topology by default. The PCAP topology now gets the worker slot which was freed up. Note that on a regular deploy, all of bro, snort and yaf topologies will be deployed by default. 

I am also looking for the issue @nickwallen mentioned about the PCAP topology appearing stopped. Nothing so far. Will continue to monitor.

MohanDV:
PCAP related properties needs to be taken out from the [rest_application](https://github.com/apache/metron/blob/master/metron-interface/metron-rest/src/main/config/rest_application.yml#L57). Also the dependency of PCAP on rest settings to be removes as Changing a REST Setting triggers Ambari to [Restart](https://hortonworks.jira.com/browse/BUG-111762) PCAP Topology .

anandsubbu:
@nickwallen suggested a great idea of making use of parser aggregation and thus starting the default parsers on full-dev as an aggregated parser topology. The latest commit makes use of PR #1215 which provides that ability.

Now, on full-dev we start a single aggregated topology for the default configured sensors viz. "bro,snort,yaf".

## Testing Done
- Spun up full-dev and noticed that a single topology named `bro__snort__yaf` is started
- Verified Alerts UI has data from all three sensors:

image

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/status_params.py
```diff
@@ -119,8 +119,8 @@
 metron_keytab_path = config['configurations']['metron-env']['metron_service_keytab']
 
 # Pcap
-metron_pcap_topology = 'pcap'
-pcap_input_topic = 'pcap'
+metron_pcap_topology = config['configurations']['metron-pcap-env']['spout_kafka_topic_pcap']
```
nickwallen: Quick fix needed.
```
metron_pcap_topology = 'pcap'
```

anandsubbu: Done, changed.

nickwallen:
+1 Thanks @anandsubbu !

anandsubbu:
Thanks much @nickwallen for giving it another spin!

Hey @MohanDV , let me know if my below explanation makes sense. @nickwallen , would you have any thoughts on this?

> @MohanDV , as a part of my commit, I have moved all the UI related elements from REST to the PCAP config tab.
> 
> In my understanding, the following parameters are not directly related to the PCAP core service, but are functionally related to REST (used by the PCAP query panel UI):
> 
> * pdml.script.path
> * base.path
> * base.interim.result.path
> * final.output.path
> * page.size
> * yarn.queue
> * finalizer.threadpool.size
> 
> For this reason, I did not move these properties out of the rest_application.yml.
> 
> > @MohanDV Changing a REST Setting triggers Ambari to Restart PCAP Topology
> 
> With my current changes, modifying any config setting in the PCAP tab will require restarting both Metron PCAP and Metron REST services--which is necessary since user can change a PCAP-service related property or a PCAP-REST related property (or both). The converse, which you noted, i.e. changing any REST config does not require restarting the PCAP topology. Here's a screenshot after making changes to REST config...
> 
> image

anandsubbu:
Cool, thanks for your opinion @nickwallen . I will wait for @MohanDV 's thoughts as well before merging.

MohanDV:
> Cool, thanks for your opinion @nickwallen . I will wait for @MohanDV 's thoughts as well before merging.

+1 Thanks @anandsubbu

anandsubbu:
Thanks @MohanDV !

PR #1213, opened 2018-09-26
METRON-1681: Decouple the ParserBolt from the Parse execution logic

merrimanr:
## Contributor Comments
The primary purpose of this PR is to create a parser container abstraction that is decoupled from Storm.  This parser container (I call it ParserRunner, anyone have a better name?) is responsible for:

- Instantiating the MessageParser and MessageFilter objects for each sensor.
- Initializing the Stellar environment.
- Accepting a raw binary message along with parser configurations and calling the MessageParser.parseOptional method for the appropriate sensor type
- Process each message.  Most of this logic was migrated from the ParserBolt.execute method.  
- Execute callbacks depending on the processing result of each message

Configuration is external to this abstraction.  A configuration supplier is passed in when initializing and a configuration object is passed in when processing each message.  I believe this was originally done because we want message processing to be atomic without the configuration unexpectedly changing.  We can easily change to a message supplier during execute if necessary.  A CuratorFramework object is also required for setting up Stellar but we could easily make this optional.  

I decided to keep writing out of the abstraction in this PR.  I can envision different platforms having different requirements or needs for sending along messages after they are parsed.  Therefore the ParserBolt still handles writing messages to Kafka.  If we do decide we want to add writing to our abstraction we could do it in a follow on PR to keep this from becoming even bigger.

Since all of the post parsing logic was in the Parserbolt, messages could be written as they were processed rather than having to wait for all messages to be processed. To maintain this behavior I added 2 callback functions in the form of Java Consumers:  onSuccess and OnError.  The other option would be to make message processing synchronous and just return a list of results.

This lifecycle of this container looks like:

1. Container is created from a collection of sensor types 
2. Container is initialized with the init method that accepts a Curator client and a configuration Supplier.  This in turn sets up Stellar and instantiates MessageParser and MessageFilter classes.
3. Container is ready and accepts messages for processing and calls the appropriate callbacks.

Because this splits the ParserBolt into 2 different classes much of the ParserBoltTest unit test didn't make sense anymore.  I ended up essentially rewriting it and also creating a unit test for ParserRunner.  I tried to represent all the original tests in ParserBoltTest and have 95%+ coverage.  If I missed any cases or made undesirable style changes, let me know.

### Changes Included

- ParserRunner abstraction that is decoupled from the ParserBolt.  The ParserBolt now initializes the ParserRunner and defers parsing to that class.  The only thing required is Metron configuration and a Curator client (which could be optional)
- Refactored ParserBolt that sets up the ParserRunner, passes message to it for processing, and writes results to Kafka.
- MessageParser and MessageFilter objects are now created when Storm calls prepare, avoiding serialization issues (https://issues.apache.org/jira/browse/METRON-1793)
- Updated unit and integration tests

### Testing
I have done basic testing in full dev:

1. Spin up full dev and verify bro and snort alerts are indexed in ES and the data looks correct.

2. Test for proper message parser error handling by producing an unparseable message to the bro topic:
```
echo 'bad message' | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic bro
```
You should see a corresponding error message in the ES error index:
```
{
        "_index" : "error_index_2018.09.25.20",
        "_type" : "error_doc",
        "_id" : "df83aebc-506d-404b-aba6-c5a740d07c57",
        "_score" : 1.0,
        "_source" : {
          "exception" : "java.lang.IllegalStateException: Unable to parse Message: test",
          "failed_sensor_type" : "bro",
          "stack" : "java.lang.IllegalStateException: Unable to parse Message: bad message
          ...
          "hostname" : "node1",
          "source:type" : "error",
          "raw_message" : "test",
          "error_hash" : "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08",
          "error_type" : "parser_error",
          "guid" : "df83aebc-506d-404b-aba6-c5a740d07c57",
          "message" : "Unable to parse Message: bad message",
          "timestamp" : 1537906408288
        }
}
```

3. Test for proper invalid message handling setting up global validation.  Add this to the global config:
```
"fieldValidations": [
    {
      "input": [
        "is_alert"
      ],
      "validation": "NOT_EMPTY"
    }
  ]
```
This should cause validation to fail on bro messages and you should see a corresponding error message in the ES error index:
```
{
        "_index" : "error_index_2018.09.25.20",
        "_type" : "error_doc",
        "_id" : "97c356a3-0fd8-4076-9e9c-0213203565c8",
        "_score" : 1.0,
        "_source" : {
          "failed_sensor_type" : "bro",
          "hostname" : "node1",
          "source:type" : "error",
          ...
          "error_hash" : "3e5bc436c661cd7df41d4624b5c7422368373f1e57d201500d731007e146c88a",
          "error_type" : "parser_invalid",
          "guid" : "97c356a3-0fd8-4076-9e9c-0213203565c8",
          "error_fields" : "is_alert",
          "timestamp" : 1537907615275
        }
}
```

 ### Next Steps
This PR contains working code but is still needs documentation.  I am planning on testing a couple different parsers in full dev (grok for example) in addition to what I've already tested.  I will be adding inline comments for some of the less obvious changes or refactors to make it easier to review.  My plan is for any discussion around specific parts of the code to get added as javadocs eventually.  I also think we should add some developer documentation to make it easier for maintaining and integrating this into other platforms.  I imagine a lot of info in this description would make it in there as well.

The intention for now is to get some feedback on the overall approach and get people thinking about it.  I'm still working on documentation and will add that soon.  Let me know what you think!

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunner.java
```diff
@@ -0,0 +1,234 @@
...
+
+  public void init(CuratorFramework client, Supplier<ParserConfigurations> parserConfigSupplier) {
+    if (parserConfigSupplier == null) {
+      throw new IllegalStateException("A parser config supplier must be set before initializing the ParserRunner.");
+    }
+    initializeStellar(client, parserConfigSupplier);
+    initializeParsers(parserConfigSupplier);
+  }
+
+  protected void initializeStellar(CuratorFramework client, Supplier<ParserConfigurations> parserConfigSupplier) {
```
merrimanr: We could make the client optional here.  The only required dependency would then be ParserConfigurations.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -242,169 +226,81 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
-    this.stellarContext = builder.build();
-    StellarFunctions.initialize(stellarContext);
-  }
-
 
   @SuppressWarnings("unchecked")
   @Override
   public void execute(Tuple tuple) {
     if (TupleUtils.isTick(tuple)) {
-      try {
```
merrimanr: Slight refactor here.  I moved this to it's own method to make execute more readable.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -242,169 +226,81 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
+    numWritten = 0;
     byte[] originalMessage = (byte[]) messageGetStrategy.get(tuple);
+    String topic = tuple.getStringByField(FieldsConfiguration.TOPIC.getFieldName());
+    String sensorType = topicToSensorMap.get(topic);
     try {
-      SensorParserConfig sensorParserConfig;
-      MessageParser<JSONObject> parser;
-      String sensor;
-      Map<String, Object> metadata;
-      if (sensorToComponentMap.size() == 1) {
```
merrimanr: I refactored this to always look up the sensor from the topic.  I think this makes it easier to read and test.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -242,169 +226,81 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
-      }
+      ParserConfigurations parserConfigurations = getConfigurations();
+      SensorParserConfig sensorParserConfig = parserConfigurations.getSensorParserConfig(sensorType);
+      RawMessage rawMessage = RawMessageUtil.INSTANCE.getRawMessage( sensorParserConfig.getRawMessageStrategy()
+              , tuple
+              , originalMessage
+              , sensorParserConfig.getReadMetadata()
+              , sensorParserConfig.getRawMessageStrategyConfig()
+      );
+      parserRunner.setOnSuccess(parserResult -> onSuccess(parserResult, tuple));
```
merrimanr: Setting onSuccess here with the current tuple.  We could also store the current tuple in a member variable if we wanted to move this call to prepare.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -242,169 +226,81 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
-        throw new RuntimeException(
-            "This should have been caught in the writerHandler.  If you see this, file a JIRA", e);
-      } finally {
-        collector.ack(tuple);
-      }
+      handleTickTuple(tuple);
       return;
     }
-
+    numWritten = 0;
```
merrimanr: This counter is implemented as a member variable and reset for each tuple.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -242,169 +226,81 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
+      }
+    } catch (Exception e) {
+      throw new RuntimeException(
+              "This should have been caught in the writerHandler.  If you see this, file a JIRA", e);
+    } finally {
+      collector.ack(tuple);
+    }
+  }
+
+  protected void handleError(String sensorType, byte[] originalMessage, Tuple tuple, Throwable ex, OutputCollector collector) {
```
merrimanr: The sensorType lookup is now outside of the try so we can pass it in.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/filters/BroMessageFilter.java
```diff
@@ -67,7 +67,7 @@ else if(protocolsObj instanceof List) {
    */
 
   @Override
-  public boolean emitTuple(JSONObject message, Context context) {
+  public boolean emit(JSONObject message, Context context) {
```
merrimanr: Renamed this because it should not be specific to Storm.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/topology/ParserComponent.java
```diff
@@ -17,26 +17,23 @@
...
   private static final long serialVersionUID = 7880346740026374665L;
 
   private MessageParser<JSONObject> messageParser;
   private MessageFilter<JSONObject> filter;
-  private WriterHandler writer;
 
-  public ParserComponents(
+  public ParserComponent(
       MessageParser<JSONObject> messageParser,
-      MessageFilter<JSONObject> filter,
```
merrimanr: WriterHandler was moved out of this class because it's handled by Storm.  MessageParser and MessageFilter is abstracted away from the streaming platform in this PR.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/topology/ParserTopologyBuilder.java
```diff
@@ -268,29 +266,14 @@ private static ParserBolt createParserBolt( String zookeeperUrl,
                                               Optional<String> securityProtocol,
                                               ParserConfigurations configs,
                                               Optional<String> outputTopic) {
-
-    Map<String, ParserComponents> parserBoltConfigs = new HashMap<>();
+    Map<String, WriterHandler> writerConfigs = new HashMap<>();
     for( Entry<String, SensorParserConfig> entry : sensorTypeToParserConfig.entrySet()) {
       String sensorType = entry.getKey();
       SensorParserConfig parserConfig = entry.getValue();
-      // create message parser
```
merrimanr: This should not be done here because it can potentially cause serialization problems:  https://issues.apache.org/jira/browse/METRON-1793.  MessageParser and MessageFilter objects should be created during initialization and not when the container is created (and before the ParserBolt is created).  We should be doing this when prepare is called on the ParserBolt.

metron-platform/metron-parsers/src/test/java/org/apache/metron/parsers/integration/ParserDriver.java
```diff
@@ -83,49 +86,21 @@ public void close() throws Exception {
...
-          , Collections.singletonMap(
-              sensorType == null ? config.getSensorTopic() : sensorType,
-              new ParserComponents(
-              ReflectionUtils.createInstance(config.getParserClassName()),
-                  null,
-                  new WriterHandler(new CollectingWriter(output))
-              )
-         )
-      );
+      super(null, parserRunner, Collections.singletonMap(sensorType, new WriterHandler(new CollectingWriter(output))));
```
merrimanr: Not sure this ShimParserBolt is necessary anymore.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunner.java
nickwallen: Agreed. A Zk connection should be optional.  Really, we might want to factor out the Stellar initialization as this most likely will need to be done slightly differently based on the execution environment.

For example, I want to use your `ParserRunner` to be able to test my parsers in the REPL.  In that case, I'm going to create my config, then test in the REPL all before it gets uploaded to Zk.  Only after I like what I have will I push the config to Zk.

nickwallen:
> @merrimanr: I decided to keep writing out of the abstraction in this PR. 

Agreed. I don't think we want this to write.  How we write out the result of parsing will depend on the execution environment.

> @merrimanr To maintain this behavior I added 2 callback functions in the form of Java Consumers: onSuccess and OnError. The other option would be to make message processing synchronous and just return a list of results.

I prefer your second suggested approach; just return a list of results.  We can add complexity (like multiple threads, caching) on top of a simple, synchronous approach should it be needed.  But I think the base functionality should be stupid-simple. 

I would make your `ParserRunner` an interface and then make your current `ParserRunner` the default implementation.  Later on we could provide alternative implementations that maintain a thread pool or do intelligent caching.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunner.java
merrimanr: Ok sounds good.  Are you ok if I leave the Stellar stuff in there as a default that only gets called if one hasn't been provided?

nickwallen: Yes, that could work.  It would make it easier to use if we have a default implementation that works in most cases.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunner.java
```diff
@@ -0,0 +1,234 @@
...
+
+  public void execute(String sensorType, RawMessage rawMessage, ParserConfigurations parserConfigurations) {
+    if (onSuccess == null) {
+      throw new IllegalStateException("An onSuccess function must be set before parsing a message.");
+    }
+    if (onError == null) {
+      throw new IllegalStateException("An onError function must be set before parsing a message.");
+    }
+    SensorParserConfig sensorParserConfig = parserConfigurations.getSensorParserConfig(sensorType);
+    if (sensorParserConfig != null) {
```
ottobackwards: Why have the api return the List AND do the callback.  Just have the callback.

merrimanr: It's going to be one or the other.  Either it's synchronous and we return a list or it's asynchronous and we expose callbacks.  The interface would have to be different in each case.

justinleet:
Re: callback vs list.  I'm pretty neutral on which one we do. I probably have a slight preference towards the callbacks, just because it more closely matches the original behavior, but I don't think it makes a huge difference in the implementation (and let me know if it does).

@merrimanr Can you look into Travis?

merrimanr:
I think the travis failure is unrelated.  I have to make changes based on feedback so I'll keep on eye on it after the next run.

Before I go implementing changes we need to decide on callback vs list.  I don't see how we can design an interface that supports both (am I wrong there?) so we need to choose.  Seems to be split at the moment.  Thoughts?

nickwallen:
> Before I go implementing changes we need to decide on callback vs list. I don't see how we can design an interface that supports both (am I wrong there?) so we need to choose. Seems to be split at the moment. Thoughts?

Agreed. We should not support both.  Its one or the other.

IMHO, Callbacks just seems overly complex and less obvious to me.  I could see the value of using callbacks if this was asynchronous, but its not.  The callbacks are made synchronously and so just returning a List seems like the most obvious approach.

If callbacks more closely aligns with what was there, then just go with that.  Better done sooner than delaying over a small nit like this.  I will assent to either.

merrimanr:
The MessageParser.parseOptional call returns a list of JSONObjects which is then processed (field transformations applied, messages validated, etc).  The difference is WHEN each processed JSONObject is written to Kafka.  In the current ParserBolt each one is sent as soon as it is processed because everything runs in ParserBolt.execute.  Since we are extracting that logic outside of the ParserBolt, we either need to write them all after processing is done by the abstraction or expose callbacks that the abstraction calls as it finishes processing each one.

I don't think it matters much either way.  I chose to use callbacks because it more closely matches the flow of the current ParserBolt.  I will assume everyone is ok with callbacks and will continue work on the suggestions made in the comments.

mmiklavc:
@merrimanr @nickwallen @justinleet - I just want to be sure that I'm following this correctly. This particular callback appears to be a synchronous, blocking callback. The crux of the implementation is that you're passing in a function pointer to be executed when the parsing either succeeds or fails in order to decouple the parsing logic from Storm itself (e.g. tuples, etc.)?

merrimanr:
That sounds correct to me.  Specifically decoupling the Kafka writing step since our writing implementation requires a tuple.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunner.java
```diff
@@ -0,0 +1,234 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class ParserRunner implements Serializable {
```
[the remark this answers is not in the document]
nickwallen: I don't get `ParserContext`.  The `ParserRunner` is something that runs/executes parsers, no?  It doesn't provide context for something else to run parsers.  But maybe I am misunderstanding why you think `ParserContext` makes sense.  Can you explain your thoughts on that?

nickwallen:
> @mmiklavc Have you looked over this by chance [2]? The reason I bring this up is because it looks like it already addresses a number of the points/questions brought up in this PR's discussion. For example, there's a strategy for handling successful messages as well as error results [3]. If we're going through the effort of refactoring, it's probably worth our while to match the other idioms to some degree. It just makes continued maintenance that much easier going forward.

Are you making the case for returning a `List` here then?  Sounds like you're thinking along these lines?
```
ParserRunner {
   List<ParserResult> execute(...);
} 
```

That would be similar to the unified enrichment work that you linked to in your comment.
```
ParallelEnricher {
  EnrichmentResult apply(...);
}
```

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunner.java
```diff
@@ -0,0 +1,234 @@
...
+
+  protected transient Consumer<ParserResult> onSuccess;
+  protected transient Consumer<MetronError> onError;
+
+  private HashSet<String> sensorTypes;
+  private Map<String, ParserComponent> sensorToParserComponentMap;
+
+  // Stellar variables
+  private transient Context stellarContext;
```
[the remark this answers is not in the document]
merrimanr: Context is not serializable.

merrimanr:
Thanks for pointing me to that @mmiklavc, I had not looked at it yet.  I am in agreement that we should do our best to match other classes and patterns we already have.  I will continue studying that to see where we can make things more consistent.  If you can provide some examples that demonstrate your ideas that would help me understand it better.  

From what I can tell, the strategy pattern was used in the UnifiedEnrichmentBolt because there are 2 separate strategies for enriching a message:  enrichment and threat intel.  Do we need the strategy pattern here since there is only one parser running strategy?  We could implement this as a strategy pattern in anticipation of one day needing another parser running strategy.  Do you think it's worth doing now or later when we need it?

The other difference I see is that Future objects are used to join the different messages after they are processed in parallel.  I believe this is done because all enrichments/threat intel must be done before the message can be pieced back together and sent along.  In this case we don't have that limitation.  The documents that are parsed do not depend on each other and can be passed along as soon as they are processed.  We are set up for post-processing these documents in parallel but I see that as a low-latency operation and I'm not sure it's worth the extra overhead.

merrimanr:
I just pushed a commit out based on some feedback from Nick: 

- The stellar context is now passed into the ParserRunner abstraction as a dependency.  I imagine different environments will require different Stellar initialization strategies so this made sense to me.  I also like it because it makes the abstraction simpler.
-  Changed ParserRunner to an interface with a single implementation.  I could change this to ParserRunnerStrategy in case we think we might want to use that pattern in the future.  Then all that would be missing is a ParserContext class (and possibly a ParserRunnerStrategies enum) which we could add now or when we need it.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -185,37 +199,12 @@ public void prepare(Map stormConf, TopologyContext context, OutputCollector coll
...
-    for (Map.Entry<String, ParserComponents> entry: sensorToComponentMap.entrySet()) {
-      String sensor = entry.getKey();
-      SensorParserConfig config = getSensorParserConfig(sensor);
-
-      if (config != null) {
-        cacheConfig.putAll(config.getCacheConfig());
-      }
-    }
-    cache = CachingStellarProcessor.createCache(cacheConfig);
+    this.parserRunner.setOnError(this::onError);
```
mmiklavc: Seems like success/error functions should go in init() if we went this route. I think it's best to choose 1 approach or the other when managing initialization and state validation for an object.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
mmiklavc: Ah, that's right. This answers my earlier comment about setsuccess/error.

mmiklavc:
@merrimanr - I think I viewed the choice of ParserRunner implementation as a strategy of sorts. I guess I don't care either way. @nickwallen's suggestion of making that an interface, that I see you've implemented, should work well there also. There's definitely some impedance mismatch between what's in parsers vs enrichment, so there won't be a direct mapping of concepts.

@nickwallen - I've been going back and forth about which approach to take in handling success/errors. I think I agree with you and prefer keeping the runner interface simpler by simply returning a `List<SomethingResulty>`. The fact that we have to jump through hoops a bit for callbacks because of needing reference to the Tuple in onSuccess suggests to me that it's probably better to manage the results in a returned list. The validation on the runner is also simpler in this case as well, ie no need to validate you've set your callbacks. Having onSucess/Error methods in the runner interface feels a little fishy to me. I also think that it makes some sense to decouple the parsing phase from writing altogether, though I recognize that callbacks are not opinionated about the "what" that happens in success/error would be.

merrimanr:
I don't feel that strongly about either approach so I'll switch it to return a list and remove the callbacks.

merrimanr:
The latest commit removes the callbacks.  Now the execute method returns a list of ParserResult objects.  I had to add a MetronError field to the ParserResult class so that both successful messages and errors are passed back.

I also noticed we are currently skipping a tuple if no configuration is found for that sensor.  I think skipping the tuple is fine but we should still log an error message.  I made a change in the ParserRunner class to throw an error on this condition.  The parser bolt will catch it and handle it similar to a parser error by creating a MetronError object and sending it to the error topic.  Does anyone think we should keep as it is now?

mmiklavc:
@merrimanr @justinleet @cestella - regarding the ignored tuple when no config for sensor - has that been longstanding or is that something intentionally added with parser aggregation and chaining? Just cautious of adding a regression. At the very least, if it is intentional a comment should be added explaining it.

merrimanr:
As far as I know it's been that way for a really long time.

justinleet:
@mmiklavc It's been like that for a long time.  https://github.com/apache/metron/blob/6fb85e124d87581c37b79ea105570f4f85b59384/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java#L118.  It's probably earlier than this, I didn't track down the actual commit, just checked a random one from awhile ago.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserResult.java
```diff
@@ -0,0 +1,75 @@
...
+  public byte[] getOriginalMessage() {
+    return originalMessage;
+  }
+
+  public boolean isError() {
+    return error != null;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
mmiklavc: Overriding equals without overriding hashCode :-O

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunnerImpl.java
```diff
@@ -0,0 +1,216 @@
...
+                .withErrorType(Constants.ErrorType.PARSER_INVALID)
+                .withSensorType(Collections.singleton(sensorType))
+                .addRawMessage(message);
+        Set<String> errorFields = failedValidators == null ? null : failedValidators.stream()
+                .flatMap(fieldValidator -> fieldValidator.getInput().stream())
+                .collect(Collectors.toSet());
+        if (errorFields != null && !errorFields.isEmpty()) {
+          error.withErrorFields(errorFields);
+        }
+        parserResult = Optional.of(new ParserResult(sensorType, error, rawMessage.getMessage()));
```
mmiklavc: @merrimanr you and @ottobackwards have landed in a similar place. I like what @ottobackwards has done with keeping the MessageParser interface backwards compatible. There will be some merge conflicts to resolve, but I think the API designs will come together nicely.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunner.java
```diff
@@ -18,20 +18,17 @@
...
 import java.util.function.Supplier;
 
 public interface ParserRunner {
 
   Set<String> getSensorTypes();
-  void setOnSuccess(Consumer<ParserResult> onSuccess);
-  void setOnError(Consumer<MetronError> onError);
   void init(Supplier<ParserConfigurations> parserConfigSupplier, Context stellarContext);
-  void execute(String sensorType, RawMessage rawMessage, ParserConfigurations parserConfigurations);
+  List<ParserResult> execute(String sensorType, RawMessage rawMessage, ParserConfigurations parserConfigurations);
```
mmiklavc: Groovy

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunnerImpl.java
```diff
@@ -101,26 +92,28 @@ public void init(Supplier<ParserConfigurations> parserConfigSupplier, Context st
...
+  public List<ParserResult> execute(String sensorType, RawMessage rawMessage, ParserConfigurations parserConfigurations) {
+    List<ParserResult> parserResults;
     SensorParserConfig sensorParserConfig = parserConfigurations.getSensorParserConfig(sensorType);
     if (sensorParserConfig != null) {
       MessageParser<JSONObject> parser = sensorToParserComponentMap.get(sensorType).getMessageParser();
       List<JSONObject> messages = parser.parseOptional(rawMessage.getMessage()).orElse(Collections.emptyList());
-      messages.forEach(message -> processMessage(sensorType, message, rawMessage, parser, parserConfigurations));
+      parserResults = messages.stream()
+              .map(message -> processMessage(sensorType, message, rawMessage, parser, parserConfigurations))
+              .filter(Optional::isPresent)
```
mmiklavc: Remind me what the difference between not present, present, and present w/exception or error is wrt the return value from processMessage.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -251,8 +249,21 @@ public void execute(Tuple tuple) {
               , sensorParserConfig.getReadMetadata()
               , sensorParserConfig.getRawMessageStrategyConfig()
       );
-      parserRunner.setOnSuccess(parserResult -> onSuccess(parserResult, tuple));
-      parserRunner.execute(sensorType, rawMessage, parserConfigurations);
+      List<ParserResult> parserResults = parserRunner.execute(sensorType, rawMessage, parserConfigurations);
+      int numWritten = 0;
+      for(ParserResult parserResult: parserResults) {
```
mmiklavc: Thanks for this Ryan, I do think it's easier to follow.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunnerImpl.java
merrimanr: I believe not present means the message was filtered out (our FilterClass mechanism).  Otherwise there will either be a message or error in the ParserResult.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserResult.java
merrimanr: Done

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunnerImpl.java
merrimanr: I am expecting some merge conflicts with https://github.com/apache/metron/pull/1184.  Happy to handle those conflicts if that goes in before this.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserResult.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+package org.apache.metron.parsers;
+
+import org.apache.metron.common.error.MetronError;
+import org.json.simple.JSONObject;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+public class ParserResult {
```
nickwallen: Can we add javadoc?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunner.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.metron.common.configuration.ParserConfigurations;
+import org.apache.metron.common.message.metadata.RawMessage;
+import org.apache.metron.stellar.dsl.Context;
+
+import java.util.List;
+import java.util.Set;
+import java.util.function.Supplier;
+
+public interface ParserRunner {
```
nickwallen: Can we add javadoc?

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunnerImpl.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class ParserRunnerImpl implements ParserRunner, Serializable {
```
nickwallen: I'd also add javadoc here.  How does this particular ParserRunner work?  Is it just the default that we should use in most cases?

justinleet:
@merrimanr Can you deconflict this?

merrimanr:
The latest commit resolves the conflicts introduced by https://github.com/apache/metron/pull/1184.  The changes in all classes except ParserRunnerImpl and ParserRunnerImplTest were fairly trivial, mainly just adjusting the return types to match the new MessageParserResult and ParserRunnerResults interfaces.

Most of the work was done in ParserRunnerImpl (the execute method specifically).  Since the MessageParser interface can now return multiple messages and exceptions (and a master exception in some cases), there needs to be a way to map them to a ParserRunnerResults object (which is a list of messages and errors).   Now the primary job of this execute method is to:
- call MessageParser.parseOptionalResult
- post process messages returned by the MessageParser
- consolidate the various errors and messages into a single ParserRunnerResults object

Let me know if anyone thinks this should work differently.  I also added more javadocs to the various classes involved, primarily ParserRunnerImpl since that's where the magic happens.  This merge was fairly significant so I am planning on testing in full dev again.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/ParserRunnerImpl.java
```diff
@@ -137,11 +208,29 @@ private void initializeParsers(Supplier<ParserConfigurations> parserConfigSuppli
...
+   * @param sensorType Sensor type of the message
+   * @param message Message parsed by the MessageParser
+   * @param rawMessage Raw message including metadata
+   * @param parser MessageParser for the sensor type
+   * @param parserConfigurations Parser configurations
+   */
   @SuppressWarnings("unchecked")
-  protected Optional<ParserResult> processMessage(String sensorType, JSONObject message, RawMessage rawMessage,
+  protected Optional<ProcessResult> processMessage(String sensorType, JSONObject message, RawMessage rawMessage,
                                                   MessageParser<JSONObject> parser,
```
ottobackwards: Is parser used?

ottobackwards: ignore the thing about Multiline... I'm mixing up PR's

merrimanr: Yes parser is used.  Line 248, `parser.validate`.

justinleet:
I'm still +1, does anybody else have any concerns they'd like addressed?

ottobackwards:
@justinleet if you take master, I'll help resolve the conflicts from #1234

justinleet:
@merrimanr I think @ottobackwards meant that for you.  Looks like #1234 is in, so this needs updating.

ottobackwards:
oops, right

merrimanr:
I merged the latest master (which includes https://github.com/apache/metron/pull/1234) into this PR.  I experimented with moving the methods in MultilineMessageParser back into MessageParser and I think it turned out well.  Makes things clearer and easier to understand in my opinion (and also made the merge much easier).

What do you think @mmiklavc and @ottobackwards?  It reverts some changes in https://github.com/apache/metron/pull/1234 so I want to get your thoughts and buy in.

mmiklavc:
@ottobackwards - @merrimanr and I had a chat about this offline and realized there was another alternative to what we had come up with in the multiline grok and syslog PRs. Basically, just making the existing methods in MessageParser `@deprecated` and with a new default implementation for `parse(..)` that throws a `NotImplementedException`. This does clean up some of the shrapnel we had discussed on 1234 such that there's no further need for a decorator.

@merrimanr - the only thing missing here I think is to mark the `parse` and `parseOptional` methods `@deprecated`. Also, I didn't see a delete of `MultilineMessageParser` in your latest commits to go with that refactor/merge. We probably want to get rid of it altogether if we take this alternate approach. Also, I think we should make a more public statement on the user and dev list about deprecating the methods if we all agree on this approach to give users time to migrate existing custom parser implementations.

ottobackwards:
That is fine

merrimanr:
The latest commit adds the Deprecated annotations to parse and parseOptional and removes MultilineMessageParser.  I also updated the `MessageParserTest` with more tests to ensure we're handling the deprecated methods appropriately.  As a result of that, I changed the logic in MessageParser.parseOptionalResult slightly to return an empty Optional rather than an empty List to match the existing test.  Let me know if anyone has a problem with this.

Once we finish this I will make an announcement on the user and dev lists.

mmiklavc:
lgtm, +1 by inspection. Thanks for the collaborative effort on this @merrimanr and @ottobackwards!

ottobackwards:
This is really nice.  My only addition would be logging.   I think it would be valuable for the worker log to state what parsers where loaded and running in a given topology.   I think we can build this out to a better user facing feature in the ui in the future as a follow on, but logging would be minimally useful for this pr

merrimanr:
Latest commit adds some logging around parser initialization.  Let me know what you think.

mmiklavc:
Good suggestion @ottobackwards. lgtm @merrimanr, still +1 from me.

ottobackwards:
+1, ship it

justinleet:
+1, thanks!

PR #1218, opened 2018-10-02
METRON-1801 Allow Customization of Elasticsearch Document ID

nickwallen:
Currently, the Metron GUID is always used as the Elasticsearch document ID. As documented in [METRON-1677](https://issues.apache.org/jira/browse/METRON-1677), using a randomized UUID like Java's `UUID.randomUUID()` can negatively impact Elasticsearch performance.  This change allows a user to customize the identifier that is used by Elasticsearch when indexing documents.

A property `es.document.id` was added that sets the message field that is used to define the document ID when a message is indexed by Elasticsearch. 

* To allow Elasticsearch to define its own document id, this property should be set to a blank or empty string. The client will not set the document ID and Elasticsearch will define its own.

* In most cases allowing Elasticsearch to define the document ID is the most performant option. This is the default behavior.

* Metron versions 0.6.0 and earlier defined the document ID using the Metron GUID, which is a randomized UUID using Java's `UUID.randomUUID()`. Using a randomized UUID can negatively impact Elasticsearch indexing performance. To maintain backwards compatibility with legacy versions of Metron use the following setting.

    ```
    es.document.id = guid
    ``` 

* To use a custom document ID, create an enrichment that defines a new message field; for example one called `my_document_id`. Then use this field to set the document ID as follows. This will set the document ID to the value of the message field `my_document_id`.
    ```
    es.document.id = my_document_id
    ```

* If a message does not contain the `es.document.id` field, a warning is issued and no document ID is set by the client.

## Changes

  * The `ElasticsearchWriter` was updated to allow the document ID to be configurable.

  * A 'search by GUID' in the REST layer was implicitly using the document ID, whereas it should be using the Metron GUID.

* Search results should use the Metron GUID as the ID returned to the UI.  All IDs visible to the user should always be the Metron GUID, not the document ID.

* The MPack was updated to allow the user to define the `es.document.id` on the Metron > Config > Index Settings tab.

* The default behavior was changed to allow Elasticsearch to set the document ID.  This is the most performant option in most cases.  I updated the `Upgrading.md` doc to describe how to revert to the legacy behavior.

## Testing

1. Spin-up a development environment.  You may need to stop the PCAP and/or Profiler topology to free-up slots to allow indexing to occur.

    ```
    vagrant up
    ```

1. Open the Alerts UI and ensure that alerts are visible.  Notice that the ID listed in the table has not changed.  This will always display the Metron GUID, no matter what ID used for the document.

    screen shot 2018-10-02 at 5 43 54 pm

1. Click on a GUID in the table to search for a single alert.

    screen shot 2018-10-02 at 12 42 54 pm

1. Create a meta-alert and ensure that alerts tied to the meta-alert are still discoverable by GUID.

    screen shot 2018-10-02 at 5 45 18 pm

1. Open Ambari and go to Metron > Configs > Index Settings.  Notice that the default setting of `es.document.id` is blank, thus allowing Elasticsearch to define its own document ID.

    screen shot 2018-10-03 at 9 13 18 am

1. Open Kibana and verify that indeed Elasticsearch is generated its own document IDs.  You will notice an `_id` field which has been generated by Elasticsearch.  This will be different than the UUID generated by Metron and stored as part of the document as `guid`.

    screen shot 2018-10-02 at 4 47 52 pm

1. Stop the indexing topologies using Ambari.

1. Login to the VM.

    ```
    vagrant ssh
    sudo su -
    ```

1. Delete the existing indices in Elasticsearch.

    ```
    curl -XDELETE http://node1:9200/bro*
    curl -XDELETE http://node1:9200/snort*
    ```

1. Open Ambari, and go to Metron > Configs > Index Settings.  Edit the "Elasticsearch Document ID Source Field" and set it to 'guid'.  This will restore the legacy behavior where the document ID is to the Metron GUID.

1. Restart the Indexing Topology in Ambari. 

1. Open the Alerts UI and ensure that alerts are visible.  Notice that the ID listed in the table has not changed.  This will always display the Metron GUID, no matter what ID used for the document.

1. Click on a GUID in the table to search for a single alert.

    screen shot 2018-10-02 at 12 42 54 pm

1. Create a meta-alert and ensure that alerts tied to the meta-alert are still discoverable by GUID.

    screen shot 2018-10-02 at 5 45 18 pm

1. Open Kibana and verify that the document ID matches the Metron GUID.

    screen shot 2018-10-03 at 5 58 49 pm

1. I would also advise running the UI e2e test suite with this change.

## Pull Request Checklist


mraliagha:
Thanks, Nick. So if es.document.id is not provided, as the default, doc id won't be send to ES indexing, right? I guess it would be also nice to provide some guidance on how document ID should be defined (in the case of custom ID). Otherwise, users may create some serious issues with the indexing and search throughput.

nickwallen:
@mraliagha I updated the README to (hopefully) better explain your options in using `es.document.id`.  I sensed by your question that what I had originally was not very clear.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriterConfig.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+public enum ElasticsearchWriterConfig {
+
+  /**
+   * Defines which message field, the document identifier is set to.
+   *
+   * <p>If defined, the value of the specified message field is set as the Elasticsearch doc ID. If
+   * this field is undefined or blank, then the document identifier is not set.
+   */
+  DOC_ID_SOURCE_FIELD("es.document.id", "", String.class);
```
nickwallen: I put this value here because I didn't see any other place for it at the time.  I then later found the other properties like `es.ip` defined in `ElasticsearchUtils`.  

Many of the property definitions there are hard coded and could use a clean-up. I'd like to do a follow-on to add `es.ip`, `es.date.format`, `es.client.class`, etc to this configuration class so that each is clearly defined.

As part of this PR, we could keep this property here or I could move it to `ElasticsearchUtils` to match those other properties.  Open to reviewer feedback on this.

mraliagha:
@nickwallen in the case of event logs and the fact that retrieval segmentation would be mostly based on timestamp, it is recommended to use timestamp as a prefix of the id. For example, something like timestamp+hash(original_string).

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriterConfig.java
merrimanr: This does leave things in an inconsistent state and will makes things confusing for anyone working with these classes.  However I think a follow up PR would be fine.

Are you aware of the [ConfigOption](https://github.com/apache/metron/blob/master/metron-platform/metron-common/src/main/java/org/apache/metron/common/configuration/ConfigOption.java) interface?      It looks like there may be some duplication with this class.

nickwallen: I didn't know we had `ConfigOption`.  I will remove `ElasticsearchWriterConfig` for now.  Then as a follow-on, I will do something similar, but reuse `ConfigOption`.  Thanks for the pointer!

merrimanr:
@mraliagha, that's a good suggestion.  I believe we can functionally achieve that be creating a custom id field in the format you suggest (with a Stellar field transform) and set that field to be the ES id with the Ambari property exposed in this PR.  Do you feel it's worth documenting as an optimization?

I spun this up in full dev and ran through all the testing instructions.  Everything worked as advertised.  I think there are just a couple open questions but this is pretty close in my opinion.

metron-platform/metron-elasticsearch/pom.xml
```diff
@@ -206,24 +206,7 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
-        <dependency>
```
merrimanr: Why were these dependencies removed?  Just curious.

nickwallen: They cause the tests to fail when executed in an IDE like IntelliJ.  I don't understand exactly why, but @justinleet pointed me in this direction.

Also, everything runs just fine without them, so they are unnecessary.  The fewer dependencies, the better.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriterConfig.java
nickwallen: The last commit removes `ElasticsearchWriterConfig`.

metron-platform/metron-elasticsearch/pom.xml
merrimanr: I have also noticed this and I end up commenting these out every time I run a test in my IDE.  Thanks for investigating and removing them.

merrimanr:
Looks good to me.  +1

mraliagha:
> @mraliagha, that's a good suggestion. I believe we can functionally achieve that be creating a custom id field in the format you suggest (with a Stellar field transform) and set that field to be the ES id with the Ambari property exposed in this PR. Do you feel it's worth documenting as an optimization?

Yes, I think it is worth documenting as people can easily create serious issues with Lucene based indexers by messing with ID. It can give users an understanding of where it is safe to play with the ID and what the recommendations are. I see if I can find any articles to share it as a part of the manual.

nickwallen:
This change was reverted [here](https://github.com/apache/metron/commit/0e037edad913955d3b6754ca9cf42b329cd84160).  A new pull request will be opened with the functionality. See also [this mailing list thread](https://github.com/apache/metron/commit/0e037edad913955d3b6754ca9cf42b329cd84160).

PR #1221, opened 2018-10-03
METRON-1805: Provide a default value for the Storm topology.max.spout.pending setting

merrimanr:
## Contributor Comments
Users have reported problems in the random and batch indexing topologies when this setting is not set.  The primary purpose of this PR is to start a discussion around providing a default value for the Storm topology.max.spout.pending setting in our topologies and implement the changes if we decide to do it.  

1. Should we provide defaults or continue to defer to the Storm default where a maximum isn't enforced?
2. Which topologies should we provide defaults for?
3. What should the default values be for the topologies we include?

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-indexing-env.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-indexing-env.xml
@@ -146,7 +146,7 @@
     <property>
         <name>ra_indexing_topology_max_spout_pending</name>
         <description>Indexing Topology Spout Max Pending Tuples</description>
-        <value/>
+        <value>300</value>
         <display-name>Indexing Max Pending for Random Access</display-name>
         <value-attributes>
             <empty-value-valid>true</empty-value-valid>
@@ -155,7 +155,7 @@
      <property>
         <name>batch_indexing_topology_max_spout_pending</name>
         <description>Indexing Topology Spout Max Pending Tuples</description>
-        <value/>
+        <value>300</value>
         <display-name>Indexing Max Pending for HDFS</display-name>
         <value-attributes>
             <empty-value-valid>true</empty-value-valid>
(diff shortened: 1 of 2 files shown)


mmiklavc:
I think this is fine. This is just the backpressure setting and having it set to the default means that the spigot is fully open and will potentially flood the topology with a high velocity of data. We recommend tuning all topologies, so starting from a default that doesn't crush the system seems reasonable to me. Since this change is only being made to indexing, can you clarify if there is/was any observable issues in the other topologies?

merrimanr:
As far as I know there was not an issue with other topologies.  Writing data tends to cause back pressure so that's why these changes were suggested for indexing topologies.

nickwallen:
For context, I set the default for Enrichments to 500.  There was some [discussion here](https://github.com/apache/metron/pull/984#discussion_r180150453).

I would echo my comments from that discussion here.  I think our defaults should focus on day 1 user experience.  Defining a default in the 300 - 700 range tends to be reasonable on most platforms for day 1.  What we don't want is to start Metron up and the topology to immediately die with an OOM because there is no back pressure.

nickwallen:
> @merrimanr: Which topologies should we provide defaults for?

To follow-up on the other topologies, we do not set a [default for the Profiler](https://github.com/apache/metron/blob/2ee6cc7e0b448d8d27f56f873e2c15a603c53917/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-profiler-env.xml#L168), which we probably should.

I am not sure about Parsers.

mmiklavc:
This seems reasonable to me all the way around. +1 by inspection.

merrimanr:
I added a default to the profiler topology.  We're probably less likely to run into a problem with the parser topologies I would think.  Should we provide defaults for those?

mmiklavc:
Outside of pcap, I've yet to see issues where parsers would have any issues handling inbound throughput. The profiler default seems reasonable to me as well. Still +1 from me.

PR #1234, opened 2018-10-11
METRON-1820 Syslog support for new api -> multiple messages and errors

ottobackwards:
This PR adds support for the Syslog parser and having multiple messages and errors.

Run a syslog 5424 source, with some invalid messages mixed in, the valid messages should still get indexed.

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/syslog/Syslog5424Parser.java
+++ b/metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/syslog/Syslog5424Parser.java
@@ -27,12 +27,17 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.metron.parsers.BasicParser;
 import org.apache.metron.parsers.DefaultMessageParserResult;
 import org.apache.metron.parsers.interfaces.MessageParserResult;
+import org.apache.metron.parsers.interfaces.MultilineMessageParser;
 import org.json.simple.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.Reader;
+import java.io.Serializable;
 import java.io.StringReader;
+import java.lang.invoke.MethodHandles;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
@@ -47,7 +52,8 @@ import java.util.Optional;
 /**
  * Parser for well structured RFC 5424 messages.
  */
-public class Syslog5424Parser extends BasicParser {
+public class Syslog5424Parser implements MultilineMessageParser<JSONObject>, Serializable {
+  protected static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   public static final String NIL_POLICY_CONFIG = "nilPolicy";
   private transient SyslogParser syslogParser;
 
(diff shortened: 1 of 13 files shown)


JonZeolla:
+1 via inspection of this and palindromicity/simple-syslog-5424#18, pending travis

ottobackwards:
I had missed support for the master throwable in the old List<JSONObject> parse function.  This function is not called by the Bolt anymore, but I corrected the issue in this and Grok.

metron-platform/metron-parsers/src/main/java/org/apache/metron/parsers/syslog/Syslog5424Parser.java
```diff
@@ -61,16 +67,37 @@ public void configure(Map<String, Object> config) {
   public void init() {
   }
 
-  @Override
   @SuppressWarnings("unchecked")
+  @Override
   public List<JSONObject> parse(byte[] rawMessage) {
+    Optional<MessageParserResult<JSONObject>> resultOptional = parseOptionalResult(rawMessage);
```
mmiklavc: Based on the changes to the ParserBolt from https://github.com/apache/metron/pull/1184, where would this still be used?

ottobackwards: It wouldn't, but the interface must be implemented.  It is awkward, I'm not sure how we want to proceed with that overall.  When we added the first parseOptional and moved the bolt to it we should have deprecated the old method or something....

ottobackwards: I'm open to a different approach, I could throw a notimplementedexception or something.

mmiklavc: Just spitballing here, but maybe an option here is to create a decorator interface around the older implementation and provide defaults when someone implements the new interface? Old api doesn't change and we handle new implementations gracefully.

```java
public interface MessageParser<T> extends Configurable {
  List<T> parse(byte[] rawMessage);
  Optional<List<T>> parseOptional(byte[] parseMessage);
}
public interface MultilineMessageParser<T> extends MessageParser {
  default List<T> parse(byte[] rawMessage) {
    throw new NotImplementedException();
  }
  default Optional<MessageParserResult<T>> parseOptionalResult(byte[] parseMessage) {
    List<T> list = new ArrayList<>();
      try {
        Optional<List<T>> optionalMessages = parseOptional(parseMessage);
        optionalMessages.ifPresent(list::addAll);
      } catch (Throwable t) {
        return Optional.of(new DefaultMessageParserResult<>(t));
      }
      return Optional.of(new DefaultMessageParserResult<T>(list));
  }
}
```

**ParserBolt:**
```java
MultilineMessageParser mmp = null;
if (!(parser instanceof MultilineMessageParser)) {
  mmp = new MultilineMessageParser<JSONObject>() {

    @Override
    public void configure(Map<String, Object> config) {
      parser.configure(config);
    }

    @Override
    public void init() {
      parser.init();
    }

    @Override
    public boolean validate(JSONObject message) {
      return parser.validate(message);
    }
  };
} else {
  mmp = (MultilineMessageParser) parser;
}

Optional<MessageParserResult<JSONObject>> results = mmp.parseOptionalResult(rawMessage.getMessage());
```

ottobackwards: I think that is a fine idea, I had thought of a new interface, but I decided to follow the precedent of parseOptional.

mmiklavc:
lgtm @ottobackwards. +1 by inspection.

PR #1239, opened 2018-10-16
METRON-1829: Large Error Message Causes Slow Search Performance

merrimanr:
## Contributor Comments
When a failure happens in the BulkWriterComponent, all pending messages in a batch are combined into a single error message.  This results in a single large objects being written to ES and causes performance problems.  This PR attempts to solve this issue by instead creating separate error messages for each message in the batch. 

### Testing
This has been tested and verified in full dev:

1. Spin up full dev and verify data is being indexed into ES
2. Stop HDFS in Ambari.  This will cause a failure in the BulkWriterComponent.  By default the batch size for OOTB sensors is set to 5.
3. After HDFS has stopped, verify that documents are being written to an ES error index
4. Retrieve a single document from the ES error index.  There should only be a single `raw_message` field.  Subsequent documents in ES should also only contain a single `raw_message` field.  Before there would have been 5 `raw_message_*` fields in a single document.
5.  Verify error documents are displayed properly in Kibana.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
@@ -140,24 +140,24 @@ public class BulkWriterComponent<MESSAGE_T> {
    * <p>Without a valid message, the JSON message cannot be added to the error.
    *
    * @param e The exception that occurred.
-   * @param tuples The tuples to error that may not contain valid messages.
+   * @param tuple The tuple to error that may not contain a valid message.
    */
-  public void error(Throwable e, Iterable<Tuple> tuples) {
-    LOG.error(format("Failing %d tuple(s)", Iterables.size(tuples)), e);
+  public void error(Throwable e, Tuple tuple) {
+    LOG.error("Failing tuple", e);
     MetronError error = new MetronError()
             .withErrorType(Constants.ErrorType.INDEXING_ERROR)
             .withThrowable(e);
-    handleError(tuples, error);
+    handleError(tuple, error);
   }
 
   /**
    * Errors a set of tuples.
    *
-   * @param tuples The tuples to error.
+   * @param tuple The tuple to error.
    * @param error
    */
-  private void handleError(Iterable<Tuple> tuples, MetronError error) {
-    tuples.forEach(t -> collector.ack(t));
+  private void handleError(Tuple tuple, MetronError error) {
+    collector.ack(tuple);
     ErrorUtils.handleError(collector, error);
   }
 
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
@@ -124,7 +124,7 @@ public class BulkWriterComponent<MESSAGE_T> {
               .withErrorType(Constants.ErrorType.INDEXING_ERROR)
               .withThrowable(e);
       error.addRawMessage(messageGetStrategy.get(t));
-      handleError(tuples, error);
+      handleError(Collections.singleton(t), error);
             });
 
   }
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
@@ -40,6 +40,7 @@ import org.apache.metron.common.writer.BulkMessageWriter;
 import org.apache.metron.common.writer.BulkWriterResponse;
 import org.apache.storm.task.OutputCollector;
 import org.apache.storm.tuple.Tuple;
+import org.apache.storm.tuple.Values;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -122,10 +123,13 @@ public class BulkWriterComponent<MESSAGE_T> {
       MetronError error = new MetronError()
               .withSensorType(Collections.singleton(sensorType))
               .withErrorType(Constants.ErrorType.INDEXING_ERROR)
-              .withThrowable(e);
-      error.addRawMessage(messageGetStrategy.get(t));
-      handleError(Collections.singleton(t), error);
-            });
+              .withThrowable(e)
+              .addRawMessage(messageGetStrategy.get(t));
+      collector.emit(Constants.ERROR_STREAM, new Values(error.getJSONObject()));
+      collector.ack(t);
+    });
+    // there is only one error to report for all of the failed tuples
+    collector.reportError(e);
 
   }
 
--- a/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
+++ b/metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
@@ -118,12 +118,15 @@ public class BulkWriterComponent<MESSAGE_T> {
 
   public void error(String sensorType, Throwable e, Iterable<Tuple> tuples, MessageGetStrategy messageGetStrategy) {
     LOG.error(format("Failing %d tuple(s); sensorType=%s", Iterables.size(tuples), sensorType), e);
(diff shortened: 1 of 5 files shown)


metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
```diff
@@ -118,12 +118,15 @@ public void commit(BulkWriterResponse response) {
 
   public void error(String sensorType, Throwable e, Iterable<Tuple> tuples, MessageGetStrategy messageGetStrategy) {
     LOG.error(format("Failing %d tuple(s); sensorType=%s", Iterables.size(tuples), sensorType), e);
-    MetronError error = new MetronError()
-            .withSensorType(Collections.singleton(sensorType))
-            .withErrorType(Constants.ErrorType.INDEXING_ERROR)
-            .withThrowable(e);
-    tuples.forEach(t -> error.addRawMessage(messageGetStrategy.get(t)));
-    handleError(tuples, error);
+    tuples.forEach(t -> {
```
nickwallen: Seems like this will ack the same tuples repetitively.  If 500 messages in a batch fail, then we will ack all 500 of them, 500 times.

There is also the nuisance that we report the error to the collector for each and every failed message, instead of just once for the batch.  There is only one `Throwable` error to report, so we should just report it once.

We may need something like this.

```suggestion
    // emit one error for each failed message
    tuples.forEach(t -> {
      MetronError error = new MetronError()
              .withSensorType(Collections.singleton(sensorType))
              .withErrorType(Constants.ErrorType.INDEXING_ERROR)
              .withThrowable(e)
              .addRawMessage(messageGetStrategy.get(t));
      collector.emit(Constants.ERROR_STREAM, new Values(error.getJSONObject()));
      collector.ack(t);
    });

    // there is only one error to report for all of the failed tuples
    collector.reportError(e);
  }
```

metron-platform/metron-writer/src/test/java/org/apache/metron/writer/BulkWriterComponentTest.java
```diff
@@ -161,9 +165,12 @@ public void writeShouldThrowExceptionWhenHandleErrorIsFalse() throws Exception {
   @Test
   public void writeShouldProperlyHandleWriterException() throws Exception {
     Throwable e = new Exception("test exception");
-    MetronError error = new MetronError()
+    MetronError expectedError1 = new MetronError()
             .withSensorType(Collections.singleton(sensorType))
-            .withErrorType(Constants.ErrorType.INDEXING_ERROR).withThrowable(e).withRawMessages(Arrays.asList(message1, message2));
+            .withErrorType(Constants.ErrorType.INDEXING_ERROR).withThrowable(e).withRawMessages(Collections.singletonList(message1));
```
nickwallen: Small nit.  While we are in here, do you think it is easier to read with newlines?  
```suggestion
            .withErrorType(Constants.ErrorType.INDEXING_ERROR)
            .withThrowable(e)
            .withRawMessages(Collections.singletonList(message1));
```

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
merrimanr: You're right!  My mistake.  I changed it to pass in just the current tuple to `handleError`.  Let me know if that works.

metron-platform/metron-writer/src/test/java/org/apache/metron/writer/BulkWriterComponentTest.java
merrimanr: I agree.  Done.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
nickwallen: It might also be useful to discuss on this PR, if the other similar method `error(Throwable, Iterable<Tuple>)`  needs updated.  That method gets called if messages are unparsable.  And in that case we send only a single error, instead of one for each error'd message.  Is that what we want?

It might be what we want, but at the very least it would be useful to comment in there as to why we treat it differently, if indeed it should be different.  Could be something like this.
```
  public void error(Throwable e, Iterable<Tuple> tuples) {
    LOG.error(format("Failing %d tuple(s) due to invalid message; sensorType unknown", Iterables.size(tuples)), e);

    // emit only a single error message, even though many failed because ...
    MetronError error = new MetronError()
            .withErrorType(Constants.ErrorType.INDEXING_ERROR)
            .withThrowable(e);
    collector.emit(Constants.ERROR_STREAM, new Values(error.getJSONObject()));
    tuples.forEach(t -> collector.ack(t));

    // there is only one error to report for all of the failed tuples
    collector.reportError(e);
  }
```

nickwallen: > You're right! My mistake. I changed it to pass in just the current tuple to handleError

Your latest commit was my first thought of a fix too; it is much simpler.  But I don't think it solves the second problem that I mentioned.  

> There is also the nuisance that we report the error to the collector for each and every failed message, instead of just once for the batch. There is only one Throwable error to report, so we should just report it once.

I think reporting the error once is both a usability and performance improvement.

I had to unravel the code base a bit more in my suggested code snippet so that it only reports the error once.

merrimanr: You're right again.  I agree, we shouldn't be reporting the same exception for every message in the batch.  I changed it to what you suggested in the initial comment and updated the tests.

For the other issue you bring up with `error(Throwable, Iterable<Tuple>)`, it looks like it gets called from only one place:  `BulkMessageWriterBolt.handleMissingMessage`.  This method only deals with a single tuple so it is misleading that the `error` method accepts multiple tuples.  Would changing that to accept a single tuple make it less confusing?

nickwallen: > This method only deals with a single tuple so it is misleading that the error method accepts multiple tuples. Would changing that to accept a single tuple make it less confusing?

Yes, agreed.  That would help clarify things.

merrimanr: Done.

nickwallen: Looks great.  I'm just going to spin it up and do some manual testing now.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
```diff
@@ -118,12 +119,18 @@ public void commit(BulkWriterResponse response) {
...
+      MetronError error = new MetronError()
+              .withSensorType(Collections.singleton(sensorType))
+              .withErrorType(Constants.ErrorType.INDEXING_ERROR)
+              .withThrowable(e)
+              .addRawMessage(messageGetStrategy.get(t));
```
nickwallen: While testing this, I am seeing a problem.  I am trying to decide whether we should call that a pre-existing condition and fix that problem after this PR goes in or whether the change here makes things any worse.  Let me know what you think.

I have described the problem in [METRON-1832](https://issues.apache.org/jira/browse/METRON-1832).  If any index destination goes down, error messages are continually recycled and grow larger and larger.  In addition, long sequence of escape characters are created.

nickwallen: Thinking more about this some more... I feel that what I have described in METRON-1832 is a pre-existing condition.  I don't think this PR makes that pre-existing condition any worse.  This PR is a solid, necessary fix.  I am in favor of merging this and I will open a discuss thread on a resolution for METRON-1832.

nickwallen:
+1 Works as advertised.  Thanks for the fix

PR #1240, opened 2018-10-17
METRON-1830: Re-implement Alerts dialog box without jQuery

sardell:
## Contributor Comments
[Link to METRON-1830 on ASF JIRA](https://issues.apache.org/jira/browse/METRON-1830)

Previously, the dialog box class in both UIs directly manipulated the DOM with jQuery. This is problematic when using a framework like Angular because it causes a disconnect with how Angular handles and tracks DOM changes. We also don't take advantage of Angular's rendering engine this way.

In this PR, I created a dialog component and service. When a component in the application needs to launch a dialog modal, they simply call a method on the singleton dialog service which determines if the dialog component should show/hide, what message to display and what type of message it is (ie Confirmation, Error, etc). 

### Changes Included

* Create a dialog component.
* Create a singleton service so that we aren't creating a new instance every time we include a dialog in a component.
* Add unit tests for both the service and component.
* Replace all MetronDialogBox class references with the DialogService.
* Remove the old MetronDialogBox.

### Testing
You can test and see the dialog menu in the following areas of the application:
* Creating a meta alert from a group in the tree view.
* Adding/removing an alert from a meta alert in the table view.
* Saving a search.
* Adding a comment to an alert.

It should behave just like it did before. If it's displaying an error message, there should only be an 'ok' button, which simply closes the dialog window. 

screen shot 2018-10-18 at 11 25 58 am

If there is a confirmation message, there should be an 'ok' button, which executes the desired behavior ie. creating a meta alert, and a 'cancel' button, which does not execute the desired behavior and closes the dialog.

screen shot 2018-10-18 at 11 25 42 am

 
To look at the test coverage, run `ng test --code-coverage`, navigate to the new `coverage` directory inside the project, and inside the folder open the index.html file. From there you can find the reports for the dialog service and component. Screenshots below for reference.

screen shot 2018-10-18 at 11 11 08 am

screen shot 2018-10-17 at 4 55 13 pm

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

- [x] N/A ~~If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?~~

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-interface/metron-alerts/src/app/alerts/alerts-list/table-view/table-view.component.ts
```diff
@@ -259,19 +259,25 @@ export class TableViewComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   deleteOneAlertFromMetaAlert($event, alert: Alert, metaAlertIndex: number) {
-    this.metronDialogBox.showConfirmationMessage('Do you wish to remove the alert from the meta alert?').subscribe(response => {
-      if (response) {
+    let confirmedSubscription = this.dialogService.confirm('Do you wish to remove the alert from the meta alert?').subscribe(r => {
+      if (r === 'Confirmed') {
         this.doDeleteOneAlertFromMetaAlert(alert, metaAlertIndex);
       }
+      if (r !== 'Initial') {
```
tiborm: What if we remove the if statement and simply unsubscribe at the end of this function?
With that, I think we could be able to remove 'Initial' value from ConfirmationType and the dispatching of 'Initial' from the first line of DialogService.confirm function. Could we keep the same functionality with this simplification?

Also, we can call variable 'r' like 'action' if we remove the coco egg.

metron-interface/metron-alerts/src/app/alerts/alerts-list/tree-view/tree-view.component.ts
```diff
@@ -421,10 +423,13 @@ export class TreeViewComponent extends TableViewComponent implements OnInit, OnC
     if (this.canCreateMetaAlert(group.total)) {
       let confirmationMsg = 'Do you wish to create a meta alert with ' +
                             (group.total === 1 ? ' alert' : group.total + ' selected alerts') + '?';
-      this.metronDialogBox.showConfirmationMessage(confirmationMsg).subscribe((response) => {
-        if (response) {
+      let confirmedSubscription = this.dialogService.confirm(confirmationMsg).subscribe(r => {
+        if (r === 'Confirmed') {
```
tiborm: I suggest to use (r === ConfirmationType.Confirmed) instead of the string literal.

sardell:
@tiborm In regard to the 'Initial' state check, I was able to do what you suggested but only after switching from a BehaviorSubject to a Subject. I did this because Subjects do not return the current value on Subscription. It only triggers with the .next() call and returns the value you pass there. That was a really great suggestion on your part and made me realize I was using the wrong tool for the job.

I also made changes based on the rest of your feedback, but they were straightforward and do not need further explanation here. Thanks for taking the time to review my PR!

tiborm:
Thanks, Shane. +1

merrimanr:
Looks good to me @sardell.  +1

PR #1242, opened 2018-10-19
METRON-1834: Migrate Elasticsearch from TransportClient to new Java REST API

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1834

This task has been a long time coming after having completed the ES upgrade in https://issues.apache.org/jira/browse/METRON-939. Motivation for completing this now is that Elasticsearch will be deprecating use of the TransportClient in v 7.x. This PR migrates the Elasticsearch client from TransportClient to the newer Java REST API. 

1. https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.6/client.html
2. https://www.elastic.co/guide/en/elasticsearch/client/java-rest/5.6/java-rest-overview.html
3. https://www.elastic.co/guide/en/elasticsearch/client/java-rest/5.6/java-rest-high-level-migration.html

This builds off and finishes work started by @cestella here - https://github.com/cestella/incubator-metron/tree/es_rest_client. I condensed his branch into 1 flattened commit and built on top of it in order to provide attribution.

I have a number of tasks I'm still working through, but I wanted to get the review process started. I've minimally validated X-Pack auth and will have some follow-up for SSL. Test plans and a breakdown of the changes will be soon to follow. For starters, full dev should continue to work as normal and you should see data flowing into indexes for bro, snort, and yaf. There are some additional changes to how this client will be configured, which I'll be documenting shortly. The new client does not take a Map of settings any longer now that it is leverage Apache HTTP Async Client https://www.elastic.co/guide/en/elasticsearch/client/java-rest/5.6/java-rest-low-usage-dependencies.html under the hood. This meant choosing a set of properties to expose and doing a translation to the builder pattern under the hood. Again, I'll have a write-up of this in the migration guide and update the README's accordingly.

NOTE: This checks off 2 items from this follow-on list https://github.com/apache/metron/pull/840#issuecomment-347281776

1. Fix Log4j logging problem - classpath issues
2. Migrate to new ES REST client

Per discussion in the Metron Slack channel, I will be updating the Jira ticket with a series of tasks to be completed prior to acceptance, including performance regression testing compared with the old API.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
[...]

PR #1247, opened 2018-10-25
METRON-1845 Correct Test Data Load in Elasticsearch Integration Tests

nickwallen:
The Elasticsearch integration tests use the Elasticsearch API to load the test data before running the tests.  Loading the test data like this does not accurately reflect how the indices will appear in a production environment.

This should be changed to use our existing `ElasticsearchUpdateDao` to write the test data.  This ensures that any changes made to the 'write' portion of our Elasticsearch code will function correctly with the 'read' portion. Said in a different way, this ensures that telemetry written into Elasticsearch by 'Indexing' can be read correctly by the Alerts UI.

- [x] Do not merge until after #1242.  This change is dependent on #1242.  Take a look at the last commit to view the changes specific to this PR.

## Testing

This only changes the integration tests.  If the integration tests pass, we're golden.

## Pull Request Checklist


metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/components/ElasticSearchComponent.java
```diff
@@ -194,35 +215,41 @@ public Client getClient() {
...
-      Map<String, Object> esDoc = JSONUtils.INSTANCE
-          .load(doc, JSONUtils.MAP_SUPPLIER);
-      indexRequestBuilder.setId((String) esDoc.get(Constants.GUID));
-      Object ts = esDoc.get("timestamp");
-      if (ts != null) {
-        indexRequestBuilder = indexRequestBuilder.setTimestamp(ts.toString());
-      }
-      bulkRequest.add(indexRequestBuilder);
-    }
+  public void add(UpdateDao updateDao, String indexName, String sensorType, Iterable<String> docs)
```
nickwallen: The Elasticsearch tests use this method to load test data.  Rather than using the Transport client here, we use the `UpdateDao` to load the test data.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchSearchIntegrationTest.java
```diff
@@ -97,48 +118,81 @@ protected static InMemoryComponent startIndex() throws Exception {
...
+
+    // write the test documents for Snort
+    List<String> snortDocuments = new ArrayList<>();
+    for (Object snortObject: (JSONArray) new JSONParser().parse(snortData)) {
+      snortDocuments.add(((JSONObject) snortObject).toJSONString());
     }
-    BulkResponse bulkResponse = bulkRequest.execute().actionGet();
-    if (bulkResponse.hasFailures()) {
-      throw new RuntimeException("Failed to index test data");
+    es.add(updateDao, snortIndex, "snort", snortDocuments);
```
nickwallen: We are now using `ElasticsearchComponent.add` to load the test data here.  This change matches what already exists in the other ES integration tests.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchSearchIntegrationTest.java
```diff
@@ -97,48 +118,81 @@ protected static InMemoryComponent startIndex() throws Exception {
...
+            .addMapping("snort_doc", JSONUtils.INSTANCE.toJSON(snortTemplate.get("mappings"), false)).get();
+
+    // setup the classes required to write the test data
+    AccessConfig accessConfig = createAccessConfig();
+    ElasticsearchClient client = ElasticsearchUtils.getClient(createGlobalConfig());
+    ElasticsearchRetrieveLatestDao retrieveLatestDao = new ElasticsearchRetrieveLatestDao(client);
+    ElasticsearchColumnMetadataDao columnMetadataDao = new ElasticsearchColumnMetadataDao(client);
+    ElasticsearchRequestSubmitter requestSubmitter = new ElasticsearchRequestSubmitter(client);
+    ElasticsearchUpdateDao updateDao = new ElasticsearchUpdateDao(client, accessConfig, retrieveLatestDao);
+    ElasticsearchSearchDao searchDao = new ElasticsearchSearchDao(client, accessConfig, columnMetadataDao, requestSubmitter);
```
nickwallen: All of this because I need an `UpdateDao` to write the tests data  and a `SearchDao` to make sure the test data is loaded.  Is there a simpler way?

metron-platform/metron-indexing/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
```diff
@@ -106,14 +106,14 @@ public void test() throws Exception {
       }}, Optional.empty());
 
       Assert.assertEquals(message0, update.getDocument());
-      Assert.assertEquals(1, getMockHTable().size());
+      Assert.assertEquals(10, getMockHTable().size());
```
nickwallen: Since I am reusing the `IndexDao` to load the tests data, the HBase table also gets this test data now.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchSearchIntegrationTest.java
```diff
@@ -97,45 +118,63 @@ protected static InMemoryComponent startIndex() throws Exception {
...
+    for (Object snortObject: (JSONArray) new JSONParser().parse(snortData)) {
+      snortDocuments.add(((JSONObject) snortObject).toJSONString());
     }
-    BulkResponse bulkResponse = bulkRequest.execute().actionGet();
-    if (bulkResponse.hasFailures()) {
-      throw new RuntimeException("Failed to index test data");
+    es.add(updateDao, snortIndex, "snort", snortDocuments);
+
+    // wait until the test documents are visible
+    assertEventually(() -> Assert.assertEquals(10, findAll(searchDao).getTotal()));
```
nickwallen: Unless I added this to hold up the test and wait for the tests data to be 'visible', some of the tests would reliably fail because the test data wasn't loaded.

nickwallen:
I have merged this with master to pull in the changes from #1242.  This is ready for review.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchSearchIntegrationTest.java
mmiklavc: @nickwallen is it possible to parameterize the call to do something similar to this? https://github.com/apache/metron/pull/1242/files#diff-99b67fd77500d6232462dc7c27ecff67R148

It's possible this is only available on batch calls, but I wanted to mention it in case we can get away with the assertEventually calls.

nickwallen: I have tried to do exactly this on a private branch that is closer to the code base in #1269.  I made the integration tests supply a RefreshPolicy that was supposed to hold them up until the loaded data is discoverable/searchable.  I was never able to get that to work consistently though despite what the ES docs say.  I still needed the `assertEventually` for the integration tests to pass consistently.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchSearchIntegrationTest.java
mmiklavc: Oh wow, do we not have a master factory for stitching these all together in the desired default manner? Not sure if this helps, but how bout this -> https://github.com/apache/metron/blob/master/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java#L97

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/components/ElasticSearchComponent.java
mmiklavc: Might it be better to just use IndexDao, which `extends UpdateDao, SearchDao, RetrieveLatestDao, ColumnMetadataDao`? To that end, if we're looking to route all of this through the ES component in that fashion, it might make sense to simply replace the internal `private Client client;` and instead use the new desired IndexDao for the proxied calls to ES. It could be setup at construction time of the component and remove the need to do the same thing for every test that uses the component class, unless they actually want to do something custom and pass in their own dao during init.

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/integration/SolrUpdateIntegrationTest.java
```diff
@@ -186,4 +195,114 @@ public void testHugeErrorFields() throws Exception {
...
+        Assert.assertEquals(2, columns.size());
+        Assert.assertEquals(message0, JSONUtils.INSTANCE.load(new String(columns.lastEntry().getValue())
+                , JSONUtils.MAP_SUPPLIER)
+        );
+        Assert.assertNotEquals(message0, JSONUtils.INSTANCE.load(new String(columns.firstEntry().getValue())
+                , JSONUtils.MAP_SUPPLIER)
+        );
+      }
+      {
+        //ensure ES is up-to-date
```
mmiklavc: Solr

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/integration/SolrUpdateIntegrationTest.java
```diff
@@ -186,4 +195,114 @@ public void testHugeErrorFields() throws Exception {
     exception.expectMessage("Document contains at least one immense term in field=\"error_hash\"");
     getDao().update(errorDoc, Optional.of("error"));
   }
+
+  @Test
+  @Override
+  public void test() throws Exception {
```
mmiklavc: What do these changes have to do with ES DAO read/write approach change? This class is testing Solr - should this be in a separate PR?

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchSearchIntegrationTest.java
mmiklavc: Are you able to get this to fail consistently? I ran it a number of times and couldn't get it to fail. I'm wondering if there are any other interesting entries in the logs when this occurs. These tests from ES might be of help - https://github.com/elastic/elasticsearch/pull/17986/files#diff-1c9f982dbd2f9ddb2853d135884621b5R112

nickwallen: > Are you able to get this to fail consistently? I ran it a number of times and couldn't get it to fail.

Can you describe what you did? You just removed line 190 and the tests continue to work reliably?

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/components/ElasticSearchComponent.java
nickwallen: > Might it be better to just use IndexDao, which extends UpdateDao, SearchDao, RetrieveLatestDao, ColumnMetadataDao? 

Why would that be better?  It needs to do updates, so it needs an `UpdateDao`.  Seemed logical to me.

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/integration/SolrUpdateIntegrationTest.java
nickwallen: This test was previously shared in UpdateIntegrationTest between ES and Solr.  With these changes, the tests don't behave exactly the same anymore.  That being said, what I did here doesn't look right.   I'll dig into this and see what is going on.

nickwallen: Ok, I found the problem. With my changes, the test data was getting loaded into both the ElasticsearchDao and the HBaseDao because I was using a MultiIndexDao to load the test data.  

In master, only Elasticsearch gets loaded with the test data.  This was causing some of the test assumptions to fail.  I corrected how the data is loaded so it only loads Elasticsearch and matches the existing behavior.

Thanks for pointing this out.

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/integration/SolrUpdateIntegrationTest.java
```diff
@@ -186,4 +195,114 @@ public void testHugeErrorFields() throws Exception {
...
+        Result r = getMockHTable().get(g);
+        NavigableMap<byte[], byte[]> columns = r.getFamilyMap(CF.getBytes());
+        Assert.assertEquals(1, columns.size());
+        Assert.assertEquals(message0
+                , JSONUtils.INSTANCE.load(new String(columns.lastEntry().getValue())
+                        , JSONUtils.MAP_SUPPLIER)
+        );
+      }
+      {
+        //ensure ES is up-to-date
```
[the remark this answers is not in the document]
nickwallen: No longer a problem.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/components/ElasticSearchComponent.java
nickwallen: > To that end, if we're looking to route all of this through the ES component in that fashion, it might make sense to simply replace the internal private Client client; and instead use the new desired IndexDao for the proxied calls to ES.

I don't think we're ready to do all that quite yet.  There is still some legacy functionality in `ElasticSearchComponent` that uses the underlying client for the old Admin API.  See [close](https://github.com/apache/metron/blob/fcd644ca77394d48d460c460b672a23d6594f49b/metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/components/ElasticSearchComponent.java#L292), [createIndexWithMapping](https://github.com/apache/metron/blob/fcd644ca77394d48d460c460b672a23d6594f49b/metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/components/ElasticSearchComponent.java#L228), and [start](https://github.com/apache/metron/blob/fcd644ca77394d48d460c460b672a23d6594f49b/metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/components/ElasticSearchComponent.java#L153).

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/ElasticsearchSearchIntegrationTest.java
mmiklavc: I meant on the es client change PR, not on this PR. Yes, I had removed assertEventually calls and it consistently passed for me in the search integration test.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/integration/components/ElasticSearchComponent.java
mmiklavc: The IndexDao handles all of the plumbing for the init that you're duplicating in this test class.

mmiklavc:
As a side note, I tend to like the approach of using the DAO layers to read/write the way you've done here @nickwallen . I've used this approach in the past, and it helped with test coverage and simplifying the amount of extra custom test code required. It might be worth an overall discussion on how we want to architect testing other similar endpoints going forward.

mmiklavc:
lgtm @nickwallen, +1 pending Travis

mmiklavc:
Still +1 on latest commit - that's a nice improvement. Glad the retry policy ended up working out!

PR #1251, opened 2018-11-02
METRON-1853: Add shutdown hook to Stellar BaseFunctionResolver

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1853

Noodling on a method to add shutdown hooks to Stellar. Modified the StellarFunction interface to include a teardown method. Added a no-op  to BaseStellarFunction to make it easy for implementations that don't need a teardown.

Rationale - some functions may have long-lived activities that include closeable resources. Adding the teardown method enables function implementers to close any resources opened during init gracefully.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunction.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunction.java
@@ -17,11 +17,17 @@
  */
 package org.apache.metron.stellar.dsl;
 
+import java.io.Closeable;
+import java.io.IOException;
 import java.util.List;
 
-public interface StellarFunction {
+public interface StellarFunction extends Closeable {
   Object apply(List<Object> args, Context context) throws ParseException;
   void initialize(Context context);
   boolean isInitialized();
-  void teardown();
+
+  @Override
+  default void close() throws IOException {
+
+  }
 }
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunction.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunction.java
@@ -23,4 +23,5 @@ public interface StellarFunction {
   Object apply(List<Object> args, Context context) throws ParseException;
   void initialize(Context context);
   boolean isInitialized();
+  void teardown(Context context);
 }
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunction.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunction.java
@@ -23,5 +23,5 @@ public interface StellarFunction {
   Object apply(List<Object> args, Context context) throws ParseException;
   void initialize(Context context);
   boolean isInitialized();
-  void teardown(Context context);
+  void teardown();
 }
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunctions.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunctions.java
@@ -30,4 +30,8 @@ public class StellarFunctions {
   public static void initialize(Context context) {
     SingletonFunctionResolver.getInstance().initialize(context);
   }
+
+  public static void teardown(Context context) {
+    SingletonFunctionResolver.getInstance().initialize(context);
+  }
 }
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunctions.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunctions.java
@@ -18,6 +18,7 @@
 
 package org.apache.metron.stellar.dsl;
 
+import java.io.IOException;
 import org.apache.metron.stellar.dsl.functions.resolver.FunctionResolver;
 import org.apache.metron.stellar.dsl.functions.resolver.SingletonFunctionResolver;
 
@@ -31,7 +32,7 @@ public class StellarFunctions {
     SingletonFunctionResolver.getInstance().initialize(context);
   }
 
-  public static void teardown(Context context) {
-    SingletonFunctionResolver.getInstance().initialize(context);
+  public static void close() throws IOException {
+    SingletonFunctionResolver.getInstance().close();
   }
 }
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/BaseFunctionResolver.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/BaseFunctionResolver.java
@@ -23,6 +23,7 @@ import static java.lang.String.format;
 import com.google.common.base.Joiner;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
+import java.io.IOException;
 import java.io.Serializable;
 import java.lang.invoke.MethodHandles;
 import java.util.Arrays;
@@ -95,12 +96,12 @@ public abstract class BaseFunctionResolver implements FunctionResolver, Serializ
   }
 
   /**
-   * Teardown the Stellar functions.
+   * Close the Stellar functions.
    */
   @Override
-  public void teardown() {
+  public void close() throws IOException {
     for (StellarFunctionInfo info : getFunctionInfo()) {
-      info.getFunction().teardown();
+      info.getFunction().close();
     }
   }
 
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/BaseFunctionResolver.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/BaseFunctionResolver.java
@@ -59,9 +59,15 @@ public abstract class BaseFunctionResolver implements FunctionResolver, Serializ
    */
   protected Context context;
 
+  /**
+   * Indicates if closed has been called on this resolver.
+   */
+  private boolean closed;
+
   public BaseFunctionResolver() {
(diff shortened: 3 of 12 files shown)


metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunctions.java
```diff
@@ -30,4 +30,8 @@ public static FunctionResolver FUNCTION_RESOLVER() {
   public static void initialize(Context context) {
     SingletonFunctionResolver.getInstance().initialize(context);
   }
+
+  public static void teardown(Context context) {
```
ottobackwards: Why is teardown initializing?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunction.java
```diff
@@ -23,4 +23,5 @@
   Object apply(List<Object> args, Context context) throws ParseException;
   void initialize(Context context);
   boolean isInitialized();
```
ottobackwards: Why not give this a default implementation so that you don't have to add do nothings to the implementations that are out there?

ottobackwards: if you are going to have a no parameter function, why not just implement closable?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunctions.java
merrimanr: Should this be:
```
public static void teardown() {
    SingletonFunctionResolver.getInstance().teardown();
  }
```

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/StellarFunction.java
mmiklavc: Yeah, was adding that as I posted.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/BaseFunctionResolver.java
```diff
@@ -94,6 +95,16 @@ public void initialize(Context context) {
     this.context = context;
   }
 
+  /**
+   * Close the Stellar functions.
+   */
+  @Override
+  public void close() throws IOException {
+    for (StellarFunctionInfo info : getFunctionInfo()) {
```
ottobackwards: Should there be an "Attempt" to close each function?  I think we want to have a try catch for each call here maybe and just throw with the functions that failed afterwards.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/FunctionResolver.java
```diff
@@ -43,4 +45,11 @@
    * @param context Context used to initialize.
    */
   void initialize(Context context);
+
+  /**
+   * Perform any cleanup necessary for the loaded Stellar functions.
+   */
```
ottobackwards: I don't understand why the override here?  Did you mean to default?

ottobackwards:
Should the resolver throw if there is a call after it is closed?
or just return nothing?  Do we have a test for that?  Either way we should.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/FunctionResolver.java
mmiklavc: Default is probably sufficient

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/BaseFunctionResolver.java
mmiklavc: On a related note, we may want to consider timeouts at some point, but we don't do this for initialization either.

mmiklavc: What about something like this?
```
public void close() throws IOException {
    Map<String, Throwable> errors = new HashMap();
    for (StellarFunctionInfo info : getFunctionInfo()) {
      try {
        info.getFunction().close();
      } catch (Throwable t) {
        errors.put(info.getName(), t);
      }
    }
    if (!errors.isEmpty()) {
      StringBuilder sb = new StringBuilder();
      sb.append("Unable to close Stellar functions:");
      for (Map.Entry<String, Throwable> e : errors.entrySet()) {
        Throwable throwable = e.getValue();
        String eText = String
            .format("Exception - Function: %s; Message: %s; Cause: %s", e.getKey(), throwable .getMessage(),
                throwable .getCause());
        sb.append(System.lineSeparator());
        sb.append(eText);
      }
      throw new IOException(sb.toString());
    }
  }
```

ottobackwards: Yes, that is great.

ottobackwards:
That is great

mmiklavc:
@ottobackwards Ok, pushed out an update for that change along with accompanying unit test.

mmiklavc:
### Testing

#### Create a custom Stellar function

See the following for reference - https://github.com/apache/metron/blob/master/metron-stellar/stellar-common/3rdPartyStellar.md

1. Setup project dirs

    ```
    mkdir custom-stellar && cd custom-stellar
    mkdir -p src/main/java/com/thirdparty/stellar
    ```

1. Create a pom.xml

    ```
    <project xmlns="http://maven.apache.org/POM/4.0.0"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
      <modelVersion>4.0.0</modelVersion>
    
      <groupId>com.thirdparty</groupId>
      <artifactId>stellar-funcs</artifactId>
      <version>1.0-SNAPSHOT</version>
      <packaging>jar</packaging>
    
      <name>Stellar Functions</name>
    
      <properties>
      </properties>
    
      <dependencies>
        <dependency>
          <groupId>org.apache.metron</groupId>
          <artifactId>stellar-common</artifactId>
          <version>0.6.1</version>
          <!-- NOTE: We will want to depend on the deployed common on the classpath. -->
          <scope>provided</scope>
        </dependency>
      </dependencies>
      <build>
        <plugins>
          <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-compiler-plugin</artifactId>
            <version>3.1</version>
            <configuration>
              <source>1.8</source>
              <target>1.8</target>
            </configuration>
          </plugin>
        </plugins>
      </build>
    </project>
    ```

1. Create a custom Stellar function

    ```
    package com.thirdparty.stellar;
    
    import org.apache.metron.stellar.common.utils.ConversionUtils;
    import org.apache.metron.stellar.dsl.Context;
    import org.apache.metron.stellar.dsl.ParseException;
    import org.apache.metron.stellar.dsl.Stellar;
    import org.apache.metron.stellar.dsl.StellarFunction;
    
    import java.io.IOException;
    import java.util.List;
    import org.jboss.netty.util.internal.ConversionUtil;
    
    public class StellarCustom {
    
      @Stellar(name = "NOW",
          description = "Simple now time",
          params = {
          "throwException - whether we should throw an exception on close()"
          },
          returns = "Time in millis"
      )
      public static class Now implements StellarFunction {
    
        private boolean isInitialized = false;
        private boolean throwException = false;
    
        public Object apply(List<Object> args, Context context) throws ParseException {
          if (args.size() == 1) {
            throwException = ConversionUtils.convert(args.get(0), Boolean.class);
          }
          return System.currentTimeMillis();
        }
    
        public void initialize(Context context) {
          try {
            System.out.println("Initializing function.");
          } finally {
            isInitialized = true;
          }
        }
    
        public boolean isInitialized() {
          return isInitialized;
        }
    
        public void close() throws IOException {
          System.out.println("Test function called close()!");
          if (throwException) {
            NullPointerException cause = new NullPointerException("Don't point at me!");
            throw new IOException("Something icky happened.", cause);
          }
        }
      }
    
    }
    ```

1. Build it

    `mvn clean install`

1. Ship it

    `scp custom-stellar/target/stellar-funcs-1.0-SNAPSHOT.jar root@node1:/tmp/`

1. Setup HDFS stellar

    ```
    sudo -u hdfs hdfs dfs -mkdir /apps/metron/stellar
    sudo -u hdfs hdfs dfs -put /tmp/stellar-funcs-1.0-SNAPSHOT.jar /apps/metron/stellar
    sudo -u hdfs hdfs dfs -chown -R metron:metron /apps/metron/stellar
    ```

1. Setup global config for Stellar HDFS location

    ```
    $METRON_HOME/bin/zk_load_configs.sh -m PULL -o $METRON_HOME/config/zookeeper -z $ZOOKEEPER -f
    vim $METRON_HOME/config/zookeeper/global.json
    # add this -> "stellar.function.paths" : "hdfs://node1:8020/apps/metron/stellar/.*.jar"
    # update global.json in ZK
    $METRON_HOME/bin/zk_load_configs.sh -m PUSH -i $METRON_HOME/config/zookeeper -z $ZOOKEEPER
    ```

1. Setup a parser or enrichment with the new function. (more specific details to come)

    Don't want an exception thrown 
    `NOW(false)`
    Restart parser or enrichment using this function
    Should get data still
    Stop the topology - should see a notice about shutdown in the logs.

1. Same as before but throw exception on shutdown. (more specific details to come)

    WANT an exception thrown 
    `NOW(true)`
    Restart parser or enrichment using this function
    Should get data still
    Stop the topology - should see our exception in the logs.

merrimanr:
+1

mmiklavc:
@ottobackwards Any other feedback on this?

ottobackwards:
+1

PR #1253, opened 2018-11-05
METRON-1857 Fix Metaalert Nested Alert Field Name in Index Template

nickwallen:
In #1049 the metaalert index template was updated to rename the alert field 'metron_alert'.  The nested alert field was missed in the original PR.

## Testing

1. Launch Full Dev.
1. Login to the Alerts UI.
1. Group alerts by IP
1. Find a grouping where there are multiple sensor types.
1. Create meta alert
1. Ensure that the meta alert is created successfully.

screen shot 2018-11-06 at 10 02 06 am

## Pull Request Checklist

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/files/metaalert_index.template
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/files/metaalert_index.template
@@ -5,7 +5,7 @@
       "dynamic_templates": [
         {
           "alert_template": {
-          "path_match": "alert.*",
+          "path_match": "metron_alert.*",
           "match_mapping_type": "string",
           "mapping": {
             "type": "keyword"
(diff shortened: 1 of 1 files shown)


merrimanr:
I'm not sure if you know the answer to this @nickwallen because it predates this PR, but is the intention to convert all `metron_alert.*` fields to keyword types?  I can see the  motivation behind doing this because we may not be aware of all field types in the various sensors that could be added to a metaalert.  Maybe @justinleet knows?

If my assumption is true then we need another small change to make that happen.  Currently the `match_mapping_type` attribute is set to `string` which will only convert string types.  If we want to convert all fields, it needs to be:
```
"dynamic_templates": [
        {
          "alert_template": {
          "path_match": "metron_alert.*",
          "match_mapping_type": "*",
          "mapping": {
            "type": "keyword"
          }
        }
```

Notice `string` has been changed to `*`.

mmiklavc:
lgtm, +1 by inspection.

PR #1254, opened 2018-11-05
METRON-1849 Elasticsearch Index Write Functionality Should be Shared

nickwallen:
The Elasticsearch index write functionality is currently duplicated between the `ElasticsearchWriter` and the `ElasticsearchUpdateDao`.  This functionality needs to be de-duplicated and shared between the two.  This will help ensure that any changes that need to be made to how alerts are written, for instance whether the client sets a document ID, are implemented and tested in the same way across Metron.

This change depends on the following PRs.
- [ ] #1242
- [ ] #1247

## Changes

* Created the `BulkDocumentWriter` and `ElasticsearchBulkDocumentWriter` that are responsible for writing messages/alerts/documents to an Elasticsearch index.

* Created the `IndexedDocument` which is what the `BulkDocumentWriter` actually writes.  This is a `Document` that contains the name of the index it is written to.  

    * I could not just add `index` to `Document` because the `HbaseDao` does not know how to populate this field.  When using a `MultiIndexDao`, the `Document` returned can come from any of the underlying DAOs, including the `HBaseDao`, so this field cannot always be populated with the current archtecture.

* The `ElasticsearchWriter` uses a `BulkDocumentWriter` to index messages in Elasticsearch. This also uses a `TupleBasedDocument` that allows it to maintain the relationship between a tuple and the document created from a tuple.  This is needed to properly ack/fail tuples.

* The scope of the `ElasticsearchWriter` tests were increased since this was now easier with the new abstractions. 

* The `ElasticsearchUpdateDao` uses a `BulkDocumentWriter` to update alerts from the Alerts UI.

* The `ElasticsearchRetrieveLatestDao` was setting the timestamp of all `Document`s returned to 0.  This was causing the tests that I added to fail.  I added logic to extract the timestamp from the message and set the `Document.timestamp` if one exists.

Instead of creating the `BulkDocumentWriter` abstraction, I also looked at a couple alternatives.

* Another approach I looked at was to simply have the `ElasticsearchWriter` use an `ElasticsearchUpdateDao`.  Unfortunately, this quickly led to a large number of changes. The `ElasticsearchUpdateDao` does not track and record individual failures in a batch.  It instead fails fast if any single update in a batch fails.  To track individual failures, as needed by the `ElasticsearchWriter`, would have required changes to the signature of `UpdateDao`.

* Another approach was to have the `ElasticsearchUpdateDao` use an `ElasticsearchWriter`.  Unfortunately, the `ElasticsearchWriter` is a `BulkMessageWriter` and requires Storm as a dependency due to its use of `TopologyContext`.  To clean this up would have required a large number of changes.

## Testing

Spin-up the Full Dev environment.  Use the Alerts UI to:
 * Search for alerts
 * Escalate alerts
 * Add comments to alerts
 * Remove comments from alerts
 * Create meta-alerts
 * Escalate meta-alerts

## Pull Request Checklist


metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/bulk/ElasticsearchBulkDocumentWriter.java
```diff
@@ -0,0 +1,184 @@
...
+    public void addDocument(D document, String index) {
+        documents.add(new Indexable(document, index));
+        LOG.debug("Adding document to batch; document={}, index={}", document, index);
+    }
+
+    @Override
+    public void write() {
+        try {
+            // create an index request for each document
+            List<DocWriteRequest> requests = documents
```
mmiklavc: Please be careful using streams. In general, but specifically here. Our ES endpoints are typically where we bottleneck in any tuning exercise and adding additional overhead may prove troublesome. I'd normally not stress too much on "premature optimization" but I think we're safely very much post-premature at this stage of the game with our Lucene stores.

https://jaxenter.com/java-performance-tutorial-how-fast-are-the-java-8-streams-118830.html

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
```diff
@@ -56,90 +58,107 @@
...
+      TupleBasedDocument document = new TupleBasedDocument(source, guid, sensorType, timestamp, tuple);
+      documentWriter.addDocument(document, indexName);
     }
 
-    BulkResponse bulkResponse = client.getHighLevelClient().bulk(bulkRequest);
-    return buildWriteReponse(tuples, bulkResponse);
+    // add successful tuples to the response
+    BulkWriterResponse response = new BulkWriterResponse();
+    documentWriter.onSuccess(docs -> {
+      List<Tuple> successfulTuples = docs.stream().map(doc -> doc.getTuple()).collect(Collectors.toList());
```
mmiklavc: Same comment on streams as above.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/bulk/ElasticsearchBulkDocumentWriter.java
nickwallen: I usually think of us as IO bound, rather than compute, in Indexing.  But I don't mind refactoring out the use of streams, just to avoid any potential problems.

nickwallen: It turned out to be much cleaner to get rid of the streams.  So double-win.  Thanks.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
nickwallen: Done too. Thanks.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/bulk/BulkDocumentWriter.java
```diff
@@ -0,0 +1,79 @@
...
+ * @param <D> The type of document to write.
+ */
+public interface BulkDocumentWriter<D extends Document> {
+
+    /**
+     * A listener that is notified when a set of documents have been
+     * written successfully.
+     * @param <D> The type of document to write.
+     */
+    interface SuccessListener<D extends Document> {
```
mmiklavc: How's this compare with the recent changes for decoupling the Parser Bolts from parser exec? At a glance, this seems very similar from an abstraction standpoint. Any reason we would want callbacks here as compared with the decisions we made for the parser bolt being more direct?

https://github.com/apache/metron/pull/1213#issuecomment-427086826

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/update/Document.java
```diff
@@ -89,46 +91,29 @@ public void setGuid(String guid) {
...
-    }
-    if (document != null ? !document.equals(document1.document) : document1.document != null) {
-      return false;
-    }
-    if (guid != null ? !guid.equals(document1.guid) : document1.guid != null) {
-      return false;
-    }
-    return sensorType != null ? sensorType.equals(document1.sensorType)
-        : document1.sensorType == null;
+    return Objects.equals(timestamp, document1.timestamp) &&
```
mmiklavc: Is this change to equals and hashcode an auto-create from IntelliJ?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
```diff
@@ -56,90 +60,111 @@
...
       }
 
-      String indexName = ElasticsearchUtils.getIndexName(sensorType, indexPostfix, configurations);
-      IndexRequest indexRequest = new IndexRequest(indexName, sensorType + "_doc");
-      indexRequest.source(esDoc.toJSONString());
-      String guid = (String)esDoc.get(Constants.GUID);
-      if(guid != null) {
-        indexRequest.id(guid);
+      // define the document id
+      String guid = String.class.cast(source.get(Constants.GUID));
```
mmiklavc: Should be using https://github.com/apache/metron/blob/master/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/ConversionUtils.java#L39 for conversions.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriterTest.java
```diff
@@ -18,170 +18,241 @@
...
-
-        BulkResponse response = mock(BulkResponse.class);
-        when(response.hasFailures()).thenReturn(true);
-
-        Exception e = new IllegalStateException();
-        BulkItemResponse itemResponse = buildBulkItemFailure(e);
-        when(response.iterator()).thenReturn(ImmutableList.of(itemResponse).iterator());
+    public void shouldWriteManySuccessfully() {
+        // create a writer where all writes will be successful
+        float probabilityOfSuccess = 1.0F;
```
mmiklavc: Hm, interesting - so the stub will definitively fail some percentage of tuples for all POS vals < 1.0F?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
```diff
@@ -196,7 +196,7 @@ public ElasticsearchDao withRefreshPolicy(WriteRequest.RefreshPolicy refreshPoli
   }
 
   protected Optional<String> getIndexName(String guid, String sensorType) throws IOException {
-    return updateDao.getIndexName(guid, sensorType);
+    return updateDao.findIndexNameByGUID(guid, sensorType);
```
mmiklavc: Is sensorType not a component to retrieving the index name? Also, would we want any parity between the updateDao's find method name vs the ElasticsearchDao's getIndexName method name?

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/update/Document.java
nickwallen: It is an auto-create from IntelliJ.  It changed when on a previous iteration, I added a field to the Document class.  I since backed that out and went with another approach.  So there really is no need for this to change now.  I will back this out.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriterTest.java
nickwallen: Right.  Just needed a way to test successes, failures, and partial successes.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
nickwallen: Agreed.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
nickwallen: > Also, would we want any parity between the updateDao's find method name vs the ElasticsearchDao's getIndexName method name?

I found the [code here confusing](https://github.com/apache/metron/blob/89a2beda4f07911c8b3cd7dee8a2c3426838d161/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchUpdateDao.java#L195-L197) and had me stuck on an issue for quite some time.  When both use `getIndexName` I have no idea what the logic is doing.  It tries one approach, then falls back to another, but since the methods are named the same, it doesn't tell me how they attempt to find the index name in a different way.

With the rename, I feel it improves understanding in a glance [what this is doing now](https://github.com/apache/metron/blob/260ccc366b79ef53595dbfd097066040444b4eda/metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchUpdateDao.java#L179) and the differences between the primary approach versus the fallback.


> Is sensorType not a component to retrieving the index name? 

So you prefer the original function name?  Or you prefer `lookupIndexName`, `findIndexNameByGUIDAndSensor`?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/bulk/BulkDocumentWriter.java
nickwallen: Yep, I think you're right.  Wish I would have noticed this before. Daft.

nickwallen: Done. See latest commit.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/writer/BulkDocumentWriterStub.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public void addDocument(D document, String index) {
+        documents.add(document);
+    }
+
+    @Override
+    public BulkDocumentWriterResults<D> write() {
+        BulkDocumentWriterResults<D> results = new BulkDocumentWriterResults<>();
+        Random random = new Random();
```
justinleet: Why are we doing this randomly?

We have the number of Documents, and we have the proportion. Why not just calculate
D * p and then just produce that many successes and the corresponding number of failures?

I don't see what we're gaining by making it probabilistic, and I can see cases where people unknowingly choose small numbers of Documents (e.g. 3 with probability .5) where suddenly it fails 1/4 of the time (all all successes or all failures when it's expecting a mix).

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchUpdateDao.java
```diff
@@ -17,103 +17,90 @@
...
+      String msg = format("Failed to update all documents; %d successes, %d failures", successes, failures);
+      LOG.error(msg);
+
+      // log each individual failure
+      for(WriteFailure<Document> failure: results.getFailures()) {
+        LOG.error(failure.getMessage(), failure.getCause());
+      }
+
+      // raise an exception
+      Throwable cause = results.getFailures().get(0).getCause();
```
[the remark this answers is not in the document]
nickwallen: The current behavior is for the `ElasticsearchUpdateDao` to throw an exception if any of the writes fail.  Inside this if statement, we know there is at least 1 exception and possibly many more.  I need to choose one of those exceptions as the root cause, so I just use the first. Make sense?

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/bulk/BulkDocumentWriterResults.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * The result of writing documents in bulk using a {@link BulkDocumentWriter}.
+ * @param <D> The type of documents to write.
+ */
+public class BulkDocumentWriterResults<D extends Document> {
+
+  private List<WriteSuccess> successes;
```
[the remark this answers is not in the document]
nickwallen: Good catch. Fixed.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/bulk/BulkDocumentWriter.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @return The number of documents waiting to be written.
+     */
+    int size();
+
+    /**
+     * Write all documents in the batch.
+     */
+    BulkDocumentWriterResults write();
```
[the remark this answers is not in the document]
nickwallen: Good catch. Fixed.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/bulk/ElasticsearchBulkDocumentWriter.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param bulkResponse The response received from Elasticsearch.
+     * @param documents The documents included in the bulk request.
+     * @param results The writer results.
+     * @return The documents that were successfully written. Failed documents are excluded.
+     */
+    private void handleBulkResponse(BulkResponse bulkResponse, List<Indexable> documents, BulkDocumentWriterResults<D> results) {
+        if (bulkResponse.hasFailures()) {
+
+            // interrogate the response to distinguish between those that succeeded and those that failed
+            Iterator<BulkItemResponse> iterator = bulkResponse.iterator();
```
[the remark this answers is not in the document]
justinleet: BulkResponse itself is `Iterable`

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchUpdateDao.java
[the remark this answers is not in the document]
justinleet: Yeah, I'm fine with that, I was just hoping there's a more elegant solution, but unfortunately I don't think there is, given the way Exception and stack traces are handled.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/bulk/ElasticsearchBulkDocumentWriter.java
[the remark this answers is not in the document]
nickwallen: Ah, gotcha. Ya, I like it.  Fixed.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/writer/BulkDocumentWriterStub.java
nickwallen: I was able to remove the stub completely and just use a mock.  Much simpler.  Thanks for the suggestion.

metron-platform/metron-elasticsearch/src/test/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriterTest.java
nickwallen: FYI - With the latest commits, I was able to completely remove this stub and just use a mock.   I couldn't figure out how to do this with mocks when I had the callbacks, but since that was refactored to be simpler, this became easier too.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchDao.java
mmiklavc: That's fine @nickwallen. I wasn't clear on the difference between the 2 functions and wanted to call it out.

mmiklavc:
+1 lgtm with the latest changes, thanks for the submission @nickwallen!

nickwallen:
Thanks for the feedback @justinleet and @mmiklavc .  It turned out better for it.  Will merge ones that latest CI build completes.

PR #1258, opened 2018-11-08
METRON-1864 fix issue where daylight savings breaks tests

ottobackwards:
Thanks to @justinleet for the fix.  I would have use a totally awesome but overdone regex.

### For all changes:

### For code changes:
- [-] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [-] Have you included steps or a guide to how the change may be verified and tested manually?
  ```
  ```

- [-] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/DateFunctionsTest.java
+++ b/metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/DateFunctionsTest.java
@@ -28,6 +28,8 @@ import org.apache.metron.stellar.dsl.StellarFunctions;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
 import java.util.Calendar;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


mmiklavc:
Thanks @ottobackwards and @justinleet!

Did you mean to completely remove the assertions for that test? It seems like we would still want to validate the results somehow.

Also, since the context for this PR came from discussions in Slack that don't show up on the dev list, can you add a comment about the motivation/origin for this fix?

justinleet:
@mmiklavc Right now the function just tests that the result is properly formatted (which is really all the original test did, but it just checked the string matched a timezone).  This will throw an exception if it's not.

The problem with validating further on this test is that it's essentially against "now", which means that we don't know exactly what the date is, other than what we get back.  Anything we choose would be slightly off.

We could probably refactor things around to make it possible to inject a given timestamp for a no arg run of the statement, but I'm not convinced that needs to happen to resolve the current unit test problem.

mmiklavc:
@justinleet We do use mock clocks in other parts of the platform, but for the sake of this fix I think this is fine. @ottobackwards Can we just throw a comment in the test so it's obvious that this is intentionally without assertions here?

ottobackwards:
I think that the test logically tests what we are concerned about, that the format of the produced string matches the desired format.  Testing that the value is explicitly x,y,z isn't testing the format, it is testing something else.

I propose, with a mind to getting master and pr's building again, that I comment as requested but not change the test, and that any other ideas for improvement and or expansion of the tests be put to a jira for later work.

justinleet:
@ottobackwards I'm good with that

mmiklavc:
+1

PR #1261, opened 2018-11-13
METRON-1860  new developer option for ansible in docker to deploy to vagrant

ottobackwards:
The goal of this PR is to provide a new "full_dev" option for new and old users that does not require as much setup and version matching to try Metron's full dev environment.

Currently, the vagrant up command runs ansible locally, on the host machine, to build and deploy metron.  This means that the user must not only have Vagrant, Virtual Box and Docker, but must also have all the tools necessary to build metron ( maven, java, c++ 11 etc ) and run ansible ( python and others ).  It has been a common source of problems for new users to get started with Metron because of version or setup problems.

This PR introduces a new metron-deployment/development option which tries to address this problem, and make it possible for the user to only have Vagrant, VirtualBox and Docker ( along with a local copy of the source ) to be able to run full dev.

The new option starts the Vagrant VM, but does not run ansible in it. Instead it runs a docker container which contains all the tools/versions necessary, and that container is what runs ansible.

##Testing
Have the correct versions of vagrant, virtual box and docker installed and running
```
cd $METRON_SRC_ROOT/metron-deployment/development/centos6_docker_build
./build_and_run.sh
```
Answer yes to building the vagrant box.
Answer yes to building the docker machine
Go grab a coffee.

The end result should be full dev running in the vagrant instance.
The logs directory will have a log for each run.

If you run a second time, you can say no to building the docker machine.

### Differences
- This does not support skip tags passed on the cli
- This does not support provision

### For all changes:

### For code changes:
- [-] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  ```

- [-] Have you written or updated unit tests and or integration tests to verify your changes?


nickwallen:
Valiant effort @ottobackwards .  I am just wondering how much easier this really makes it?

Have you thought about just publishing a Metron Demo image to Vagrant Cloud?  Does that scratch the same itch? 
* Everything is already pre-installed (removes one source of potential problems for new users.)
* All a user needs is Vagrant and VirtualBox.

mmiklavc:
@ottobackwards thanks for the submission. Per recent community comments, it sounds like we could use some improvements to how our build/deploy dep versions interact with other tooling that may require other versions of things. 

Is this the correct dependency listings for each host/container?

The Docker container will be pre-configured with:

- Java 8
- Ansible 2.4.0+
- Python 2.7
- Maven 3.3.9
- C++11 compliant compiler, like GCC

And the developer host machine would now need to manage versions of:

- Vagrant 2.0+
- Vagrant Hostmanager Plugin
- Virtualbox 5.0+
- Docker

The deployment goes through these steps:

1. You mount m2 repo and your code dir in the Docker instance.
2. build_and_run optionally spins up Vagrant in VirtualBox.
3. build_and_run optionally creates Docker instance with pre-reqs for building and deploying Metron.
4. build_and_run runs the build and then calls the Ansible deployment scripts from within Docker.

At the end, you have an ephemeral Docker instance + Vagrant instance that has the running Metron instance?

ottobackwards:
@nickwallen I did not think of that.  I was improving the process that stands today.  I think in a world where the posted image exists, we would still want the ability to try the latest ( to verify a fix pre-release etc ).

ottobackwards:
The integration test failure has to do with the profiler tests and seem unrelated.

nickwallen:
If we go this approach, why not just replace the existing "Full Dev" environments (both centos and ubuntu) rather than add new environments to support, test, and keep in-sync?

nickwallen:
How does the build time compare to the current approach (`mvn clean install -DskipTests`)?  Does it change at all since it is now run in a Docker container?  

How does the time it takes to get Full Dev running compare (whatever is equivalent to `vagrant up`)?

JonZeolla:
I'm going to take a stab at a further look next week.  For now I gave it a quick run-up and it was successful.

justinleet:
In the interest of pure (and almost certainly incredibly ignorant) speculation, along the lines of the vagrant cloud, is it theoretically possible to save off a base image that has the Hadoop and non-Metron installs done? Then if said base image is missing (or possibly has changed properties, e.g. HDP version) rebuild it, and if found just build Metron and install mpack?  Or alternatively, don't load the end result to Vagrant Cloud, just upload the base image we install on top of.  Then only update it on a base image update.

What I'm getting at is that it would be nice to be able to build off of master, in a way that doesn't require an external dependency, and still lets us cache off the majority of the install (setting up HDP and Ambari). At that point, full dev would essentially be (for most builds) "build metron, build RPMs, run up image, do Metron install and setup".  Which is probably 20 minutes faster and would make me a substantially happier person.

ottobackwards:
It is possible to imagine a number of scenarios, including that, but also needing to build with new hadoop versions ( can't loose build from scratch ).

There are a number of things we can do down the road.

I think this work is going to help people enough in the near term to land it, while we discuss longer term refactoring and workflow.

ottobackwards:
If you create an issue for your vagrant base machine with our hadoop / ambari already in it, you can assign it to me. @justinleet

nickwallen:
hey @ottobackwards - I am still wrapping my head around this, but one small nit is that I don't like all the confirmation prompts.  With the prompts I have to constantly check back to ensure it is not stuck on another prompt waiting for me to do something.  

As much as possible it should just be fire-and-forget, so I can run it, work on something else, and hopefully come back some time later with a functioning Metron install.

ottobackwards:
@nickwallen, yeah,  I did prompts as I went along debugging.  I was thinking that folks may not like them.
I'll parameterize things.

ottobackwards:
@nickwallen Done

metron-deployment/development/centos6_docker_build/build_and_run.sh
```diff
@@ -0,0 +1,144 @@
...
+   help
+   exit 0
+   shift # past argument with no value
+  ;;
+
+ #
+ # Unknown option
+ #
+  *)
+   UNKNOWN_OPTION="${i#*=}"
```
JonZeolla: Why `#*=`?

metron-deployment/development/centos6_docker_build/build_and_run.sh
```diff
@@ -0,0 +1,144 @@
...
+if [[ ! -d ~/.m2 ]]; then
+ mkdir ~/.m2
+fi
+
+DATE=`date`
+LOG_DATE=${DATE// /_}
+LOGNAME="metron-build-${LOG_DATE}.log"
+
+# get the node1 ip address so we can add it to the docker hosts
+NODE1_IP=$(awk '/^\s*hosts/{flag=1; next} /}]/{flag=0} flag' ${VAGRANT_PATH}/Vagrantfile | grep  "^\s*ip:" | awk -F'"' '{print $2}')
```
JonZeolla: `s/"^\s/"^\\s/`

metron-deployment/development/centos6_docker_build/build_and_run.sh
```diff
@@ -0,0 +1,144 @@
...
+if [[ "$FORCE_DOCKER_BUILD" = true ]]; then
+ echo "docker build"
+ docker build -t metron-build-docker:latest .
+fi
+
+if [[ ! -d ~/.m2 ]]; then
+ mkdir ~/.m2
+fi
+
+DATE=`date`
```
JonZeolla: Shellcheck prefers `$(date)` ¯\_(ツ)_/¯

metron-deployment/development/centos6_docker_build/build_and_run.sh
```diff
@@ -0,0 +1,144 @@
...
+docker run -it \
+ -v ${VAGRANT_PATH}/../../..:/root/metron \
+ -v ~/.m2:/root/.m2 \
+ -v ${VAGRANT_PATH}:/root/vagrant \
+ -v ${ANSIBLE_PATH}:/root/ansible_config \
+ -v ${VAGRANT_KEY_PATH}:/root/vagrant_key \
+ -v ${VAGRANT_PATH}/logs:/root/logs \
+ -e ANSIBLE_CONFIG='/root/ansible_config/ansible.cfg' \
+ -e ANSIBLE_LOG_PATH="/root/logs/${LOGNAME}" \
+ -e ANSIBLE_SKIP_TAGS=${A_SKIP_TAGS} \
```
JonZeolla: From shellcheck.

```
In build_and_run.sh line 130:
 -v ${VAGRANT_PATH}/../../..:/root/metron \
    ^-- SC2086: Double quote to prevent globbing and word splitting.


In build_and_run.sh line 132:
 -v ${VAGRANT_PATH}:/root/vagrant \
    ^-- SC2086: Double quote to prevent globbing and word splitting.


In build_and_run.sh line 133:
 -v ${ANSIBLE_PATH}:/root/ansible_config \
    ^-- SC2086: Double quote to prevent globbing and word splitting.


In build_and_run.sh line 134:
 -v ${VAGRANT_KEY_PATH}:/root/vagrant_key \
    ^-- SC2086: Double quote to prevent globbing and word splitting.


In build_and_run.sh line 135:
 -v ${VAGRANT_PATH}/logs:/root/logs \
    ^-- SC2086: Double quote to prevent globbing and word splitting.


In build_and_run.sh line 138:
 -e ANSIBLE_SKIP_TAGS=${A_SKIP_TAGS} \
                      ^-- SC2086: Double quote to prevent globbing and word splitting.
```

metron-deployment/development/centos6_docker_build/docker_run_ansible.sh
```diff
@@ -0,0 +1,46 @@
...
+
+
+#
+# This script runs IN the docker container
+#
+
+cd /root/metron || exit 1
+
+# make sure we have the right c++ tools
+source /opt/rh/devtoolset-6/enable
```
JonZeolla: Consider adding `# shellcheck disable=SC1091` prior.

metron-deployment/development/centos6_docker_build/run_docker_only.sh
```diff
@@ -0,0 +1,39 @@
...
+cd ../docker || exit 1
+pwd
+
+echo "===============Running Docker==============="
+docker run -it \
+ -v  ${VAGRANT_PATH}/../../..:/root/metron \
+ -v ~/.m2:/root/.m2 \
+ -v ${VAGRANT_PATH}:/root/vagrant \
+ -v ${ANSIBLE_PATH}:/root/ansible_config \
+ -v ${VAGRANT_KEY_PATH}:/root/vagrant_key \
```
JonZeolla: From shellcheck.
```
In run_docker_only.sh line 34:
 -v  ${VAGRANT_PATH}/../../..:/root/metron \
     ^-- SC2086: Double quote to prevent globbing and word splitting.


In run_docker_only.sh line 36:
 -v ${VAGRANT_PATH}:/root/vagrant \
    ^-- SC2086: Double quote to prevent globbing and word splitting.


In run_docker_only.sh line 37:
 -v ${ANSIBLE_PATH}:/root/ansible_config \
    ^-- SC2086: Double quote to prevent globbing and word splitting.


In run_docker_only.sh line 38:
 -v ${VAGRANT_KEY_PATH}:/root/vagrant_key \
    ^-- SC2086: Double quote to prevent globbing and word splitting.
```

metron-deployment/development/centos6_docker_build/run_docker_only.sh
```diff
@@ -0,0 +1,39 @@
...
+#  limitations under the License.
+#
+
+VAGRANT_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+echo "setting the ansible configuration path"
+ANSIBLE_PATH=${VAGRANT_PATH}/ansible
+echo ${ANSIBLE_PATH}
+echo "setting the ssh key"
+VAGRANT_KEY_PATH=`pwd`/.vagrant/machines/node1/virtualbox
+echo ${VAGRANT_KEY_PATH}
```
JonZeolla: From shellcheck
```
In run_docker_only.sh line 26:
echo ${VAGRANT_KEY_PATH}
     ^-- SC2086: Double quote to prevent globbing and word splitting.
```

metron-deployment/development/centos6_docker_build/run_docker_only.sh
```diff
@@ -0,0 +1,39 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+VAGRANT_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+echo "setting the ansible configuration path"
+ANSIBLE_PATH=${VAGRANT_PATH}/ansible
+echo ${ANSIBLE_PATH}
+echo "setting the ssh key"
+VAGRANT_KEY_PATH=`pwd`/.vagrant/machines/node1/virtualbox
```
JonZeolla: Shellcheck prefers `$(pwd)` ¯_(ツ)_/¯

metron-deployment/development/centos6_docker_build/run_docker_only.sh
```diff
@@ -0,0 +1,39 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+VAGRANT_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+echo "setting the ansible configuration path"
+ANSIBLE_PATH=${VAGRANT_PATH}/ansible
+echo ${ANSIBLE_PATH}
```
JonZeolla:
```
In run_docker_only.sh line 23:
echo ${ANSIBLE_PATH}
     ^-- SC2086: Double quote to prevent globbing and word splitting.
```

metron-deployment/development/docker/README.md
```diff
@@ -45,12 +45,4 @@ This provides a Docker Container containing all of the prerequisites required to
     ```
     cd /root/metron
     mvn clean package -DskipTests
-    ```
```
JonZeolla: This should be reinstated.

metron-deployment/development/centos6_docker_build/README.md
```diff
@@ -0,0 +1,89 @@
...
+    ```
+
+### Deploy Metron
+
+1. Ensure that the Docker service is running.
+
+1. Deploy Metron
+
+ ```bash
+    cd metron-deployment/development/centos6_docker
```
JonZeolla: `s/centos6_docker/centos6_docker_build/`

metron-deployment/development/centos6_docker_build/README.md
```diff
@@ -0,0 +1,89 @@
...
+ ```
+ ```bash   
+    usage: ./build_and_run.sh
+        --skip-vagrant-up               skip vagrant up
+        --force-docker-build            force build docker machine
+        --skip-tags='tag,tag2,tag3'     the ansible skip tags
+        -h/--help                       Usage information.
+
+    example: to skip vagrant up and force docker build with two tags
+        build_and_run.sh -skip-vagrant-up --force-docker-build --skip-tags='solr,sensors'
```
JonZeolla: `s/-skip-vagrant-up/--skip-vagrant-up/`

metron-deployment/development/centos6_docker_build/README.md
```diff
@@ -0,0 +1,89 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+Metron on CentOS 6 Built in Docker
+==================================
+
+This project fully automates the provisioning and deployment of Apache Metron and all necessary prerequisites on a single, virtualized host running CentOS 6.
+It utilizes Vagrant for the virtual machine, and Docker for the build and deployment.  Therefore lessens the burden on the user to have the correct versions of the build and deployment tools in order to try Metron.
+
+Metron is composed of many components and installing all of these on a single host, especially a virtualized one, will greatly stress the resources of the host.   The host will require at least 8 GB of RAM and a fair amount of patience.  It is highly recommended that you shut down all unnecessary services.
```
JonZeolla: I'd suggest at most 2 spaces between sentences, if not one.

metron-deployment/development/centos6_docker_build/build_and_run.sh
```diff
@@ -0,0 +1,144 @@
...
+function help {
+ echo " "
+ echo "usage: ${0}"
+ echo "    --skip-vagrant-up               skip vagrant up"
+ echo "    --force-docker-build            force build docker machine"
+ echo "    --skip-tags='tag,tag2,tag3'     the ansible skip tags"
+ echo "    -h/--help                       Usage information."
+ echo " "
+ echo "example: to skip vagrant up and force docker build with two tags"
+ echo "   build_and_run.sh -skip-vagrant-up --force-docker-build --skip-tags='solr,sensors'"
```
JonZeolla: `s/-skip-vagrant-up/--skip-vagrant-up/`

metron-deployment/development/centos6_docker_build/ansible/inventory/group_vars/all
```diff
@@ -0,0 +1,86 @@
...
+java_home: /usr/jdk64/jdk1.8.0_77
+
+# hbase
+pcap_hbase_table: pcap
+tracker_hbase_table: access_tracker
+threatintel_hbase_table: threatintel
+enrichment_hbase_table: enrichment
+
+# metron
+metron_version: 0.6.1
```
JonZeolla: Just bringing awareness that this may get changed to `0.7.0` prior to the next release.  It's possible this could get missed depending on the order of operations.

metron-deployment/development/centos6_docker_build/ansible/inventory/group_vars/all
```diff
@@ -0,0 +1,86 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+# only need to build the RPMs for CentOS
+metron_build_packages_cmd: "shell cd {{ metron_build_dir }}/metron-deployment && mvn clean package -DskipTests -Pbuild-rpms"
```
JonZeolla: I see this is in our current ubuntu and centos `all` files, but what is shell, is this an ansible thing?

metron-deployment/development/centos6_docker_build/run_docker_only.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "setting the ssh key"
+VAGRANT_KEY_PATH=`pwd`/.vagrant/machines/node1/virtualbox
+echo ${VAGRANT_KEY_PATH}
+
+# move over to the docker area
+cd ../docker || exit 1
+pwd
```
JonZeolla: Is this still useful?

metron-deployment/development/centos6_docker_build/docker_run_ansible.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# give the option to skip building metron, in case they have already done so
+#read -p "  build metron? [yN] " -n 1 -r
+#echo
+#if [[ $REPLY =~ ^[Yy]$ ]]; then
+# USE TAGS
+#fi
```
JonZeolla: We should clean up lines 30-35.

metron-deployment/development/centos6_docker_build/build_and_run.sh
ottobackwards: To make sure we consume the option and value

ottobackwards: This is kind of boilerplate

metron-deployment/development/centos6_docker_build/build_and_run.sh
ottobackwards: ok

metron-deployment/development/centos6_docker_build/build_and_run.sh
ottobackwards: ok

metron-deployment/development/centos6_docker_build/docker_run_ansible.sh
ottobackwards: ok

metron-deployment/development/centos6_docker_build/build_and_run.sh
ottobackwards: ok

metron-deployment/development/centos6_docker_build/ansible/inventory/group_vars/all
ottobackwards: yes

metron-deployment/development/centos6_docker_build/run_docker_only.sh
ottobackwards: Not sure what you mean

ottobackwards: oh, the pwd.... I guess not, this is a debugging leftover.  I'll remove

metron-deployment/development/centos6_docker_build/build_and_run.sh
JonZeolla: I would have thought this would only provide back the value, filtering out the option.

metron-deployment/development/centos6_docker_build/docker_run_ansible.sh
ottobackwards: yes

metron-deployment/development/centos6_docker_build/build_and_run.sh
ottobackwards: this change breaks the statement.  I disabled the warning.

mmiklavc:
@JonZeolla - would it make sense to add ShellCheck to Travis? Seems like a lot of potentially useful detail here.

metron-deployment/development/centos6_docker_build/build_and_run.sh
JonZeolla: Hmm, I successfully executed it locally.  Did you do it on the awk?  It should only get applied to the grep.

JonZeolla:
@mmiklavc probably at some point.  I assume it would require some cleanup for various scripts at the same time.

ottobackwards:
@mmiklavc I have it integrated into IntelliJ, though it isn't as obvious as java errors

ottobackwards:
I think spellcheck is integrated into travis, so what we _can_ do is make shellcheck a build dependency and execute it through maven in the poms

metron-deployment/development/centos6_docker_build/build_and_run.sh
ottobackwards: ahhh, yes, I did it for both.

ottobackwards: Fixed

mmiklavc:
@ottobackwards I like @justinleet's suggestion on having a base option. I'd imagine what we could have is a mechanism that allows you to build an image with just HDP of a specific version, and snapshot that if you like for general use, along with the option of just running against a standard image. I dunno, have we just re-hashed quickdev here? I can't recall specifically how or why it became problematic for us to maintain. It seems like there should be a path through this that allows us to not go stale, but also save the 20 min building the base installation, as Justin pointed out.

mmiklavc:
@ottobackwards - yeah, that's probably reasonable. The base HDP image create option would be an additional improvement to existing functionality.

ottobackwards:
anyone have any other thoughts?  Have you tried it?

ottobackwards:
I have created METRON-1967 for the base image issue.  Where are we at with this pr?

nickwallen:
Is this not still a WIP?  If not, maybe update the PR title.

ottobackwards:
sorry

nickwallen:
There are a few things holding me back here. I like the end goal, but I don't fully understand the approach.

(1) Why do we need a new VM definition in `metron-deployment/development/` to achieve this?  Why do we have to maintain a parallel set of artifacts in `metron-deployment/development/centos6_docker_build/`?  

(2) I don't grok the need for all the scripts like `build_and_run.sh`, `docker_run_ansible.sh`, `run_docker_only.sh`.  That's just a lot of bash that I'd prefer to not maintain.  Can we simplify to match just the functionality we have today?  Why all the special handling for Cypress?

ottobackwards:
for

1) I did it as a 'new' think to make it more clear and reviewable, and leave open the question of having multiple approaches, I'm open to whatever in this regard, if we want to replace both the ubuntu and centos machine folders with docker versions, that is fine.  The ubuntu change would be a follow on.  The longer term goal would be to have a script that let you chose the os and options and just started or built whatever you wanted.

2) Without specifying the cypress cache, the build downloads a ton of cypress stuff every time.  I wrote a separate script to just run docker, to cover other use cases and things that came up during development, like if you wanted to make changes before you ran ansible etc.  I guess the scripts could be merged and parameterized.  My docker scripts for bro plugin are better, but came after this.

ottobackwards:
What would you propose @nickwallen?  

- replacing the existing centos6 in place?
- parameterizing the scripts so that there is on run_* script?

nickwallen:
> @ottobackwards: replacing the existing centos6 in place?

Yeah, let's just change how the existing VMs are built.  We want everyone to be building Metron the same way, not two different mechanisms.  I think that makes the most sense.

And we need to reconcile this with the existing build mechanism that is driven by that Ansible 'metron-builder' role.  Both Ubuntu and Centos are built using that `metron-builder` role, so I expect that to no longer be needed.

> @ottobackwards: parameterizing the scripts so that there is on[e] run_* script?

Sure.  Let's not over-complicate things. If `vagrant up` no longer works, then one simple script that does the same thing works for me. `./run.sh` is pretty self-explanatory.

ottobackwards:
OK, I'll flag when I've refactored.  Thanks!

mmiklavc:
Just wanted to chime in that I'm in support of this effort. We've had, and continue to have, numerous folks running into environmental issues on everything from Ansible versions to the version of sed on their cli paths.

* https://lists.apache.org/thread.html/5e0cf029f026113ae74b4cd1f0f370515999319fcd27c18be2458008@%3Cuser.metron.apache.org%3E
* https://lists.apache.org/thread.html/5bfade813bc5c95635be26656e8bb87ec9780dd070c27859c6a13472@%3Cdev.metron.apache.org%3E

metron-deployment/development/centos6_docker_build/ansible/inventory/group_vars/all
mmiklavc: Version will minimally need updated to 0.7.0 or 0.7.1 depending on when next release gets rolling.

ottobackwards:
@mmiklavc @nickwallen should we align this with possible work with Centos7 support?

nickwallen:
What do we need to align?  If you just replace the existing build mechanism, then there is no overlap between this and needing CentOS 7.  Can you describe more what you are suggesting?

ottobackwards:
I mean, instead of replacing centos6, do we want to use this method when building centos7 from the start

nickwallen:
Correct me if I am wrong.  My understanding is that you are in-progress on refactoring this so that the existing build mechanism (using the `metron-builder` Ansible role) is updated to build via Docker, instead of adding a different build mechanism.  

If we follow that approach, then CentOS 7 will seamlessly use your Docker build changes, just like Ubuntu would.

ottobackwards:
no, that is not correct.  Ansible + the build run in Docker.

mmiklavc:
@ottobackwards This PR won't change the Vagrant image we base full dev off of, will it? Correct me if I'm misunderstanding something, but I thought that's the only place where we need to change the CentOS version.

ottobackwards:
My understanding from looking at centos 7 and full dev in the past was that it was not trivial wrt ansible changes.  Is that no longer true?

nickwallen:
> no, that is not correct. Ansible + the build run in Docker.

Which part of what I said is not correct?  Can you elaborate?  Sounds like you are saying is that Ansible + Docker is what is going to change, which is also my understanding.

> My understanding from looking at centos 7 and full dev in the past was that it was not trivial wrt ansible changes. Is that no longer true?

I already have a branch with Full Dev running on CentOS 7.

ottobackwards:
@nickwallen , I took your statement to mean that only that role runs in or executes docker, when I'm running the entire ansible run in docker.  If that is not what you meant, then we are just misunderstanding and are on the same page.

Glad that you have ansible with centos7 working.

So the question is:
Since I'm going to refactor this to _replace_ the centos6, maybe in light of what we are going to do with centos7 there would be a different better way.

ottobackwards:
Like having a top level script, where you pick ubuntu or centos7 that just sets it all up instead having multiple top level areas

nickwallen:
@ottobackwards Does this accurately describe the changes?  You are effectively adding the blue "Docker 2" box?

Metron Builds and METRON-1860

ottobackwards:
No, it is not accurate exactly, I'll post an updated version of that

ottobackwards:
Copy of Metron Builds and METRON-1860


> NOTE:  I added 'awareness' to not run the rpm docker if we where in the ansible docker a while ago

ottobackwards:
@nickwallen @mmiklavc 
I have ->
- replaced the old centos6 with the new dockerized build and deploy
- renamed the script to metron-up.sh
- removed the spurious script to get a docker shell
- taken master

ottobackwards:
It is much easier to see the changes this way, @nickwallen ++

nickwallen:
This is looking real nice.  Isn't the script supposed to build the container for me?
```
nallen : ~/tmp/metron-pr1261/metron-deployment/development/centos6
$ ./metron-up.sh
Running with
SKIP_VAGRANT_UP    = false
FORCE_DOCKER_BUILD = false
SKIP_TAGS          = sensors,solr
===================================================
Bringing machine 'node1' up with 'virtualbox' provider...
==> node1: Importing base box 'centos/6'...
==> node1: Matching MAC address for NAT networking...
==> node1: Checking if box 'centos/6' is up to date...
==> node1: Setting the name of the VM: centos6_node1_1556575069291_69092
==> node1: Clearing any previously set network interfaces...
==> node1: Preparing network interfaces based on configuration...
    node1: Adapter 1: nat
    node1: Adapter 2: hostonly
==> node1: Forwarding ports...
    node1: 22 (guest) => 2222 (host) (adapter 1)
==> node1: Running 'pre-boot' VM customizations...
==> node1: Booting VM...
==> node1: Waiting for machine to boot. This may take a few minutes...
    node1: SSH address: 127.0.0.1:2222
    node1: SSH username: vagrant
    node1: SSH auth method: private key
    node1: Warning: Connection reset. Retrying...
    node1: Warning: Remote connection disconnect. Retrying...
    node1: Warning: Connection reset. Retrying...
    node1: Warning: Remote connection disconnect. Retrying...
    node1: Warning: Connection reset. Retrying...
    node1: Warning: Remote connection disconnect. Retrying...
    node1:
    node1: Vagrant insecure key detected. Vagrant will automatically replace
    node1: this with a newly generated keypair for better security.
    node1:
    node1: Inserting generated public key within guest...
    node1: Removing insecure key from the guest if it's present...
    node1: Key inserted! Disconnecting and reconnecting using new SSH key...
==> node1: Machine booted and ready!
==> node1: Checking for guest additions in VM...
    node1: No guest additions were detected on the base box for this VM! Guest
    node1: additions are required for forwarded ports, shared folders, host only
    node1: networking, and more. If SSH fails on this machine, please install
    node1: the guest additions and repackage the box to continue.
    node1:
    node1: This is not an error message; everything may continue to work properly,
    node1: in which case you may ignore this message.
==> node1: Setting hostname...
==> node1: Configuring and enabling network interfaces...
    node1: SSH address: 127.0.0.1:2222
    node1: SSH username: vagrant
    node1: SSH auth method: private key
==> node1: Rsyncing folder: /Users/nallen/tmp/metron-pr1261/metron-deployment/development/centos6/ => /vagrant
==> node1: [vagrant-hostmanager:guests] Updating hosts file on active guest virtual machines...
==> node1: [vagrant-hostmanager:host] Updating hosts file on your workstation (password may be required)...
Password:
Log will be found on host at /Users/nallen/tmp/metron-pr1261/metron-deployment/development/centos6/logs/metron-build-Mon_Apr_29_17:58:48_EDT_2019.log
Using NODE1 IP 192.168.66.121
Cypres Cache is set to '/Users/nallen/.Library/caches/Cypress'
===============Running Docker===============
Unable to find image 'metron-build-docker:latest' locally
docker: Error response from daemon: pull access denied for metron-build-docker, repository does not exist or may require 'docker login'.
See 'docker run --help'.
```

metron-deployment/development/centos6/docker_run_ansible.sh
```diff
@@ -14,15 +16,25 @@
...
-log_path = ./ansible.log
-
-# fix for "ssh throws 'unix domain socket too long' " problem
-[ssh_connection]
-control_path = ~/.ssh/ansible-ssh-%%h-%%r
+
+
+#
+# This script runs IN the docker container
+#
```
nickwallen: Since this script is not intended to be run by the user, what if we put this in a sub-directory?  I think this would make it more clear as to its intended purpose.
```
metron-deployment/development/centos6/docker/run_ansible.sh
````

This would also match how we have all the Ansible stuff living under `metron-deployment/development/centos6/ansible`.

And then there is only one "top-level" script so its blatantly obvious what I'm supposed to run.

File: metron-deployment/development/centos6/docker_run_ansible.sh [hunk 20]
```diff
@@ -14,15 +16,25 @@
...
+
+cd /root/metron || exit 1
+
+# make sure we have the right c++ tools
+
+#shellcheck disable=SC1091
+source /opt/rh/devtoolset-6/enable
+
+sed -i -e '/^node1.*/d' ~/.ssh/known_hosts
```
nickwallen: It my test run, I am getting this error.  
```
sed: can't read /root/.ssh/known_hosts: No such file or directory
```

I understand this won't be a problem on the second run, but maybe we can guard this with a file existence check.

metron-deployment/development/centos6/docker_run_ansible.sh
ottobackwards: Good idea, that is how I do it in the plugin, which I did after this, will do

ottobackwards:
@nickwallen please check the latest

nickwallen:
Hey @ottobackwards , I have been testing the changes and I am noticing that it *seems* to take much longer to spin-up Full Dev with these changes.  For example, the last run took 87 minutes.  Is this unique to my environment?  How long does it take for you?

ottobackwards:
what command are you running?

nickwallen:
Those 87 minutes do not include building the Docker image itself.  I already had that built and just ran the following.
```
time ./metron-up.sh
```

ottobackwards:
Let me try it and see.  Full Dev is kind of a kick it off and then go run an errand thing for me, I don't watch it

ottobackwards:
Can you check the log in the logs/ directory and see if anything stands out as taking the most time?
Is it building or settings up the cluster etc etc?

ottobackwards:
I would think it would have to be a difference in the build portion, since that is now in docker.  the deploy was always and still is run in the vagrant machine

JonZeolla:
Is it your CPU/Memory preferences in docker?

mmiklavc:
I was going to say, we may need to bump the mem resources for that Docker container. The build is a beast, and it takes 10x as long for me when I've got a Vagrant instance running along with a bunch of browser tabs.

nickwallen:
> @JonZeolla: Is it your CPU/Memory preferences in docker?

Doubt it. CPUs: 6, Mem: 12 GB, Swap 3 GB

Is it running sufficiently fast for you @JonZeolla ?

nickwallen:
> @mmiklavc: ...and it takes 10x as long for me when I've got a Vagrant instance running along with a bunch of browser tabs.

I was thinking about this the other day actually.  Since with these changes we no longer rely on Vagrant to kick-off Ansible, we could alter the steps so that it first completes the build and packaging and then only after that is done, it launches the VM.  That way we don't have the two fighting for memory at the same time.

First off, does that sound like a reasonable thing to do @ottobackwards ?  If so, is it a heavy lift we should hold off on for a follow-on PR?  Or is it just a simple re-ordering of some of the actions in `metron-up.sh`?

ottobackwards:
./metron-up.sh  6.30s user 3.09s system 0% cpu 1:20:47.77 total

@nickwallen I think that is worth doing definitely!   I'll get on it.

nickwallen:
Can we do it as a follow-on PR?

So I can't tell from your message. How long did it take for you?  Wall clock time?

ottobackwards:
I think it will take me a couple of hours

ottobackwards:
Ok, I'm going to do that as a follow on

ottobackwards:
https://issues.apache.org/jira/browse/METRON-2098

nickwallen:
Hey @ottobackwards - What are your thoughts on the duration of the build?  Do you have any thoughts on how we could improve that?

ottobackwards:
You mean other than not run the vm?

nickwallen:
I don't know where this stands.  Do you think this should be merged in spite of the time it takes to spin-up?  Or are you looking at ways to improve the build time?  What is the path forward here?

ottobackwards:
I have a secondary branch and I'm working through building without the vm running
I'll have time comparisons soon, let's see how that pan's out

ottobackwards:
This latest merge contains a refactoring that makes the resource use much better, by delaying the vagrant machine start until after the builds are done.

This involves refactoring the playbooks and the scripts.

Screen Shot 2019-05-06 at 15 51 11

ottobackwards:
Some thoughts for next steps.  

- try setting volume sync options to :delegate ( I tried and it didn't make much difference, but we could look more )
- build the docker image, but add a last layer with a COPY of all the source into the image.  That way no volume penalty, and caching will make the docker build fast.

- new vagrant base image that has the cluster already installed, just without metron to make deploy faster

mmiklavc:
@ottobackwards Thanks for that comparison info, that makes it really easy to compare. 

It's a fractionally small part of the build, but what's going on with clean taking 8x as long?

Those time drops are pretty substantial - is that purely resource contention?

Thinking through this build process a bit more, just curious, are you sharing local m2 repo from the host? I'm assuming this doesn't pull every dep down local to the Docker container. That's probably worth adding to the diagram for clarity.

ottobackwards:
I believe the issue with clean is due to slowness with docker doing writes to osx hosted volumes.
I think resource contention is a big part of this on my machine at least.
I _do_ share the repo.
https://github.com/apache/metron/pull/1261/files#diff-80771b861babcdefd8d7bbc81c412ff1R110

metron-deployment/development/centos6/host_scripts/docker_run_build_container.sh
```diff
@@ -0,0 +1,143 @@
...
+  Darwin*)    CYPRESS_CACHE=~/.Library/caches/Cypress;;
+esac
+
+# Build the docker command line
+declare -a DOCKER_CMD_BASE
+#DOCKER_CMD="bash -c /root/scripts/docker_run_ansible_install.sh"
+DOCKER_CMD="bash"
+DOCKER_CMD_BASE[0]="docker run -d -t --name MetronBuild "
+DOCKER_CMD_BASE[1]="-v \"${VAGRANT_PATH}/../../..:/root/metron\" "
+DOCKER_CMD_BASE[2]="-v ~/.m2:/root/.m2 "
```
mmiklavc: 👍

ottobackwards:
@mmiklavc Somone else needs to try it obviously

ottobackwards:
@mmiklavc as I stated above, if we need to, another change would be to create a docker layer that was a copy of the source, such that we didn't need to use the volumes at all....  but I'd have to do some research on that as a follow on

nickwallen:
@ottobackwards I ran your latest code up a few times and I am not able to replicate your results unfortunately.  

This is the overall duration to go from source code to a fully deployed dev environment.
* Master; `time vagrant up` = 65 - 68 minutes
* #1261 : `time ./metron-up.sh` = 85 - 90 minutes

Your results differ significantly from mine... 
* Master takes you 103 minutes, which is 35 minutes longer my experience.
* #1261 takes 67 minutes, which is 18 minutes less than my experience.

(1) Any ideas why we are getting such different results?  I have allocated Docker 12G of RAM, 6 cores, and 3G of swap.  Are there any other settings that you have adjusted on your build machine?

(2) Which logs did you use to break out the build, clean, package times?  Since Ansible runs on Docker, we no longer have that log persisted on the host.  Once the Docker container is shut-down the logs are lost, unless I am missing something.

ottobackwards:
In this PR, the ansible log goes to the /logs directory

mmiklavc:
@ottobackwards - I want to spend some time in the next week resurrecting this and taking another look to see where we're at.

ottobackwards:
@mmiklavc I think we are doing docker wrong, wrt building.
I think we may want to have the build docker be at the root of the source and use docker ignore to set that up, that way the src is automagically in the image etc.

mmiklavc:
@luozhenwei - it's generally best to ask these questions on the user or dev list, but here's the list of jars we depend on in our full dev (e.g. https://github.com/apache/metron/tree/master/metron-deployment/development/centos6) env:
```
[root@node1 ~]# ls -1 /usr/metron/0.7.2/lib
metron-common-0.7.2.jar
metron-data-management-0.7.2.jar
metron-elasticsearch-storm-0.7.2-uber.jar
metron-enrichment-common-0.7.2-uber.jar
metron-enrichment-storm-0.7.2-uber.jar
metron-maas-service-0.7.2-uber.jar
metron-management-0.7.2.jar
metron-parsers-0.7.2-uber.jar
metron-parsers-common-0.7.2-uber.jar
metron-parsing-storm-0.7.2-uber.jar
metron-pcap-backend-0.7.2.jar
metron-performance-0.7.2.jar
metron-profiler-repl-0.7.2.jar
metron-profiler-spark-0.7.2.jar
metron-profiler-storm-0.7.2-uber.jar
metron-rest-0.7.2.jar
metron-solr-storm-0.7.2-uber.jar
```

17 jars. The Docker deploy you linked to hasn't been maintained/updated in quite some time, so I'm not entirely sure what its current state is. @merrimanr may have more detail on this. If you're just trying to explore Metron, I would run up full dev on centos6 via the instructions in the link I provided above.

PR #1259, opened 2018-11-13
METRON-1867 Remove `/api/v1/update/replace` endpoint

nickwallen:
The `/api/v1/update/replace` endpoint is no longer used.  This is dead code and should be removed.

This is dependent on the following PRs.
- [x] #1242
- [ ] #1247
- [ ] #1254 

## Testing

1. Spin-up Full Dev.
1. Open the Alerts UI.
    * View alerts.
    * Escalate an alert.
    * Comment on an alert.
    * Create a meta-alert.
    * Escalate a meta-alert.
1. Run the UI e2e tests.

## Pull Request Checklist

- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel).
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
  ```
  mvn -q clean integration-test install && dev-utilities/build-utils/verify_licenses.sh 
  ```
- [x] Have you written or updated unit tests and or integration tests to verify your changes?
- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [x] Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

PR #1269, opened 2018-11-16
METRON-1677 UUIDv4 GUID is not Lucene friendly

nickwallen:
With this change, when documents are written to Elasticsearch the document ID is no longer set as the Metron GUID, but instead left unset so that Elasticsearch can auto-generate it.  Doing this improves write performance into Elasticsearch.

This will also be the case for any Lucene based Indexer, including Solr.  This work only covers Elasticsearch, but the same should be done for Solr as part of a separate effort.

While the default table view looks the same, in the following screenshot I customized the table to show both the document ID and the GUID. 

This change is dependent on the following open pull requests.
- [ ] #1247
- [ ] #1254 
- [ ] #1259  

## Changes

* The `ElasticsearchRetrieveLatestDao` was updated since the GUID is no longer the document ID.  This instead does a terms query on the GUID field instead of an ID query.

* The `Document` class now contains an optional documentID field.  If the `Document` is retrieved from one of the DAOs this field will be populated.  When creating a new document, this field will be empty.

* Many of the integrations tests had to be updated because the GUID and document ID are now different.

* The Alert UI was updated so that it visually looks the same. By default, the Metron GUID is still shown as one of the first columns in the table.  

    * The table is actually showing the document's GUID field instead of the document ID as it was before.  The ID field remains, which contains the document ID generated by Elasticsearch.  The user can choose to add this to the table, if they like.

## Testing

1. Spin-up Full Dev.

1. Open up the Alerts UI and perform the following basic actions.
    * Search for alerts
    * Escalate an alert
    * Comment on an alert
    * Delete a comment from an alert
    * Create a meta-alert
    * Escalate a meta-alert

1. Click on the configure wheel and add the 'id' field to the table view.  This will now display both the GUID and document ID in the table. They of course will be different.

1.  Click on the 'guid' field in any row to filter the search results by the guid.
    screen shot 2018-11-14 at 2 44 21 pm

1. Click on the 'id' field to filter the search results by the document ID.
    screen shot 2018-11-14 at 2 44 08 pm

1. Group by some fields to drill into the data.  In the tree view, click on the 'guid' column and ensure the data sorts correctly.  Do the same for the 'id' column that was added.
    screen shot 2018-11-14 at 2 46 43 pm

## Pull Request Checklist


nickwallen:
I was doing some final testing and noticed that the 'guid' field for metaalerts is showing the document ID, rather than the guid.  I need to fix the presentation of metaalerts in the UI.

screen shot 2018-11-19 at 10 56 19 am

nickwallen:
This issue has been fixed.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
```diff
@@ -56,90 +58,107 @@
...
+      String guid = String.class.cast(source.get(Constants.GUID));
+      if(guid == null) {
+        LOG.info("Missing '{}' field; document ID will be auto-generated.", Constants.GUID);
       }
 
-      Object ts = esDoc.get("timestamp");
-      if(ts != null) {
-        indexRequest.timestamp(ts.toString());
+      // define the document timestamp
+      Long timestamp = Long.class.cast(source.get(Constants.Fields.TIMESTAMP.getName()));
```
nickwallen: @anandsubbu pointed out to me that this previously worked when the timestamp is a String.  The code now requires a Long.  I need to address this point, although I am not sure how it should be right now.  I need to do a little research.  Thanks @anandsubbu !

mmiklavc:
Refresh me on this - prior to this change, did we provide both a doc id and a guid that were identical? And now with this change you would expect to see guid != doc id (bc guid generated by Metron, doc id now generated by ES), except where otherwise configured for backwards compatibility?

nickwallen:
Prior to this change we use the Metron-generated GUID as the documentID.  They were always the same.  There were various places in the code where the name `guid`, `docID`, or `id` were used that implicitly meant the Metron GUID.

With this change, the Metron-generated guid will never be equal to the Elasticsearch-generated doc ID. You can view and use either the guid or the doc ID in the UI.  Both are available as separate fields in the index.

This PR has pulled on those strings and tried to make clear what is a Metron GUID and what is a document ID.  In the Alerts UI you can see and/or use either the GUID or the document ID.  It is up to the user which one they care to see.  Although most users will not care what the document ID is.

From my PR description:

> This change is backwards compatible. The Alerts UI should continue to work no matter if some of the underlying indices were written with the Metron GUID as the document ID and some are written with the auto-generated document ID.

There is no option to continue to use the GUID as the documentID.  I can't think of a use case worthy of the additional effort and testing needed to support that.

It is backwards compatible in that the Alerts UI will work when searching over both "legacy" indices where guid = docID and "new" indices where guid != docID.  All places in the code where a docID is needed, that docID is actually first retrieved from Elasticsearch, rather than making an assumption about what it is.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/update/UpdateDao.java
```diff
@@ -73,23 +74,28 @@ default Document patch(RetrieveLatestDao retrieveLatestDao, PatchRequest request
...
+    String documentID = null;
+    Long timestamp = optionalTimestamp.orElse(System.currentTimeMillis());
+
+    Map<String, Object> originalSource = request.getSource();
+    if (originalSource == null) {
+      // no document source provided, lookup the latest
+      Document toPatch = retrieveLatestDao.getLatest(guid, sensorType);
+      if(toPatch != null && toPatch.getDocument() != null) {
+        originalSource = toPatch.getDocument();
+        documentID = toPatch.getDocumentID().orElse(null);
```
nickwallen: We need to set the doc ID when patching an already indexed document.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/MultiIndexDao.java
```diff
@@ -226,18 +257,30 @@ public void init(AccessConfig config) {
...
+    DocumentContainer container;
+    try {
+      Document document = indexDao.getLatest(guid, sensorType);
+      container = new DocumentContainer(document);
+      LOG.debug("Found latest document; indexDao={}, guid={}, sensorType={}, document={}",
+              ClassUtils.getShortClassName(indexDao.getClass()), guid, sensorType, document);
+
+    } catch (Throwable e) {
+      container = new DocumentContainer(e);
+      LOG.error("Unable to find latest document; indexDao={}, error={}",
```
nickwallen: We had no logging to indicate when a particular IndexDao fails.  I ran into this and it took some effort to trace since we lacked logging.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/MultiIndexDao.java
```diff
@@ -145,18 +164,30 @@ public Document removeCommentFromAlert(CommentAddRemoveRequest request) throws I
...
+    DocumentContainer container;
+    try {
+      Document document = indexDao.removeCommentFromAlert(request, latest);
+      container = new DocumentContainer(document);
+      LOG.debug("Removed comment from alert; indexDao={}, guid={}, sensorType={}, document={}",
+              ClassUtils.getShortClassName(indexDao.getClass()), document.getGuid(), document.getSensorType(), document);
+
+    } catch (Throwable e) {
+      container = new DocumentContainer(e);
+      LOG.error("Unable to remove comment from alert; indexDao={}, error={}",
```
nickwallen: We had no logging to indicate when a particular IndexDao fails.  I ran into this and it took some effort to trace since we lacked logging.

metron-platform/metron-indexing/src/main/java/org/apache/metron/indexing/dao/MultiIndexDao.java
```diff
@@ -117,18 +124,30 @@ public Document addCommentToAlert(CommentAddRemoveRequest request) throws IOExce
...
+    DocumentContainer container;
+    try {
+      Document document = indexDao.addCommentToAlert(request, latest);
+      container = new DocumentContainer(document);
+      LOG.debug("Added comment to alert; indexDao={}, guid={}, sensorType={}, document={}",
+              ClassUtils.getShortClassName(indexDao.getClass()), document.getGuid(), document.getSensorType(), document);
+
+    } catch (Throwable e) {
+      container = new DocumentContainer(e);
+      LOG.error("Unable to add comment to alert; indexDao={}, error={}",
```
nickwallen: We had no logging to indicate when a particular IndexDao fails.  I ran into this and it took some effort to trace since we lacked logging.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchRetrieveLatestDao.java
```diff
@@ -102,54 +92,47 @@ public Document getLatest(String guid, String sensorType) throws IOException {
...
-      idsQuery = QueryBuilders.idsQuery(types);
-    } else {
-      idsQuery = QueryBuilders.idsQuery();
-    }
 
-    for (String guid : guids) {
-      query = idsQuery.addIds(guid);
+    // should match any of the guids
+    // the 'guid' field must be of type 'keyword' or this term query will not match
+    BoolQueryBuilder guidQuery = boolQuery().must(termsQuery(Constants.GUID, guids));
```
nickwallen: We can no longer use an id query, since the doc ID is no longer the same as the GUID.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/dao/ElasticsearchMetaAlertUpdateDao.java
```diff
@@ -172,6 +177,13 @@ public Document update(Document update, Optional<String> index) throws IOExcepti
     }
   }
 
+  private Document toDocument(SearchResult result, Long timestamp) {
+    Document document = Document.fromJSON(result.getSource());
+    document.setTimestamp(timestamp);
+    document.setDocumentID(result.getId());
```
nickwallen: Need to set the document ID when retrieving the document.

justinleet:
This is probably relatively minor, but we seem to map between the underlying "_id" and "id" on the UI.  However, in the search box, I can only do "_id:<blah>" instead of "id:blah", probably because we pass that stuff directly.

Making that work in the general case (given that there could be other modifiers), is probably not really worth it, given that it's an id field.

However, are we okay with the user having to know that to find a particular id in the alerts UI they need to do "_id"?  It's particularly confusing, because the results of a findOne don't include "id" at all (because it's an ES added field, not a Metron document field), and the results of search include "id", e.g.
```
"results": [
    {
      "id": "AWejNX8L98Y5WuImUxxU",
      "source": {
          ...
      }
    }
]
```

I'm a bit confused why the search endpoint returns with "id', when directly hitting ES returns "_id". I don't see anything here that would obviously cause it, but maybe I'm not looking in the right place.
For example
```
curl -X GET "localhost:9200/bro*/_search?pretty=1" -H 'Content-Type: application/json' -d'
{
  "query": {
    "terms": {
      "_id": [ "AWejNX8L98Y5WuImUxxU" ]
    }
  }
}
'
```
Returns
```
...
      "hits" : [
      {
        ...
        "_id" : "AWejNX8L98Y5WuImUxxU",
        ...
      }
```

nickwallen:
> This is probably relatively minor, but we seem to map between the underlying "_id" and "id" on the UI. However, in the search box, I can only do "_id:" instead of "id:blah", probably because we pass that stuff directly.

The UI code has a few of places where it does special treatment for a field called "id" that I had to remap.

One related thing I did ensure we handle is that if you add the "id" field to the table, then click on it, it will correctly populate the search bar so that it filters for that document ID.  That is shown in the screenshot of step 5 in the testing instructions.  From that screenshot, you see that it actually adds _id to the search bar.  And that just confirms what you are seeing.

Logically, it does make sense to me that a user should be able to type "id:XYZ" into the search bar and it filters by document ID, since the column is labelled "id".  Either that or we should get rid of this special thing called "id" and represent it consistently in the UI as "_id".  Either way it really should be one way or the other.

I can look into this and see how much effort it would take.

> However, are we okay with the user having to know that to find a particular id in the alerts UI they need to do "_id"?

What is your opinion on this? Do you think it is worth the effort?

nickwallen:
Also note that this inconsistency exists in master too.    

(1) The table has a column called "id" that when you click on it, it populates the search bar with "guid:fff6c243-dedc-4b46-b005-9a187dbb9d24".  

screen shot 2018-12-12 at 1 58 08 pm

(2) If I manually type in "id:fff6c243-dedc-4b46-b005-9a187dbb9d24", the UI returns no results because it doesn't map ID correctly to the guid.

screen shot 2018-12-12 at 2 00 48 pm

That being said, the behavior in this PR is rather similar to master in this way. Or is that just me trying to make a case to be lazy and not address this issue? :)

mmiklavc:
Was the intent for the id mapping to make our client searching doc index tech agnostic? It seems like a reasonable thing to want our client-facing querying interface to be shielded from changes to our underlying dependencies.

mmiklavc:
@nickwallen are you able to search for "_id" or is that restricted by the UI due to special chars? I think we might want to encapsulate a formal concept of an "id" field that maps to whatever the underlying doc store is - solr/es/other, and it sounds like we already do this to some extent. But I might also expect that users could search on fields specific to the doc store that we don't explicitly handle, e.g. id maps to _id for ES, but maybe you could also search for "_id" explicitly, if desired. Again, this may cause problems with syntax parsing in the UI so I'm just thinking out loud.

nickwallen:
> @mmiklavc: are you able to search for "_id" or is that restricted by the UI due to special chars?

Yes, see the screenshot in step 5 of the test instructions.

justinleet:
If it exists in master, I'm fine leaving it working that way here (maybe document it somewhere?).  I'd like to see a follow-on Jira to clean this up though. It seems like we wanted some special casing, but it's not clear for what and it's obviously not a clean implementation

justinleet:
+1, assuming @mmiklavc is fine with the "id". Ran it up and it worked really well. Thanks for the contribution!

mmiklavc:
+1 by inspection. 

On the subject of the id vs _id vs _uid vs guid - I do think we need to clean this up at some point. _uid, which is _type/doctype + _id, is used for sorting bc you can't sort on _id in ES. In 6.x _uid is deprecated https://www.elastic.co/guide/en/elasticsearch/reference/6.0/mapping-uid-field.html. We've touched on this before, but we should consider a translation layer that manages variations between our optional downstream indexing engines. I'm not a big fan of this implementation bleed into the UI, but I recognize that this is bigger than just the id field.

PR #1270, opened 2018-11-17
METRON-1880 Use Caffeine for Profiler Caching

nickwallen:
Caffeine is a more performant cache when cache sizes are larger.  We've already seen a significant improvement elsewhere in Metron based on the work done in #947. 

The Profiler should use Caffeine for caching to ensure that the caches are not a performance bottleneck.

## Testing

1. Spin-up Full Dev.
1. Create a profile using the Stellar REPL.
1. Deploy the profile for the Storm-based Profiler.
1. Ensure valid profile data is persisted in HBase.

## Pull Request Checklist


justinleet:
Not being particularly familiar with the caching discussions from last time, is there anything we should be testing that we don't during this switch. I seem to recall this being pretty much drop-in, but wanted to double check.

nickwallen:
It is 'drop-in' in my opinion.  I just suggested sanity checks on the Profiler in the 3 different execution environments; Storm, REPL, and Spark.

justinleet:
Sounds good.

+1 by inspection.

PR #1274, opened 2018-11-19
METRON-1887: Add logging to the ClasspathFunctionResolver

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1887

From the Jira description:
We had a user reporting non-deterministic NPE's here - https://github.com/apache/metron/blob/master/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/resolver/ClasspathFunctionResolver.java#L250. The purpose of this ticket is to more gracefully handle the NPE and add some additional debugging information to better track down the source of this issue should it come up again.

### For all changes:

### For code changes:
  ```
  ```

- n/a Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via `site-book/target/site/index.html`:

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/VFSClassloaderUtil.java
+++ b/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/VFSClassloaderUtil.java
@@ -112,14 +112,18 @@ public class VFSClassloaderUtil {
    * @throws FileSystemException
    */
   public static Optional<ClassLoader> configureClassloader(String paths) throws FileSystemException {
+    LOG.debug("Configuring class loader with paths = {}", paths);
     if(paths.trim().isEmpty()) {
+      LOG.debug("No paths provided. Not returning a ClassLoader.");
       return Optional.empty();
     }
     FileSystemManager vfs = generateVfs();
     FileObject[] objects = resolve(vfs, paths);
     if(objects == null || objects.length == 0) {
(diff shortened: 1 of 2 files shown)


metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/VFSClassloaderUtil.java
```diff
@@ -112,14 +112,18 @@ public static FileSystemManager generateVfs() throws FileSystemException {
    * @throws FileSystemException
    */
   public static Optional<ClassLoader> configureClassloader(String paths) throws FileSystemException {
+    LOG.debug("Configuring class loader with paths = {}", paths);
     if(paths.trim().isEmpty()) {
```
nickwallen: Should we add a null check here, just in case?  Or use StringUtils.isBlank?

nickwallen:
One nit on a null check, but that is a pre-existing condition and does not seem to be directly related with the problem at hand.  Dealer's choice on that one.

Either way, +1 looks good.

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/common/utils/VFSClassloaderUtil.java
mmiklavc: We haven't encountered any instances with path being null that I'm aware of - I'd like to start with this and add more checks as needed.

PR #1276, opened 2018-11-20
METRON-1888 Default Topology Settings in MPack Cause Profiler to Stall

nickwallen:
The default max spout pending value set by the Mpack for the Profiler topology is 300. This value is too low and causes the topology to stall and stop consuming messages. 

This is an annoying issue to debug because I am unable to find anything logged by Storm indicating that the max spout pending limit is breached.  When this limit is exceeded it just appears that the KafkaSpout stops consuming messages from Kafka and everything else would indicate the topology is healthy.

The Profiler topology is the only Metron topology that uses Storm's windowing support.  Storm will queue up messages for the window (by default 30 seconds) and process each of the tuples in the window as one unit.  This causes Storm to need to "hang on" to tuples longer than the other topologies and this requires a greater max spout pending value.

This PR restores the original default value prior to #1221 that changed this value.  There really is no need to define this value by default.  I was a reviewer on #1221 and should have caught this. Doh!

## Testing

1. Spin-up Full Dev.
2. Create a profile using the Streaming Profiler in Storm.
3. Allow the topology to run for at least 5 to 30 minutes.
4. Ensure that the Profiler topology continues to consume messages at the end of this period. 

## Pull Request Checklist

### DIFF
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-profiler-env.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-profiler-env.xml
@@ -165,7 +165,7 @@
     <name>profiler_topology_max_spout_pending</name>
     <description>Profiler Topology Spout Max Pending Tuples</description>
     <display-name>Spout Max Pending Tuples</display-name>
-    <value>300</value>
+    <value/>
     <value-attributes>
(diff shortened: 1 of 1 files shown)


justinleet:
+1 by inspection.  I'm okay with leaving this undefined by default. 

Should we have a jira for the ability to detect (and maybe in a perfect world, correct/mitigate) the max spout pending limit issue you mention above?

nickwallen:
Thanks @justinleet .  I added https://issues.apache.org/jira/browse/METRON-1897 for this.

PR #1289, opened 2018-12-05
METRON-1810 Storm Profiler Intermittent Test Failure

nickwallen:
This PR hopefully resolves some of the more frequent, intermittent `ProfilerIntegrationTest` failures.  

### Testing

Run the integration tests and they should not fail.

I have repeatedly run the integration tests on my laptop and have yet to see a failure.  I am also repeatedly triggering Travis CI builds on this branch to see if any failures occur there.  I can't prove the problem is solved, but I am hoping this helps.

### Changes

* This change uses Caffeine's CacheWriter in place of the RemovalListener for profile expiration.  This is explained more in the next section.

* Removed the ability to define the cache maintenance executor for the `DefaultMessageDistributor`.  This was only needed for testing RemovalListeners, which no longer exist.

* I re-jiggered some of the integration tests to provide more information when they do fail. I removed the use of `waitOrTimeout` and replaced it with `assertEventually`.  I am also using a different style of assertion; Hamcrest-y.  These changes should provide additional information and also be more consistent with the rest of the code base.

* After removing the dependency that provided `waitOrTimeout`, I had to rework some of the project dependencies because multiple versions of`httpclient` lib was being pulled in.  This caused the new REST function in `stellar-common` to blow up when a Stellar execution environment is loaded during the integration test.

* Added additional debug logging for the caches including an estimate of the cache size.

#### RemovalListener to CacheWriter

Profiles are designed to expire and flush after a fixed period of time; the TTL.  The integration tests rely on some of the profile values being flushed by this TTL mechanism.  

The TTL mechanism is driven by cache expiration. There is a cache of "active" profiles.  This cache is configured to have values timeout after they have not been accessed for a fixed period of time.  Once they timeout from the active cache, they are placed on an "expired" cache.  Messages are not applied to expired profiles, but these expired profiles hang around for a fixed period of time to allow them to be flushed.

Previously, a Caffeine [RemovalListener](https://github.com/ben-manes/caffeine/wiki/Removal#removal-listeners) was used so that when a profile expires from the active cache, it is placed into the expired cache. When the tests fail, it seems that the `RemovalListener` is not notified in a timely fashion, so the profile doesn't make it to the expired cache and so never flushes to be read by the integration test.

 In Caffeine, these listeners are notified asynchronously, on a separate thread (via ForkJoinPool.commonPool()), not inline with cache reads or writes.  For running tests that depend on RemovalListener's it is recommended to set the cache maintenance executor to something like `MoreExecutors.sameThreadExecutor()` so that cache maintenance is executed on the main execution thread when `cleanUp` is called.  This was done for the unit tests, but was not done for the integration tests. 

The Caffeine Wiki mentions that when notification should be performed synchronously, which logically works for this use case, to use a [CacheWriter](https://github.com/ben-manes/caffeine/wiki/Writer) instead.

> Removal listener operations are executed asynchronously using an Executor. The default executor is ForkJoinPool.commonPool() and can be overridden via Caffeine.executor(Executor). When the operation must be performed synchronously with the removal, use [CacheWriter](https://github.com/ben-manes/caffeine/wiki/Writer) instead.

This does not negatively impact production performance because the `ActiveCacheWriter` only does work on a delete which occurs only rarely when a profile stops receiving messages, not on a write.  In addition, these caches are very read-heavy as in most cases the cache is only written to when a new profile is defined or on topology start-up.

## Pull Request Checklist
  ```
  ```

### DIFF
--- a/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageDistributor.java
+++ b/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageDistributor.java
@@ -239,13 +239,11 @@ public class DefaultMessageDistributor implements MessageDistributor, Serializab
   private void cacheMaintenance() {
     activeCache.cleanUp();
     LOG.debug("Active cache maintenance triggered: cacheStats={}, size={}",
-            activeCache.stats().toString());
-            //activeCache.estimatedSize());
+            activeCache.stats().toString(), activeCache.estimatedSize());
 
     expiredCache.cleanUp();
     LOG.debug("Expired cache maintenance triggered: cacheStats={}, size={}",
-            expiredCache.stats().toString());
-            //expiredCache.estimatedSize());
+            expiredCache.stats().toString(), expiredCache.estimatedSize());
   }
 
   /**
--- a/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageDistributor.java
+++ b/metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageDistributor.java
@@ -21,11 +21,11 @@
 package org.apache.metron.profiler;
 
 import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.CacheWriter;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.github.benmanes.caffeine.cache.RemovalListener;
 import com.github.benmanes.caffeine.cache.Ticker;
-import com.google.common.util.concurrent.MoreExecutors;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.metron.common.configuration.profiler.ProfileConfig;
 import org.apache.metron.stellar.dsl.Context;
@@ -41,7 +41,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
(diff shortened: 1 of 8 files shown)


metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/DefaultMessageDistributor.java
```diff
@@ -315,41 +308,51 @@ public DefaultMessageDistributor withPeriodDuration(int duration, TimeUnit units
...
     @Override
-    public void onRemoval(@Nullable Integer key, @Nullable ProfileBuilder expired, @Nonnull RemovalCause cause) {
-      LOG.warn("Profile expired from active cache; profile={}, entity={}",
-              expired.getDefinition().getProfile(),
-              expired.getEntity());
+    public void delete(@Nonnull Integer key, @Nullable ProfileBuilder value, @Nonnull RemovalCause cause) {
+      if(cause.wasEvicted()) {
+        // add the profile to the expired cache
+        expiredCache.put(key, value);
+        LOG.info("Profile expired from active cache due to inactivity; profile={}, entity={}, cause={}",
```
justinleet: Will this LOG.info cause a lot of churn in the logs? In a production scenario, there could be a lot of entities per profile, right? And these could be expiring pretty often.  It seems like you could easily hit scenarios where thousands of entities are expiring and then clogging the logger up.  This seems like more a debugging log than anything else.

nickwallen: Yep, I will change it to debug.

metron-analytics/metron-profiler-common/src/main/java/org/apache/metron/profiler/ProfilePeriod.java
```diff
@@ -151,6 +152,8 @@ public String toString() {
     return "ProfilePeriod{" +
             "period=" + period +
             ", durationMillis=" + durationMillis +
+            ", startTime=" + Instant.ofEpochMilli(getStartTimeMillis()).toString() +
+            ", endTime=" + Instant.ofEpochMilli(getEndTimeMillis()).toString() +
```
nickwallen: I will remove this as it is not necessary.

nickwallen:
I have run it up in Full Dev successfully.  

I ran all the tests in `ProfilerIntegrationTest` in a continuous loop on my desktop for maybe 3 or 4 hours yesterday.  Doing the same on master fails within 20 minutes or so.

I have manually triggered it in Travis 10 times.

justinleet:
+1, glad to see it working well.

PR #1295, opened 2018-12-10
METRON-1931 Moved from git-wip-us to gitbox

rlenferink:
As of the move to GitBox, the URLs needed to be changed in the different scripts.


justinleet:
+1 by inspection.  Thanks for the help with the relocation!

PR #1307, opened 2018-12-19
METRON-1944: Unable to Delete a Comment in Alerts UI

ruffle1986:
## Contributor Comments

Link to the original ASF Jira ticket: https://issues.apache.org/jira/browse/METRON-1944

**The problem**

The problem and its solution are quite trivial in this case. The code which is responsible for the deletion of the comment tried to address the deleted comment in an array after it was removed from it.

`splice` returns with an array of the removed elements so we can keep working with that to persist the operation for the given comment on the server.

**Description from the ASF Jira ticket:**

I am unable to delete a comment in the Alerts UI. While it appears that the comment is deleted, it is only deleted locally, but not server side. An exception is thrown in the browser console.

```
ERROR TypeError: Cannot read property 'alertComment' of undefined at SafeSubscriber._next (alert-details.component.ts:258) at SafeSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.SafeSubscriber.__tryOrUnsub (Subscriber.js:207) at SafeSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.SafeSubscriber.next (Subscriber.js:145) at Subscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber._next (Subscriber.js:80) at Subscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (Subscriber.js:55) at Subject.push../node_modules/rxjs/_esm5/internal/Subject.js.Subject.next (Subject.js:47) at DialogService.push../src/app/service/dialog.service.ts.DialogService.approve (dialog.service.ts:49) at MetronDialogComponent.push../src/app/shared/metron-dialog/metron-dialog.component.ts.MetronDialogComponent.confirm (metron-dialog.component.ts:54) at Object.handleEvent (login-guard.ts:40) at handleEvent (core.js:10258)
```

**Steps To Replicate**

1. Find an Alert
2. Comment on the alert.
3. Delete the comment.  
4. Close the side panel.
5. Search for the alert again.
6. Click on the alert and the comment remains.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.ts
@@ -42,7 +42,7 @@ export enum Tabs {
   DETAILS, COMMENTS
 }
 
-class AlertCommentWrapper {
+export class AlertCommentWrapper {
   alertComment: AlertComment;
   displayTime: string;
 
--- a/metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.ts
@@ -260,9 +260,7 @@ export class AlertDetailsComponent implements OnInit {
         commentRequest.timestamp = deletedCommentWrapper.alertComment.timestamp;
         commentRequest.sensorType = this.alertSourceType;
         this.updateService.removeComment(commentRequest).subscribe(
-            () => {
-              this.alertCommentsWrapper.map(alertsWrapper => alertsWrapper.alertComment)
-            },
+            null,
             () => {
               // add the deleted comment back
               this.alertCommentsWrapper.unshift(deletedCommentWrapper);
--- a/metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.ts
@@ -255,9 +255,9 @@ export class AlertDetailsComponent implements OnInit {
         let deletedCommentWrapper = this.alertCommentsWrapper.splice(index, 1)[0];
         let commentRequest = new CommentAddRemoveRequest();
         commentRequest.guid = this.alertSource.guid;
-        commentRequest.comment = this.alertCommentsWrapper[index].alertComment.comment;
-        commentRequest.username = this.alertCommentsWrapper[index].alertComment.username;
-        commentRequest.timestamp = this.alertCommentsWrapper[index].alertComment.timestamp;
+        commentRequest.comment = deletedCommentWrapper.alertComment.comment;
+        commentRequest.username = deletedCommentWrapper.alertComment.username;
+        commentRequest.timestamp = deletedCommentWrapper.alertComment.timestamp;
         commentRequest.sensorType = this.alertSourceType;
         this.updateService.removeComment(commentRequest).subscribe(
             () => {
--- a/metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.html
+++ b/metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.html
@@ -17,12 +17,22 @@
             <div class="nav-container" *ngIf="!isMetaAlert">
                 <ul class="nav flex-column">
                     <li class="nav-item">
-                        <a class="nav-link" [ngClass]="{'active': activeTab === tabs.DETAILS}" (click)="activeTab=tabs.DETAILS">
+                        <a
+                            class="nav-link"
+                            [ngClass]="{'active': activeTab === tabs.DETAILS}"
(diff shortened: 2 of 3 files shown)


sardell:
@ruffle1986 It looks like Travis is failing, but it's failing during the integration tests that don't have to do with the UI (where your work happened). Specifically:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.18:test (integration-tests) on project metron-pcap-backend: Execution integration-tests of goal org.apache.maven.plugins:maven-surefire-plugin:2.18:test failed: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
```

From the looks of it, the VM just randomly crashed? Could you try closing/reopening this PR to re-run Travis?

justinleet:
Is there any way to add a unit test for this? I know a lot of the testing code and infra has been a bit in flux, so I'm not quite sure how much of an ask that is.

ruffle1986:
Good point. Shall I open a JIRA ticket on this and add the tests to another PR? Or just simply add them here? @tiborm @sardell

sardell:
@ruffle1986 If the amount of mocking and setup you need to do is minimal, I think writing a unit test specific to the comment functionality of the alert details component should be done in this PR. However, I do see there are no unit tests currently setup for the alert details component. If the scope of setting up these tests is very time consuming, I think you should create a separate JIRA to cover this component with unit tests. That way, we aren't delaying the bug fix getting into master.

sardell:
+1. Glad to see the unit test for the comment deletion functionality. Hopefully in the near future, we can extend the unit test coverage for this component, but for now I think this is good. Thanks for the contribution, @ruffle1986!

ruffle1986:
@sardell Thanks for the report! Nice catch! 

Yes, it was there before probably. In order the enable comment deletion, I had to slightly modify the existing code. However the issue you're struggling with is critical, I respectfully think it's out of the scope of this PR. 

When it comes to removing comments (and adding as well!!), the app deals with HTTP requests which are not organized properly. When you successfully removed or added new comments, the server responses with the fresh new lest of the comments attached to the given alert. But if you're fast, adding/removing a bunch of comments then immediately closing and opening the comments pane again could cause issues. The problem is that the API endpoints for the deletion/addition returns with a list of comments and when you open the pane, it also returns the list of comments and the two are not necessarily the same at a certain point of time. They're not synced yet.

Long story short, the current design and the implementation is not prepared for this kind of user quickness, the solution for this is not trivial and needs to have a further discussion for sure how to deal with it.

If you don't think otherwise, I would open a Jira ticket about the aforementioned syncing issue and ship the patch in a different PR.

sardell:
@ruffle1986 I see. After reading your explanation, I agree that resolving this issue is going to take more thought, and will likely result in a pretty big change in implementation. I also agree that this is an issue bigger in scope than the original task of resolving a front-end specific bug which prevented a user from deleting a comment.

+1 pending @ruffle1986 creating and sharing a JIRA ticket to track the syncing issue described above.

ruffle1986:
I created the ASF Jira ticket which describes the aforementioned problem. I also made a video and added the scenario to reproduce the problem.

Here you are: https://issues.apache.org/jira/browse/METRON-2001

metron-interface/metron-alerts/src/app/alerts/alert-details/alert-details.component.ts
```diff
@@ -260,9 +260,7 @@ export class AlertDetailsComponent implements OnInit {
         commentRequest.timestamp = deletedCommentWrapper.alertComment.timestamp;
         commentRequest.sensorType = this.alertSourceType;
         this.updateService.removeComment(commentRequest).subscribe(
-            () => {
-              this.alertCommentsWrapper.map(alertsWrapper => alertsWrapper.alertComment)
-            },
+            null,
```
ruffle1986: I removed the next callback here because it makes no sense to map over an array and does nothing with the mapped array. 

[map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#Return_value) does not mutate the array on which you call it. It creates a brand new array with the mapped elements. The code above does nothing with the return value of map so it can be removed.

sardell: Yeah, that makes sense to remove since we do not know the original intention of the line of code.

@justinleet It looks like this line of code was added here. Is it okay we remove it? Asking because I'm not sure what the original intention of the code was. Since there isn't a PR attached I'm not even sure you were the author of the code: 
https://github.com/apache/metron/commit/e0fc47582e65e9daf1a0c85dc1b1657166ff14f5#diff-bce5ae5c6be579250a35b6ff4f6da594R271

sardell:
+1. Thanks for the contribution, @ruffle1986!

PR #1309, opened 2018-12-20
METRON-1950: Site-book generation broken in master

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-1950

This is currently breaking the ability the run `mvn clean site` in the site-book module. The reason for  this issue is that we recently had a PR make it in that likely looked good via Github markdown, but didn't end up getting tested via Doxia using our site-book build. The errors end up being really esoteric, e.g. it will complain about opening/closing tags. In reality, this stemmed from a few nested problems. First, indentation around some code samples (triple backticks) within bullet lists was incorrect. This cause the open/close tag error. Even if you fix that, the html page will render horribly. You can try all the nested indexing for 2nd level bullet list code examples you want, but it will never work until you correct the full chain. The Doxia plugin is extremely opinionated about having 4 spaces for nesting indented bullet list along with any other comments or code examples that should be indented at the same level. Items that should end up at the same indention level as a bullet point actually need another 4 spaces of indentation. You can see examples in the README's in this PR. A sample follows:

```
* Test1
    * Test1a
        
        I'm a comment within Test1a

        ```
        { 
          "foo" : "bar",
          "this" : "is a code example with json"
        }
        ```

    I'm at the same level as Test1a, a non-bullet item under Test1
* Test2
    * Test2a
```

This would render as follows:
* Test1
    * Test1a

        I'm a comment within Test1a

        ```
        { 
          "foo" : "bar",
          "this" : "is a code example with json"
        }
        ```

    I'm at the same level as Test1a, a non-bullet item under Test1
* Test2
    * Test2a

I generally add 1 extra blank line above and below the indented code snippets for better spacing, but I believe that's optional and shouldn't affect the overall document formatting, unlike the indentation.

I also fixed a couple broken images in the parsers README and use-cases README for parser-chaining. The upshot to this part of the fix is that you need to make sure you update the rewriting rules to include the original image path, and then rewrite and project-relative links accordingly in each markdown page using the relative path to "images", which is where the site-book copies them all.

**Testing**

Run the site-book documentation validation steps in our PR checklist from below.
* Check formatting and images on the main parser page - metron-platform/metron-parsing/index.html
* Check broken images on metron-platform/metron-parsing/metron-parsers-common/ParserChaining.html
* Check I didn't break any images in use-cases/parser_chaining/index.html

### For all changes:

### For code changes:
  ```
  ```

- n/a Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

  ```
  cd site-book
  mvn site
  ```


anandsubbu:
+1 

Ran `mvn clean site` without the change and did ran into the build failure. 
Applied the fix and the sitebook was generated fine. Validated the metron-parsing README in the sitebook and the contents looked fine. Thanks for the fix, @mmiklavc !

PR #1312, opened 2018-12-21
METRON-1953 Stellar REPL Fails on Start

nickwallen:
The Stellar REPL will fail to launch.  See full stack trace in [METRON-1953](https://issues.apache.org/jira/browse/METRON-1953)

There was a small typo in the Storm Parsers jar.  Instead of "parsers" it needs to be "parsing".

## Testing

1. Launch Full Dev.
    ```
    vagrant up
    vagrant ssh
    ```

1. Launch the REPL.  Ensure that it successfully is started.
    ```
    sudo su -
    source /etc/default/metron
    cd $METRON_HOME
    bin/stellar -z $ZOOKEEPER
    ```

1. Run some commands to validate the REPL.

## Pull Request Checklist

### DIFF
--- a/metron-platform/metron-common/src/main/scripts/stellar
+++ b/metron-platform/metron-common/src/main/scripts/stellar
@@ -38,7 +38,7 @@ set -u
 
 export METRON_PARSERS_PLATFORM="${METRON_PARSERS_PLATFORM:-storm}"
 export HBASE_CONFIGS=$(hbase classpath)
-export STELLAR_LIB=$(find $METRON_HOME/lib/ -name metron-parsers-*-uber.jar | grep -v "common")
+export STELLAR_LIB=$(find $METRON_HOME/lib/ -name metron-parsing-${METRON_PARSERS_PLATFORM}*.jar)
 export MANAGEMENT_LIB=$(find $METRON_HOME/lib/ -name metron-management*.jar)
 export PROFILER_LIB=$(find $METRON_HOME/lib/ -name metron-profiler-repl*.jar)
 java $METRON_JVMFLAGS -cp "${CONTRIB:-$METRON_HOME/contrib/*}:$STELLAR_LIB:$MANAGEMENT_LIB:$PROFILER_LIB:$HBASE_CONFIGS" org.apache.metron.stellar.common.shell.cli.StellarShell "$@"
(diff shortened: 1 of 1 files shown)


metron-platform/metron-common/src/main/scripts/stellar
```diff
@@ -38,7 +38,7 @@ set -u
 
 export METRON_PARSERS_PLATFORM="${METRON_PARSERS_PLATFORM:-storm}"
 export HBASE_CONFIGS=$(hbase classpath)
-export STELLAR_LIB=$(find $METRON_HOME/lib/ -name metron-parsers-${METRON_PARSERS_PLATFORM}*.jar)
+export STELLAR_LIB=$(find $METRON_HOME/lib/ -name metron-parsers-*-uber.jar | grep -v "common")
```
mmiklavc: Could this just be?:
`export STELLAR_LIB=$(find $METRON_HOME/lib/ -name metron-parsers-common-*-uber.jar`

e.g. Running that find command on our parsers dir returns the common uber jar for parsers:
```
find metron-platform/metron-parsing -name metron-parsers-common-\*-uber.jar
metron-platform/metron-parsing/metron-parsers-common/target/metron-parsers-common-0.7.0-uber.jar
```

nickwallen: It is my understanding that we don't want common.  We want metron-parsers because that includes metron-parsers-common.

nickwallen: The `grep -v` is inverting the search, so that we find anything other than common.

nickwallen: And in a follow-on PR, we can probably exclude metron-parsers-common from being deployed.

nickwallen: Updated the PR description to better clarify these points.

ottobackwards:
testing

nickwallen:
Actually, looking at this again, the intention is to depend on the Parsers Storm Uber jar.  I think I should restore that behavior, not depend on metron-parsers.

nickwallen:
Ok, I fixed it.  This follows the intention of what is supposed to happen.  The only problem was "parsers" needed to be "parsing".

mmiklavc:
+1 by inspection

ottobackwards:
I think I verified the first fix and not the second.
The second doesn't work for me now.
https://github.com/apache/metron/tree/master/metron-stellar/stellar-common/src/main/scripts/deployed
Is the file that I think needs to be changed, not metron-common

PR #1328, opened 2019-01-31
METRON-1986 Batch Profiler Fails to Resolve Stats Stellar Functions

nickwallen:
The Batch Profiler does not seem to see the STATS_* functions. When I attempt to run a profile using any of those functions, it fails. It is able to 'see' a limited subset of functions, but not all those that I would expect.

## Root Cause

The ClassIndex library is used to create an index of all Stellar functions at compile-time. This index is packaged in the JAR as `META-INF/annotations/org.apache.metron.stellar.dsl.Stellar`.

When creating a shaded jar using the Maven Shade Plugin, the index files created by ClassIndex are overwritten by default. ClassIndex provides a special transformer for Maven which merges the index files instead of overwriting them.

Without using this transformer, the index files are overwritten and Stellar function resolution will not work. This is why Stellar function resolution is not working.

https://github.com/atteo/classindex#making-shaded-jar

## Testing

1. Spin-up this profile in the Batch Profiler.  It should write successfully to HBase.

```
{
  "profiles": [
    {
      "profile": "profile-using-stats",
      "foreach": "ip_src_addr",
      "update": { "s": "STATS_ADD(s, length)" },
      "result": "STATS_SUM(s)"
    }
  ],
  "timestampField":"timestamp"
}
```

## Pull Request Checklist

### DIFF
--- a/metron-analytics/metron-profiler-spark/pom.xml
+++ b/metron-analytics/metron-profiler-spark/pom.xml
@@ -52,13 +52,17 @@
             <groupId>org.apache.metron</groupId>
             <artifactId>metron-profiler-client</artifactId>
             <version>${project.parent.version}</version>
-            <scope>test</scope>
             <exclusions>
                 <exclusion>
                     <groupId>org.antlr</groupId>
                     <artifactId>antlr4-runtime</artifactId>
                 </exclusion>
             </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.metron</groupId>
+            <artifactId>stellar-common</artifactId>
+            <version>${project.parent.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


nickwallen:
FYI - I added an integration test for this just to try and help avoid future issues like this.  Unfortunately though, the integration test will *not* catch this issue.  Our integration tests do not run from the shaded jars that we build.  That is a gap in our testing capabilities currently.

mmiklavc:
As a side note, I found that this plugin has issues if not placed last in the plugins list while working through the ES REST client migration.

https://github.com/apache/metron/pull/1242/files#diff-1f250467ce2f882bfb71cc0928813df8R301

mmiklavc:
This looks good, thanks for the fix @nickwallen. +1 by inspection.

PR #1330, opened 2019-02-05
METRON-1968: Messages are lost when a parser produces multiple messages and batch size is greater than 1

merrimanr:
## Contributor Comments
This PR represents a fairly significant shift in the Writer class architecture.  Currently these classes do not support tuples that result in multiple messages, mainly due to a limitation in the `BulkMessageWriter` interface.  The `write` method accepts separate lists of tuples and messages so there is no way to know which tuples are associated with which message.  It has worked so far with parsers that only emit a single message from a tuple because a 1 to 1 relationship is assumed in these classes.

I experimented with several different approaches to fixing this and tried to follow a strategy that would avoid having to significantly rewrite multiple classes (this was unavoidable in certain places).  I change the `BulkMessageWriter.write` interface from:
```
BulkWriterResponse write(String sensorType, WriterConfiguration configurations, Iterable<Tuple> tuples, List<MESSAGE_T> messages) throws Exception;
```
to:
```
BulkWriterResponse write(String sensorType, WriterConfiguration configurations, Map<String, MESSAGE_T> messages) throws Exception;
```
The messages being passed in are now represented as a map where the keys are message ids and the values are the messages.  This made changes to the `BulkMessageWriter` implementations easier because they continue to assume a 1 to 1 relationship between message ids (formerly tuples) and messages.  Now the writers report which message ids succeeded/failed instead of tuples.  The `BulkWriterComponent` continues track batch sizes and timeouts but no longer manages tuples or error handling.  Instead it calls `BulkWriterResponseHandler.handleFlush` when messages are flushed.  This interface is injected into `BulkWriterComponent` and allows us to properly commit messages according to the requirements of the platform the classes are running on.  For now a Storm implementation of `BulkWriterResponseHandler` is setup and passed into the `BulkWriterComponent`.   Managing tuple to message relationships as well as tuple acking and error handling are now done in a single class.  The bolts now have to setup a `StormBulkWriterResponseHandler` object and add tuples/messages to it.

### Changes Included

- The `BulkMessageWriter` is updated and all implementations have been adjusted to conform to the new interfaces.  Changes to the implementations are small and straightforward.
- Tuple acking and error handling is moved out of `BulkWriterComponent` and into `StormBulkWriterResponseHandler`. 
- Bolts have been updated with the new `StormBulkWriterResponseHandler` pattern
- I noticed `HBaseWriter` and `PcapWriter are no longer being used.  Rather than go to the trouble of updating their tests I removed them.
- Removed `WriterHandler.handleAck` since acks are no longer handled here.
- Added logging to make troubleshooting writer issues easier

There were also several significant changes that needed to be made to the tests:

- The writer tests are updated to match the new `BulkMessageWriter.write` interface.  In most cases these changes were simple however I noticed there were no unit tests for `KafkaWriter.write` so I added them.
- The parser integration tests now verify the all tuples were acked.
- There is now a parser integration test that simulates the use case described in this Jira.  The `jsonMapQuery` parser integration test now produces multiple messages from a single tuple and sets the batch size to 5.
- A test was added for `StormBulkWriterResponseHandler` that also simulates the use case described in this Jira.  Error handling tests that were originally in `BulkWriterComponentTest` were migrated here.

### Testing

This has been tested in full dev both for regression and for the use case described in this Jira.  There are 3 test cases:

- bro sensor with a batch size set to 5
- snort sensor with a batch size set to 1
- jsonMapQuery sensor that produces multiple messages from a tuple with a batch size set to both less than and greater than the number of messages 

#### Setup
1. Spin up full dev and verify data is landing in both Elasticsearch and HDFS.   
2. Stop all sensors with `service sensor-stubs stop`.
3. Create the jsonMapQuery sensor topic:
```
/usr/hdp/current/kafka-broker/bin/kafka-topics.sh --zookeeper node1:2181 --create --topic jsonMapQuery --partitions 1 --replication-factor 1
```
4. Copy the jsonMapQuery sample data to full dev:
```
cd /metron-deployment/development/centos6
vagrant scp ../../../metron-platform/metron-integration-test/src/main/sample/data/jsonMapQuery/raw/jsonMapExampleOutput /tmp
```
5. Stop Metron Parsers in Ambari.
6. Navigate to Ambari > Metron > Configs > Parsers and change "Metron Parsers" from `bro,snort,yaf` to `bro,snort,jsonMapQuery`.
7. Start Metron Parsers in Ambari.
8. Clear the Elasticsearch indices:
```
curl -XDELETE http://node1:9200/bro_index*
curl -XDELETE http://node1:9200/snort_index*
```
9. Stop Metron Indexing in Ambari. 
10. Clear the HDFS files:
```
su hdfs -c "hdfs dfs -rm /apps/metron/indexing/indexed/bro/*"
su hdfs -c "hdfs dfs -rm /apps/metron/indexing/indexed/snort/*"
```
11. Start Metron Indexing in Ambari.

12. Verify there is no bro or snort data in Elasticsearch or HDFS:
```
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep bro
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep snort
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/bro/*.json | wc -l
cat: `/apps/metron/indexing/indexed/bro/*.json': No such file or directory
0
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/snort/*.json | wc -l
cat: `/apps/metron/indexing/indexed/snort/*.json': No such file or directory
0
```
13. Each topology (parsing, enrichment, indexing) can be configured with a batch size and timeout.  The batch timeout defaults to 1/2 the tuple timeout so tests described next should be done reasonably quick (before the timeout happens).  To make the test results easier to understand we also need to adjust the batch sizes to 1 for topologies we are not directly testing.  Set the bro parser batch size to 1:
```
  "parserClassName": "org.apache.metron.parsers.bro.BasicBroParser",
  "sensorTopic": "bro",
  "parserConfig": {
    "batchSize": 1
  }
}' 'http://user:password@node1:8082/api/v1/sensor/parser/config/bro'
```
14. Set the snort parser batch size to 1:
```
  "parserClassName": "org.apache.metron.parsers.snort.BasicSnortParser",
  "sensorTopic": "snort",
  "parserConfig": {
    "batchSize": 1
  }
}' 'http://user:password@node1:8082/api/v1/sensor/parser/config/snort'
```
15. Set the enrichment batch size to 1:
```
  "es.clustername": "metron",
  "es.ip": "node1:9200",
  "update.hbase.cf": "t",
  "es.client.settings": {},
  "profiler.client.period.duration": "15",
  "profiler.client.period.duration.units": "MINUTES",
  "user.settings.hbase.table": "user_settings",
  "user.settings.hbase.cf": "cf",
  "bootstrap.servers": "node1:6667",
  "source.type.field": "source:type",
  "threat.triage.score.field": "threat:triage:score",
  "enrichment.writer.batchSize": "1",
  "profiler.writer.batchSize": "15",
  "profiler.writer.batchTimeout": "0",
  "geo.hdfs.file": "/apps/metron/geo/default/GeoLite2-City.tar.gz",
  "asn.hdfs.file": "/apps/metron/asn/default/GeoLite2-ASN.tar.gz"
}' 'http://user:password@node1:8082/api/v1/global/config'
```
16. Set the jsonMapQuery indexing batch size to 1 (since we will test this sensor in parsers):
```
  "hdfs": {
    "index": "jsonmapquery",
    "batchSize": 1,
    "enabled": true
  },
  "elasticsearch": {
    "index": "jsonmapquery",
    "batchSize": 1,
    "enabled": true
  },
  "solr": {
    "index": "jsonmapquery",
    "batchSize": 1,
    "enabled": false
  }
}' 'http://user:password@node1:8082/api/v1/sensor/indexing/config/jsonMapQuery'
```
It can be helpful to keep a Kafka console consumer running for each topic:
```
/usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh --zookeeper node1:2181 --topic bro
/usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh --zookeeper node1:2181 --topic snort
/usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh --zookeeper node1:2181 --topic jsonMapQuery
/usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh --zookeeper node1:2181 --topic enrichments
/usr/hdp/current/kafka-broker/bin/kafka-console-consumer.sh --zookeeper node1:2181 --topic indexing
```
#### bro
1. Bro has a batch size of 5 in indexing by default.  Write 4 bro messages:
```
shuf -n 4 /opt/sensor-stubs/data/bro.out | sed -e "s/\"ts\"\:[0-9]\+\./\"ts\"\:`date +%s`\./g" | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic bro
```
There should be no data in Elasticsearch or HDFS:
```
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep bro
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/bro/*.json | wc -l
cat: `/apps/metron/indexing/indexed/bro/*.json': No such file or directory
0
```
2. Write an additional bro message:
```
shuf -n 1 /opt/sensor-stubs/data/bro.out | sed -e "s/\"ts\"\:[0-9]\+\./\"ts\"\:`date +%s`\./g" | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic bro
```
There should now be 5 messages in Elasticsearch and HDFS:
```
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep bro
bro_index_2019.02.07.22    5
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/bro/*.json | wc -l
5
```

#### snort
1. Snort has a batch size of 1 in indexing by default:  Write a snort message:
```
shuf -n 1 /opt/sensor-stubs/data/snort.out | sed -e "s/[^,]\+ ,/`date +'%m\/%d\/%y-%H:%M:%S'`.000000 ,/g" | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic snort
```
There should now be 1 message in Elasticsearch and HDFS:
```
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep snort
snort_index_2019.02.07.22    1
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/snort/*.json | wc -l
1
```
#### jsonMapQuery
1. The jsonMapQuery sensor produces 10 messages for every tuple.  Set the batch size to 5 in the parser topology:
```
  "parserClassName": "org.apache.metron.parsers.json.JSONMapParser",
  "sensorTopic": "jsonMapQuery",
  "parserConfig": {
    "jsonpQuery": "$.foo",
    "batchSize": 5
  }
}' 'http://user:password@node1:8082/api/v1/sensor/parser/config/jsonMapQuery'
```
2. Writer a jsonMapQuery message:
```
cat /tmp/jsonMapExampleOutput | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic jsonMapQuery
```
This should cause the issue described in the Jira.  Before there was only a single message being written for each tuple.  With this PR there should be 10 messages in Elasticsearch and HDFS:
```
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep jsonmapquery
jsonmapquery_index_2019.02.07.22   10
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/jsonMapQuery/*.json | wc -l
10
```
3. Change the batch size to 15:
```
  "parserClassName": "org.apache.metron.parsers.json.JSONMapParser",
  "sensorTopic": "jsonMapQuery",
  "parserConfig": {
    "jsonpQuery": "$.foo",
    "batchSize": 15
  }
}' 'http://user:password@node1:8082/api/v1/sensor/parser/config/jsonMapQuery'
```
4. Writer another jsonMapQuery message:
```
cat /tmp/jsonMapExampleOutput | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic jsonMapQuery
```
There should still be 10 messages in Elasticsearch and HDFS (no additional messages written):
```
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep jsonmapquery
jsonmapquery_index_2019.02.07.22   10
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/jsonMapQuery/*.json | wc -l
10
```
5. Writer another jsonMapQuery message:
```
cat /tmp/jsonMapExampleOutput | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic jsonMapQuery
```
This should cause a batch to flush so we should see 15 additional messages:
```
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep jsonmapquery
jsonmapquery_index_2019.02.07.22   25
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/jsonMapQuery/*.json | wc -l
25
```
6. Writer another jsonMapQuery message:
```
cat /tmp/jsonMapExampleOutput | /usr/hdp/current/kafka-broker/bin/kafka-console-producer.sh --broker-list node1:6667 --topic jsonMapQuery
```
This should cause another batch to flush since there were 5 messages still in the batch.  We should see another 15 messages added:
```
[root@node1 vagrant]# curl -s -XGET http://node1:9200/_cat/indices?h=index,docs.count | grep jsonmapquery
jsonmapquery_index_2019.02.07.22   40
[root@node1 vagrant]# hdfs dfs -cat /apps/metron/indexing/indexed/jsonMapQuery/*.json | wc -l
40
```

The various topologies can be tested by adjusting their batch sizes and timeouts.  The previous instructions focus on testing batch sizes for bro in indexing, snort in indexing and jsonMapQuery in parsing.  Other topologies and scenarios can and should be tested with different batch sizes and timeouts.

### Feedback Requested

Outside of the standard code review and function testing, they are a few areas that are not 100% clear and I would like feedback on:

- Is the fundamental architectural approach solid?  Are there any holes I'm not thinking of?
- The message ids generated by the bolts and used to track messages in the writer classes are Java UUIDs.  Is this good enough?  Any id that uniquely identifies a message could be used here.  Is there something that would perform better? 
- There is the potential for messages to pile up in `StormBulkWriterResponseHandler` (and probably other `BulkWriterComponent` too) when tuples time out and are replayed.  I think this is also an issue now but should we explore some kind of cache that can evict messages after timeouts?
- I have added javadocs and am planning on taking another pass and adding more.  Are there areas you feel need more/better explanation?  My goal is for these classes to be easy to understand and navigate.  Is the interaction between the bolts and response handler clear?

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -326,6 +329,10 @@ protected void handleError(String sensorType, byte[] originalMessage, Tuple tupl
     ErrorUtils.handleError(collector, error);
   }
```
ottobackwards: Why are we hashing here?  This seems like it would be something that would maybe impact performance that we can do another way.  Is there no existing field or type we can use?  Why not just pass the Tuple itself?

metron-platform/metron-parsing/metron-parsing-storm/src/test/java/org/apache/metron/parsers/bolt/WriterBoltTest.java
```diff
@@ -84,32 +86,39 @@ public SensorParserConfig getSensorParserConfig(String sensorType) {
   public void testBatchHappyPath() throws Exception {
     ParserConfigurations configurations = getConfigurations(5);
     String sensorType = "test";
+    WriterBolt bolt = spy(new WriterBolt(new WriterHandler(batchWriter), configurations, sensorType));
     List<Tuple> tuples = new ArrayList<>();
+    List<String> messageIds = new ArrayList<>();
     for(int i = 0;i < 5;++i) {
       Tuple t = mock(Tuple.class);
```
ottobackwards: the format string could be a static final String constant

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterResponseHandler.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+package org.apache.metron.writer;
+
+import org.apache.metron.common.writer.BulkWriterResponse;
+
+/**
+ * This interface is used by the {@link org.apache.metron.writer.BulkWriterComponent} to report that a queue for a
+ * sensor type has been flushed.  Different frameworks may have different requirements for committing processed messages
+ * so this abstraction provides a way to pass in the appropriate commit logic for the framework in use.
```
ottobackwards: s/abstraction/Interface

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
merrimanr: After thinking about this more, hashing the message is definitely not the right answer.  The most recent commit uses a java UUID instead.  We can explore even more efficient options but this should be an improvement.

ottobackwards: @merrimanr I think that is a better change, unless we can find a consistent and already existing id to use ( which I don't see either ).
We should just move ahead with this approach

ottobackwards:
This looks really good.  I think that the way this works is important, maybe important enough to be in the architecture documentation.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
```diff
@@ -227,110 +118,93 @@ public void write( String sensorType
...
-      sensorMessageMap.put(sensorType, messageList);
-    }
-    messageList.add(message);
+    messageList.put(messageId, message);
 
     //Check for batchSize flush
-    if (tupleList.size() >= batchSize) {
-      flush(sensorType, bulkMessageWriter, configurations, messageGetStrategy, tupleList, messageList);
+    if (messageList.size() >= batchSize) {
+      LOG.debug("Batch size of {} reached. Flushing {} messages for sensor {}.", batchSize, messageList.size(), sensorType);
```
nickwallen: It might make a lot of sense to extract the "Should I flush?" logic into separate policy classes.  

For example, we have some kind of `Policy` interface that the `BulkWriterComponent` interacts with.   The `BulkWriterComponent` can be initialized with zero or more Policy implementations.  It simply asks each `Policy`, "Hey, do I need to flush?" 

We would then have two policies; one that flushes when a max size is exceeded and another that flushes when a max timeout is exceeded.  Keeping these separate is going to make the code much simpler and easier to test.

nickwallen: I definitely appreciate the clean-up you did, like adding the `timeoutReached` method.  You were definitely heading in the same direction.  But breaking this out into policies is going to make the whole thing much simpler (IMHO).

justinleet:
@merrimanr Looked like the build failed.  Could you look into it when you have a chance?

merrimanr:
The latest commit converts `Map<String, MESSAGE_T>` to `List<BulkWriterMessage<MESSAGE_T>>`  in the `BulkMessageWriter.write` method (where `BulkWriterMessage` contains and id and message).  This should make it clearer what the `BulkMessageWriter` interface is expecting.

ottobackwards:
really like that last change @merrimanr, nice

merrimanr:
The latest commit implements @nickwallen's suggestion of refactoring the flush decision logic in `BulkWriterComponent.write` method into separate policies.  I think it was a great suggestion and I like it better than what we had.

I also added several javadocs to the various classes included in this PR.  I am still reviewing but if you see classes that need some javadoc love, just let me know.

Since this is a fairly significant change I tested everything again in full dev and things are still working.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
merrimanr: The latest commit adds this.

metron-platform/metron-parsing/metron-parsing-storm/src/test/java/org/apache/metron/parsers/bolt/WriterBoltTest.java
merrimanr: Done

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterResponseHandler.java
merrimanr: Done

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/FlushPolicy.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.metron.common.configuration.writer.WriterConfiguration;
+import org.apache.metron.common.writer.BulkMessageWriter;
+import org.apache.metron.common.writer.BulkWriterMessage;
+
+import java.util.List;
+
+/**
+ * This interface is used by the {@link org.apache.metron.writer.BulkWriterComponent} to determine if a batch should be flushed.
+ * @param <MESSAGE_T> Message type
+ */
```
ottobackwards: Can I ask how the MESSAGE_T is relevant to this interface?  Don't you just need the size?  Why does this need this?

merrimanr: Yeah I guess you're right.  My thinking was we might want a policy that depends on message contents at some point.  We don't have an actual use for that now so it is unnecessary.  If we do have a need in the future we can just change the interface since it's internal.

ottobackwards: Well, there is nothing about the function of the interface ( the call itself ) that needs the type.  You can always implement the interface with an implementing class that is specialized.

merrimanr: I think your initial reaction was correct.  Let's keep it simple for now instead of trying to predict the future.  Latest commit uses batch size instead.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
```diff
@@ -18,315 +18,159 @@
...
-
-  public BulkWriterComponent(OutputCollector collector, boolean handleCommit, boolean handleError) {
-    this(collector);
-    this.handleCommit = handleCommit;
-    this.handleError = handleError;
+  public BulkWriterComponent(BulkWriterResponseHandler bulkWriterResponseHandler) {
+    this.bulkWriterResponseHandler = bulkWriterResponseHandler;
+    flushPolicies = new ArrayList<>();
+    flushPolicies.add(new BatchSizePolicy());
+    batchTimeoutPolicy = new BatchTimeoutPolicy();
```
mmiklavc: Small nit - instead of having a special instance variable for batchTimeoutPolicy that later has a special dep injection method for clock, could we instead add a protected constructor that allows you to pass in the default flush policies, or maybe just the clock? When I see the constructor that has special handling for one of the policies as a private member variable, it leads me to wondering why it's treated special. I don't have a strong preference for either approach, though passing in the clock is a bit more clear that we're testing with default policies and only injecting variation with the provided clock. 

In general, I have a slight preference for constructors over setter methods for dep injection because the options for test injections are very clearly documented by the additional constructors vs having to look for n setter methods in the API. It also forces the object to be fully configured with an expected valid state _prior_ to calling any methods under test, which I think is easier for future developers making changes to the same classes and tests.

Here are a couple examples:

1. 
```
// Using clock as constructor arg
/**
 * Main constructor
 **/
public BulkWriterComponent(BulkWriterResponseHandler bulkWriterResponseHandler) {
    this(bulkWriterResponseHandler, new Clock());
}

public BulkWriterComponent(BulkWriterResponseHandler bulkWriterResponseHandler, Clock clock) {
    this.bulkWriterResponseHandler = bulkWriterResponseHandler;
    flushPolicies = new ArrayList<>();
    flushPolicies.add(new BatchSizePolicy());
    flushPolicies.add(new BatchTimeoutPolicy(clock));
}
```
2.
```
// Using flush polilcies as constructor arg
/**
 * Main constructor
 **/
public BulkWriterComponent(BulkWriterResponseHandler bulkWriterResponseHandler) {
    List<FlushPolicy> flushPolicies = new ArrayList<>();
    flushPolicies.add(new BatchSizePolicy());
    flushPolicies.add(new BatchTimeoutPolicy(new Clock()));
    this(bulkWriterResponseHandler, flushPolicies);
}

public BulkWriterComponent(BulkWriterResponseHandler bulkWriterResponseHandler, List<FlushPolicy> flushPolicies) {
    this.bulkWriterResponseHandler = bulkWriterResponseHandler;
    this.flushPolicies = flushPolicies;
}
```

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -77,17 +85,6 @@ public ParserBolt( String zookeeperUrl
     super(zookeeperUrl);
     this.parserRunner = parserRunner;
     this.sensorToWriterMap = sensorToWriterMap;
```
merrimanr: This was removed because the `WriterHandler.handleAck` method no longer exists.  Acking tuples is not the responsibility of the `StormBulkWriterResponseHandler` class.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
merrimanr: The reason I did this is because of `batchTimeoutPolicy`.  There is a `BulkWriterComponent.setDefaultBatchTimeout` method and it's not obvious it's only used for testing (like clock is).  I didn't want to assume this can only be set at `BulkWriterComponent` creation time.  It looks like it's only set in the bolt prepare methods so I believe we can safely move it to the constructor like you're suggesting.  Do we want to make this assumption, that `batchTimeoutPolicy` is only set once?  I will make the clock change either way if you think it makes the code easier to follow.

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
```diff
@@ -326,6 +329,14 @@ protected void handleError(String sensorType, byte[] originalMessage, Tuple tupl
     ErrorUtils.handleError(collector, error);
   }
 
+  protected List<String> getMessageIds(int size) {
+    List<String> messageIds = new ArrayList<>();
+    for(int i =0; i < size; i++) {
+      messageIds.add(UUID.randomUUID().toString());
```
nickwallen: Why do we have to generate the UUID, instead of just using the message's GUID?  I thought everything coming out of the `ParserRunner` has to have a GUID.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
mmiklavc: I see what you're saying. Honestly, I think the batch timeout value belongs to the policy, not the component. I get providing a default via the component, but BulkWriterComponent.setDefaultBatchTimeout implies a hard-coded dependency on batch timeout policy, whereas at the same time we're trying to abstract the component away from specific policy implementations. I think we should either provide the timeout externally as an arg (like my other recommendations), or provide config that's pluggable for the policies. E.g. if we need to update that timeout on the fly (I think we do, actually), then have the class that provided the policies in the first place update the config for that policy. I think that might be in the bolt, and this sounds like it's related to the config supplier stuff we do with Curator. 

Another idea for managing default policies could be to put the construction of that list into a default factory of sorts. Just a thought, I'm ok either way.

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
mmiklavc: With HBaseWriter and PcapWriter now gone (they were legacy leftovers, I believe), is there a reason to keep MessageWriter around? If there are other side effects, e.g. Flux files, Ambari configs, docs, etc. I'm happy to have it left for now and make that a follow-on Jira, but figured it's worth mentioning.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BatchTimeoutPolicy.java
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Returns the configured timeout for a sensor type in milliseconds.  The default timeout will be used if the configured timeout is
+   * set to 0 or greater than the default timeout.
+   * @param sensorType
+   * @param configurations
+   * @return
+   */
+  protected long getBatchTimeout(String sensorType, WriterConfiguration configurations) {
+    int batchTimeoutSecs = configurations.getBatchTimeout(sensorType);
+    if (batchTimeoutSecs <= 0 || batchTimeoutSecs > defaultBatchTimeout) {
```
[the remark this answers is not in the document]
merrimanr: This was preexisting.  I don't know the exact reason but I suspect it was a guard against setting the batch timeout to be greater than the tuple timeout.  @mmiklavc also brought up a question around default batch timeout and I'm not quite sure if what we have is correct or what the requirements are.  I know that the default batch timeout is 1/2 the tuple timeout when running in a bolt and 6 otherwise (test scenarios I think).  I also know the default batch timeout is only set once at the beginning and isn't tied to anything dynamically. 

How should default batch timeout work?  I think we need to keep the default to be 1/2 the tuple time in storm but what about other scenarios?  Should this setting be exposed in say global config?  Should we be able to change it at runtime?

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterResponseHandler.java
```diff
@@ -0,0 +1,35 @@
...
+ * so this interface provides a way to pass in the appropriate commit logic for the framework in use.
+ */
+public interface BulkWriterResponseHandler {
+
+  /**
+   * Called immediately after a batch has been flushed.
+   * @param sensorType The type of sensor generating the messages
+   * @param response A response containing successes and failures within the batch
+   */
+  void handleFlush(String sensorType, BulkWriterResponse response);
```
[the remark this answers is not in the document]
merrimanr: I think they are different because they are called at different times and have different purposes:  `'FlushPolicy` classes are called before a flush to determine if a flush is necessary and `BulkWriterResponseHandler` is called after a flush happens with the results.  Maybe an example can help me understand?

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterMessage.java
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.metron.common.writer;
+
+import java.util.Objects;
+
+public class BulkWriterMessage<MESSAGE_T> {
```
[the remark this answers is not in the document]
merrimanr: Yes it is preexisting.  I believe `JSONObject` is the only type that is ever used..

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BatchTimeoutPolicy.java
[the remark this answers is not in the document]
nickwallen: I think we should just do what the user sets it to here.  Changing it under the covers like this is just going to cause confusion.  Especially since this code will also be run outside of Storm, it shouldn't be bound by subtle Storm constraints.

Perhaps there is somewhere in Storm-specific code that we can add a check and clear log statement or exception, if the configured batch timeout exceeds some percentage of the tuple timeout?  `ParserBolt` maybe or elsewhere?

If this is current behavior, I'd also be open to just tabling this for a separate JIRA/PR.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
mmiklavc: @merrimanr - Thought about this a bit more in context of @nickwallen's recent comments. Yeah, while the actual timeout setting itself can change, the _default_ should definitely not change on the fly. I can't think of a good reason why we would ever want a default to dynamically change. That's more of an upgrade/new release type of change imo that has documentation and release note implications.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterResponseHandler.java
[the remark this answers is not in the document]
nickwallen: It may just be a bad idea. :)  They just seem so similar to me.  I'll flesh out something and see if it makes sense.  Don't worry about it for now.

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
merrimanr: I checked and you are correct.  I will change it to use the GUID instead.  Is it worth generating the UUID in case the GUID is missing for some reason?

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
merrimanr: I experimented with this and the changes were just significant enough to make me want to push it off to a follow on.  If you can handle reviewing it here I'm happy to make the changes :)

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterComponent.java
merrimanr: Thanks that helps.  I will play around with this and see if I can come up with a solution based on your and @nickwallen's feedback.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BatchTimeoutPolicy.java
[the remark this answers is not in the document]
mmiklavc: Hrm, that's a good catch @nickwallen. And I believe @merrimanr is right. The reason we want to restrict the timeout range is so that if the tuple timeout ends up being short enough with respect to our batch timeout, we'll fail a lot of tuples due to timeout. I'm not sure if there's a better way to capture that reasoning in the code, but we should definitely refactor or add comments to make this clearer.

It seems likely that the only reason this was configurable (the default, that is) in the first place was for testing. And oddly enough, that capability and flexibility made this less clear for us now. I think ultimately in Storm world we should have that default lockstep with the Storm concept of tuple timeout. So if it's 1/2 the tuple timeout, let's leave that be. Considering other streaming frameworks, that concept of where we derive the default should be more general. 

I would be in favor of renaming "defaultBatchTimeout" to "maxBatchTimeout" and then having a method "setMaxBatchTimeout" or constructor arg that clearly sets that value. Any reliance on tuples or Storm-specific config I think should be kept out of the timeout policy implementation and provided externally. In this case, this is an option that is not dynamic bc changing the tuple timeout requires a topology restart. We should continue to pull the current tuple timeout from the Storm Bolt config and pass that in per the calculation as the "max timeout." Our Storm code cares about the max being 1/2 tuple timeout whereas the policy class only cares that there's the concept of a max value in the first place. i.e. let's leave the construction of that value to the classes that are instantiating the policies. How's that sound to you gents?

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
mmiklavc: @merrimanr Sounds good to me - Would you mind creating a Jira for that and linking it to our comments here for reference and background?

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BatchTimeoutPolicy.java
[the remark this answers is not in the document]
merrimanr: The latest commit changes "defaultBatchTimeout" to "maxBatchTimeout", moves that setting to the constructor(s) and removes the set methods.  A default value no longer lives in `BatchTimeoutPolicy` and instead an exception is thrown if the value is less than or equal to 0.  The bolts now pass down the value into the `BulkWriterComponent` constructor which uses it to create the policy.

There was a default value of 6 being used previously.  As I was making these changes I realized the `WriterBolt` class relies on this default instead of calculating what it should be.  I moved the default to this class instead so it continues to function as it does now.  Rather than going to the trouble of figuring out the right way to set this I think we should replace it with the `BulkMessageWriterBolt`.  They do the same thing and are redundant.  Planning on opening a separate PR for that anyways.  If anyone disagrees with this and would like to discuss the correct way to set "maxBatchTimeout" in this bolt let me know.

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
merrimanr: This is done.

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterResponse.java
```diff
@@ -29,38 +28,38 @@
 import java.util.Map;
 
 public class BulkWriterResponse {
-    private Multimap<Throwable, Tuple> errors = ArrayListMultimap.create();
-    private List<Tuple> successes = new ArrayList<>();
+    private Multimap<Throwable, String> errors = ArrayListMultimap.create();
+    private List<String> successes = new ArrayList<>();
```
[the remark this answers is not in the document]
merrimanr: This is done in the latest commit.

metron-platform/metron-elasticsearch/src/main/java/org/apache/metron/elasticsearch/writer/ElasticsearchWriter.java
```diff
@@ -65,11 +64,11 @@
   /**
    * Responsible for writing documents.
    *
-   * <p>Uses a {@link TupleBasedDocument} to maintain the relationship between
+   * <p>Uses a {@link MessageIdBasedDocument} to maintain the relationship between
    * a {@link Tuple} and the document created from the contents of that tuple. If
```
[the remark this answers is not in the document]
merrimanr: Done

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkMessageWriter.java
```diff
@@ -32,18 +29,16 @@
...
 
   /**
-  * Writes the messages to a particular output (e.g. Elasticsearch). Exceptions trigger failure of the entire batch.
-  * @param sensorType The type of sensor being generating the messages
+  * Writes the messages to a particular output (e.g. Elasticsearch).  A response is returned with successful and failed message ids.
+  * @param sensorType The type of sensor generating the messages
   * @param configurations Configurations that should be passed to the writer (e.g. index and
-  * @param tuples The Tuples that produced the message to be written
-  * @param messages  The message to be written
+  * @param messages  The message to be written in the form of a message id to message map.  Message ids are used in the response to report successes/failures.
```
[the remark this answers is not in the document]
merrimanr: Done

metron-platform/metron-writer/src/test/java/org/apache/metron/writer/BatchTimeoutPolicyTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+package org.apache.metron.writer;
+
+import org.apache.metron.common.configuration.writer.WriterConfiguration;
+import org.apache.metron.common.system.Clock;
+import org.apache.metron.common.writer.BulkWriterMessage;
+import org.json.simple.JSONObject;
+import org.junit.Test;
+
+import java.util.ArrayList;
```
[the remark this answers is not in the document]
merrimanr: Done

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/BulkWriterResponseHandler.java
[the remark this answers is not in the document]
nickwallen: Just to flush out the idea, i opened a PR against your branch.  This is only to flesh out the idea, I didn't attempt to get the tests all working.  I added some comments in there to help emphasize the changes, since a bunch of master commits showed up in the PR.

https://github.com/merrimanr/incubator-metron/pull/40

Feel free to take it or leave it.  It is a subtle difference that in my mind makes things simpler to test and reason about.  But I would not be against keeping what you have.  It is completely your call.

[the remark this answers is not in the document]
merrimanr: The latest commit includes your suggestion.  Let me know what you think.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/FlushPolicy.java
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * This method is called whenever messages are passed to {@link BulkWriterComponent#write(String, BulkWriterMessage, BulkMessageWriter, WriterConfiguration)}.
+   * Each implementation of {@link org.apache.metron.writer.FlushPolicy#shouldFlush(String, WriterConfiguration, int)} will be called in order
+   * and the first one to return true will trigger a flush and continue on.
+   * @param sensorType sensor type
+   * @param configurations configurations
+   * @param batchSize batch size of messages written
+   * @return true if batch should be flushed
+   */
+  boolean shouldFlush(String sensorType, WriterConfiguration configurations, int batchSize);
```
nickwallen: Why pass in the `batchSize`, rather than just pass in the batch itself?  It seems like passing the entire 'batch' of messages would give policies more information, at little to no extra cost.
```
boolean shouldFlush(
  String sensorType, 
  WriterConfiguration configs, 
  List<BulkWriterMessage<MESSAGE_T>> messages)
```

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkMessageWriter.java
```diff
@@ -32,18 +29,16 @@
...
-  * @param messages  The message to be written
+  * @param messages  A list of messages to be written.  Message ids are used in the response to report successes/failures.
   * @return A response containing successes and failures within the batch.
   * @throws Exception If an unrecoverable error is made, an Exception is thrown which should be treated as a full-batch failure (e.g. target system is down).
   */
   BulkWriterResponse write(String sensorType
             , WriterConfiguration configurations
-            , Iterable<Tuple> tuples
-            , List<MESSAGE_T> messages
+            , List<BulkWriterMessage<MESSAGE_T>> messages
```
nickwallen: Would it makes sense for `BulkWriterMessage` to also contain the `sensorType`?

Seems like that might simplify the signature of a number of classes including `FlushPolicy`, `BulkMessageWriter`, and `BulkWriterComponent`. But maybe there is some interaction that I am not thinking about.

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterMessage.java
```diff
@@ -0,0 +1,58 @@
...
```
nickwallen: Can you add javadoc to this class?

nickwallen: Maybe it is just me and my dyslexic tendencies, but having both a `BulkWriterMessage` and `BulkMessageWriter` is really confusing me as I navigate through the code.  

Would naming this class a `BulkMessage` be more clear?  It makes sense to me that a `BulkMessageWriter` is responsible for writing `BulkMessages`.

I am totally willing to accept that this is just some weird short circuit in my brain and the name you have works just fine. :)

merrimanr: I agree with you, it's confusing.  I like your idea better.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/FlushPolicy.java
merrimanr: I did it that way initially but changed it based on this [feedback](https://github.com/apache/metron/pull/1330#pullrequestreview-203467469) from @ottobackwards.  I see both points so I could go either way.

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkWriterMessage.java
mmiklavc: @nickwallen I had the same double-take when I first read that. Good suggestion.

metron-platform/metron-writer/src/main/java/org/apache/metron/writer/FlushPolicy.java
nickwallen: I think it just makes sense to provide the information in as raw a form as possible to the policies, rather than assuming that they need certain bits of information, like the batch size. 

I don't understand the downside to that approach in @ottobackwards 's original comment.

merrimanr: The latest commit changes it back to the way I had it initially.  The `FlushPolicy.shouldFlush` method now accepts a list of messages.

metron-platform/metron-common/src/main/java/org/apache/metron/common/writer/BulkMessageWriter.java
merrimanr: I experimented with this and it's not as straightforward as it seems.  Throughout all these classes it is assumed that messages are batched per sensor.  If we move sensorType into each message than each class that consumes a `List<BulkMessage>` now has to handle it differently.

For example the `ElasticsearchWriter` class looks up an index for the given sensor type and writes all messages to that index.  With this change we would now need to lookup an index for each sensor type.  Another example is `BatchTimeoutPolicy`.  The list of messages isn't even used because it relies solely on the timeout configured for each sensor type.  Do we get the sensor type from the list of messages now?  Do we just use the sensor type of the first message or do we need to ensure the sensor type matches for all messages in the list?

I feel like this is a significant change and wouldn't provide much benefit.  If we do want to simplify these interfaces I would prefer to move the sensor type and list of messages into a wrapper object (`BulkWriterRequest` or something).

nickwallen: Sounds legit.  I am good with not touching it.  Thanks for taking a look.

justinleet:
The implementation of this seems pretty reasonable, from taking a look at things.  I didn't dig as deep as the other reviewers, but I'm good with where this is at code-wise.

@ottobackwards requested some architecture documentation and I didn't see any updates there. This feels like a good opportunity to get a core part of our architecture well documented and understood to make issues like this easier to understand in the future. Would you be able to write up some docs around this?

merrimanr:
@justinleet yes was planning on circling back to that once we made it through the code review.  Now that this PR is in a good spot I'll start on that documentation.  Thanks.

ottobackwards:
If you want to put out a discuss for how to document this and like things and control scope, I would not be opposed to a follow on, as long as it get's done before next release.

merrimanr:
Thanks @ottobackwards I appreciate that.  Would save me from the difficult task of keeping this big PR up to date with master.  I will start a discuss thread now on the topic.  If there are no objections there I will create a follow on Jira and commit to getting it done within a few days.

ottobackwards:
I think we need some way to document architectural concepts that is not as onerous as a full architecture document.  I can imagine almost an architectural FAQ type of thing:

Q. How to the Indexers batch writes?
A. [index_batching.md](?)

merrimanr:
@ottobackwards just started a discuss thread on the architectural documentation topic.

merrimanr:
Follow on Jira is [here](https://issues.apache.org/jira/browse/METRON-2014).

justinleet:
I'm +1, with the follow-on for documentation (and can you post the Jira here for posterity?)

merrimanr:
https://issues.apache.org/jira/browse/METRON-2014

ottobackwards:
I am +1 as well. Thanks for working through this with us @merrimanr

mmiklavc:
Nice work @merrimanr, I am +1 pending agreement from @JonZeolla in the DISCUSS thread put out about documentation. Also, @merrimanr would you be able to link that thread to the Jira and tag it as required for our next release?

metron-platform/metron-parsing/metron-parsing-storm/src/main/java/org/apache/metron/parsers/bolt/ParserBolt.java
mmiklavc: @merrimanr I may have missed the Jira/comment on it - did this ticket for the MessageWriter cleanup get created?

merrimanr: https://issues.apache.org/jira/browse/METRON-2015

merrimanr:
Here is the Jira from the DISCUSS thread:  https://issues.apache.org/jira/browse/METRON-2014

PR #1334, opened 2019-02-08
METRON-1997: Replace Threat Triage Score Field Slider with Text Box

ruffle1986:
## Contributor Comments

This PR replaces the score slider with an ace editor to highlight Stellar syntax. From now on users can set the score by adding a Stellar expression. It's backward compatible so the user still can add numbers as scores but they have to type it instead of setting the slider.

Since score has become a string and isn't a number anymore we cannot group or sort rules by that. I removed the two components which were responsible for sorting and grouping.

This is what I'm talking about:

screen shot 2019-02-28 at 16 15 16

This PR also adds syntax highlighting to the `rule` field. The label "TEXT" was rewritten to "RULE" to make it clear what we're editing.

**A few things to keep in mind**

- If you add two or more rules with the exact same `rule` it just simply doesn't add it to the list (it should be more responsive by letting the user know)

**Testing**

1. Go to `/sensors`
2. Click on a sensor's edit button (the pen on the right)
3. Look for the `Threat Triage` section in the right floating pane
3. Click on the Threat Triage section's expander (it's the icon on the right)
4. In the second pane, click on the big `+` button at the bottom
5. In the 3rd panel add a name, a rule and a score (a score can be a Stellar expression like `value * 10`)
6. Click on the `Save` button. It'll validate the expression for you. If it's not, it'll let you know.
7. Check the newly added rule in the rule list in the 2nd pane
9. Take a quick look at the Threat Triage section in the 1st pane if it's ok
8. Don't forget to click on the Save button in the 1st pane

9. Open the read-only sensor pane (clicking on anywhere else but the pen on the row)
10. Find the Threat Triage section
11. Check if it's ok

threat-triage

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


sardell:
@ruffle1986 Currently, the slider allows for an input of 0 - 100. No negative numbers or values over 100. Also, we should probably add some validation in the case of Stellar expressions.

ruffle1986:
@sardell Good point! 

@nickwallen @merrimanr Is there an API endpoint with which we can validate the expression typed by the user? Validating a number is not a big deal but since it accepts stellar expressions as well, IMHO, that would be great if it was validated by the backend.

What do you think?

nickwallen:
@ruffle1986 This might do it `POST /api/v1/stellar/validate/rules`.

https://metron.apache.org/current-book/metron-interface/metron-rest/index.html#POST_apiv1stellarvalidaterules

ruffle1986:
@nickwallen Perfect! Thanks for the quick reply.

I'm going to wire the validation on the client side if we all agree it's necessary (it is IMHO). I'm also adding a syntax highlighter to the text box.

ruffle1986:
@merrimanr @nickwallen as we discussed earlier I removed `score` from the triage rule as it was a deprecated property but the server still responds with `400 Bad Request`.

This is the request body I'm sending to the server (`/api/v1/sensor/enrichment/config/syslog5424`):

```json
{
   "enrichment":{
      "fieldMap":{

      },
      "fieldToTypeMap":{

      },
      "config":{

      }
   },
   "threatIntel":{
      "fieldMap":{

      },
      "fieldToTypeMap":{

      },
      "config":{

      },
      "triageConfig":{
         "riskLevelRules":[
            {
               "name":"qwe",
               "comment":"",
               "rule":"match{ var1 < 10 => 'warn', var1 >= 10 => 'critical', default => 'info'}",
               "scoreExpression":"match{ var1 < 10 => 'warn', var1 >= 10 => 'critical', default => 'info'}"
            }
         ],
         "aggregator":"MAX",
         "aggregationConfig":{

         }
      }
   }
}
```

sardell:
@ruffle1986 Just a few more things:

- I noticed that when I add a threat triage rule without a name, the rule and score adjustment are shown in the list of rules. If I add a rule with a name, the name and score adjustment are shown and the rule is hidden. I think this can be confusing and hard to parse if a user has many rules because of the way we are displaying the data right now. I think it might be better to use a table to display the rule information. If there isn't a name for a rule, that rule's name column  remains blank.

- I also noticed that in the Threat Triage Rules in the read-only pane there is a word break. 
Screen Shot 2019-03-28 at 1 49 34 PM
This can be quickly resolved by removing the bootstrap col classes from the form labels and values and applying it to the wrapping div. 

- Your testing instructions include a test button, but it looks like that was removed. Could you just strikeout or remove that part of your instructions?

ruffle1986:
@sardell 

> I noticed that when I add a threat triage rule without a name, the rule and score adjustment are shown in the list of rules. If I add a rule with a name, the name and score adjustment are shown and the rule is hidden. I think this can be confusing and hard to parse if a user has many rules because of the way we are displaying the data right now. I think it might be better to use a table to display the rule information. If there isn't a name for a rule, that rule's name column remains blank.

You're right. I'll add the table headers for each column to make it clear what's going on. I'm also stretching the table. By this, the icons on the right goes to the right side of the pane. It gives more space to display the name for each rule in an individual column.
As a bonus, I'll replace the "whatever icon it is" next to the trash icon with a pencil icon to use the same icons for the same purpose consistently. 

> I also noticed that in the Threat Triage Rules in the read-only pane there is a word break.

Nice catch! 👍 

> Your testing instructions include a test button, but it looks like that was removed. Could you just strikeout or remove that part of your instructions?

Thanks! Description updated.

merrimanr:
I ran this up in full dev and everything worked as expected.  I was able to edit and save rules and verified syntax highlighting and Stellar validation worked properly.  +1

PR #1340, opened 2019-02-15
METRON-2007: Management UI not loading grok statements correctly

merrimanr:
## Contributor Comments
I believe these regressions happened when upgrading Angular:

- The `responseType` needs to be set to `string` for any REST calls that return a String type.  I also fixed `kafka.service.ts` even though it's not directly related to this issue.
- The `grokStatement` type was incorrectly changed from string to Object in `sensor-parser-config.component.ts` so I changed it back and handled casting it in the callback.

### Testing
This has been tested in full dev.  Before this fix opening either a squid or yaf sensor would throw an error about not being able to load the grok statement from HDFS or the classpath.  After this fix you should be able to open either sensor with the grok statement correctly loaded and displayed.  You should also be able to create a new Grok sensor and have the pattern label automatically populated with the sensor name.

### Outstanding Issues
- Fixing this subsequently causes an "Expression has changed after it was checked"  error in the browser console.  I believe the offending line is [here](https://github.com/apache/metron/blob/master/metron-interface/metron-config/src/app/sensors/sensor-parser-config/sensor-parser-config.component.html#L200).
- I have not updated or run the unit tests yet

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-config/src/app/service/hdfs.service.ts
+++ b/metron-interface/metron-config/src/app/service/hdfs.service.ts
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 import { Injectable, Inject } from '@angular/core';
-import {HttpClient, HttpHeaders, HttpParams} from '@angular/common/http';
+import {HttpClient, HttpParams} from '@angular/common/http';
 import { Observable } from 'rxjs';
 import { map, catchError } from 'rxjs/operators';
 import { HttpUtil } from '../util/httpUtil';
--- a/metron-interface/metron-config/src/app/service/hdfs.service.ts
+++ b/metron-interface/metron-config/src/app/service/hdfs.service.ts
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 import { Injectable, Inject } from '@angular/core';
-import { HttpClient, HttpParams } from '@angular/common/http';
+import {HttpClient, HttpHeaders, HttpParams} from '@angular/common/http';
 import { Observable } from 'rxjs';
 import { map, catchError } from 'rxjs/operators';
 import { HttpUtil } from '../util/httpUtil';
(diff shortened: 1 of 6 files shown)


metron-interface/metron-config/src/app/service/hdfs.service.ts
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 import { Injectable, Inject } from '@angular/core';
-import { HttpClient, HttpParams } from '@angular/common/http';
+import {HttpClient, HttpHeaders, HttpParams} from '@angular/common/http';
```
ruffle1986: Thank you very much for the quick fix and sorry for the inconveniences.

I see a small possible tweak here but it's not critical at all.

You're importing the `HttpHeaders` object but you're not using it in the code.

sardell:
+1. Built and tested with full dev.

metron-interface/metron-config/src/app/service/hdfs.service.ts
merrimanr: Done.

merrimanr:
Thanks for the review @sardell and @ruffle1986.  Any idea on the console error?

sardell:
@merrimanr I'm not 100% sure. Maybe the value of configValid is updated with an asynchronous operation and the value is returned after Angular finishes its lifecycle hook with the previous configValid value. I would have to look into it more, but that's usually the cause of these types of errors.

merrimanr:
The latest commit resolves the console error.  The issue is that we are registering an `onGrokStatementChange` function that updates `grokStatementValid` which is then used to update the `configValid` field.  I believe this is generally incorrect because we are trying to directly manage state rather than letting Angular do it.  

This fix in the latest commit:
- removes the `configValid` field and replaces it with a `isConfigValid` function call in the html template
- removes the `grokStatementValid` and `configValid` fields
- removes the `onGrokStatementChange` from the `grokStatement` input control
- adds a new `isGrokStatementValid` method that computes status directly from the `grokStatement` field

sardell:
+1. Tested against full dev after the latest changes to fix the "Expression has changed" error. Thanks, @merrimanr!

PR #1341, opened 2019-02-16
METRON-614: Eliminate use of the default Charset

justinleet:
## Contributor Comments
Most of the motivation here is to spur discussion about what we want to do.  I changed everything to UTF-8. Because UTF-8 is usually a solid default.

Everything with default charset now uses `StandardCharsets.UTF_8` (with the occasional variant of `StandardCharsets.UTF_8.name())`.

I've done a few things here
* An IntelliJ driven automatic fix for "Implicit usage of the platform's default charset". This fixes the vast majority of issues.
* An IntelliJ driven automatic fix for "Standard Charset object can be used". This replaces a few "UTF-8" strings that were floating.
* Manual replacements for `PrintStream`, `FileReader`, and `FileWriter`.
** `PrintStream` just requires modifying the args slightly (need to include `autoFlush` in params).
** `FileReader` and `FileWriter` do not accept encoding arguments and are drop in replaced with variants that do.
* A pass for issues that error-prone caught that IntelliJ didn't. These were typically something like, `"".getBytes()` that were in our tests.
* Changing the POM to throw compiler errors on default charset issues. Otherwise, it's gonna keep cropping up, at least until our build is substantially cleaner.

As a nice side effect, the build logs are cleaner now that error prone isn't squawking on every default charset.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
+1 by inspection.  Thanks!

nickwallen:
I noticed that your work removes all [DefaultCharset](http://errorprone.info/bugpattern/DefaultCharset) warnings from Error Prone.  

Can we turn on the error level setting for that check, so that if anyone accidentally re-introduces this the build will fail?  I think the following change to pom.xml will do that.
```
diff --git a/pom.xml b/pom.xml
index 8b660c6c5..187cbde43 100644
--- a/pom.xml
+++ b/pom.xml
@@ -238,6 +238,9 @@
                     <compilerArgument>-Xlint:unchecked</compilerArgument>
                     <target>${global_java_version}</target>
                     <showWarnings>true</showWarnings>
+                    <compilerArgs>
+                      <arg>-Xep:DefaultCharset:ERROR</arg>
+                    </compilerArgs>
                 </configuration>
                 <dependencies>
```

justinleet:
@nickwallen That change is already included in the pom

https://github.com/apache/metron/pull/1341/files#diff-600376dffeb79835ede4a0b285078036R248

nickwallen:
Ah, very good.  +1  Thanks for addressing this.

nickwallen:
Hey @justinleet - Can you clean-up the conflicts when you get a chance?

mmiklavc:
I like this PR - aside from the conflicts is there anything left to resolve?

justinleet:
Fixed the conflicts and ran back up in full dev.  Having said that, I'd like to hold off on merging this until the release is over, since I don't want to try to change something like this or work around it if something comes up.

justinleet:
@mmiklavc @nickwallen @ottobackwards I merged master post-release and ran up in full dev again. Is everyone still good with the changes?

mmiklavc:
Probably worth referencing this - https://lists.apache.org/thread.html/55e57410cb8cd467a51545e4ae0f9f67d32312cc3f9e1afa144552f4@%3Cdev.metron.apache.org%3E

Just to follow up - are there any concerns with, for example, non-UTF-8 inbound data charsets? I think it makes sense for us to normalize, however we may run into trouble in our parsers, e.g https://github.com/apache/metron/pull/1341/files#diff-e5a5c182fb529e5e2d93fc9f06ce1012.

justinleet:
Ahh good call, it's been long enough that I'd forgotten about that discussion. There should definitely at least be a README addition that I'll add.

Re: non-UTF-8 inbound data sets, that's potentially a fair problem, although I don't personally know what the circumstances would be where non-UTF-8 string data is coming from (maybe Latin-1?).  Seems like the only real way to deal with this is to make it configurable at the parser level, or otherwise mixing incoming charset encodings is a problem (Which I think it would be right now, if everything is just using platform default, right? Double check my thinking on that). Then the parser itself just reads with whatever character encoding.

At that point, stuff like say GrokParser would need do something like `new InputStreamReader(commonInputStream, getEncoding());` or similar.

mmiklavc:
I think setting a default to `UTF-8` in the parsers and documenting it would be the way to go. Provide a per-sensor config option, e.g. `inputDataCharset` that lets users configure it for the edge case. Emphasis on per-sensor because 99/100 sensors will probably be `UTF-8`, and then one will be something wild like `EBCDIC` because hey, why not.

In general, I agree that it would be odd for any network sensors to be set to anything other than `UTF-8`. We're probably looking at other sources of mischief, though. A couple examples could be streaming and bulk loaded enrichments. I would not be surprised to find someone at some point loading `ISO-8859-1` or `Windows-1252`. In multiple big data projects prior to Metron I had to deal with encodings like this.

nickwallen:
> I think setting a default to UTF-8 in the parsers and documenting it would be the way to go. Provide a per-sensor config option, e.g. inputDataCharset

What is the status of this?  I assume the suggestion here to provide additional configuration is a feature enhancement request and not something we would tackle on this PR.  Is that right @mmiklavc?

What do we need to do to get this PR merged?

mmiklavc:
> > I think setting a default to UTF-8 in the parsers and documenting it would be the way to go. Provide a per-sensor config option, e.g. inputDataCharset
> 
> What is the status of this? I assume the suggestion here to provide additional configuration is a feature enhancement request and not something we would tackle on this PR. Is that right @mmiklavc?
> 
> What do we need to do to get this PR merged?

I thought this could be handled pretty trivially here. What if we modified this ever so slightly to include a default charset for the parsers that can be overridden as needed? Here's an example.

```
# MessageParser interface

public interface MessageParser<T> extends Configurable {
...
    default Charset getReadCharset() {
        return StandardCharsets.UTF_8;
    }
...
}

// then we can do this for the default case

# CSVParser

public class CSVParser extends BasicParser {
...
  @Override
  public List<JSONObject> parse(byte[] rawMessage) {
    try {
      String msg = new String(rawMessage, getReadCharset());
...

// And for implementations that might want something else

public class CSVParser extends BasicParser {
...
  private Charset readCharset;

  @Override
  public void configure(Map<String, Object> parserConfig) {
    converter = new CSVConverter();
    converter.initialize(parserConfig);
    Object tsFormatObj = parserConfig.get(TIMESTAMP_FORMAT_CONF);
    if(tsFormatObj != null) {
      timestampFormat = new SimpleDateFormat(tsFormatObj.toString());
    }
    if (parserConfig.containsKey(READ_CHARSET)) {
      readCharset = Charset.forName((String) parserConfig.get(READ_CHARSET));
    }
  }

  @Override
  public Charset getReadCharset() {
    return readCharset != null ? readCharset : super.getReadCharset();
  }
...
```

We could even push that config option into the BasicParser if we wanted. The upshot is that this seems like a really trivial change. Rather than opening another PR to address it, can we just add the option with our desired default now and be done with it?

ottobackwards:
I think that is reasonable, but I would want to see tests with alternate charsets

mmiklavc:
> I think that is reasonable, but I would want to see tests with alternate charsets

e.g. test that we can read in ISO-8859-1 and properly convert to UTF-8? Yeah, that makes sense. @justinleet I'm happy to work on that and submit a PR against your PR. I think we just need some conflict resolution before I tackle that.

justinleet:
@mmiklavc Sure, I'll deconflict and get you set up. Definitely would appreciate the help; I've been too busy to come back to this and that's not gonna clear up for a couple weeks at least.

mmiklavc:
> @mmiklavc Sure, I'll deconflict and get you set up. Definitely would appreciate the help; I've been too busy to come back to this and that's not gonna clear up for a couple weeks at least.

Bump - any chance you can deconflict this @justinleet? Again, I'm happy to contribute the parser changes discussed above, but I need a clean PR to work from.

justinleet:
@mmiklavc I'll try to get to it by the end of the week. I had a personal emergency come up last week, and unfortunately this fell pretty hard in my personal priorities. Sorry about the delay on it

justinleet:
@mmiklavc Merged in master, fixed the incoming issues that would have caused problems, built successfully, and ran up full dev and made sure data flowed through.

justinleet:
Apparently I built skipping tests, I'll take a look when I have a chance

mmiklavc:
Thanks @justinleet! I'll take a look at this again.

mmiklavc:
fyi, I'm also seeing this failing unit test `shouldPopulateMessagesOnProcessMessage`. Looking into it - it's probably some flakiness with Mockito pattern matching on `when().then()` constructs.

justinleet:
@mmiklavc I merged in your PR against this branch. I'll add any commentary here as I review, but obviously feel encouraged to run through anything you want on here.

metron-platform/metron-parsing/metron-parsers-common/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
```diff
@@ -81,4 +81,11 @@
    */
   boolean validate(T message);
 
+  /**
+   * Provides a hook to override the default charset parsers use to read data.
+   * @return Charset to use for for reading
+   */
+  default Charset getReadCharset() {
+    return StandardCharsets.UTF_8;
```
justinleet: Should this potentially be a global configuration that defaults to `UTF_8`?, which is then overridden by parser specific charset?

I.e. if a user has all (or almost all) of their parsers in another charset, right now they'd have to set it for each and every parser. I'm not sure if this an actual or theoretical issue to be honest.

metron-platform/metron-parsing/metron-parsers-common/src/main/java/org/apache/metron/parsers/ParserRunnerImpl.java
```diff
@@ -255,7 +256,7 @@ private void initializeParsers(Supplier<ParserConfigurations> parserConfigSuppli
     if (!message.containsKey(Constants.GUID)) {
       message.put(Constants.GUID, UUID.randomUUID().toString());
     }
-    message.putIfAbsent(Fields.ORIGINAL.getName(), new String(rawMessage.getMessage()));
+    message.putIfAbsent(Fields.ORIGINAL.getName(), new String(rawMessage.getMessage(), StandardCharsets.UTF_8));
```
justinleet: This needs to care about the underlying parser's format, right? This is pulling the raw message, but that might not be `UTF_8`

metron-platform/metron-parsing/metron-parsers-common/src/main/java/org/apache/metron/parsers/json/JSONMapParser.java
```diff
@@ -202,7 +204,7 @@ public void init() {
       }
       return Collections.unmodifiableList(parsedMessages);
     } catch (Throwable e) {
-      String message = "Unable to parse " + new String(rawMessage) + ": " + e.getMessage();
+      String message = "Unable to parse " + new String(rawMessage, StandardCharsets.UTF_8) + ": " + e.getMessage();
```
justinleet: This needs to be the underlying parser charset, I think. Otherwise we'll be trying to decode with the wrong thing and not outputting the message in a useful manner.

This seems like it occurs a couple times in these messages, so make sure I'm not misinterpreting it.

metron-platform/metron-parsing/metron-parsers-common/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
mmiklavc: I made a compromise on not handling this also with a global option. As we had talked above about needing to change read charsets likely being an edge case, it seemed reasonable to add the capability to the parsers individually.

metron-platform/metron-parsing/metron-parsers-common/src/main/java/org/apache/metron/parsers/json/JSONMapParser.java
mmiklavc: I need to look into this a bit more. You're correct, that should be the readCharset also. I may be missing something, but our parser configure(...) method doesn't seem to get the high level parser configuration detail, e.g. all the options noted here that are siblings of the `parserConfig` key - https://github.com/apache/metron/tree/master/metron-platform/metron-parsing#parser-configuration. It seems better to make this option general - sibling to `parserConfig` rather than within `parserConfig`, but I don't think the API will allow it in its current form.

mmiklavc: Ok @justinleet - I posted an updated PR against your PR. Please review and see if that addresses this.

mmiklavc: Update - sorry, I made this comment on the wrong comment from you. I meant this regarding the parserrunnerimpl class.

mmiklavc: Both issues are now addressed. See https://github.com/justinleet/metron/pull/22

metron-platform/metron-parsing/metron-parsers-common/src/main/java/org/apache/metron/parsers/ParserRunnerImpl.java
mmiklavc: See https://github.com/justinleet/metron/pull/22

metron-platform/metron-parsing/metron-parsers-common/src/main/java/org/apache/metron/parsers/interfaces/MessageParser.java
justinleet: Sounds good. I'd honestly just forgotten that conversation existed at this point and missed it when I too quickly looked.

justinleet:
The PR from @mmiklavc (https://github.com/justinleet/metron/pull/22) got merged to address the couple `readCharset` issues.

justinleet:
@mmiklavc Unit test looks like it fails because a mocked parser NPEs on the `getReadCharset`. Could just do a `when(broParser.getReadCharset()).thenReturn(StandardCharsets.UTF_8);` to fix the the issue.

mmiklavc:
Well that's odd - apparently our version of Mockito doesn't call the provided concrete method for default interface methods. Submitting another fix now.

justinleet:
Merged. Adds some docs and the test fix.

@mmiklavc Assuming the build succeeds, are we good for @nickwallen + @ottobackwards to follow up on their reviews?

ottobackwards:
Ran this up in full-dev, everything seems to be working.

Ship it.  +1

justinleet:
@mmiklavc We should probably add a note to the Upgrading.md.  There is a backwards compatibility concern when upgrading:

When a current user's default Charset is not UTF_8, they'll need to update their parser configs to use the appropriate one, or parsing will run into issues.

mmiklavc:
I assume you mean if they were using the system default, which is Java's default when a charset is not provided when converting from bytearray to String? Just to be clear, this is an upgrade concern based off the fundamental nature of this PR change as a whole, not specifically the specific parser configs.

justinleet:
@mmiklavc If they're using a system default that isn't UTF-8 (say ISO_8859_1), all the parsers that currently work would silently start using the new default of UTF-8 and no longer be able to parse the data. Let me know if this understanding is incorrect.

For most, if not all, users this shouldn't be a concern, but I just want to make sure it's communicated.

mmiklavc:
That is correct @justinleet

justinleet:
@mmiklavc I've added an upgrading note. Let me know if I'm missing anything there.

As a reminder to whoever merges this in (admittedly probably me), it'll be important to manually squash the commits in order to maintain attribution for both Mike and myself.

mmiklavc:
lgtm. I'm a non-binding +1 (due to my contributions to the PR).

justinleet:
@nickwallen anything you need to see before we wrap up this PR?

nickwallen:
No, I am good.  Thanks for all the hard work.  Please proceed.

mmiklavc:
Just confirming, are you still +1 @ottobackwards with the recent Upgrading.md changes?

ottobackwards:
yup, let's get it out

PR #1343, opened 2019-02-19
METRON-2010 Unable to Build Metron Due to Inaccessible Repository

nickwallen:
It appears that the repository that we have configured to pull in the dependency `multiline-string` is no longer accessible.  I am unable to build Metron because of this.

* The repository at `https://raw.github.com/benelog/multiline/master/maven-repository` is no longer accessible. I am not sure how long this repository has been inaccessible.

* This likely went unnoticed for some period of time because the dependency is cached in most developer environments.  Once I cleared my Maven cache, I ran into this problem.

* The Multiline project's documentation indicates that the package should be pulled from the [JCenter repository](http://jcenter.bintray.com).  See the [Multiline project's documentation for Maven.](https://github.com/benelog/multiline/wiki/Maven-project-with-Eclipse#step-2--add-the-dependency-in-maven).

* We already have JCenter configured as a repository for the project.  We should be able to pull the dependency from there directly.

* I simply removed the inaccessible repository and everything builds as you would expect.  The `multiline-string` dependency is pulled from JCenter as you would expect.

## Testing

1. Clear your Maven cache at `~/.m2/repository/`.

1. Attempt to build Metron.

1. If you are able to build Metron, this PR has worked.

1. Optionally, deploy "Full Dev" and ensure that telemetry is consumed. (I validated this in "Full Dev".)

## Pull Request Checklist

### DIFF
--- a/metron-analytics/pom.xml
+++ b/metron-analytics/pom.xml
@@ -83,13 +83,6 @@
 			<url>http://clojars.org/repo</url>
 		</repository>
 		<repository>
-			<id>multiline-release-repo</id>
-			<url>https://raw.github.com/benelog/multiline/master/maven-repository</url>
-			<snapshots>
-				<enabled>false</enabled>
-			</snapshots>
-		</repository>
-		<repository>
 			<id>sonatype.org</id>
 			<url>https://oss.sonatype.org/content/groups/public/</url>
 			<snapshots>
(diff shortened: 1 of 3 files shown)


justinleet:
+1 by inspection. Thanks!

PR #1345, opened 2019-02-21
METRON-2012 Unable to Execute Stellar Functions Against HBase in the REPL

nickwallen:
When running any functions that attempt to access HBase from the REPL, an IllegalAccessError exception is thrown. This can be replicated with Stellar functions like ENRICHMENT_GET and PROFILE_GET that attempt to read from HBase.

To replicate, start the Stellar REPL with HBase and Zookeeper running and accessible.
  ```
  [root@node1 ~]# source /etc/default/metron
  [root@node1 ~]# cd $METRON_HOME
  [root@node1 0.7.1]# bin/stellar -z $ZOOKEEPER
  [Stellar]>>> ENRICHMENT_GET("example","192.168.1.1","example","E")
  2019-01-30 08:51:31 ERROR SimpleHBaseEnrichmentFunctions:251 - Unable to call exists: java.lang.IllegalAccessError: tried to access method com.google.common.base.Stopwatch.<init>()V from class org.apache.hadoop.hbase.zookeeper.MetaTableLocator
  org.apache.hadoop.hbase.DoNotRetryIOException: java.lang.IllegalAccessError: tried to access method com.google.common.base.Stopwatch.<init>()V from class org.apache.hadoop.hbase.zookeeper.MetaTableLocator
   at org.apache.hadoop.hbase.client.RpcRetryingCaller.translateException(RpcRetryingCaller.java:229)
   at org.apache.hadoop.hbase.client.RpcRetryingCaller.callWithRetries(RpcRetryingCaller.java:140)
   at org.apache.hadoop.hbase.client.HTable.get(HTable.java:879)
   at org.apache.hadoop.hbase.client.HTable.get(HTable.java:845)
   at org.apache.metron.enrichment.lookup.EnrichmentLookup$Handler.get(EnrichmentLookup.java:70)
   at org.apache.metron.enrichment.lookup.EnrichmentLookup$Handler.get(EnrichmentLookup.java:52)
   at org.apache.metron.enrichment.lookup.Lookup.get(Lookup.java:68)
   at org.apache.metron.enrichment.stellar.SimpleHBaseEnrichmentFunctions$EnrichmentGet.apply(SimpleHBaseEnrichmentFunctions.java:245)
   at org.apache.metron.stellar.common.StellarCompiler.lambda$exitTransformationFunc$13(StellarCompiler.java:664)
   at org.apache.metron.stellar.common.StellarCompiler$Expression.apply(StellarCompiler.java:259)
[...]

PR #1351, opened 2019-03-05
METRON-2026: Remove Storm dependency from metron-common

merrimanr:
## Contributor Comments
This PR includes the necessary changes to move Storm-dependent code to a separate module.  In most cases classes were moved unchanged with the exception of the package changing from `org.apache.metron.common.utils.*` to `org.apache.metron.storm.common.utils.*`.  More complex changes are listed in the next section.

### Changes Included

- Removal of Storm and Flux Maven dependencies from the metron-common pom file.
- Additional of the `org.ow2.asm:asm:5.0.3` Maven dependency to metron-common pom file.  This dependency was inherited from the Storm dependencies.
- Created a new metron-storm-common module for classes migrated out of metron-common.
- Refactored the `RawMessageUtil` and `MetadataUtil` classes to separate Storm dependent code.  The `MetadataUtil.extractMetadata` method's purpose is to extract metadata from a tuple and should be in metron-storm-common while the other methods in `MetadataUtil` I believe are reusable and should be kept in metron-common.  For now I moved `RawMessageUtil` to metron-storm-common and moved the method to `RawMessageUtil.extractMetadata`. 
- Removed the `TopologyContext` parameter from `BulkMessageWriter.init` and adjusted implementations to match.
- All `BulkMessageWriter` implementations did not use `TopologyContext` except for `HdfsWriter`.  This class is tightly coupled to Storm dependencies and will likely need to be rewritten without them for other streaming platforms.  In this PR, it's treated as a special case in `BulkMessageWriterBolt` with the `TopologyContext` call in a separate `HdfsWriter` method.
- Storm-dependent classes and tests were moved from metron-common to metron-storm-common including:
  - Classes in the `org.apache.metron.common.bolt` package
  - `MessageGetStrategy` interface and implementations
  - `RawMessageUtil` including the `extractMetadata` method

### Testing
I have done some basic testing in full dev.  I've spun it up and made some basic assertions:
- Data is flowing to ES and HDFS
- No topologies are reporting errors
- All components are showing green in Ambari
- REST is functioning including all endpoints
- UIs are working
- Streaming enrichments are working in Stellar (https://github.com/apache/metron/pull/1345)
- Streaming enrichments are working in the Storm enrichment topology

To test streaming enrichments in the enrichment topology:
1. Add an ip address that matches the bro data set to the 'user' enrichment set up in https://github.com/apache/metron/pull/1345:
```
[Stellar]>>> msgs := ["user1,192.168.66.1"]
[user1,192.168.66.1]
[Stellar]>>> KAFKA_PUT("user", msgs)
1
```
2. Add the enrichment to the bro enrichment config:
```
  "enrichment": {
    "fieldMap": {
      "geo": [
        "ip_dst_addr",
        "ip_src_addr"
      ],
      "host": [
        "host"
      ],
      "stellar": {
        "config": {
          "enrichment": "ENRICHMENT_GET('user', ip_src_addr, 'enrichment', 't')"
        }
      }
    },
    "fieldToTypeMap": {},
    "config": {}
  },
  "threatIntel": {
    "fieldMap": {
      "hbaseThreatIntel": [
        "ip_src_addr",
        "ip_dst_addr"
      ]
    },
    "fieldToTypeMap": {
      "ip_src_addr": [
        "malicious_ip"
      ],
      "ip_dst_addr": [
        "malicious_ip"
      ]
    },
    "config": {},
    "triageConfig": {
      "riskLevelRules": [],
      "aggregator": "MAX",
      "aggregationConfig": {}
    }
  },
  "configuration": {}
}' 'http://user:password@node1:8082/api/v1/sensor/enrichment/config/bro'
```
3. Verify there are no errors in the enrichment topology and that the new `enrichment` field appears in the bro index.

### Outstanding Issues
- I think we could improve on the `MetadataUtil` refactor.  Should we rename `RawMessageUtil` to something more appropriate?  Should we reorganize that code differently?
- Is the `HdfsWriter` approach good enough?  How else could we do this?
- Is `metron-platform/metron-storm-common` the right module for common Storm code?

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


mmiklavc:
@merrimanr - thanks for taking this on! These refactoring PR's can get pretty deep, and your work here is greatly appreciated. 

After a quick pass, I have a couple initial Q's/thoughts

1. I see the writer interface change to remove `TopologyContext topologyContext`. It doesn't look like we actually used that context anywhere explicitly, other than maybe the HDFSWriter. Seems like this was a pretty clean change for us in that regard. Is that the case, or can you call out any specific places where this may have caused drama?
2. `metron-storm-common` is a desirable split. What do you think about exploring following a similar packaging/module approach as was taken in `metron-parsing`. ie
```
* metron-common/
    * metron-common-storm/
    * metron-common-spark/ (upcoming)
```
I think that might organize things for us just a bit neater.

merrimanr:
The `HDFSWriter` class was the only special case.  Let me know if my explanation about how that was handled doesn't make sense or if you see a better way.

Sure I can move `metron-storm-common` there.  Are you thinking like this:
```
* metron-common/
    * metron-common/
    * metron-common-storm/
    * metron-common-spark/ (upcoming)
```

merrimanr:
Actually we can't have 2 modules with the same name.  Would it be `metron-common-common`?

mmiklavc:
@merrimanr - I believe a module can be an aggregator, a parent, and a jar-producing pom. So you could still keep it as:
```
* metron-common/
    * metron-common-storm/
    * metron-common-spark/ (upcoming)
```

It doesn't fit the layout of our other modules, however. What do you think of keeping metron-common by itself and creating another parent module for the streaming/platform code? e.g.

```
* metron-common/
* metron-common-streaming/
    * metron-common-storm/
    * metron-common-spark/ (upcoming)
```

merrimanr:
@mmiklavc latest commit updates the module structure to match your suggestion.  Let me know what you think.

metron-platform/metron-common-streaming/metron-common-storm/src/test/java/org/apache/metron/storm/common/bolt/ConfiguredEnrichmentBoltTest.java
```diff
@@ -64,15 +64,15 @@ public void reloadCallback(String name, ConfigurationType type) {
   public void setupConfiguration() throws Exception {
     TestingServer testZkServer = new TestingServer(true);
     this.zookeeperUrl = testZkServer.getConnectString();
-    byte[] globalConfig = ConfigurationsUtils.readGlobalConfigFromFile(TestConstants.SAMPLE_CONFIG_PATH);
+    byte[] globalConfig = ConfigurationsUtils.readGlobalConfigFromFile("../" + TestConstants.SAMPLE_CONFIG_PATH);
```
mmiklavc: Can we make all of these relative paths changes a local test constant? We could probably spend more time to make this less brittle, but for now I think it's a reasonable compromise. See https://github.com/apache/metron/blob/master/metron-platform/metron-enrichment/metron-enrichment-storm/src/test/java/org/apache/metron/enrichment/integration/EnrichmentIntegrationTest.java#L94, for example.

metron-platform/metron-writer/src/test/java/org/apache/metron/writer/hdfs/HdfsWriterTest.java
```diff
@@ -76,7 +76,8 @@ public void setup() throws IOException {
   public void testGetHdfsPathNull() {
     WriterConfiguration config = new IndexingWriterConfiguration(WRITER_NAME, new IndexingConfigurations());
     HdfsWriter writer = new HdfsWriter().withFileNameFormat(testFormat);
-    writer.init(new HashMap<String, String>(),createTopologyContext(), config);
+    writer.init(new HashMap<String, String>(), config);
+    writer.initFileNameFormat(createTopologyContext());
```
mmiklavc: Just for reference later - this is not part of the metron-common refactoring, but we should remove our Storm dependencies from the writer infrastructure as well.

mmiklavc:
@merrimanr Save for a small nit with the relative paths "../" duplicated in multiple places within individual tests, I think this is pretty solid. Also looks like a small conflict with metron-platform/pom.xml. Thanks for the effort on this!

You mentioned doing some light testing. Considering the classpath issues we've recently worked through, do you think it's worth running through this test script before we bring this in? https://github.com/apache/metron/pull/1368#issuecomment-477798665

metron-platform/metron-common-streaming/metron-common-storm/src/test/java/org/apache/metron/storm/common/bolt/ConfiguredEnrichmentBoltTest.java
merrimanr: Done

merrimanr:
I will spin this up in full dev again and run through testing.

merrimanr:
I ran through the testing instructions again and everything worked as expected.

mmiklavc:
@merrimanr - looks good. Just to confirm, did you run through the script from 1368? I'm +1 pending confirmation on that as we need to be extremely careful about regressions.

merrimanr:
I ran the tests again and did not notice any regressions.

mmiklavc:
Cool, just to be clear, +1 stands - go ahead and merge this.

PR #1356, opened 2019-03-13
METRON-2029: Configure Table should have filter

sardell:
## Contributor Comments
Link to original ASF ticket: https://jira.apache.org/jira/browse/METRON-2029?filter=-2

Previously, when a user opened the configure table panel, they would have to scroll down to see all the available column choices. This list can be really long and require a lot of scrolling and parsing for a user to find what they are looking for.

I added a simple filter so a user can more easily find the column they are looking for by either their original name or the display name they can optionally give a column.

## Testing
1. Clone this branch locally.
1. Navigate to `metron-interface/metron-alerts`, run `npm ci` and then `scripts/scripts/start-dev.sh`
1. Open the Alerts UI at localhost:4201.
1. Click on the gear icon to open the configure table panel.
1. Type in the filter to narrow the choices down.
1. To test the display name, add a display name to an existing column choice, click save at the bottom of the configure table panel, reopen the panel and try filtering with the term being the display name. It should be included in the results.
1. To clear the filter and reset the filtered results to include all available columns, click on the 'x' on the right hand side of the filter input.
1. You can find unit tests for this feature inside the `configure-table.component.spec.ts` file and run them with `ng test`.

Click on the gif below to see an expanded view of the filter in action.

filter-example

## Further Considerations

Currently, if a display name exists on a column, the filtering logic will only check against the display name. This was intentionally written on my part because I felt if a user renames a column with a display name, they are going to be searching for that column by display name.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

- n/a ~[ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?~

- n/a ~[ ] Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via `site-book/target/site/index.html`:~

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-alerts/src/app/alerts/configure-table/configure-table.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/configure-table/configure-table.component.ts
@@ -103,30 +103,18 @@ export class ConfigureTableComponent implements OnInit, AfterViewInit {
       });
   }
 
-  filterColumns(val) {
+  filterColumns(val: string) {
     const words = val.trim().split(' ');
     this.filteredColumns = this.allColumns.filter(col => {
-      if (col.displayName) {
-        if (this.colMissingFilterKeyword(words, col, true)) {
-          return false;
-        } else {
-          return true;
-        }
-      } else {
-        if (this.colMissingFilterKeyword(words, col)) {
-          return false;
-        } else {
-          return true;
-        }
-      }
+      return !this.isColMissingFilterKeyword(words, col, col.displayName);
     });
   }
 
-  colMissingFilterKeyword(words, col, displayName = false) {
+  isColMissingFilterKeyword(words: string[], col: ColumnMetadataWrapper, displayName?: string) {
     if (displayName) {
-      return words.map(word => col.displayName.toLowerCase().includes(word.toLowerCase())).includes(false);
+      return !words.every(word => col.displayName.toLowerCase().includes(word.toLowerCase()));
     } else {
-      return words.map(word => col.columnMetadata.name.toLowerCase().includes(word.toLowerCase())).includes(false);
+      return !words.every(word => col.columnMetadata.name.toLowerCase().includes(word.toLowerCase()));
     }
   }
 
--- a/metron-interface/metron-alerts/src/app/alerts/configure-table/configure-table.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/configure-table/configure-table.component.ts
@@ -116,7 +116,10 @@ export class ConfigureTableComponent implements OnInit, AfterViewInit {
     });
   }
 
-
+  clearFilter() {
+    this.filterColResults.nativeElement.value = '';
+    this.filteredColumns = this.allColumns;
+  }
 
   onSelectDeselectAll($event) {
     let checked = $event.target.checked;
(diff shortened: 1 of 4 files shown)


metron-interface/metron-alerts/src/app/alerts/configure-table/configure-table.component.ts
```diff
@@ -91,6 +95,46 @@ export class ConfigureTableComponent implements OnInit {
...
+    });
+  }
+
+  colMissingFilterKeyword(words, col, displayName = false) {
+    if (displayName) {
+      return words.map(word => col.displayName.toLowerCase().includes(word.toLowerCase())).includes(false);
+    } else {
+      return words.map(word => col.columnMetadata.name.toLowerCase().includes(word.toLowerCase())).includes(false);
+    }
+  }
```
ruffle1986: Well done. The only thing is (for me at least) that it took me a while to understand this part of the code.

So the code above basically checks whether there's any word in the words array that is missing from the col name. If so, it returns true which means yes, `col(is)MissingFilterKeyword`.  Amiright?

My suggestion is that we should make it more naturally readable.

First, I'd start the function with a verb to immediately make it clear that this function returns with a bool. Since we don't use type annotation here, and the function begins with a `map`, it's really difficult to notice it immediately. 

Function name starting with a verb:
`isColMissingFilterKeyword`

Secondly, at first glance, I thought this function returned an array because of the map function at the beginning of the line. In my opinion, it's more natural to read, but again, that is just my opinion.

```
colMissingFilterKeyword(words, col, displayName = false) {
  if (displayName) {
    return !words.every(word => col.displayName.toLowerCase().includes(word.toLowerCase()))
  } else {
    return !words.every(word => col.columnMetadata.name.toLowerCase().includes(word.toLowerCase()))
  }
}
```

In the example above it almost says "Not every word is included in col.displayName or col.columnMetadata.name"

metron-interface/metron-alerts/src/app/alerts/configure-table/configure-table.component.ts
```diff
@@ -91,6 +95,46 @@ export class ConfigureTableComponent implements OnInit {
...
+        }
+      } else {
+        if (this.colMissingFilterKeyword(words, col)) {
+          return false;
+        } else {
+          return true;
+        }
+      }
+    });
+  }
```
ruffle1986: It's way explicit, which is a good thing in general but in my opinion, we can save a few extra lines here:

```
filterColumns(val) {
  const words = val.trim().split(' ');
  this.filteredColumns = this.allColumns.filter(col => {
    return !this.colMissingFilterKeyword(words, col, col.displayName)
  });
}
```

metron-interface/metron-alerts/src/app/alerts/configure-table/configure-table.component.ts
sardell: I like both of these suggestions, especially the second one. Makes it much more readable. I'm going to update with these changes asap. Thanks for the feedback, @ruffle1986!

tiborm:
Changeset reviewed, tested locally. Thanks, @sardell! +1

nickwallen:
+1 by inspection.  Very useful feature!

PR #432, opened 2017-02-01
METRON-682: Unify and Improve the Flat File Loader

cestella:
Currently the flat file loader is deficient in a couple ways:
* It only supports importing local data despite there being a separate, poorly named, application which supports importing enrichment via MapReduce called threat_intel_loader.sh
* It does not support local imports from HDFS
* It does not support local imports from URLs
* It does not support importing zipped archives locally
* You cannot import more than one file at once

This JIRA will:
* Unify the MapReduce and local imports into one program and allow the user to specify the import mode with a CLI flag
* Support local imports from HDFS and URLs
* Support local imports from zipped files
* Support importing more than one file at once

Testing plan in comments

PR #1358, opened 2019-03-14
METRON-2038 Enrichment Loader Fails When Run as MR Job

nickwallen:
The enrichment loader fails when run as an MR job on YARN. It runs successfully when run in local mode.  See [METRON-2038](https://issues.apache.org/jira/browse/METRON-2038) for more information.

## Changes

#1345 unexpectedly changed how Maven packages the `metron-data-management` JAR even though  #1345 was completely unrelated to `metron-data-management`.

#### Why did #1345 impact `metron-data-management`?  

This is due to the problematic dependency on `metron-enrichment` from other projects like `metron-data-management`.  This can cause dependency changes in unrelated projects like `metron-profiler-client` to impact completely unrelated projects like `metron-data-management`.  This problematic dependency chain is in the process of being cleaned up as part of other ongoing work.

#### Why did the enrichment loader not work in MR, but work in local mode?

The problem occurs because Maven packaged `org.apache.hadoop.hbase.HBaseConfiguration` in the `metron-data-management` jar, even though HBase is marked as a `provided` dependency.  This really should not be packaged in the JAR, but I have been unable to determine why Maven was doing this.

When the enrichment loader is executed by the `flatfile_loader.sh` script, a large number of HBase dependencies are added at runtime.  The underlying MR code requires a different, conflicting version of this class. Unfortunately, it would find the conflicting `org.apache.hadoop.hbase.HBaseConfiguration` class definition in the `metron-data-management` jar rather than in its runtime dependencies. 

#### How was this fixed?

Rather than load the HBase dependency at runtime, now we package the HBase dependency in the JAR and do not add any of these dependencies at runtime.  This approach avoids the conflict and helps ensure that the integration tests run more similarly to how the tool is used in production.

## Testing

Follow the steps to replicate from [METRON-2038](https://issues.apache.org/jira/browse/METRON-2038) in Full Dev.  Ensure that the enrichments are loaded to the enrichments table in HBase.

## Pull Request Checklist

### DIFF
--- a/dependencies_with_url.csv
+++ b/dependencies_with_url.csv
@@ -194,6 +194,7 @@ com.google.guava:guava:jar:18.0:compile,ASLv2,
 com.google.inject.extensions:guice-servlet:jar:3.0:compile,ASLv2,
 com.google.inject:guice:jar:3.0:compile,ASLv2,
 com.googlecode.disruptor:disruptor:jar:2.10.4:compile,The Apache Software License, Version 2.0,http://code.google.com/p/disruptor/
+com.lmax:disruptor:jar:3.3.0:compile,The Apache Software License, Version 2.0,https://github.com/LMAX-Exchange/disruptor/
 com.lmax:disruptor:jar:3.3.2:compile,The Apache Software License, Version 2.0,https://github.com/LMAX-Exchange/disruptor/
 com.googlecode.json-simple:json-simple:jar:1.1:compile,The Apache Software License, Version 2.0,http://code.google.com/p/json-simple/
 com.googlecode.json-simple:json-simple:jar:1.1.1:compile,The Apache Software License, Version 2.0,http://code.google.com/p/json-simple/
@@ -301,6 +302,10 @@ org.jboss.resteasy:resteasy-jaxrs:jar:3.0.1.Final:compile,ASLv2,
 org.joda:joda-convert:jar:1.2:compile,Apache 2,http://joda-convert.sourceforge.net
 org.mortbay.jetty:jetty-util:jar:6.1.26:compile,ASLv2,
 org.mortbay.jetty:jetty:jar:6.1.26:compile,ASLv2,
+org.mortbay.jetty:jetty-sslengine:jar:6.1.26:compile,ASLv2,
+org.mortbay.jetty:jsp-2.1:jar:6.1.14:compile,ASLv2,
+org.mortbay.jetty:jsp-api-2.1:jar:6.1.14:compile,ASLv2,
+org.mortbay.jetty:servlet-api-2.5:jar:6.1.14:compile,ASLv2,
 org.noggit:noggit:jar:0.6:compile,Apache License, Version 2.0,http://github.com/yonik/noggit
 org.scannotation:scannotation:jar:1.0.3:compile,Apache License V2.0,http://scannotation.sf.net
 org.slf4j:log4j-over-slf4j:jar:1.6.6:compile,Apache Software Licenses,http://www.slf4j.org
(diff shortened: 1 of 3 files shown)


mmiklavc:
We should probably reference this test script for regression testing
https://github.com/apache/metron/pull/432#issuecomment-276733075

mmiklavc:
Ah, this seems like it might be related - https://issues.apache.org/jira/browse/MSHADE-267. I also saw that HBaseConfiguration is included in the uber jar. It looks like it's getting this from hbase-common, which is marked as provided, but may be sneaking in due to transitive dependencies of test jars that include it in compile scope.

mmiklavc:
@nickwallen - Thanks for tackling this. I happened to also see the rogue HBaseConfiguration along with a number of other deps that come from hbase-common getting pulled into the metron-data-management uber jar. Bootstrapping off of your findings, I thought it worthwhile to dig into this a bit more. It seemed like there might be a bug in the Maven shade plugin which would potentially cause us more trouble down the line and I wanted to see what we were in for. For starters, here's the class I'm using as a tracer bullet.

```
jar tvf /Users/mmiklavcic/.m2/repository/org/apache/metron/metron-data-management/0.7.1/metron-data-management-0.7.1.jar|grep HBaseConfiguration
  7195 Mon Mar 18 12:02:08 MDT 2019 org/apache/hadoop/hbase/HBaseConfiguration.class
```

I went through a similar series of inclusions and exclusions in the mdm pom.xml only to find no change in the rogue deps appearing. Also, regardless of their scope tag, I continued to see that classes in the uber jar. I got to the point where I had excluded all transitive dependencies for hbase-common and *still* was seeing the classes. Either there was a bug in the shade plugin or we were missing something.

The next step I took was to look over the dependency:tree from the CLI - all deps checked out accordingly. Default scope included by the Shade plugin? - yep, scope "provided" is not included by default. I also examined the build output from the Shade plugin - very clearly, there is no hbase-common dependency included in the shading. This is starting to feel like The Adventures of Buckaroo Banzai Across the 8th Dimension or The Adventures of Baron Munchausen. Either way, it's a trip.

Ok, let's go more brute force on this and explicitly copy dependencies. I looked at the compile scope dependencies first, but that didn't look quite right. I landed on runtime scope most closely representing what the Shade plugin bundles. I then iterated over all the jars looking for the config class.

```
mvn dependency:copy-dependencies -DoutputDirectory=/tmp/mikeout -DincludeScope=runtime
for jar in /tmp/mikeout/*.jar; do echo $jar; jar tvf $jar|grep HBaseConfiguration; done
...
/tmp/mikeout/metron-profiler-client-0.7.1.jar
  7161 Fri Mar 15 15:45:58 MDT 2019 org/apache/hadoop/hbase/HBaseConfiguration.class
...
```

Ok, so there it is. But why??? I looked over our dependency graph and see that metron-data-management depends on metron-enrichment depends on metron-profiler-client. Ok, there's the link, but what about hbase-common? metron-profiler-client has a dependency on hbase-common, scope=compile. Getting closer. The client's pom _also_ has a reference to the maven shade plugin and an assembly.xml. I explored dependencies on metron-profiler-client from a shell script and deployment perspective and did not find any circumstances where we would need to deploy this module stand-alone. I bumped the hbase-common dependency to provided (debatable) and removed all references to shading/relocating and a tarball assembly. I rebuilt the project and no longer see the HBaseConfiguration, or any other hbase-common classes for that matter, in the metron-data-management jar.

I'm currently undertaking testing from https://github.com/apache/metron/pull/432#issuecomment-276733075 as indicated above, but I think this should solve the root problem without requiring us to modify our license file or eliminate the core Hadoop libs from our shell script classpath.

mmiklavc:
As I worked through the full gamut of testing, all of the flatfile loader tests work, but it re-introduces the exception found in https://github.com/apache/metron/pull/1345. 

I would like to explore this a bit further in a separate PR to see if I can get this working, but for now I'm +1 on the current work.

nickwallen:
Wow - nice investigation.  I think we definitely need to clean-up what you've identified.  I think a follow-on for that effort will work.

PR #1365, opened 2019-03-25
METRON-2050: Automatically populate a list of enrichments from HBase

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-2050

We currently hard-code the list of enrichments available to the management UI within the REST API itself. This PR removes the hard-coded list and replaces it with a new approach using HBase coprocessors.

Testing and additional documentation to follow.

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


metron-platform/metron-hbase-server/src/main/java/org/apache/metron/hbase/coprocessor/EnrichmentCoprocessor.java
```diff
@@ -0,0 +1,198 @@
...
+    String zkUrl = config.get(ZOOKEEPER_URL);
+    if (null == zkUrl) {
+      throw new IllegalStateException(
+          "Enrichment coprocessor requires property '" + ZOOKEEPER_URL
+              + "' to be provided at startup.");
+    }
+    return zkUrl;
+  }
+
+  private GlobalConfigService getGlobalConfigService(String zkUrl) {
```
merrimanr: We have a `org.apache.metron.common.zookeeper.ZKConfigurationsCache` abstraction that does the same thing.  Does it make sense to reuse that here?  I wouldn't say it's a requirement because this is just a one time load but could be useful in the future if we do need to access the global config after a coprocessor is started.  Just something for you to consider.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
```diff
@@ -121,6 +121,14 @@
             <empty-value-valid>true</empty-value-valid>
         </value-attributes>
     </property>
+    <property>
```
merrimanr: Why do we have a property for the HDFS url?  Shouldn't we get this property from Ambari similar to how we get Zookeeper and Kafka urls?

metron-platform/metron-hbase/src/main/java/org/apache/metron/hbase/client/HBaseClient.java
```diff
@@ -302,4 +303,35 @@ private static String tableName(HTableInterface table) {
...
+  /**
+   * Scans an entire table returning all row keys as a List of Strings.
+   *
+   * <p>
+   * <b>**WARNING**:</b> Do not use this method unless you're absolutely crystal clear about the performance
+   * impact. Doing full table scans in HBase can adversely impact performance.
+   *
+   * @return List of all row keys as Strings for this table.
+   */
+  public List<String> readRecords() throws IOException {
```
merrimanr: Would it make sense to add some kind of configurable guard here or do you think a warning is enough?

merrimanr:
Great work so far.  Just a couple minor suggestions to consider before we start testing.

metron-platform/metron-hbase-server/src/main/java/org/apache/metron/hbase/coprocessor/EnrichmentCoprocessor.java
mmiklavc: I had considered that, but I wanted to be careful because, as you mentioned, this is a one time load and there could be adverse side effects without explicitly changing it to handle changes to the configured parameters. The other thing is that the ZKConfigurationsCache abstraction is tightly coupled to the concept of our topology types, e.g. 
* ENRICHMENT
* PARSER
* INDEXING
* PROFILER

Any reference to the global config has to be through one of those configuration types. There is probably some refactoring that could be done there if we wanted, but it felt a bit ham-fisted to shoehorn that in for what amounts to about 10 fairly concise lines of code.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
mmiklavc: It took me a bit of effort to figure out how we were doing that, actually. The way we grab the existing properties for Zookeeper and Kafka is via the service_advisor code. See:
* https://github.com/apache/metron/blob/96a66c88e7a62867c6287106a6646d198e2e6a81/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/service_advisor.py#L130
* https://github.com/apache/metron/blob/96a66c88e7a62867c6287106a6646d198e2e6a81/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/service_advisor.py#L142

You effectively grab the values at cluster install time and provide those as default recommended values for properties that we've defined in the MPack. Here is the path through that maze:
* https://github.com/apache/metron/blob/96a66c88e7a62867c6287106a6646d198e2e6a81/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml#L133
* https://github.com/apache/metron/blob/96a66c88e7a62867c6287106a6646d198e2e6a81/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/status_params.py#L111
* https://github.com/apache/metron/blob/96a66c88e7a62867c6287106a6646d198e2e6a81/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py#L157
* https://github.com/apache/metron/blob/96a66c88e7a62867c6287106a6646d198e2e6a81/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/service_advisor.py#L144

Defined in the env file, loaded into status_params first (so the status commands can use it), inherited by the params os-specific file (for use during install), modified during install by the service_advisor script.

Hopefully it makes sense why I did this (though not necessarily why this is all so esoteric)

metron-platform/metron-hbase/src/main/java/org/apache/metron/hbase/client/HBaseClient.java
mmiklavc: What were you thinking? I had considered what to do about this. One option was simply putting this in its own class only useable by the coprocessor, but I felt like that might be a bit too heavy handed and redundant. Adding an aggressive note was where I landed as a hopefully reasonable compromise, but I'm definitely open to suggestions.

metron-platform/metron-hbase-server/src/main/java/org/apache/metron/hbase/coprocessor/EnrichmentCoprocessor.java
merrimanr: As long as you are aware of that and made a conscious decision to not use it then it's fine with me.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
merrimanr: I did not realize you were populating that in service_advisor.py.  The zookeeper and kafka urls are actually populated in params_linux.py [here](https://github.com/apache/metron/blob/master/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py#L129) and [here](https://github.com/apache/metron/blob/master/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py#L129).  I'm not sure which approach is better or what the tradeoffs are.

metron-platform/metron-hbase/src/main/java/org/apache/metron/hbase/client/HBaseClient.java
merrimanr: I was thinking a configurable limit that would break out of the scan loop after a threshold is reached.  Looks like there would be more work involved though to pass in a config so I think a note is fine.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
mmiklavc: Hm, that's a good find - I did not know that existed. We also seem to be grabbing Zookeeper from [here](https://github.com/apache/metron/blob/master/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/service_advisor.py#L162) in the service_advisor.py for Solr. But that's apparently independent from our other zk url. Let me see if there's a way to grab HDFS in the same way. I prefer the approach you found.

mmiklavc:
FYI, I just rebased from the latest master to get the enrichment refactoring changes from https://github.com/apache/metron/pull/1368.

metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-env.xml
mmiklavc: @merrimanr - I played around with this a bit. The one benefit to the way I have it is that it allows the user to change the HDFS url if there's an issue with it for any reason - it will pull the default recommended value from the property `fs.defaultFS` on install. My main concern with this property at all is namenode HA. I can't exactly use the same approach as was done for storm and zookeeper because I can't get the namenode port, i.e.  `hdfs_url = default("/clusterHostInfo/namenode_host", [])` only gives me a host, no port, and there's no other property in clusterHostInfo to obtain that. I was, however, able to find a modification to the service_advisor style that allowed me to grab it from inside the params files, `config["configurations"]["core-site"]["fs.defaultFS"]`. That gives me the full url `hdfs://node1:8020` as desired. The shortcoming here is that there is no way to change this whatsoever without modifying the python files if there's an issue. But I believe this should still work with namenode HA just fine, per [https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/HDFSHighAvailabilityWithNFS.html](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/HDFSHighAvailabilityWithNFS.html)

> fs.defaultFS - the default path prefix used by the Hadoop FS client when none is given
> 
> Optionally, you may now configure the default path for Hadoop clients to use the new HA-enabled logical URI. If you used “mycluster” as the nameservice ID earlier, this will be the value of the authority portion of all of your HDFS paths. This may be configured like so, in your core-site.xml file:

```
<property>
  <name>fs.defaultFS</name>
  <value>hdfs://mycluster</value>
</property>
```

I guess I don't have a strong opinion either way - both can work. Exposing the property means that the user can change it if they need, but they will also need to manually manage it in cases where the namenode nameservice ID changes. Not exposing it means they cannot change it at all, however it will always pull the latest ID from any changes with namenode HA, etc. Which do you think is better? In light of the HA probably working fine with fs.defaultFS, I'm leaning towards not exposing the property.

merrimanr: I agree, I don't think we should expose the property.  I think we should go with an approach where Ambari can consistently and reliably populate that setting with or without namenode HA enabled.  As long as that condition is satisfied, I'm good with how you choose to do it.

mmiklavc: @merrimanr - thanks for talking through this - I decided get rid of it in favor of the backend approach. Testing now and getting the test script together for you to start looking at as well.

mmiklavc:
# Testing Plan

We need to verify

1. The enrichment coprocessor loads as expected
1. Normal sensor data flow through the system, from parsers through to indexing, still functions as normal
1. The new enrichment list table is populated when new enrichment types are added to the enrichment HBase table

## TOC

* Setup Test Environment
* Verify Basics
* Flatfile loader
* Streaming enrichment
* Final check

## Setup Test Environment

1. Build full dev `metron/metron-deployment/development/centos6$ vagrant up`
1. Login to full dev `ssh root@node1`, password "vagrant"
1. Set some environment variables
    ```
    # the root and metron users will need to do this - add to the user's ~/.bashrc or source each time you switch to the user
    source /etc/default/metron
    ```

## Verify Basics

### HBase enrichment table setup with coprocessor

1. Run the following command from the CLI - you should see the coprocessor in the table attributes. Ambari should set this up as part of the MPack installation.
    ```
    # echo "describe 'enrichment'" | hbase shell
    HBase Shell; enter 'help<RETURN>' for list of supported commands.
    Type "exit<RETURN>" to leave the HBase Shell
    Version 1.1.2.2.6.5.1050-37, r897822d4dd5956ca186974c10382e9094683fa29, Tue Dec 11 
    02:04:10 UTC 2018

    describe 'enrichment'
    Table enrichment is ENABLED
    enrichment, {TABLE_ATTRIBUTES => {METADATA => {'Coprocessor$1' => 
    'hdfs://node1:8020/apps/metron/coprocessor/metron-hbase-server-0.7.1-uber.jar|org.apache.metron.hbase.coprocessor.EnrichmentCoprocessor||zookeeperUrl=node1:2181'}
    }
    COLUMN FAMILIES DESCRIPTION
    {NAME => 't', BLOOMFILTER => 'ROW', VERSIONS => '1', IN_MEMORY => 'false', KEEP_DELETED_CELLS => 'FALSE', DATA_BLOCK_ENCODING => 'NONE', TTL => 'FOREVER', COMPRESSION => 'NONE', MIN_VERSIONS => '0', BLOCKCACHE => 'true', BLOCKSIZE => '65536', REPLICATION_SCOPE => '0'}
    1 row(s) in 0.3790 seconds 
    ```

### Pipeline still processes to indexing

Verify data is flowing through the system, from parsing to indexing

1. Open Ambari and navigate to the Metron service http://node1:8080/#/main/services/METRON/summary
2. Open the Alerts UI
3. image
4. Verify alerts show up in the main UI - click the search icon (you may need to wait a moment for them to appear)
image
5. Head back to Ambari and select the Kibana service http://node1:8080/#/main/services/KIBANA/summary
6. Open the Kibana dashboard via the "Metron UI" option in the quick links
7. image
8. Verify the dashboard is populating
9. image

## Flatfile loader

### Preliminaries

1. Before we start adding enrichments, let's verify the enrichment_list table is empty
1. Go to Swagger
image
1. Click the `sensor-enrichment-config-controller` option.
image
1. Click the `GET /api/v1/sensor/enrichment/config/list/available/enrichments` option.
1. And finally click the "Try it out!" button. You should see an empty array returned in the response body.
image
1. Now, let's perform an enrichment load. We'll do this as the metron user
    ```
    su - metron
    source /etc/default/metron
    ```

1. Download the alexa 1m dataset:
    ```
    wget http://s3.amazonaws.com/alexa-static/top-1m.csv.zip
    unzip top-1m.csv.zip
    ```
1. Stage import file
    ```
    head -n 10000 top-1m.csv > top-10k.csv
    # plop it on HDFS
    hdfs dfs -put top-10k.csv /tmp
    ```
1. Create an extractor.json for the CSV data by editing `extractor.json` and pasting in these contents:
    ```
    {
      "config": {
        "columns": {
          "domain": 1,
          "rank": 0
        },
        "indicator_column": "domain",
        "separator": ",",
        "type": "alexa"
      },
      "extractor": "CSV"
    }
    ```

The extractor.json will get used by flatfile_loader.sh in the next step

### Import from HDFS via MR

```
# truncate hbase
echo "truncate 'enrichment'" | hbase shell
# import data into hbase 
$METRON_HOME/bin/flatfile_loader.sh -i /tmp/top-10k.csv -t enrichment -c t -e ./extractor.json -m MR
# count data written and verify it's 10k
echo "count 'enrichment'" | hbase shell
```

You should see a 10k count in the enrichment table. We'll add one more source of enrichment type before checking our enrichment list.

## Streaming Enrichment

1. Switch back to root if you're still the metron user.
    ```
    [metron@node1 ~]$ exit
    ```

1. Pull down latest config from Zookeeper
    ```
    $METRON_HOME/bin/zk_load_configs.sh -m PULL -o ${METRON_HOME}/config/zookeeper -z $ZOOKEEPER -f
    ```

1. Create a file named `user.json` in the parser directory.
    ```
    touch ${METRON_HOME}/config/zookeeper/parsers/user.json
    ```

1. Enter these contents:
    ```
    {
      "parserClassName" : "org.apache.metron.parsers.csv.CSVParser" ,
      "writerClassName" : "org.apache.metron.writer.hbase.SimpleHbaseEnrichmentWriter",
      "sensorTopic":"user",
      "parserConfig": {
        "shew.table" : "enrichment",
        "shew.cf" : "t",
        "shew.keyColumns" : "ip",
        "shew.enrichmentType" : "user",
        "columns" : {
          "user" : 0,
          "ip" : 1
        }
      }
    }
    ```

1. Push the changes back up to Zookeeper
    ```
    $METRON_HOME/bin/zk_load_configs.sh -m PUSH -i $METRON_HOME/config/zookeeper/ -z $ZOOKEEPER
    ```

1. Create the user Kafka topic
    ```
    ${HDP_HOME}/kafka-broker/bin/kafka-topics.sh --create --zookeeper $ZOOKEEPER --replication-factor 1 --partitions 1 --topic user
    ```

1. Start the topology
    ```
    ${METRON_HOME}/bin/start_parser_topology.sh -s user -z $ZOOKEEPER
    ```

1. Create a simple file with named `user.csv` with user mapping to IP, e.g.
    ```
    echo "mmiklavcic,192.168.138.158" > user.csv
    ```

1. Push the data to Kafka
    ```
    tail user.csv | ${HDP_HOME}/kafka-broker/bin/kafka-console-producer.sh --broker-list $BROKERLIST --topic user
    ```

1. Verify data makes it to the enrichment table.
    ```
    echo "count 'enrichment'" | hbase shell
    ```

    There should be 10,001 records now.

### Final check

1. Check the Swagger UI again with our earlier steps. You should now see an "alexa" and a "user" enrichment type returned in the enrichment list results
image

merrimanr:
I ran through the test plan and everything works as expected.  This is a really nice addition.  +1

mmiklavc:
@merrimanr - as I reviewed your global config doc change PR (https://github.com/apache/metron/pull/1376), I went to double check that I hadn't missed anything in this PR and ended up tweaking the docs a bit. Can you look over my latest changes when you have a moment and confirm whether your +1 stands?

merrimanr:
Latest change looks good to me.  My +1 stands.

mmiklavc:
Gah, looks like my recent metron-common changes from master conflict. Fixing.

mmiklavc:
Ok, latest commit fixes merge conflict with master - metron-common README improvements from https://github.com/apache/metron/pull/1376. I checked the site-book again just to be sure and the links all work.

merrimanr:
I check the metron-common README and everything looks ok. +1

PR #1367, opened 2019-03-26
METRON-2022: Metron rest creates large number of connections to ZK which causes subsequent connection to zk fail

merrimanr:
## Contributor Comments
I was able to track down the problem in the `SolrMetaAlertRetrieveLatestDao` class.  Every time the latest meta alert is retrieved in `SolrMetaAlertRetrieveLatestDao.getLatest`, a new `SolrClient` is being created and never closed.  This issue also happens when updating a meta alert because the latest version of the meta alert is retrieved first.

To fix this, the `SolrClient` created in `SolrDao` is now shared with all child Dao classes.  Some slight refactoring was needed to achieve this and I tried to keep it as consistent with other `SolrDao` classes as possible.

### Testing
1. Spin up full dev and create a meta alert:
```
  "alerts": [
    {
      "guid": "f130166c-ba0d-4ecd-8395-eaea8aa950fe",
      "sensorType": "bro"
    }
  ],
  "groups": [
    "string"
  ]
}' 'http://user:password@node1:8082/api/v1/metaalert/create'
``` 
2. Record the number of active zookeeper connections by first getting the REST PID:
```
[root@node1 tmp]# ps -ef | grep metron-rest
metron   18151     1  7 00:59 ?        00:01:36 /usr/jdk64/jdk1.8.0_112/bin/java -Dhdp.version=2.6.5.0-292 -cp /etc/hadoop/conf/:/usr/hdp/current/hbase-client/conf:/usr/metron/0.7.1/lib/metron-rest-0.7.1.jar:/usr/metron/0.7.1/lib/metron-parsing-storm-0.7.1-uber.jar:/usr/metron/0.7.1/lib/metron-solr-0.7.1-uber.jar org.apache.metron.rest.MetronRestApplication --spring.config.location=/usr/metron/0.7.1/config/rest_application.yml,classpath:/application.yml --server.port=8082 --spring.profiles.active=dev --index.dao.impl=org.apache.metron.solr.dao.SolrDao,org.apache.metron.indexing.dao.HBaseDao --meta.dao.impl=org.apache.metron.solr.dao.SolrMetaAlertDao --index.writer.name=solr
root     20411   566  0 01:21 pts/0    00:00:00 grep metron-rest
```
Then print the number of connections for that PID:
```
[root@node1 tmp]# netstat -nape | awk '{if ($5 == "::ffff:127.0.0.1:2181") print $9;}' | sort | uniq -c | grep 18151
      2 18151/java
```
3. Update the created meta alert `alert_status` field  several times with different values:
```
  "guid": "f8da66bc-0d26-4f90-9d93-4689e70e9ce3",
  "index": "metaalert",
  "patch": [
   {"op":"add","path":"/alert_status","value":"DISMISSED"}
  ],
  "sensorType": "metaalert"
}' 'http://user:password@node1:8082/api/v1/update/patch'
```
4. Printing the number of zookeeper connections should continue to be 2.  Before this the number of connections would continue to grow as updates are sent.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
@@ -59,15 +59,16 @@ public class SolrClientFactory {
    * Closes the SolrClient connection and releases the reference.
    */
   public static void close() {
-    if (solrClient != null) {
-      try {
-        solrClient.close();
-      } catch (IOException e) {
-        LOG.error(e.getMessage(), e);
-      } finally {
-        solrClient = null;
+    synchronized (SolrClientFactory.class) {
+      if (solrClient != null) {
+        try {
+          solrClient.close();
+        } catch (IOException e) {
+          LOG.error(e.getMessage(), e);
+        } finally {
+          solrClient = null;
+        }
       }
-
     }
   }
 
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
@@ -1,0 +1,35 @@
+package org.apache.metron.solr.client;
+
+import com.google.common.base.Splitter;
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.client.solrj.impl.CloudSolrClient;
+
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.metron.solr.SolrConstants.SOLR_ZOOKEEPER;
+
+/**
+ * Factory for creating a SolrClient.  The default implementation of SolrClient is CloudSolrClient.
+ */
+public class SolrClientFactory {
+
+  /**
+   * Creates a SolrClient.
+   * @param globalConfig Global config
+   * @return SolrClient
+   */
+  public static SolrClient create(Map<String, Object> globalConfig) {
+    return new CloudSolrClient.Builder().withZkHost(getZkHosts(globalConfig)).build();
+  }
+
+  /**
+   * Retrieves zookeeper hosts from the global config and formats them for CloudSolrClient instantiation.
+   * @param globalConfig Global config
+   * @return A list of properly formatted zookeeper servers
+   */
+  protected static List<String> getZkHosts(Map<String, Object> globalConfig) {
+    return Splitter.on(',').trimResults()
+            .splitToList((String) globalConfig.getOrDefault(SOLR_ZOOKEEPER, ""));
+  }
+}
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
@@ -21,7 +21,11 @@ package org.apache.metron.solr.client;
 import com.google.common.base.Splitter;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.impl.CloudSolrClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
 import java.util.List;
 import java.util.Map;
 
@@ -32,13 +36,39 @@ import static org.apache.metron.solr.SolrConstants.SOLR_ZOOKEEPER;
  */
 public class SolrClientFactory {
 
+  private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private static SolrClient solrClient;
+
   /**
    * Creates a SolrClient.
    * @param globalConfig Global config
    * @return SolrClient
    */
   public static SolrClient create(Map<String, Object> globalConfig) {
-    return new CloudSolrClient.Builder().withZkHost(getZkHosts(globalConfig)).build();
+    if (solrClient == null) {
+      synchronized (SolrClientFactory.class) {
+        if (solrClient == null) {
+          solrClient = new CloudSolrClient.Builder().withZkHost(getZkHosts(globalConfig)).build();
+        }
+      }
+    }
+    return solrClient;
+  }
+
+  /**
+   * Closes the SolrClient connection and releases the reference.
+   */
+  public static void close() {
+    if (solrClient != null) {
+      try {
+        solrClient.close();
+      } catch (IOException e) {
+        LOG.error(e.getMessage(), e);
+      } finally {
+        solrClient = null;
+      }
+
+    }
   }
 
   /**
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.apache.metron.solr.client;
 
 import com.google.common.base.Splitter;
--- a/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
+++ b/metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
@@ -17,9 +17,6 @@
  */
 package org.apache.metron.solr.dao;
 
-import static org.apache.metron.solr.SolrConstants.SOLR_ZOOKEEPER;
-
-import com.google.common.base.Splitter;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.util.List;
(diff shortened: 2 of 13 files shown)


metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
```diff
@@ -168,18 +168,21 @@ public Document removeCommentFromAlert(CommentAddRemoveRequest request, Document
...
+    return this.client;
   }
 
   /**
    * Builds a Solr client using the ZK hosts specified.
    * @return SolrClient
    */
-  public SolrClient getSolrClient(List<String> zkHosts) {
+  protected SolrClient getSolrClient(List<String> zkHosts) {
     return new CloudSolrClient.Builder().withZkHost(zkHosts).build();
```
mmiklavc: Why would we ever want to create new clients from within a DAO class? Can we get rid of this altogether and/or get the client creation code into a proper factory? I think having `getSolrClient()` return an existing client and `getSolrClient(List<String> zkHosts)` opens us up to a relapse of this problem.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrMetaAlertRetrieveLatestDao.java
```diff
@@ -36,8 +37,10 @@
     MetaAlertRetrieveLatestDao {
 
   private SolrDao solrDao;
+  private SolrClient solrClient;
 
-  public SolrMetaAlertRetrieveLatestDao(SolrDao solrDao) {
+  public SolrMetaAlertRetrieveLatestDao(SolrClient client, SolrDao solrDao) {
```
mmiklavc: I like this approach of constructor dep injection for all DAOs. Independent of a larger refactoring effort, I think we should standardize on at least passing in a solrclient via constructor for all DAOs. We really shouldn't be passing around connection/client objects like stolen candy between them. If we're decorating/composing objects, then let's wire this up via constructors.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrMetaAlertUpdateDao.java
```diff
@@ -59,12 +59,13 @@
...
    */
-  public SolrMetaAlertUpdateDao(SolrDao solrDao,
+  public SolrMetaAlertUpdateDao(SolrClient solrClient,
+          SolrDao solrDao,
       SolrMetaAlertSearchDao metaAlertSearchDao,
       SolrMetaAlertRetrieveLatestDao retrieveLatestDao,
       MetaAlertConfig config) {
     super(solrDao, retrieveLatestDao, config);
-    this.solrClient = solrDao.getSolrClient(solrDao.getZkHosts());
+    this.solrClient = solrClient;
```
mmiklavc: Thumbs up

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/integration/SolrMetaAlertIntegrationTest.java
```diff
@@ -113,12 +114,12 @@ protected String getDefaultSourceTypeField() {
...
 
-
+    SolrClient solrClient = solrDao.getSolrClient();
     SolrMetaAlertSearchDao searchDao = new SolrMetaAlertSearchDao(
-        solrDao.getSolrClient(solrDao.getZkHosts()),
+        solrClient,
         solrDao.getSolrSearchDao(), config);
-    SolrMetaAlertRetrieveLatestDao retrieveLatestDao = new SolrMetaAlertRetrieveLatestDao(solrDao);
-    SolrMetaAlertUpdateDao updateDao = new SolrMetaAlertUpdateDao(solrDao, searchDao,
+    SolrMetaAlertRetrieveLatestDao retrieveLatestDao = new SolrMetaAlertRetrieveLatestDao(solrClient, solrDao);
```
mmiklavc: Yeah, I think this is much clearer. Then we don't have to dig through a morass of code just to figure out who is creating what client objects and who is secretly depending on them under the hood.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
merrimanr: I moved the SolrClient creation to a factory.  This should closely match the way we do it for the `ElasticsearchDao` classes.  The `SolrDao` and `SolrMetaAlertDao` are created separately in `IndexConfig` so I had to keep the `SolrDao.getSolrClient` method.  Now it only returns the current client and throws an exception if `SolrDao` hasn't been initialized.

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/client/SolrClientFactoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static org.apache.metron.solr.SolrConstants.SOLR_ZOOKEEPER;
+import static org.junit.Assert.assertEquals;
+
+public class SolrClientFactoryTest {
+
+  @Test
+  public void testGetZkHostsSingle() {
+    Map<String, Object> globalConfig = new HashMap<String, Object>() {{
+      put(SOLR_ZOOKEEPER, "   zookeeper:2181   ");
```
mmiklavc: Are the spaces intentional?

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrMetaAlertDao.java
```diff
@@ -144,10 +144,11 @@ protected String getDefaultSourceTypeField() {
       }
     };
 
-    SolrClient solrClient = solrDao.getSolrClient(solrDao.getZkHosts());
+    SolrClient solrClient = solrDao.getSolrClient();
```
mmiklavc: I'm still confused as to why we are passing the solr client around in this fashion. I thought Spring is smart enough to create singletons so that all the DAO constructors can take a SolrClient arg. See note on https://github.com/apache/metron/pull/1367/files#r269692863

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/client/SolrClientFactoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+    List<String> actual = SolrClientFactory.getZkHosts(globalConfig);
+    List<String> expected = new ArrayList<>();
+    expected.add("zookeeper:2181");
+    assertEquals(expected, actual);
+  }
+
+  @Test
+  public void testGetZkHostsMultiple() {
+    Map<String, Object> globalConfig = new HashMap<String, Object>() {{
+      put(SOLR_ZOOKEEPER, "   zookeeper:2181    ,   zookeeper2:2181    ");
```
mmiklavc: Same Q about the spaces.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrMetaAlertDao.java
justinleet: Well, as the original implementer, the technical reasoning is that I didn't know better. There's no compelling reason to leave it this way.

metron-platform/metron-solr/src/test/java/org/apache/metron/solr/client/SolrClientFactoryTest.java
merrimanr: I did not change this test, just moved it.  I imagine the test is ensuring correct parsing even when spaces are present.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrMetaAlertDao.java
merrimanr: The `SolrDao` and `SolrMetaAlertDao` are created separately through reflection in the REST `IndexConfig` class.  Solr is a runtime dependency so the `SolrClient` instance cannot be created in REST.  There must be some way for the client to be shared between them.  The `getSolrClient` is simply returning an existing client (not creating anything) so I feel it's pretty safe.  I'm happy to move it to the factory class if you think that makes more sense.

mmiklavc: Per @justinleet's comments, I think we should get rid of `getSolrClient()` from the SolrDao. It's not "dangerous" specifically in this case, but it makes it harder for someone else to come along later and understand where the client connections are coming from, how they're being created, and how many there will be. We just need to share a single instance at construction time, or per the issue you cited regarding runtime deps, inject singletons.

The real solution here is to refactor this using traditional GOF patterns and simply use the abstract factory pattern as that's exactly its purpose - https://en.wikipedia.org/wiki/Abstract_factory_pattern. I think it's arguably out of scope for a bug fix if you don't want to do it, but that's my cents on what the best approach would be. Mind you, the DAO combining part might need some thought, but I suspect we can simplify it just fine.

merrimanr: The latest commit removes the `getSolrClient` method from the `SolrDao` class.  The SolrClient instance is now managed by the `SolrClientFactory` class.  I had to also add a `close` method to close the connection and release the reference so that a new SolrClient can be instantiated as needed.  This was necessary for integration tests because multiple in memory Solr servers are spun up in the same JVM and a new client is needed for each.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+    return solrClient;
+  }
+
+  /**
+   * Closes the SolrClient connection and releases the reference.
+   */
+  public static void close() {
+    if (solrClient != null) {
```
mmiklavc: Probably need the synchronized here as well.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
```diff
@@ -168,29 +165,14 @@ public Document removeCommentFromAlert(CommentAddRemoveRequest request, Document
   }
 
   /**
-   * Builds a Solr client using the ZK hosts from the global config.
+   * Returns the SolrClient.
    * @return SolrClient
    */
   public SolrClient getSolrClient() {
```
mmiklavc: Wasn't this method supposed to go way?

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/client/SolrClientFactory.java
merrimanr: Done.

metron-platform/metron-solr/src/main/java/org/apache/metron/solr/dao/SolrDao.java
merrimanr: Done.

mmiklavc:
@merrimanr - the final round of changes look good. I think we're making some good iterative progress on refactoring the DAO layer. +1 by inspection pending the Travis build passing.

PR #1374, opened 2019-04-04
METRON-2061: Solr documents with date fields cannot be updated with Dao classes

merrimanr:
## Contributor Comments
The initial purpose of this PR is to explain the cause of the problem described in https://issues.apache.org/jira/browse/METRON-2061 and facilitate discussion on an optimal solution.  A solution is presented here but I expect we will explore other solutions as well.  

### Problem

Date type fields are a problem for our `UpdateDao` classes because of the logic used to update documents.  Currently the process is this:

1. Get the latest document using either the Elasticsearch or Solr client API.  This document is returned as a Java object.
2. Convert both the document object and patch object to a `JsonNode` object using Jackson.  This includes serializing/deserializing the objects internally.
3. Apply the patch using the `JsonPatch` API.
4. Convert the patched `JsonNode` back to a document object.
5. Index the document object into Elasticsearch or Solr.

The problem is that the date type is lost during serialization.  Jackson automatically converts it to epoch long by default: https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/SerializationFeature.html#WRITE_DATES_AS_TIMESTAMPS.  There is also an option of converting the value to a date string.  The error happens because the original date field now contains a long when a document is indexed.

### Possible Solutions
I think there are several ways to solve this problem and I will present some options here.  Others are welcome to propose additional solutions.

1. Add special handling for date fields - We could detect date fields and handle the conversion.  We would need to manage the Jackson date format and ensure it lines up with the date format Elasticsearch or Solr expects.
2. Handle the patch with a custom utility (the initial solution in this PR) - The JSON Patch specification (https://tools.ietf.org/html/rfc6902) is fairly simple and straightforward.  Instead of using a 3rd party library that requires `JsonNode` objects and serialization/deserialization, we could apply the patches directly to the objects.
3. Partial updates - I believe both Elasticsearch and Solr provide support for partial updates.  In this case we would only need to handle date type fields when those specific fields are being updated.

I chose option #2 because it is simple and avoids any future document field type issues.  I believe we should move away from the `JsonPatch` API and serialization/deserialization altogether because there is potential for other Jackson issues and it is inefficient.  This rules out option #1 for me.  Option #3 is attractive for obvious reasons but may require significant changes to our API.

### Testing
The initial solution included here has been tested in full dev.  We can use this test script regardless of the solution we choose.

1. Spin up full dev and enable Solr using the instructions in the metron-solr README.
2. Stop the Storm topologies.  This will make it easier to isolate our test to a single document.
3. Pick a schema and add a date type field to it:
```
<field name="timestamp_solr" type="iso_timestamp" indexed="true" stored="true" />
<fieldType name="iso_timestamp" stored="true" indexed="true" multiValued="false" class="solr.TrieDateField" sortMissingLast="false" docValues="true"/>
```
4. Clear out the collection corresponding to the changed schema and index a document with a `timestamp_solr` value:
```
{
  "guid":"some-guid",
  "source.type":"yaf",
  "timestamp_solr":"2019-04-04T00:00:00Z"
}
```
I did this using the "Documents" menu item in the Solr UI at `http://node1:8983`.

5. Attempt to patch an unrelated field:
```
  "guid": "some-guid",
  "patch": [
    {"op":"add","value":"DISMISS","path":"/alert_status"}
  ],
  "sensorType": "yaf"
}' 'http://user:password@node1:8082/api/v1/update/patch'
```
This request should succeed without error and the updated document should be returned with the new field.  Before this PR a 500 error was returned with a message similar to:
```
Error from server at http://node1:8983/solr/yaf: Invalid Date String:'75231198000'
```
6. This can also be tested in the Alerts UI by filtering on the guid and changing the status in the details panel.

### Outstanding Issues
Once we land on a solution I will add comprehensive tests and proper Java documentation.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


mmiklavc:
How is this DAO and jsonpatch being used by the meta-alert infrastructure? This manual test is picking a fairly narrow interpretation of all that can be done with JsonPatch, and I'm unclear whether we're introducing a feature regression here or not. I know we're using json patch (separately) for handling Ambari modifications to global config to ensure we don't trample user-added or other type of updates that aren't managed by Ambari. I'm concerned about what's not covered by switching from json patch to our own private version of it, specifically here. Very likely, the existing unit and integration tests are only verifying some basic functionality wrt the jsonpatch library (we are depending on the jsonpatch project to handle the fine grained testing of all the RFC details for us), so if we're going to replace it I think we need to do one of the following:
1. (preferred) Own all of the test scenarios as defined in https://tools.ietf.org/html/rfc6902 to ensure we meet the spec and do not introduce any regressions.
2. Modify the documentation, whether in our REST API implementation details for UI devs or in user-facing docs, to be explicit about what patching functionality we are or are not planning to support.

If we are walking back functionality by choosing option 2, I think this needs to be accompanied with a proper DISCUSS thread and a feature change notice in the release/upgrading notes. In addition, we would need to:
1. Expand the manual test cases to include any existing features in the UI to ensure we we haven't removed or altered what is expected from RFC 6902
2. Add additional unit and integration tests to cover cases not already handled by the automated testing framework.

mmiklavc:
Looks like this also needs conflict resolution. And can you mark any checklist items currently left empty as "n/a" rather than leaving the brackets `[ ]` unchecked so it's clear what is not going to be done vs what is left remaining as a TODO before wrapping up the PR?

merrimanr:
I was under the impression all checkboxes need to be done before a PR is complete.  You can assume an empty bracket means it's pending.

To be clear, the global config patch is unaffected by this change.  The only thing affected is the REST Patch endpoint.

I'm assuming you are ok with the solution I chose (number 2 in the PR description)?  I agree that your suggestion (number 1) is ideal.  Think the only thing missing from https://tools.ietf.org/html/rfc6902 is the "test" operation which is easy enough to add.  I will proceed with writing tests for all scenarios in https://tools.ietf.org/html/rfc6902.

mmiklavc:
> I was under the impression all checkboxes need to be done before a PR is complete. You can assume an empty bracket means it's pending.

Sort of - sometimes they're not relevant and it helps to remove ambiguity:
1. doc fix will not have integration tests, manual testing in full dev, test script
2. code fix may or may not require manual validation in full dev or documentation depending on the scope of what's being done
3. If you haven't added new dependencies, it's also n/a
4. The changes may not contain any new or updates to existing architecture diagrams

mmiklavc:
> To be clear, the global config patch is unaffected by this change. The only thing affected is the REST Patch endpoint.
Yes, I'm clear on this.
 
> I'm assuming you are ok with the solution I chose (number 2 in the PR description)? I agree that your suggestion (number 1) is ideal. Think the only thing missing from https://tools.ietf.org/html/rfc6902 is the "test" operation which is easy enough to add. I will proceed with writing tests for all scenarios in https://tools.ietf.org/html/rfc6902.

Yes, that sounds reasonable to me.

merrimanr:
Latest commit adds support for the "test" operation and unit tests for `PatchUtils`.  I believe it's 100% covered but happy to add more tests if there are cases I missed.

mmiklavc:
@merrimanr your updates look pretty good. The only minor test addition I think we need here is for paths nested more than depth 1. e.g. can you add tests to validate something like "/foo/bar/baz/blah" still works?

merrimanr:
Latest commit adds a test for nested paths.

mmiklavc:
Looks good @merrimanr. +1

PR #1380, opened 2019-04-10
METRON-1253: Manual pasting of timestamps into the timestamp picker

ruffle1986:
## Contributor Comments

Link to the original ASF Jira ticket: https://issues.apache.org/jira/browse/METRON-1253

Currently, it's not possible to write or paste a valid date to a date selector powered input field. This PR has been intended to solve this inconvenience. Please make sure you add a valid date format which is `YYYY-MM-DD HH:mm:ss`

Testing locally
==========

1. Start the Metron Alerts UI
2. Go to `http://localhost:4201/alerts-list` in the browser
3. Open the date range selector floating pane at the upper right
4. In the floating window, try to manually write or paste valid date formats within the accepted time range (In the two input fields on the left side of the floating window)
5. It should be nicely done. 🙌

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


sardell:
@ruffle1986 This looks good to me, but I do want to bring up one thing that I noticed that's existed before but is even more pronounced now that we can enter the timestamp manually.

In the date picker template, there is a click event that toggles the menu on the input element. I think if a user is manually entering a timestamp, they won't need to see a calendar. In addition, let's say I want to changed the default populated 'now' value to my own custom timestamp. I would double click on the field in order to highlight the 'now' value and replace it with my own value. This quickly opens and closes the calendar. When I go to click on the Apply button, the calendar appears again. I included a gif below to show what I'm talking about.

date-picker-click-toggle

I think that instead of having a click event on the input, we should maybe have it on the calendar icon. That way, the date picker behaves more like this one from Angular Material.

angular-material-date-picker

What do you think?

ruffle1986:
@sardell Good point and doesn't sound like a big effort. Let me check this out.

ruffle1986:
@sardell Alright, I put the trigger on the calendar icon, but in that case I had to remove the toggle listener from the component. The reason is because it has conflicted with Pikaday and Pikaday doesn't have this "toggling" capability by by default and there's no option to turn it on so we lost that behaviour. But I think it's not a big deal so I ended going without the toggle.

sardell:
@ruffle1986 I see what you mean after testing locally. However, when add the toggle function and the click listener on the calendar icon back into the code, I can see the calendar toggling visibility if I step through with a debugger. After further inspection and testing, I was able to override this behavior by adding `bound: false` to the Pikaday configuration (this allows us to manually handle the open/close functionality with Angular's event binding) and adding `this.picker.hide();` in the ngOnInit function after Pikaday is instantiated (this is what prevents the re-opening of the calendar when toggling open/close).

Screen Shot 2019-05-20 at 1 43 27 PM


In addition, I added `this.picker.hide();` at the end of both if statements inside of ngOnChanges so the calendar closes after a user chooses a date. 

Screen Shot 2019-05-20 at 1 43 37 PM

I also noticed that the calendar defaults back to `position: static` with the `bound: false` configuration, so I added the following to styles.scss:
```
.pika-single {
  position: absolute !important;
}
```

After doing all that, it seems to work fine. Try it out and let me know what you think.

pikaday-open-close-solution1

ruffle1986:
@sardell Thank you for the suggestions, they are all good points. But, with your changes, I'm still experiencing the "flash of date picker" after manually pasting a date and clicking outside of the input element. I think it's not a big deal especially when I look at the other disadvantages I'm experiencing also. 

However, with the patch we can enable this toggling functionality which is great, but now, this is the only way to hide the date picker if I don't want to set a date from the picker. So for me, it hasn't make it any better or user friendlier. 

It also breaks the cypress tests and adds a lot of complexity in order to add something that should be convenient but not really is. It also breaks the date pickers on the PCAP panel so additional css hacks are needed to get rid of the bad parts that are introduced by the suggested patch.

So long story short, I think these changes are not necessary and are not worth the effort.

I'd like you to reconsider the PR based on my concerns.

Thank you in advance.

sardell:
I do think my suggestion is creating more work that is clearly out of scope of this specific fix for the datepicker. I respectfully disagree that these changes aren't necessary or worth the effort. Maybe I'm oversimplifying this, but you could add a HostListener to listen for clicks outside of the datepicker. For example: https://stackoverflow.com/questions/40107008/detect-click-outside-angular-component. The Cypress tests are probably broken with the suggested changes because they are targeting the input vs. the icon to open the calendar.

All that said, I'm going to give this a +1, as this work clearly addresses the problem you set out to fix (copy and pasting dates in the datepicker). I've created a separate ticket to address the UX issues with the datepicker: https://issues.apache.org/jira/browse/METRON-2146

PR #1385, opened 2019-04-15
METRON-2071 Add MAP_PUT and MAP_MERGE to Stellar

mmiklavc:
## Contributor Comments

https://issues.apache.org/jira/browse/METRON-2071

Adds 2 new functions, MAP_PUT and MAP_MERGE, to Stellar. I had originally looked at MAP_PUT alone, and when reviewing some of the other data type features such as Sets and Multisets, MAP_MERGE also seemed like a reasonable addition.

**Testing**

I've added pretty extensive unit tests for this, and there are no new dependencies outside of core Java, so I believe manual testing would be optional for this change. If you wish to perform manual validation, spin up full dev (or the Stellar shell in stand-alone mode locally, see https://github.com/apache/metron/tree/master/metron-stellar/stellar-common#stellar-shell-stand-alone) and run through the expressions in the new unit tests for MAP_PUT and MAP_MERGE.

### For all changes:

### For code changes:
  ```
  ```

- n/a Have you verified the basic functionality of the build by building and running locally with Vagrant full-dev environment or the equivalent?

  ```
  cd site-book
  mvn site
  ```


metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MapFunctions.java
```diff
@@ -82,4 +83,67 @@ public Object apply(List<Object> objects) {
...
+      if (list.size() < 1) {
+        return null;
+      }
+      LinkedHashMap<Object, Object> ret = new LinkedHashMap<>();
+      Object o = list.get(0);
+      if (o != null) {
+        if (!(o instanceof Iterable)) {
+          throw new IllegalArgumentException("Expected an Iterable, but " + o + " is of type " + o.getClass());
+        }
+        Iterable<? extends Map> maps = (Iterable<? extends Map>) o;
```
cstella-stripe: Perhaps `if(Iterables.isEmpty(maps)) { return Iterables.getFirst(maps);)`?

cestella: I concur with my alter ego ;) . Anyway, this isn't strictly required, just a small thing that may be of use.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/BasicStellarTest.java
```diff
@@ -641,6 +645,74 @@ public void testMapGet() {
...
+      assertThat(result.size(), equalTo(6));
+      assertThat(result.get("a"), equalTo(1));
+      assertThat(result.get("b"), equalTo(2));
+      assertThat(result.get("c"), equalTo(3));
+      assertThat(result.get("d"), equalTo(4));
+      assertThat(result.get("e"), equalTo(5));
+      assertThat(result.get("f"), equalTo(6));
+    }
+    {
+      String query = "MAP_MERGE( [ { 'a' : '1', 'b' : '2' }, { 'c' : '3', 'd' : '4' }, { 'e' : '5', 'f' : '6' } ] )";
```
cestella: Can we add a test case where the maps being merged overlap in the keyspace?
e.g.
```
String query = "MAP_MERGE( [ { 'a' : '1', 'b' : '2' }, { 'c' : '3', 'd' : '4' }, { 'a' : '5', 'd' : '6' } ] )";
Map result = (Map) run(query, new HashMap<>());
...
assertThat(result.get("a"), equalTo("5"));
```

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MapFunctions.java
mmiklavc: Hey @cestella, thanks for the review! I'm not sure I completely follow - `getFirst` also requires a default. I initialize the return value on line 132, but it's outside of the `if` block. Maybe you meant `if (Iterables.size(maps) == 1){ return Iterables.getFirst(maps);)`?

Side note, I had initially wanted to use `Iterables.addAll()` for consistency , however this only accepts `Collections`, which the `Map` interface does not extend. https://github.com/apache/metron/blob/master/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/SetFunctions.java#L137

mmiklavc:
Thanks @cestella - I just kicked out some additional changes based on your feedback. Also, I don't think we have a formal process around this, but GitHub added a review feature that will indicate "changes requested." If you don't mind also marking it resolved when you're happy with a +1.

metron-stellar/stellar-common/src/test/java/org/apache/metron/stellar/dsl/functions/BasicStellarTest.java
mmiklavc: Done

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MapFunctions.java
cestella: yep!  looks good to me

cestella:
+1

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MapFunctions.java
```diff
@@ -82,4 +84,71 @@ public Object apply(List<Object> objects) {
...
+           returns = "The original map modified with the key/value. If the map argument is null, a new map will be created and returned that contains the provided key and value - note: if the 'map' argument is null, only the returned map will be non-null and contain the key/value."
+          )
+  public static class MapPut extends BaseStellarFunction {
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Object apply(List<Object> objects) {
+      if (objects.size() < 3) {
+        throw new IllegalArgumentException("Must pass a key, value, and map");
+      } else {
```
ottobackwards: Do we not want type checking at all here?  This is a pretty specific function

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MapFunctions.java
```diff
@@ -82,4 +84,71 @@ public Object apply(List<Object> objects) {
...
+           returns = "A Map. null if the list of maps is empty."
+          )
+  public static class MapMerge extends BaseStellarFunction {
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Object apply(List<Object> list) {
+      if (list.size() < 1) {
+        return null;
+      }
```
ottobackwards: Type checking?

metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/MapFunctions.java
mmiklavc: What kind of type checking do you think we should add? The map should accept any types for keys and values, and the map argument itself will throw a ClassCastException if it's not actually a map. I tried to make the functionality pretty consistent with our examples, e.g. SET_ADD https://github.com/mmiklavc/metron/blob/a8628f1610a40cd0c790eb0b3caf6d2585db3018/metron-stellar/stellar-common/src/main/java/org/apache/metron/stellar/dsl/functions/SetFunctions.java#L64

ottobackwards: If you search for `instanceof` in stellar common you will see that we do handle it, if only to return null at times.  In many places.
We are certainly inconsistent on this point.  My preference would be to handle it, since we do not have explicitly typed parameters or parameter checking

PR #1346, opened 2019-02-26
METRON-2016: Parser aggregate groups should be persisted and available through REST

merrimanr:
## Contributor Comments
This PR adds a REST endpoint that manages parser groups in the global config.  The endpoint exposes CRUD operations for a new `SensorParserGroup` object used to store a group name, description and sensor list.

### Changes Included
- `SensorParserGroup` type was added.
- Method added to `ParserConfigurations` for converting the objects stored in the global config to an array of `SensorParserGroup` objects.
- REST classes added to support CRUD operations (`SensorParserGroupController`, `SensorParserGroupService`, `SensorParserGroupServiceImpl`).
- Logic added to the existing `StormAdminServiceImpl` and `StormStatusServiceImpl` classes for converting a `SensorParserGroup` to the corresponding Storm topology name.
- The `parser_commands.py` Ambari mpack script was updated to sort the sensors before starting/stopping topologies.  This is needed so that both Ambari and REST can correctly generate a topology name from a list of sensors.
- Slightly refactored `ParserTopologyCLI` to use a constant sensor separator.
- Added tests and javadocs where appropriate.

### Changes NOT Included
- The `ParserTopologyCLI` still expects a comma-separated list of sensors when starting a topology.  The concept of a `SensorParserGroup` is limited to REST at this point.  We could add group name support to `ParserTopologyCLI` but I don't see a requirement.  If desired I believe this should be a follow on.

### Testing
This has been tested in full dev.

1. Spin up full dev and verify data is flowing to ES.
2. Create a sensor group:
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{
"description": "group1 description",
"name": "group1",
"sensors": [
"bro","snort","yaf"
]
}' 'http://user:password@node1:8082/api/v1/sensor/parser/group'
```
A 201 status with the new group should be returned.

3. Create another group:
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{
"description": "group2 description",
"name": "group2",
"sensors": [
"jsonMap","jsonMapQuery"
]
}' 'http://user:password@node1:8082/api/v1/sensor/parser/group'
```
4. Get all groups and verify both groups are returned:
```
curl -X GET --header 'Accept: application/json' 'http://user:password@node1:8082/api/v1/sensor/parser/group'
```
5. The groups should also be visible in the global config:
```
curl -X GET --header 'Accept: application/json' 'http://user:password@node1:8082/api/v1/global/config'
```
[...]

PR #1396, opened 2019-04-29
METRON-2034: REST endpoint for getting all parser topology status should return group name

merrimanr:
## Contributor Comments
This PR fixes a bug in the REST application that fails to convert a parser topology id to a group name when calling the `http://node1:8082/api/v1/storm` endpoint.  The `StormStatusServiceImplTest` was also updated to included this case.  

### Testing
This has been tested in full dev:

1. Spin up full dev and verify data is flowing to Elasticsearch.
2. Define a group name for the OOTB bro, snort, yaf parser topology that is included in full dev:
```
    "name": "group1",
    "description": "group1",
    "sensors": [
      "bro",
      "snort",
      "yaf"
    ]
  }' 'http://user:password@node1:8082/api/v1/sensor/parser/group'
```
3. Request a storm status summary for all topologies:
```
curl -X GET --header 'Accept: application/json' 'http://node1:8082/api/v1/storm'
```
The `bro__snort__yaf` Storm topology should be returned with the name set to `group1`:
```
{
    "id": "bro__snort__yaf-4-1556557869",
    "name": "group1",
    "status": "ACTIVE",
    "latency": 2751.941,
    "throughput": 6.045,
    "emitted": 3710,
    "acked": 3627
  }
```

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-platform/metron-common/src/main/config/zookeeper/global.json
+++ b/metron-platform/metron-common/src/main/config/zookeeper/global.json
@@ -6,5 +6,16 @@
   "update.hbase.table": "metron_update",
   "update.hbase.cf": "t",
   "es.client.settings": {
-  }
+  },
+  "parser.groups": [
+    {
+      "name": "group",
+      "description": "group",
+      "sensors": [
+        "bro",
+        "snort",
+        "yaf"
+      ]
+    }
+  ]
 }
--- a/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/StormStatusServiceImpl.java
+++ b/metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/StormStatusServiceImpl.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import org.apache.metron.common.configuration.SensorParserGroup;
@@ -83,9 +84,14 @@ public class StormStatusServiceImpl implements StormStatusService {
   public List<TopologyStatus> getAllTopologyStatus() {
     List<TopologyStatus> topologyStatus = new ArrayList<>();
     for (TopologyStatus topology : getTopologySummary().getTopologies()) {
-      topologyStatus.add(restTemplate
-          .getForObject(getStormUiProperty() + TOPOLOGY_URL + "/" + topology.getId(),
-              TopologyStatus.class));
+      TopologyStatus status = restTemplate
+              .getForObject(getStormUiProperty() + TOPOLOGY_URL + "/" + topology.getId(),
+                      TopologyStatus.class);
+      if (status != null) {
+        Optional<String> groupName = topologyNameToGroupName(status.getName());
+        groupName.ifPresent(status::setName);
+        topologyStatus.add(status);
+      }
     }
     return topologyStatus;
   }
@@ -149,15 +155,10 @@ public class StormStatusServiceImpl implements StormStatusService {
       String topologyName = topology.getName();
 
       // check sensor group
-      if (topologyName.contains(ParserTopologyCLI.STORM_JOB_SEPARATOR)) {
-        Set<String> sensors = new HashSet<>(Arrays.asList(topologyName.split(ParserTopologyCLI.STORM_JOB_SEPARATOR)));
-        SensorParserGroup group = sensorParserGroupService.findOne(name);
-        if (group == null) {
-          break;
-        } else if (sensors.equals(group.getSensors())){
-          id = topology.getId();
-          break;
-        }
+      Optional<String> groupName = topologyNameToGroupName(topologyName);
+      if (groupName.isPresent() && groupName.get().equals(name)) {
+        id = topology.getId();
+        break;
       }
 
       if (topologyName.equals(name)) {
@@ -166,5 +167,19 @@ public class StormStatusServiceImpl implements StormStatusService {
       }
     }
     return id;
+  }
+
+  public Optional<String> topologyNameToGroupName(String topologyName) {
+    Optional<String> groupName = Optional.empty();
+    if (topologyName.contains(ParserTopologyCLI.STORM_JOB_SEPARATOR)) {
+      Set<String> sensors = new HashSet<>(Arrays.asList(topologyName.split(ParserTopologyCLI.STORM_JOB_SEPARATOR)));
+      for (SensorParserGroup group: sensorParserGroupService.getAll().values()) {
+        if (sensors.equals(group.getSensors())) {
+          groupName = Optional.of(group.getName());
+          break;
+        }
+      }
+    }
+    return groupName;
   }
 }
--- a/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/service/impl/StormStatusServiceImplTest.java
+++ b/metron-interface/metron-rest/src/test/java/org/apache/metron/rest/service/impl/StormStatusServiceImplTest.java
@@ -148,7 +148,9 @@ public class StormStatusServiceImplTest {
       add("bro");
       add("snort");
     }});
-    when(sensorParserGroupService.findOne("group")).thenReturn(group);
+    when(sensorParserGroupService.getAll()).thenReturn(new HashMap() {{
+      put("group", group);
+    }});
     when(environment.getProperty(STORM_UI_SPRING_PROPERTY)).thenReturn(HTTP_STORM_UI);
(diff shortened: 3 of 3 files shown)


ruffle1986:
@merrimanr I've tested it with full dev and now the UI is able to display if a group of sensors is running or not properly. So this is fine but there's another issue. The reason why it has become a blocker for the latest release candidate is that `bro`, `snort`, and `yaf` is already aggregated.

And the topology info for the three of them coming back from the `storm` endpoint is just fine. It says it's `ACTIVE` but since the UI doesn't get any information about their "group", it's unable to render the three sensors as they are grouped together. 

So currently, the name of the topology item regarding the three aforementioned sensors is "bro__snort__yaf" but there is no group with the same name so the UI can't figure it's a group and the three sensors belong to it.

merrimanr:
You're correct @ruffle1986.  The group should already be added when full dev spins up rather than having to add it in a separate step.  I am testing a change that should fix this and will push it out once that's done.

merrimanr:
The latest commit should setup the parser group in full dev automatically.

metron-platform/metron-common/src/main/config/zookeeper/global.json
```diff
@@ -6,5 +6,16 @@
   "update.hbase.table": "metron_update",
   "update.hbase.cf": "t",
   "es.client.settings": {
-  }
+  },
+  "parser.groups": [
```
mmiklavc: What is this option in global config? I don't recall seeing this before, and it's not mentioned anywhere from what I can tell, e.g. https://github.com/apache/metron/tree/master/metron-platform/metron-common#global-configuration.

merrimanr: Parser groups are stored in the global config.  There are REST endpoints for managing parser groups, a user would not directly edit this in the global config.  I can add it to that README but I'm not sure it's useful to a user.  I'm happy to add more documentation wherever it makes sense.

ottobackwards: why would we store them there and not in zookeeper?

mmiklavc: > why would we store them there and not in zookeeper?

@ottobackwards not sure I completely understand the Q here - global.json is stored in ZK. Is your question why this is stored in global.json?

Part of the reason I asked about the docs is because I don't have a good sense of the implications to storing this here. This is the original PR - https://github.com/apache/metron/pull/1346. @merrimanr what happens when I start a parser group from the CLI, will global.json be out of sync for the UI then?

merrimanr: There is no change to starting a parser group from the CLI.  You would still pass them in as a comma-separate list of parsers if you wanted to group parsers together.  Same goes for Ambari.  This setting is only used if parser topologies are managed in the Management UI.  The REST service handles converting parser groups to the correct CLI arguments and Storm job names.

ottobackwards: Yes, I'm asking why they are in the global configuration and not just a new ZK key/node

mmiklavc: I think the question here then becomes why this config change is made in this PR and not part of the UI PR that supports aggregation? If this is literally a "the 2 PR's are tightly coupled and should be merged together" that's probably fine. But I'm currently confused as to what the grouping change is doing here since there doesn't seem to be an existing endpoint/feature that this is impacting.

merrimanr: I'm happy to resubmit this against the feature branch if that makes reviewing easier.

mmiklavc: Just to be clear, I think the overall PR is fine, it's just the recent parser group change to global.json that looks like it belongs against the feature branch. Does that work?

merrimanr: Yes no problem.

mmiklavc: > Parser groups are stored in the global config. There are REST endpoints for managing parser groups, a user would not directly edit this in the global config. I can add it to that README but I'm not sure it's useful to a user. I'm happy to add more documentation wherever it makes sense.

Can you add documentation for this feature? It looks like it was missed in the original PR.

merrimanr: What documentation do you want me to add?  Do you want a dedicated section on parser groups somewhere?  Just add the setting to the global config?

mmiklavc: Something to explain what it is if I've never seen this config option before, but am familiar with global config. From what you said earlier, this is purely a sharing mechanism, not something a user should ever modify. Some of this may be defined implicitly from the parser aggregation feature, mostly concerned with how this additional feature functions in concert with it.

This is just off the top of my head skimming through, so I imagine you'll have a lot more to add to it as the implementer. Some of this may already be documented out there somewhere, just want to make sure we put a bow on it.

**User-level concerns**

1. If a user does modify it manually, whether intentionally or by accident, what will happen?
2. What processes manage this config? Will Ambari or the config UI manage or overwrite these values?
3. If a user adds a group manually, will the values remain, or get overwritten/removed? (e.g. what's the interaction with our global config JSON Patching?)
4. How is a group different from a parser aggregation as defined in the rest of our documentation?

**Developer concerns**

1. Same as above, plus
2. What is the unique ID for a group, is there one?
3. What if I group the same sensors in multiple groups?
4. What are each of the fields and how are they used? i.e. If I go modify this feature in the UI, what is my API and what are the guarantees it provides?
5. What fields are required vs optional? Related to the "guarantees" in 4. Important for null checks, type checks, etc. etc.

merrimanr: I spent some time trying to document this and I realized the documentation doesn't make much sense because the UI component hasn't been done yet.  I find myself guessing how it is going to work because I'm not 100% clear on how the UI will function and I don't have anything to reference because that work is still being submitted.  Almost all of the requests in your list involve the UI in some way.  Parser group and parser aggregation are synonymous so that's something else we'll want to fix. 

I think I may put this on hold until the UI part is further along unless you would accept that work as a follow on Jira that blocks the acceptance of this feature branch.  The benefit of committing this now is it will resolve a bug that was found earlier.

mmiklavc:
@merrimanr now that https://github.com/apache/metron/tree/feature/METRON-1856-parser-aggregation is under way, do we have enough detail to bridge the gap between what the UI needs and the backend piece, in particular the documentation? I'm not completely sure if global config is the best place for it or not, since this only really seems to be leverage by REST application (ie, it's not used globally), but I don't have a particularly excellent alternative to recommend currently.

merrimanr:
We are not at a point in the review where we can spin up or test anything.  Agreed, documentation belongs in the REST README.

mmiklavc:
Let's hold off on this item until the end, or whenever we can reasonably start doing some validation with the actual UI components in this feature branch. It sounds like this may have come from some offline collaboration and testing, so we should make sure that this accomplishes the desired results now that the aggregation feature is going through review.

merrimanr:
I filed this as a bug back in March based on feedback from the contributors working on the parser aggregation feature branch.  This fix is required for that feature to work correctly, which is why I think we should move documentation to a follow on Jira.

mmiklavc:
Can you link to what you're describing? I'm not seeing much on the ticket - https://issues.apache.org/jira/browse/METRON-2034

merrimanr:
Are you saying you want a more detailed description the bug as it relates to the UI?  This might be easier to demonstrate when the UI part is further along.

mmiklavc:
> Are you saying you want a more detailed description the bug as it relates to the UI? This might be easier to demonstrate when the UI part is further along.

I'm not seeing where the "based on feedback from the contributors working on the parser aggregation feature branch" interaction happened. Was that a private message/interaction? All I'm saying is that if there was a comment on another PR, or a discuss thread, or a convo in Slack - just add it here for context.

merrimanr:
It was a Slack conversation.  @ruffle1986 was testing and noticed group names were not working for the endpoint that gets status for all topologies (vs the endpoint that gets status for a single topology).  He brought it to my attention and I created this fix as a result.

mmiklavc:
> I filed this as a bug back in March based on feedback from the contributors working on the parser aggregation feature branch. This fix is required for that feature to work correctly, which is why I think we should move documentation to a follow on Jira.

I'm not sure I see the point of splitting up a documentation task here - seems like extra overhead with an unclear benefit, but if that's your preference, you're welcome to add a sub-task to the master  feature branch Jira that would need to be checked off before the FB is completed.

PR #1400, opened 2019-05-06
METRON-2106: Escalation topic setting in Ambari has no effect

merrimanr:
## Contributor Comments
This PR fixes a small bug in the Metron REST start script that incorrectly configures the property files.  Properties defined in `rest_application.yml` (where Ambari-managed properties are written) are being overridden by the defaults when the opposite should happen.  This fixes that by updating the configuration to load `rest_application.yml` before the defaults.

### Testing
This has been tested in full dev.  Spin up full dev and configure a custom escalation topic in Ambari (Ambari > Services > Metron > Configs > REST > Metron escalation topic).  Escalate a message:
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: */*' -d '[
  {"field":"value"}
]' 'http://user:password@node1:8082/api/v1/alerts/ui/escalate'
```
The message should appear in the custom escalation topic rather than the default `escalation` topic.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-rest/src/main/scripts/metron-rest.sh
+++ b/metron-interface/metron-rest/src/main/scripts/metron-rest.sh
@@ -83,7 +83,7 @@ echo "METRON_SPRING_PROFILES_ACTIVE=${METRON_SPRING_PROFILES_ACTIVE}"
 
 # the vagrant Spring profile provides configuration values, otherwise configuration is provided by rest_application.yml
 if [[ !(${METRON_SPRING_PROFILES_ACTIVE} == *"vagrant"*) ]]; then
-    METRON_CONFIG_LOCATION=" --spring.config.location=$METRON_HOME/config/rest_application.yml,classpath:/application.yml"
+    METRON_CONFIG_LOCATION=" --spring.config.additional-location=file:$METRON_HOME/config/rest_application.yml"
     echo "METRON_CONFIG_LOCATION=${METRON_CONFIG_LOCATION}"
     METRON_SPRING_OPTIONS+=${METRON_CONFIG_LOCATION}
 fi


metron-interface/metron-rest/src/main/scripts/metron-rest.sh
```diff
@@ -83,7 +83,7 @@ echo "METRON_SPRING_PROFILES_ACTIVE=${METRON_SPRING_PROFILES_ACTIVE}"
 
 # the vagrant Spring profile provides configuration values, otherwise configuration is provided by rest_application.yml
 if [[ !(${METRON_SPRING_PROFILES_ACTIVE} == *"vagrant"*) ]]; then
-    METRON_CONFIG_LOCATION=" --spring.config.location=$METRON_HOME/config/rest_application.yml,classpath:/application.yml"
```
mmiklavc: The ordering/appending via `--spring.config.additional-location` makes sense. But I see you've removed `classpath:/application.yml` altogether. Was that intentional? How is application.yml loaded now, or is it not needed at all?

merrimanr: Using `--spring.config.location` replaces the default locations.  This is why `classpath:/application.yml` was originally included even though it matches the default locations.  Using `--spring.config.additional-location` adds to the list of configs instead of replacing the default locations so having `classpath:/application.yml` in the list is no longer necessary.  Doing it this way enforces that anything added on top with `--spring.config.additional-location` will take priority over the default.  I think this is the correct way to do it because you don't have to worry about ordering them correctly.

mmiklavc: @merrimanr - perfect, that makes sense. Thanks for the explanation! +1 by inspection.

PR #1431, opened 2019-05-23
METRON-2102: [UI] Adding click-through navigation to Alerts table

tiborm:
## Contributor Comments

Click Through Navigation is a feature makes Metron Users able to reach other web services via dynamically created URLs by clicking link item in a context menu.
This context menu (aka. click-through menu) is attached to the alerts table and the links are populated with alert data from the specific row of the table. 

Items and URLs in the context menu based on a configuration (this is currently a JSON file). A configuration could be attached to a cell or a row.
If you like to attach a menu configuration to a cell of a column you should use the field id (what field of the alert populates the column) to target the particular column.

If you like to learn more about the feature or the possible ways of configuration please check the description in the [Jira ticket](https://issues.apache.org/jira/browse/METRON-2102).

click-trough-menu

Screen Shot 2019-05-23 at 7 35 56 PM

Screen Shot 2019-05-23 at 7 35 35 PM

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ repodata/
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+metron-interface/metron-alerts/TODO
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,3 @@ repodata/
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
-metron-interface/metron-alerts/TODO
--- a/metron-interface/metron-alerts/cypress/integration/alert-list/context-menu.spec.js
+++ b/metron-interface/metron-alerts/cypress/integration/alert-list/context-menu.spec.js
@@ -59,7 +59,7 @@ context('Context Menu on Alerts', () => {
     cy.get('.ace_value').should('contain', '192.168.66.121');
   });
 
-  it('clicking on "Add to search bar" should colose the dropdown of context menu', () => {
+  it('clicking on "Add to search bar" should close the dropdown of context menu', () => {
     cy.get('[data-qe-id="row-5"] > :nth-child(6) > a').click();
     cy.get('[data-qe-id="cm-dropdown"]').should('be.visible');
     cy.contains('Add to search bar').click();
--- a/metron-interface/metron-alerts/cypress/integration/alert-list/context-menu.spec.js
+++ b/metron-interface/metron-alerts/cypress/integration/alert-list/context-menu.spec.js
@@ -1,0 +1,89 @@
+/// <reference types="Cypress" />
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as appConfigJSON from '../../../src/assets/app-config.json';
+
+context('Context Menu on Alerts', () => {
+
+  beforeEach(() => {
+    cy.server();
+    cy.route({
+      method: 'GET',
+      url: '/api/v1/user',
+      response: 'user'
+    });
+    cy.route({
+        method: 'POST',
+        url: '/api/v1/logout',
+        response: []
+    });
+
+    cy.route('GET', '/api/v1/global/config', 'fixture:config.json');
+    cy.route('POST', 'search', 'fixture:search.json');
+
+    cy.route('GET', appConfigJSON.contextMenuConfigURL, 'fixture:context-menu.conf.json');
+
+    cy.visit('login');
+    cy.get('[name="user"]').type('user');
+    cy.get('[name="password"]').type('password');
+    cy.contains('LOG IN').click();
+
+    cy.get('[data-qe-id="alert-search-btn"]').click();
+  });
+
+  it('clicking on a table cell should show context menu', () => {
+    cy.get('[data-qe-id="row-5"] > :nth-child(6) > a').click();
+    cy.get('[data-qe-id="cm-dropdown"]').should('be.visible');
+  });
+
+  it('clicking on "Add to search bar" should apply value to filter bar', () => {
+    cy.get('[data-qe-id="row-5"] > :nth-child(6) > a').click();
+    cy.get('[data-qe-id="cm-dropdown"]').should('be.visible');
+    cy.contains('Add to search bar').click();
+    cy.get('.ace_keyword').should('contain', 'ip_src_addr:');
+    cy.get('.ace_value').should('contain', '192.168.66.121');
+  });
+
+  it('clicking on "Add to search bar" should colose the dropdown of context menu', () => {
+    cy.get('[data-qe-id="row-5"] > :nth-child(6) > a').click();
+    cy.get('[data-qe-id="cm-dropdown"]').should('be.visible');
+    cy.contains('Add to search bar').click();
+    cy.get('[data-qe-id="cm-dropdown"]').should('not.be.visible');
+  });
+
+  it('dynamic items should be rendered', () => {
+    cy.get('[data-qe-id="row-5"] > :nth-child(6) > a').click();
+    cy.get('[data-qe-id="cm-dropdown"]').should('be.visible');
+    cy.get('[data-qe-id="cm-dropdown"]').contains('IP Investigation Notebook').should('be.visible');
+  });
+
+  // this use case was a former bug caused by the behaviour of the rxjs Subject class
+  // here we pinning down the fix with a test
+  it('dynamic items should be rendered after a clicking a predefined item', () => {
+    cy.get('[data-qe-id="row-5"] > :nth-child(6) > a').click();
+    cy.get('[data-qe-id="cm-dropdown"]').should('be.visible');
+    cy.contains('Add to search bar').click();
+
+    cy.wait(300);
+
+    cy.get('[data-qe-id="row-5"] > :nth-child(6) > a').click();
+    cy.get('[data-qe-id="cm-dropdown"]').should('be.visible');
+    cy.get('[data-qe-id="cm-dropdown"]').contains('IP Investigation Notebook').should('be.visible');
+  });
+
+})
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/scripts/params/params_linux.py
@@ -465,6 +465,7 @@ knox_group = config['configurations']['knox-env']['knox_group']
 metron_knox_root_path = '/gateway/metron'
 metron_rest_path = '/api/v1'
 metron_alerts_ui_login_path = '/login'
+metron_alerts_ui_context_menu_config_url = '/assets/context-menu.conf.json'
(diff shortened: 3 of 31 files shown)


.gitignore
```diff
@@ -32,3 +32,4 @@ repodata/
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+metron-interface/metron-alerts/TODO
```
sardell: Should this be here?

tiborm: No. I going to remove it. It was just for my todo file while I was developed this.

metron-interface/metron-alerts/cypress/integration/alert-list/context-menu.spec.js
```diff
@@ -0,0 +1,89 @@
...
+
+  it('clicking on "Add to search bar" should apply value to filter bar', () => {
+    cy.get('[data-qe-id="row-5"] > :nth-child(6) > a').click();
+    cy.get('[data-qe-id="cm-dropdown"]').should('be.visible');
+    cy.contains('Add to search bar').click();
+    cy.get('.ace_keyword').should('contain', 'ip_src_addr:');
+    cy.get('.ace_value').should('contain', '192.168.66.121');
+  });
+
+  it('clicking on "Add to search bar" should colose the dropdown of context menu', () => {
```
sardell: Typo. 'close' instead of 'colose'.

sardell:
+1 pending a small typo fix. Tested with full dev and the instructions left on the ASF Jira ticket.

metron-interface/metron-alerts/cypress/integration/alert-list/context-menu.spec.js
tiborm: Thanks, fixed

mmiklavc: @tiborm This line still shows "colose" in your latest version. Did you push that commit?

mmiklavc:
@tiborm this looks like a pretty cool feature, thanks for the contribution! I'd like to address a couple things before we commit this. I am on board with providing a flag for this feature and leaving it disabled by default - that looks reasonable to me, and I also coincidentally just kicked off a DISCUSS thread along these lines as well - https://lists.apache.org/thread.html/e24e4709fbc06473977a0a7f0c0ce92eb02cfcc8c1c2bc41be08a698@%3Cdev.metron.apache.org%3E. I do think we need to resolve some outstanding questions about how we handle that as a community - see below.

I looked over your instructions on the Jira ticket and I think we're pretty close. I'd like to see the following:

1. What's the plan for handling this feature flag? We haven't had any discussion in the community about enabling feature flags to this point. I suspect we might consider them an option for lowering the bar to testing, but we should address this. I don't see any non-happy path tests on this PR, for example. I think that either we need to put this PR through more rigorous testing or come to a community understanding (and document it in the dev guidelines) about what a reasonable level of quality and feature polish will be for code committed under a feature flag.
1. Non-happy path test scenarios? e.g.
    1. What happens to the UI if you misspell a field name? What errors/indication does the user get? Will the UI fail to come up? Will it only fail when a field is visible in the UI column listing?
    2. What happens if you type "fasle" for `isEnabled`? Again, what's the default behavior for misconfiguration and where can operations people managing this feature look for logs that detail the exceptions?
    3. What happens for fields not present in a particular record that are used for token replacement? e.g. per the Jira instructions - "But in the configuration, any available alert property field could be referenced like the following:" Also, is this impacted in any way by what fields are currently displayed/enabled in the alerts UI column listing? If I've hidden ip_src_addr from display, will it still render the URL correctly?
1. These instructions should have a dedicated section in the documentation/README for the Alerts UI.
    1. Please provide a detailed explanation of the configuration lifecycle - will the changes be picked up automatically, or only after a restart? Provide instructions for whatever those steps are (e.g. restart Alerts UI using command foo)
    1. Provide a detailed schema description for the JSON you provided in your testing examples. I think you've done a decent job of explaining what `alertEntry` and `metaAlertEntry` mean, this just needs to be expanded on a bit and laid out similar to what we've done for parsers here - [general purpose parser config options](https://github.com/apache/metron/tree/master/metron-platform/metron-parsing#introduction) and here - [config details for all parsers](https://github.com/apache/metron/tree/master/metron-platform/metron-parsing#parser-configuration)

sardell:
> What's the plan for handling this feature flag?

@mmiklavc After thinking about it, I feel like this implementation is less like a feature flag and more like a configuration option. I say this because, from how I understand it, feature flags are meant to eventually be included to all users. From what I can tell, this is not a feature that all users will want to turn on because of the required configuration (@tiborm correct me if I'm wrong here).

> What happens if you type "fasle" for isEnabled?

Good point. I just built the Alerts app with Angular CLI and it completed successfully with that typo. I haven't been able to test the other non-happy/failure path scenarios you mention, but these are all good points.

> These instructions should have a dedicated section in the documentation/README for the Alerts UI.

@tiborm Let me know if I can help with this.

mmiklavc:
> After thinking about it, I feel like this implementation is less like a feature flag and more like a configuration option
@sardell makes sense to me. That being said, I think we need to polish it a bit further per my earlier comments before committing.

tiborm:
> @tiborm this looks like a pretty cool feature, thanks for the contribution! I'd like to address a couple things before we commit this. I am on board with providing a flag for this feature and leaving it disabled by default - that looks reasonable to me, and I also coincidentally just kicked off a DISCUSS thread along these lines as well - https://lists.apache.org/thread.html/e24e4709fbc06473977a0a7f0c0ce92eb02cfcc8c1c2bc41be08a698@%3Cdev.metron.apache.org%3E. I do think we need to resolve some outstanding questions about how we handle that as a community - see below.
> 
> I looked over your instructions on the Jira ticket and I think we're pretty close. I'd like to see the following:
> 
> 1. What's the plan for handling this feature flag? We haven't had any discussion in the community about enabling feature flags to this point. I suspect we might consider them an option for lowering the bar to testing, but we should address this. I don't see any non-happy path tests on this PR, for example. I think that either we need to put this PR through more rigorous testing or come to a community understanding (and document it in the dev guidelines) about what a reasonable level of quality and feature polish will be for code committed under a feature flag.

I agree with @sardell on this. The isEnable config parameter here is not a feature flag.
IMHO feature flags are for development teams.  In this case, the isEnabled parameter is for our end users (more precisely for those who maintain deployed Metron on a service level). Even if the click through navigation was put behind a feature flag, I would keep the isEnabled flags for our clients to turn on and off based according to their preference. Having this feature on by default would weaken the user experience if users don’t want to use it. 
I know you already agreed with @sardell briefly pointing this out, but I wanted to clarify my  use of the isEnabled parameter. 

On a related note, I added these two followup Jira here:
[Create REST endpoint to persist Click Through navigation configuration](https://issues.apache.org/jira/browse/METRON-2104)
[Impl Click Through navigation admin interface on Management UI](https://issues.apache.org/jira/browse/METRON-2103)
With these two additional changes, isEnabled would be fully exposed to our end users (trough a toggle control) and they would be able to turn it on or off through the Management UI.

> 2. Non-happy path test scenarios? e.g.
>    
>    1. What happens to the UI if you misspell a field name? What errors/indication does the user get? 

Will the UI fail to come up? Will it only fail when a field is visible in the UI column listing?
If the field name is misspelled then the isEnabled field is missing. It will evaluate as not true, therefore the feature stays turned off. I’ll extend the implementation with warnings in the console log if any required configuration is missing. I’m also going to cover these scenarios with unit tests. 
If a field name is misspelled it will be considered missing. None of these scenarios causing errors but I’ll cover them with unit tests.

>    2. What happens if you type "fasle" for `isEnabled`? Again, what's the default behavior for misconfiguration and where can operations people managing this feature look for logs that detail the exceptions?

That would actually evaluate as true because of Javascript’s auto type coercion. I’m going to fix this and cover it with unit tests.

>    3. What happens for fields not present in a particular record that are used for token replacement? e.g. per the Jira instructions - "But in the configuration, any available alert property field could be referenced like the following:" Also, is this impacted in any way by what fields are currently displayed/enabled in the alerts UI column listing? If I've hidden ip_src_addr from display, will it still render the URL correctly?

Visibility of a column has no impact on the token replacement. 
Referencing a non-existing token causes no error. However, the URL will not be resolved properly. I’ll cover this scenario with unit tests as well. And add log messages to help configuration.

> 3. These instructions should have a dedicated section in the documentation/README for the Alerts UI.
>    
>    1. Please provide a detailed explanation of the configuration lifecycle - will the changes be picked up automatically, or only after a restart? Provide instructions for whatever those steps are (e.g. restart Alerts UI using command foo)
>    2. Provide a detailed schema description for the JSON you provided in your testing examples. I think you've done a decent job of explaining what `alertEntry` and `metaAlertEntry` mean, this just needs to be expanded on a bit and laid out similar to what we've done for parsers here - [general purpose parser config options](https://github.com/apache/metron/tree/master/metron-platform/metron-parsing#introduction) and here - [config details for all parsers](https://github.com/apache/metron/tree/master/metron-platform/metron-parsing#parser-configuration)

Will do. This should be easy using the description in Jira as a basis.

Thanks @mmiklavc

mmiklavc:
@tiborm The follow on Jiras make sense. Thanks for responding to these questions and adding the documentation and tests. You probably don't want every edge case validation conditional spelled out in the docs, but I think it's worth letting the user know what to expect, e.g. "Be aware that if you misspell a property, it will be treated as non-existent and the default value of 'X' will be used. You will not get an exception/error."

As a general soft rule around adding features, they can typically come in these flavors, in quasi-descending order of desirability:

1. Feature added with full Management UI or Ambari support for enabling/disabling and configuring in single PR. (if it's a huge PR, it may be better to go with option 2)
2. Feature added with UI config support, but split into multiple PRs. Indicate in the description what you're doing and link the PRs to one another in the description. You can request individual review of each PR and say something like "don't merge until all K PRs in this work have been +1'ed," and then commit each PR in sequence. Ideally, the last PR will include all K-1 PRs and will have been tested to ensure the entire unit of work works. See parser chaining as an example. https://github.com/apache/metron/pull/1084. I would take this a step further and explicitly add the links to the PRs.
    > Note, this PR depends on METRON-1643 and METRON-1642, so those should be reviewed prior to this.
3. Feature added without UI support. **MUST** have documentation for manual configuration. Create Jiras to track the follow-on work and link them to the original.
4. Feature flags - yet to be formally adopted as an approach in Metron, but there is open discussion on it.

Not really applicable here, but one other option in situations like this is a feature branch. We typically use that for more disruptive and/or larger changes that we want more flexibility with on the individual commits, but it's worth mentioning. In this case, the gating factor for the individual PRs in the feature branch is a bit lower. You might provide a PR that breaks something temporarily in the FB, for example. Or you save documentation until after you've iterated on the feature enough to be comfortable with the implementation. You indicate what the necessary follow-on work will be arising from the PR against the feature branch, create a sub-task for it on the feature Jira, and it becomes a gating factor for final vote and acceptance on merging the FB into master.

This PR fits option 3, which is perfectly fine. We just need a bit of expansion on testing and documentation as we've already touched on.

Apache Metron community standards around automated and manual testing apply the same in all cases, with the exception of feature branches where only the final unit of work is reviewed by that standard. Any committer can veto a PR and the approach taken at any time if they think it will not work well for the product and its users, so ymmv. Communicating the full feature delivery lifecycle (e.g. justifying/explaining the why and how of follow-ons) obviously helps grease the skids on this and increases the likelihood of an expedient review and ultimately a +1.

Ok, enough on dev guidelines for now. I should probably get these thoughts into a discuss thread and our dev guidelines at some point, but it seemed useful here as we discussed the implementation choices.

tiborm:
@mmiklavc @sardell I was extended the validation of context-menu.conf.json and covered all scenarios with unit tests. Alerts UI validates possible issues with config JSON and provides error log in the browser console.

I also extended app-config.service with validation, logging and unit tests. The reason why I did that in the scope of this changeset is bc app-config contains **contextMenuConfigURL** entry. Alerts UI's app config is managed by Ambari. I found better to check the existence of required app config entries in the client and provide error logs if something missing. This makes debugging easier in a deployed environment if it's needed.

 Also extended and added feature documentation as a [README.md](https://github.com/tiborm/metron/blob/METRON-2102-click-through-menu/metron-interface/metron-alerts/src/app/shared/context-menu/README.md)

I hope covered everything we discussed. Thank you for reviewing!

mmiklavc:
Nice improvements @tiborm. Thanks for the contribution! +1 by inspection. @sardell are you still good with a +1 as well?

sardell:
I am still good with this. +1

PR #1451, opened 2019-06-28
METRON-2168 Elasticsearch Updates Not Tested in Integration Test

nickwallen:
The `ElasticsearchUpdateIntegrationTest` is not testing that the `ElasticsearchDao` can update and retrieve values in a manner similar to what would occur in production.

### What?
Within the Elasticsearch index, the test fails to define the 'guid' field to be of type 'keyword', instead the type is defaulted to 'text'.  In a production setting this mistake would prevent any documents from being found by guid.  Unfortunately, the test passes despite this. The test needs to match the behavior of what a user would experience in production.

### Why? 
These problems arise because of the way the test is setup.  Instead of directly testing an `ElasticsearchDao` as you might expect this test runs against a `MultiIndexDao` initialized with both an `ElasticseachDao` and an `HBaseDao`. On retrievals the `MultIndexDao` will return the document from whichever index responds first.

With the current test setup, the underlying `ElasticsearchDao` will never retrieve the document that the test case is expecting.  In all cases where the test passes, the document is actually being returned from the `HBaseDao` which is actually just interacting with a mock backend.  The test needs to actually test that we can update and retrieve documents from Elasticsearch.

### Proof?
 If you alter the test to run against just an `ElasticsearchDao` the test will fail as follows in the attached log file.

## Testing

This only changes the integration tests.  Run the integration tests to validate these changes.

## Pull Request Checklist

### DIFF
--- a/metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
+++ b/metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
@@ -26,7 +26,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <guava_version>${global_guava_version}</guava_version>
+        <guava_version>${global_hbase_guava_version}</guava_version>
     </properties>
     <dependencies>
         <dependency>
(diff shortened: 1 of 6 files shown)


nickwallen:
One point that reviewers should consider is do we have enough test coverage for the `MultiIndexDao` now?  

I don't think the `MultiIndexDao` warrants an integration test.  It doesn't integrate with any external systems.  I think it can be tested sufficiently with unit testing.  But is what we have in  `MultiIndexDaoTest` enough?

nickwallen:
I added more unit tests for MultiIndexDao.

metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
```diff
@@ -26,7 +26,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <guava_version>${global_hbase_guava_version}</guava_version>
```
mmiklavc: Is this safe to change? ie do we have an hbase enrichment running to verify the classpath doesn't have problems?

nickwallen: It looks like I don't need that change for this PR, so I reverted it. I may need to revisit this in #1454.

mmiklavc:
lgtm, +1

PR #1454, opened 2019-07-02
METRON-2172 Solr Updates Not Tested in Integration Test

nickwallen:
The `SolrUpdateIntegrationTest` is not testing that the `SolrDao` can update and retrieve values.

- [x] This PR builds on #1451 .  The diffs here will include the changes in #1451 until that PR is merged.

### What?
This gap in the integration test is hiding a few existing bugs in the `SolrUpdateIntegrationTest`.  This fix is similar to what was done for Elasticsearch in #1451.

### Why? 
These problems arise because of the way the test is setup.  Instead of directly testing a `SolrDao` this test runs against a `MultiIndexDao` initialized with both a `SolrDao` and an `HBaseDao`. On retrievals the `MultIndexDao` will return the document from whichever index responds first.

With the current test setup, the underlying `SolrDao` will never retrieve the document that the test case is expecting.  In all cases where the test passes, the document is actually being returned from the `HBaseDao` which is actually just interacting with a mock backend.  The test needs to actually test that we can update and retrieve documents from Elasticsearch.

### Bugs?
After correcting the test setup in `SolrUpdateIntegrationTest` the following bugs were found.
1. A timestamp is not being populated in the documents returned from Solr.
1. The comment field is not removed from the document when the last comment is removed.
1. A NullPointerException occurs if a document does not contain a sensor type.

### Changes

1. The `SolrUpdateIntegrationTest` was changed so that it specifically tests a `SolrDao`.

1. The logic to serialize and deserialize a document to/from Solr was moved into a new abstraction called a `DocumentBuilder`.  The Solr DAOs now all use a `SolrDocumentBuilder` to do this.  Most of the existing logic from `SolrUtilities` was moved to `SolrDocumentBuilder`, except for portions that needed corrected.  

    This allowed me to add unit tests for this logic in `SolrDocumentBuilderTests`. Specifically, I was able to add a test to ensure that the document's timestamp is being set correctly and that a missing sensor type does not cause a NullPointerException.

1. Duplicate logic to add, remove, and (de)serialize comments existed in `UpdateIntegrationTest`, `SolrUpdateDao`, `SolrUtilities`, `HBaseUpdateDao`, and `ElasticsearchUpdateDao`. This logic was also not covered by unit tests.

    The comment logic was moved to `Document` so that it could be reused.  This made it so that no additional "reformatting" logic was needed for the comments. This also made it easy to add unit tests to test this functionality in `DocumentTest`.

1. After the previous change the logic for adding and removing comments is exactly the same for Solr, HBase, and Elasticsearch.  So there is no need for custom `addCommentToAlert` and `removeCommentFromAlert` in `HBaseUpdateDao`, `SolrUpdateDao`, and `ElasticsearchUpdateDao`.  The shared implementation for these was moved up to the `UpdateDao` interface as default methods.

### Acceptance Testing

- [ ] Validate against Elasticsearch.
    1. Spin-up Full Dev with Elasticsearch running.
    1. Ensure alerts are visible in the Alerts UI.
    1. Add a comment to an alert.

- [ ] Validate against Solr.
    1. Spin-up Full Dev with Solr running.
    1. Ensure alerts are visible in the Alerts UI.
    1. Add a comment to an alert.

## Pull Request Checklist

### DIFF
--- a/metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
+++ b/metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
@@ -26,7 +26,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <guava_version>${global_guava_version}</guava_version>
+        <guava_version>${global_hbase_guava_version}</guava_version>
     </properties>
     <dependencies>
         <dependency>
--- a/metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
+++ b/metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
@@ -26,7 +26,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <guava_version>${global_hbase_guava_version}</guava_version>
+        <guava_version>${global_guava_version}</guava_version>
     </properties>
     <dependencies>
         <dependency>
@@ -59,7 +59,21 @@
                     <groupId>log4j</groupId>
                     <artifactId>log4j</artifactId>
                 </exclusion>
+                <exclusion>
+                    <artifactId>commons-httpclient</artifactId>
+                    <groupId>commons-httpclient</groupId>
+                </exclusion>
             </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.4.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
+            <version>4.4.9</version>
         </dependency>
         <dependency>
             <groupId>com.google.guava</groupId>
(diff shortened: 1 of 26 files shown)


metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
```diff
@@ -26,7 +26,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <guava_version>${global_hbase_guava_version}</guava_version>
+        <guava_version>${global_guava_version}</guava_version>
```
nickwallen: I will take a look at this more carefully and see if we need to alter the Guava version here.

merrimanr:
These changes look good to me.  Thank you for fixing this.  +1 by inspection

metron-platform/metron-elasticsearch/metron-elasticsearch-common/pom.xml
nickwallen: This is not needed.  Removed it.

nickwallen:
I found a problem with this change.  The Alerts UI does not properly display comments when indexing into Elasticsearch.  See screenshot.

Screen Shot 2019-07-02 at 3 02 15 PM

The problem is that we serialize comments into Elasticsearch differently than we do Solr.  For Elasticsearch we write the comments as a list of maps.  For Solr, we write a list of JSON strings.  I do not understand why we serialize these differently, but I definitely don't want to change that behavior in this PR.

With this change, it wrote the comments as a list of JSON strings for both Solr and Elasticsearch.  When running against Elasticsearch, the Alerts UI is then passed a raw JSON string that it does not know how to handle; hence the"Invalid Date" error in the Alerts UI.

To really fix this, I would need to refactor deeper into the Elasticsearch side of the DAOs and have them also use the `DocumentBuilder` abstraction.  I would also need to add a test to catch this issue.

As mentioned in the PR description we (de)serialize comments in many different places and this should ultimately be cleaned up and more thoroughly tested.  I am going to shelve this refactor for later.  Until then, the minimal necessary fixes to test current behavior is in #1456 .  Its less than ideal, but it gets the job done.

PR #1457, opened 2019-07-03
METRON-2176 Upgrade REST for HBase 2.0.2

nickwallen:
This change updates the REST service to work with HBase 2.0.2
* This PR is for the `feature/METRON-2088-support-HDP-3.1` feature branch.
* This PR is dependent on #1456 and #1454 . The diff will show those changes here until that PR is merged.

## Changes

* Modified the `UserSettingsClient` to use the `HBaseClient` abstraction.
* This drove some of the additional operations that were added to the `HBaseClient` in #1456.
* Made the `IndexDao` extend `Closeable` so that when the DAOs are instantiated via Spring, there is a way to clean-up resources, like HBase connections.
* Modified the `HBaseDao` and related tests to work with HBase 2.0.2.
* Created the `FakeHBaseClient` which makes testing anything that interacts with HBase, like `HBaseSettingsClient` possible.
   * While this class is more complex than I'd like, I think its pros outweigh the cons.  It will be used in quite a few places to enable testing of classes that would normally need to interact with HBase. It is also an improvement over the previous "Mock HBase" code because its implementation does not rely on any HBase APIs.

## Acceptance Testing

This should not impact the deployment and functioning of the system.  As a sanity check, spin-up Full Dev and ensure alerts are visible.

In addition, you can follow these steps to save and retrieve a user setting.

1. Go to the Swagger UI.

1. Retrieve the current user's settings using `GET /api/v1/alerts/ui/settings`. This should return 404 initially indicating there are no settings for the current user.

1. Save a user setting.  Pass the following as the `alertsUIUserSettings` body to `POST /api/v1/alerts/ui/settings`.
    ```
    {
      "facetFields": [
        "ip_src_addr"
      ]
      "user": "admin"
    }
    ```

1. Retrieve the setting again using `GET /api/v1/alerts/ui/settings`.  This should return 200 with the following response body
    ```
    {
      "user": "admin",
      "tableColumns": null,
      "savedSearches": null,
      "facetFields": [
        "ip_src_addr"
      ]
    }
    ```

## Pull Request Checklist


metron-interface/metron-rest/src/main/java/org/apache/metron/rest/service/impl/AlertsUIServiceImpl.java
```diff
@@ -136,12 +136,15 @@ public void saveAlertsUIUserSettings(AlertsUIUserSettings alertsUIUserSettings)
   }
 
   @Override
-  public boolean deleteAlertsUIUserSettings(String user) {
-    boolean success = true;
+  public boolean deleteAlertsUIUserSettings(String user) throws RestException {
+    boolean success = false;
     try {
-      userSettingsClient.delete(user, ALERT_USER_SETTING_TYPE);
+      if(userSettingsClient.findOne(user, ALERT_USER_SETTING_TYPE).isPresent()) {
```
nickwallen: This was needed to maintain the behavior of returning false if you attempt to delete settings that don't exist.  When issuing the delete you don't get any indication of whether something was deleted or not.

metron-interface/metron-rest/src/test/java/org/apache/metron/rest/service/impl/AlertsUIServiceImplTest.java
```diff
@@ -126,63 +127,63 @@ public void escalateAlertShouldSendMessageToKafka() throws Exception {
...
-    expectedAlertsUIUserSettings.setTableColumns(Collections.singletonList("user1_field"));
-    assertEquals(expectedAlertsUIUserSettings, alertsUIService.getAlertsUIUserSettings().get());
-    verify(userSettingsClient, times(1)).findOne(user1, AlertsUIServiceImpl.ALERT_USER_SETTING_TYPE);
-    verifyNoMoreInteractions(userSettingsClient);
+  public void shouldGetActiveProfile() throws Exception {
+    AlertsUIUserSettings expected = new AlertsUIUserSettings();
+    expected.setTableColumns(Collections.singletonList("user1_field"));
+
+    // save a profile for current user
+    userSettingsClient.save(user1, ALERT_USER_SETTING_TYPE, toJSON(expected));
```
nickwallen: Being able to use a real `HBaseSettingsClient` instead of a mock made these tests much easier to grok and maintain in my opinion.  Previously we had to use a lot of mockito-magic to test this.

merrimanr:
I ran through the test instructions and everything works as expected.  I also tested the admin user settings endpoints and those also work as expected.

After seeing the `FakeHBaseClient` class in context, I agree there is no easy way to do this.  This approach is nice because we can use the same underlying classes for multiple tests.  Otherwise we would need to mock each test individually.  I'm am satisfied with the tradeoff you have chosen.  +1

nickwallen:
Thanks for the review.  This has been merged into `feature/METRON-2088-support-hdp-3.1`.

PR #1460, opened 2019-07-11
METRON-2179: Make navigation in both UIs consistent

sardell:
## Contributor Comments
READ FIRST: This Pull Request seems very large, but a majority of the line changes are in the auto-generated package-lock.json file. There are maybe a one or two hundred lines of code changes outside of that auto-generated file. Now, breathe a sigh of relief. :smile:

Link to original JIRA ticket here: https://issues.apache.org/jira/browse/METRON-2179

Currently, both UIs approach navigation in two different ways. This makes the interfaces seem inconsistent and can be confusing for new users. Both ways are also not collapsible. We also cannot navigate between both UIs (this was intentional in the original implementation, but feedback from users over time seems to indicate separating the personas is not necessarily beneficial).

This is how the navigations currently look on master:

Screen Shot 2019-07-11 at 11 42 38 AM

Screen Shot 2019-07-11 at 11 42 50 AM

In this PR, I added a collapsible sidebar navigation and removed the previous implemented navigations.

Screen Shot 2019-07-11 at 10 18 17 AM

Screen Shot 2019-07-11 at 10 18 27 AM

In order to do this, I utilized the Angular-version of [Ant Design's component library](https://ng.ant.design/docs/introduce/en). I did this because bootstrap does not contain a collapsible side navigation like this, and Ant Design allowed me to quickly implement this feature.

I also had to upgrade our version of Angular to the latest stable 7 release. Upgrading from 6 to 7 did not introduce any breaking changes.

## Testing

Open either the Alerts or Management UI. A navigation should exist on the left-hand side of the screen (see above screenshots), and you should be able to minimize it by clicking on the arrow at the bottom of the menu. Besides navigating within the app, you should be able to navigate between both UIs.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel).
[...]

PR #1464, opened 2019-07-16
METRON-2179: Make navigation in both UIs consistent

sardell:
## Contributor Comments
Link to original JIRA ticket here: https://issues.apache.org/jira/browse/METRON-2179

A PR was already opened for this a couple of days ago, but after some discussion it was decided that I should split out the Angular upgrade work from the central navigation implementation. To catch up on the discussion, you can check it out here: https://github.com/apache/metron/pull/1460

Currently, both UIs approach navigation in two different ways. This makes the interfaces seem inconsistent and can be confusing for new users. Both ways are also not collapsible. We also cannot navigate between both UIs (this was intentional in the original implementation, but feedback from users over time seems to indicate separating the personas is not necessarily beneficial).

This is how the navigations currently look on master:

Screen Shot 2019-07-11 at 11 42 38 AM

Screen Shot 2019-07-11 at 11 42 50 AM

In this PR, I added a collapsible sidebar navigation and removed the previous implemented navigations.

Screen Shot 2019-07-11 at 10 18 17 AM

Screen Shot 2019-07-11 at 10 18 27 AM

In order to do this, I utilized the Angular-version of [Ant Design's component library](https://ng.ant.design/docs/introduce/en). I did this because bootstrap does not contain a collapsible side navigation like this, and Ant Design allowed me to quickly implement this feature.

## Testing

Open either the Alerts or Management UI. A navigation should exist on the left-hand side of the screen (see above screenshots), and you should be able to minimize it by clicking on the arrow at the bottom of the menu. Besides navigating within the app, you should be able to navigate between both UIs.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

- [x] N/A ~Have you ensured that format looks appropriate for the output in which it is rendered by building and verifying the site-book? If not then run the following commands and the verify changes via `site-book/target/site/index.html`:~

  ```
  cd site-book
  mvn site
  ```

- [x] N/A ~Have you ensured that any documentation diagrams have been updated, along with their source files, using [draw.io](https://www.draw.io/)? See [Metron Development Guidelines](https://cwiki.apache.org/confluence/display/METRON/Development+Guidelines) for instructions.~

#### Note:


metron-interface/metron-alerts/src/app/app.component.ts
```diff
@@ -27,6 +27,40 @@ import { environment } from 'environments/environment';
...
+        }
+      ]
+    },
+    {
+      linkName: 'Management',
+      iconClass: 'tool',
+      subLinks: [
+        {
+          linkName: 'Sensors',
+          routerLink: ':4200/sensors',
```
mmiklavc: Is this the right place to hardcode a port number?

metron-interface/metron-alerts/src/app/app.component.ts
```diff
@@ -27,6 +27,40 @@ import { environment } from 'environments/environment';
...
+          externalLink: true
+        },
+        {
+          linkName: 'General Settings',
+          routerLink: ':4200/general-settings',
```
mmiklavc: Is this the right place to hardcode a port number?

metron-interface/metron-config/src/app/app.component.ts
```diff
@@ -28,6 +28,40 @@ import {AuthenticationService} from './service/authentication.service';
...
+  isCollapsed = false;
+  hostname = window.location.hostname;
+  centralNavLinks = [
+    {
+      linkName: 'Alerts',
+      iconClass: 'warning',
+      subLinks: [
+        {
+          linkName: 'Overview',
+          routerLink: ':4201/alerts-list',
```
mmiklavc: Is this the right place to hardcode a port number?

metron-interface/metron-config/src/app/app.component.ts
```diff
@@ -28,6 +28,40 @@ import {AuthenticationService} from './service/authentication.service';
...
+          externalLink: true
+        },
+        {
+          linkName: 'PCAP',
+          routerLink: ':4201/pcap',
```
mmiklavc: Is this the right place to hardcode a port number?

sardell:
@mmiklavc Good point. Let me take a look and update this weekend.

sardell:
@mmiklavc After taking a moment to look at Swagger, I do not see a REST endpoint where I can access the dynamic port values for both UIs. I figured this might be the case since we don't need those in the current UI implementations. So, now my question is if I cannot obtain these values currently, how much work is required to make this happen?

Including @merrimanr probably has some insight on this as well.

tiborm:
@mmiklavc is there a way to get the port numbers from Ambari?

nickwallen:
@tiborm Yes, you can.  See the [Ansible scripts](https://github.com/apache/metron/blob/1245fd160a5ae826a8e2900528ab12c64350fdc6/metron-deployment/ansible/roles/ambari_gather_facts/tasks/main.yml#L122) for an example.

mmiklavc:
Did a bit of sleuthing here @tiborm and @sardell. Here's the chain of variable resolutions that should be able to provide this in some way:

* Gets values from Ambari - https://github.com/apache/metron/blob/master/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/package/templates/management_ui.yml.j2
* The config UI startup script, which pulls in the env vars from the defaults file as well as leverages the yml config file - https://github.com/apache/metron/blob/master/metron-interface/metron-config/scripts/metron-management-ui

That management_ui.yml should provide you access to the port in the same way it provides access to the REST API host and port.

sardell:
@mmiklavc I think I'm having trouble following what you're proposing above. How would I be able to handle links between interfaces with the values in that configuration? For example, links to the Management UI from the Alerts UI and vice versa.

mmiklavc:
@sardell - we're writing out the host:port info in those config files via Ambari. Probably the easiest thing to do here is expose the additional properties in the respective UI configs. So each UI config would have 1..n config details for connections to the other service(s). If you want to tag-team that, maybe @merrimanr or I can walk you through it. Should be pretty easy to test, also - it's either going to work, or it's not, ha.

This UI menu approach looks like it has the potential to start moving us towards a more pluggable UI architecture where we have more than just 2 UI services. There has been some loosely guided discussion in the past about microservices as well as containerizing certain components/services. I spoke a bit offline with @merrimanr about this. I'll let him talk about this in more detail, but one option for moving in this direction is to use a gateway to manage the endpoints. Of course, there are other options as well, e.g. web containers have had application isolation for ages now. I don't think any of that's required for this PR, but it seems like a good time to start some community discussion around it.

sardell:
Alright, @mmiklavc. I think we are now covered for customized ports thanks to @merrimanr's contribution. Thank you both for helping me out with that task!

mmiklavc:
Nice work guys! I'm +1 pending a successful Travis build.

merrimanr:
Looks like the build timed out.  I noticed this same issue in the PR I submitted against this.  Could it be specific to this PR?  @sardell you might want to try closing and reopening.

sardell:
@merrimanr I closed and reopened, but it looks like we are still timing out. It's odd because it looks to me like the build is successful in the Travis logs.

https://travis-ci.org/apache/metron/jobs/566665244

sardell:
According to traviscistatus.com, there were some network issues last week that could've caused the mvn download error in the latest Travis build. I'm going to close and rebuild since it looks like they've resolved the issue.

sardell:
Well, it looks everything builds fine now. However, upon one last inspection, I noticed that the menu is still rendering _sometimes_ when a user is logged out (it's hard to catch as it's mostly invisible). See the screenshot below.

Screen Shot 2019-08-15 at 4 28 35 PM

Obviously, this needs to be fixed before merging. So far, I can only make this happen when I toggle between UIs on full dev. I haven't been able to reproduce this when serving the UIs manually from a terminal window. Keeping my fingers crossed that I can figure this out through some debugging in dev tools.

mmiklavc:
@sardell good catch. Thank you for the attention to detail on this.

sardell:
@mmiklavc After clearing my browser's cache and testing again for quite a while, I'm unable to reproduce the bug mentioned previously. If you're still a +1, I'll merge this in.

mmiklavc:
Yes, I'm still +1. Good work @sardell

PR #1465, opened 2019-07-16
METRON-2172 Solr Updates Not Tested in Integration Test

nickwallen:
The `SolrUpdateIntegrationTest` is not testing that the `SolrDao` can update and retrieve values. This fix is similar to what was done for Elasticsearch in #1451.

This replaces #1454 which was a more involved refactor that fixed some issues with code duplication and provided additional unit tests.  I am going to break-off just the bare minimum of what we need to fix the broken tests, rather than fix and refactor in the same PR.

### What?
This gap in the integration test is hiding a couple bugs.  

1. A timestamp is not being populated in the documents returned from Solr.

1. The integration test is not accounting for the fact that comments are serialized differently when written to Solr versus Elasticsearch.  When comments are written to Solr, they are serialized as a list of JSON strings.  When comments are written to Elasticsearch, they are serialized as a list of maps.

### Why? 
These problems arise because of the way the test is setup.  Instead of directly testing a `SolrDao` this test runs against a `MultiIndexDao` initialized with both a `SolrDao` and an `HBaseDao`. On retrievals the `MultIndexDao` will return the document from whichever index responds first.

With the current test setup, the underlying `SolrDao` will never retrieve the document that the test case is expecting.  In all cases where the test passes, the document is actually being returned from the `HBaseDao` which is actually just interacting with a mock backend.  The test needs to actually test that we can update and retrieve documents from Elasticsearch.

### Changes

1. The `SolrUpdateIntegrationTest` was changed so that it specifically tests a `SolrDao`.

1. The `UpdateIntegrationTest` was changed so that it can handle comments stored as either strings or maps.  This allows the same integration tests to continue to work when run against either Solr or Elasticsearch.

### Acceptance Testing

- [x] Validate against Elasticsearch.
    1. Spin-up Full Dev with Elasticsearch running.
    1. Ensure alerts are visible in the Alerts UI.
    1. Add a comment to an alert.

- [x] Validate against Solr.
    1. Spin-up Full Dev with Solr running.
    1. Ensure alerts are visible in the Alerts UI.
    1. Add a comment to an alert.

## Pull Request Checklist

### DIFF
--- a/metron-platform/metron-indexing/metron-indexing-common/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
+++ b/metron-platform/metron-indexing/metron-indexing-common/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
@@ -303,14 +303,14 @@ public abstract class UpdateIntegrationTest {
   }
 
   /**
-   * Reformats the format of stored comments.
+   * Normalizes the format of stored comments.
    *
    * <p>Comments are serialized differently when stored in Elasticsearch and Solr.  Comments
    * are stored as maps in Elasticsearch and JSON strings in Solr.  This reformats all comments
    * as maps, so they look the same when validation occurs in the integration tests.
-   * @param fields
+   * @param fields The fields of a document that may contain comments.
    */
-  protected static void reformatComments(Map<String, Object> fields) {
+  protected static void normalizeCommentsAsMap(Map<String, Object> fields) {
     @SuppressWarnings("unchecked")
     List<Object> commentValues = (List<Object>) fields.get(COMMENTS_FIELD);
     if (commentValues != null) {
@@ -338,7 +338,7 @@ public abstract class UpdateIntegrationTest {
   protected Document findUpdatedDoc(Map<String, Object> expected, String guid, String sensorType)
       throws InterruptedException, IOException, OriginalNotFoundException {
     // comments are stored differently in Solr and Elasticsearch
-    reformatComments(expected);
+    normalizeCommentsAsMap(expected);
 
     for (int t = 0; t < MAX_RETRIES; ++t, Thread.sleep(SLEEP_MS)) {
       Document found = getDao().getLatest(guid, sensorType);
--- a/metron-platform/metron-indexing/metron-indexing-common/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
+++ b/metron-platform/metron-indexing/metron-indexing-common/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
@@ -17,12 +17,12 @@ package org.apache.metron.indexing.dao;
 import org.adrianwalker.multilinestring.Multiline;
 import org.apache.commons.collections.MapUtils;
 import org.apache.metron.common.Constants;
-import org.apache.metron.hbase.mock.MockHTable;
 import org.apache.metron.indexing.dao.search.AlertComment;
 import org.apache.metron.indexing.dao.update.CommentAddRemoveRequest;
 import org.apache.metron.indexing.dao.update.Document;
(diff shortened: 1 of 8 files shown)


metron-platform/metron-solr/metron-solr-common/src/test/java/org/apache/metron/solr/dao/SolrSearchDaoTest.java
```diff
@@ -510,5 +512,13 @@ public void getGroupResultsShouldProperlyReturnGroupResults() {
     assertNull(level2GroupResults.get(1).getGroupResults());
   }
 
-
+  private SolrDocument createSolrDocument(String sensorType, Long timestamp) {
```
mmiklavc: Thank you for this - simple types like this should not be mocked. Good improvement.

metron-platform/metron-solr/metron-solr-common/src/test/java/org/apache/metron/solr/dao/SolrSearchDaoTest.java
```diff
@@ -510,5 +512,13 @@ public void getGroupResultsShouldProperlyReturnGroupResults() {
...
+    SolrDocument solrDocument = new SolrDocument();
+    solrDocument.addField(SolrDao.VERSION_FIELD, 1.0);
+    solrDocument.addField(Constants.GUID, UUID.randomUUID().toString());
+    solrDocument.addField(Constants.SENSOR_TYPE, sensorType);
+    solrDocument.addField(Constants.Fields.TIMESTAMP.getName(), timestamp);
+    solrDocument.addField("ip_src_addr", "192.168.1.1");
```
mmiklavc: `Constants.Fields.SRC_ADDR`

https://github.com/apache/metron/blob/3754ff33f6cd149ffca57474d744e0298d4c172a/metron-platform/metron-common/src/main/java/org/apache/metron/common/Constants.java#L52

metron-platform/metron-indexing/metron-indexing-common/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
```diff
@@ -272,29 +273,81 @@ protected Document createDocument(String guid, Long timestamp) {
...
-  protected Document findUpdatedDoc(Map<String, Object> message0, String guid, String sensorType)
+  /**
+   * Reformats the format of stored comments.
+   *
+   * <p>Comments are serialized differently when stored in Elasticsearch and Solr.  Comments
+   * are stored as maps in Elasticsearch and JSON strings in Solr.  This reformats all comments
+   * as maps, so they look the same when validation occurs in the integration tests.
+   * @param fields
+   */
+  protected static void reformatComments(Map<String, Object> fields) {
```
mmiklavc: How about `normalizeCommentsAsMap` for a name?

mmiklavc:
Looks pretty good. A couple small nits and I think this is good to go.

metron-platform/metron-solr/metron-solr-common/src/test/java/org/apache/metron/solr/dao/SolrSearchDaoTest.java
nickwallen: Done. Thanks

metron-platform/metron-indexing/metron-indexing-common/src/test/java/org/apache/metron/indexing/dao/UpdateIntegrationTest.java
nickwallen: That is more descriptive. Done.

mmiklavc:
+1 thanks @nickwallen

PR #1472, opened 2019-07-29
METRON-2195 lazy logging

tigerquoll:
## Contributor Comments
This PR illustrates a proof on concept for a logging wrapper that supports lazy argument evaluation.   Lazy parameter evaluation technique is automatically supported in Java 8 native logging (which we don't use), and log4j 2.4+ (which we appear to use via SLF4J wrappers). 
Unfortunately the lazy parameter evaluation provided by log4j 2.4+ is not currently exposed in the interfaces provided by the current SLF4J logging framework (support is scheduled for release with SLF4J 2.x)  So to support lazy evaluation while maintaining use of the current SLF4j logging libraries a wrapper is required.

## Rationale for lazy parameter evaluation
Even though the logging framework will not construct a string unless the logging level indicates a log message will be needed, all parameter arguments are still evaluated by the JVM as part of the standard java parameter passing process.  

Lazy parameter evaluation is a useful technique for when some arguments are very expensive to calculate (eg. Serialising a collection object into a JSON string).   When disabled logging calls are deeply nested under multiple levels of iteration, then there can be a real hidden cost to this unneeded parameter evaluation overhead.

This PR illustrates a technique of utilising a logging wrapper that supports lambdas as logging parameters. Typical scenarios where this could be used include code scenarios such as:

<code>
LOG.trace("Printed result of an expensive operation: {}", expensiveOperation().toString())

</code>

Where ```expensiveOperation().toString``` is executed regardless of what logging level is enabled.

The suggested code to use with this wrapper would be
<code>
private static final LazyLogger LOG = LazyLoggerFactory.getLogger(MethodHandles.lookup().lookupClass());

LOG.trace("Expensive operation only if trace logging enabled: {}", () -> expensiveOperation().toString()
</code>

As an example of usage  
```metron-platform/metron-writer/metron-writer-storm/src/main/java/org/apache/metron/writer/hdfs/HdfsWriter.java```
has been updated to utilise this wrapper module.

The general guidance would be that any logging use that involves anything more than extremely basic string construction should use the lambda versions of the logging functions for performance reasons.  
Error-level logging would be exempt from the requirement to use lambda versions of the logging functions due to error logging usually never being disabled, and thus no performance benefit is likely to come from forcing lazy parameter evaluation in this case.

If the lazy evaluation logging concept is approved I will add some unit testing before finalising the PR.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:


ottobackwards:
Can you rebase this on master?

ottobackwards:
I would like to see the results from an extended version of @mmiklavc 's performance test as applied after this change, IE a test using the lazy logger.

I think after the initial warm up, the times shouldn't be bad.

ottobackwards:
I like this a lot,  I think you should continue with the PR.

If you are going to limit the scope of this pr, please state that in the PR.

tigerquoll:
Rebasing caused a sea of errors. It was easier just to regenerate a new pull request.

PR #1515, opened 2019-09-19
METRON-2261 Isolate Curator Dependencies

nickwallen:
As part of the HDP 3.1 upgrade, we need to upgrade to Curator 4.x.  There was a [discuss thread](https://lists.apache.org/thread.html/265a76f6071229af662135b92df6db652d8ee2b17dbec8af7e8bf1ba@%3Cdev.metron.apache.org%3E)  covering the need for this.

Currently, Curator is being pulled in as a transitive of Hadoop, HBase and Storm.  At the current version levels of these dependencies this does not cause a problem.  But when upgrading Hadoop, different conflicting versions of Curator will get pulled in.  We need to ensure that only a single version of Curator is pulled in after upgrading Hadoop.  

This change is preparing us for follow-on PRs including...
1. [Upgrade to Curator 4.x](https://issues.apache.org/jira/browse/METRON-2262) #1516 
2. [Upgrade to Hadoop 3.1.1](https://issues.apache.org/jira/browse/METRON-2232).

## Changes

* This change will maintain the current versions in use including Curator 2.7.1 and Zookeeper 3.4.6 (see note 1).

* Defines Zookeeper as a direct dependency instead of relying on it being pulled in as a transitive of Curator (see note 2).

* Defines Curator as a direct dependency instead of relying on it being pulled in as a transitive.

* Introduces a separate `${global_curator_test_version}` to allow the Curator test dependencies to differ from the main Curator version `${global_curator_version}`.  This will be needed when upgrading to Curator 4.x (see note 3).

#### Notes

1.  In upgrading to Curator 4.x, there is a breaking change that causes the MaasIntegrationTest to fail.  The fix for that plus the actual upgrade to Curator 4.x will be performed under a separate Jira to aid the review process.

2. When upgrading to Curator 4.x, we will need to continue to run against Zookeeper 3.4.6.   By default Curator 4.x will pull in Zookeeper 3.5.x.

3. To run with Curator 4.x and Zookeeper 3.4.6, we need to use the 2.12.0 version of curator-test based on [this information provided by the Curator community](http://curator.apache.org/zk-compatibility.html).

## Acceptance Testing

This PR should be tested using the centos7 development environment.  

1. Start up the centos7 dev environment.
    ```
    cd metron-deployment/development/centos7
    vagrant destroy -f
    vagrant up
    ```

### Basics

Ensure that we can continue to parse, enrich, and index telemetry.  Verify data is flowing through the system, from parsing to indexing

[...]

PR #1516, opened 2019-09-19
METRON-2262: Upgrade to Curator 4.2.0

merrimanr:
## Contributor Comments
This PR is based on [METRON-2261](https://github.com/apache/metron/pull/1515) and includes those changes.  It must be merged before this.

This PR upgrades Curator to version 4.2.0 and fixes a couple bugs in `MaasIntegrationTest` that resulted from the upgrade.  There was an extra service unregister in the `ContainerRequestListener` class (should only be unregistered once) and another more complicated bug (described in the next section).

### Bug Fix
After upgrading to Curator version 4.2.0, both tests in `MaasIntegrationTest` fail with this error:
```
testMaaSWithoutDomain(org.apache.metron.maas.service.MaasIntegrationTest)  Time elapsed: 214.791 sec  <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at org.apache.metron.maas.service.MaasIntegrationTest.testDSShell(MaasIntegrationTest.java:269)
	at org.apache.metron.maas.service.MaasIntegrationTest.testMaaSWithoutDomain(MaasIntegrationTest.java:113)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
``` 
The service is not properly unregistered, the test [waits for the service to be removed from the list of endpoints](https://github.com/apache/metron/blob/master/metron-analytics/metron-maas-service/src/test/java/org/apache/metron/maas/service/MaasIntegrationTest.java#L261) and then timeouts and fails.

The root cause is a change in how the `org.apache.curator.x.discovery.details. ServiceDiscoveryImpl` class unregisters a service after the upgrade.  The list of services is maintained in 2 places:  Zookeeper and an [internal Map](https://github.com/apache/curator/blob/apache-curator-4.2.0/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java#L64) in `ServiceDiscoveryImpl`.  There are several read functions in `ServiceDiscoveryImpl` but then don't refresh the internal Map.  The internal Map is only updated when `registerService` and `unregisterService` are called. 

Before the upgrade, a service is [removed from Zookeeper](https://github.com/apache/curator/blob/apache-curator-2.7.0/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java#L218) then [removed from the internal Map](https://github.com/apache/curator/blob/apache-curator-2.7.0/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java#L224).

After the upgrade, the order is reversed.  The service is [removed from the internal Map](https://github.com/apache/curator/blob/apache-curator-4.2.0/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java#L255) then [removed from Zookeeper](https://github.com/apache/curator/blob/apache-curator-4.2.0/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java#L524).  However there is a check that [ensures the service existed in the internal Map](https://github.com/apache/curator/blob/apache-curator-4.2.0/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java#L511) before removing it for Zookeeper.  Services are registered in `org.apache.metron.maas.service.runner.Runner` and unregistered in `org.apache.metron.maas.service.callback.ContainerRequestListener`.  The latter runs in a different process so registered services are not in the internal Map.  This was not a problem before the upgrade because Zookeeper was updated regardless of the internal Map state.

The fix proposed in this PR is to register a service immediately before unregistering it in `org.apache.metron.maas.discovery.ServiceDiscoverer`.  This populates the internal Map in `ServiceDiscoveryImpl` with the service so that the subsequent unregister call functions correctly.  A side effect of this would be writing duplicate data to Zookeeper.  The only other solution I can think of would be to rearchitect how and where services are registered/unregistered.

### Testing
The `MaasIntegrationTest` (and all tests) should pass.  Upgrading Curator without the fixes in this PR should cause the error above.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-analytics/metron-maas-common/pom.xml
+++ b/metron-analytics/metron-maas-common/pom.xml
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- 
-  Licensed to the Apache Software 
-  Foundation (ASF) under one or more contributor license agreements. See the 
-  NOTICE file distributed with this work for additional information regarding 
-  copyright ownership. The ASF licenses this file to You under the Apache License, 
-  Version 2.0 (the "License"); you may not use this file except in compliance 
-  with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0 
-  Unless required by applicable law or agreed to in writing, software distributed 
-  under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES 
-  OR CONDITIONS OF ANY KIND, either express or implied. See the License for 
-  the specific language governing permissions and limitations under the License. 
+<!--
+  Licensed to the Apache Software
+  Foundation (ASF) under one or more contributor license agreements. See the
+  NOTICE file distributed with this work for additional information regarding
+  copyright ownership. The ASF licenses this file to You under the Apache License,
+  Version 2.0 (the "License"); you may not use this file except in compliance
+  with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software distributed
+  under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+  OR CONDITIONS OF ANY KIND, either express or implied. See the License for
+  the specific language governing permissions and limitations under the License.
   -->
 
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
@@ -54,11 +54,25 @@
       <groupId>org.apache.curator</groupId>
       <artifactId>curator-recipes</artifactId>
       <version>${global_curator_version}</version>
+      <exclusions>
+        <exclusion>
+          <!-- need to exclude zookeeper 3.5.x http://curator.apache.org/zk-compatibility.html -->
+          <groupId>org.apache.zookeeper</groupId>
+          <artifactId>zookeeper</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.curator</groupId>
       <artifactId>curator-x-discovery</artifactId>
       <version>${global_curator_version}</version>
+      <exclusions>
+        <exclusion>
+          <!-- need to exclude zookeeper 3.5.x http://curator.apache.org/zk-compatibility.html -->
+          <groupId>org.apache.zookeeper</groupId>
+          <artifactId>zookeeper</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
@@ -83,8 +97,15 @@
     <dependency>
       <groupId>org.apache.curator</groupId>
       <artifactId>curator-test</artifactId>
-      <version>${curator.version}</version>
+      <version>${global_curator_test_version}</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <!-- need to exclude zookeeper 3.5.x http://curator.apache.org/zk-compatibility.html -->
+          <groupId>org.apache.zookeeper</groupId>
+          <artifactId>zookeeper</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
   </dependencies>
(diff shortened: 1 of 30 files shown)


mmiklavc:
@merrimanr This is pretty complicated so I want to talk through this a bit and make sure this is clear. 

It looks like the upshot is that previously the functionality only depended on shared state in Zookeeper. You could reference a `ServiceDiscoveryImpl` in 2 separate processes (1 in the integration test process, the other in the application master container process spawned by YARN) with no issue. Now, the internal map matters, but unlike the zookeeper path this state is not distributed. So that verification fails when there's no item in the internal map.

The integration test creates a `ServiceDiscover` here https://github.com/apache/metron/blob/master/metron-analytics/metron-maas-service/src/test/java/org/apache/metron/maas/service/MaasIntegrationTest.java#L218

The YARN app creates a `ServiceDiscover` here when it creates the `ContainerRequestListener ` https://github.com/apache/metron/blob/master/metron-analytics/metron-maas-service/src/main/java/org/apache/metron/maas/service/ApplicationMaster.java#L454

merrimanr:
Almost all correct.  There are other `ServiceDiscoverer` instances created [here](https://github.com/apache/metron/blob/master/metron-analytics/metron-maas-service/src/main/java/org/apache/metron/maas/service/runner/Runner.java#L193) and [here](https://github.com/apache/metron/blob/master/metron-analytics/metron-maas-common/src/main/java/org/apache/metron/maas/functions/MaaSFunctions.java#L229).  Here's what each one does:

`MaasIntegrationTest` - Used to test which endpoints are registered.  Runs in test.
`MaaSFunctions` - Used to get endpoint.  Runs in a Stellar context.
`Runner` - Registers service endpoint when container starts. Runs in model container.
`ContainerRequestListener ` - Unregisters services based on container callbacks. Runs in app master.

mmiklavc:
> Almost all correct. There are other `ServiceDiscoverer` instances created [here](https://github.com/apache/metron/blob/master/metron-analytics/metron-maas-service/src/main/java/org/apache/metron/maas/service/runner/Runner.java#L193) and [here](https://github.com/apache/metron/blob/master/metron-analytics/metron-maas-common/src/main/java/org/apache/metron/maas/functions/MaaSFunctions.java#L229). Here's what each one does:
> 
> `MaasIntegrationTest` - Used to test which endpoints are registered. Runs in test.
> `MaaSFunctions` - Used to get endpoint. Runs in a Stellar context.
> `Runner` - Registers service endpoint when container starts. Runs in model container.
> `ContainerRequestListener ` - Unregisters services based on container callbacks. Runs in app master.

Oh, that's actually more than I thought without having run the lookup on uses. Glad I asked. 

One manual test we might want to add is attempting to remove a non-existent endpoint. Looking at the code, I don't see a whole lot of risk in the "add then remove" approach other than if the operation were to fail in between add/remove. For a service that doesn't exist, you could end up with an item in zookeeper that doesn't actually exist.

ottobackwards:
So the bug is that we register and un-register in different processes. 

I think we could think about 

- Implementing our own ServiceDiscovery<T> based on the existing impl with our reqs.
- Filing a jira and maybe a PR to curator about the use case.  Although I think registration and unregistration from different processes is kind of a corner case.

ottobackwards:
- another idea would be for them to support multiple implementations with the builder and have a non-caching, distributed friendly version etc.

Assuming the logic for using this way makes sense

nickwallen:
Hey @merrimanr - We don't need the following changes on this PR.  I know these are just due to the iterations and changes I went through on #1515.  
* metron-platform/metron-indexing/metron-indexing-storm/pom.xml
* metron-platform/metron-enrichment/pom.xml
* metron-stellar/stellar-zeppelin/pom.xml

merrimanr:
Ah makes sense @nickwallen.  Done with latest commit.

PR #1527, opened 2019-10-07
METRON-2276: [UI] Performance: Switching back from manual filtering fires a "full query"

ruffle1986:
## Contributor Comments

Link to the original Jira: https://issues.apache.org/jira/browse/METRON-2276

The problem
=========

Every time the user switches back from "manual query mode" (to the query builder), we clear the filter and remove everything from the search input field. Even when there were something typed in the query builder previously.

The solution
=========

It would be great if we save the value of the search field no matter if it's the query builder of the manual query mode. 
So instead of firing a search event every time the user switches back from manual mode, let's just simply not fire it so we can see the previous value for the query builder. And vica versa.

Previously, when the ace editor fired a "textChanged" event, we immediately called the search method. From now on, there's a different callback on the textChanged event which puts the application state in stale mode. 
This is how it's supposed to work currently. On every change, the app should be in stale mode (there's a yellow warning message in the middle of the screen). The only way, to see the result of a new query is pressing the magnifier button (search button) on the right. 

Previously that wasn't true if the user removed a query item in the query builder by pressing the tiny x icon on the right side of the query item. With this patch, this issue is going to be fixed too which means every time you remove a query item, the app is going to be in stale mode instead of performing a search immediately. 

Although it's not the scope of this PR, I've found an issue with the manual query mode. The app generates the "raw" representation of the query builder's result when the user switches to manual mode BUT only for the first time when there's no previous manual mode value:

https://github.com/apache/metron/blob/master/metron-interface/metron-alerts/src/app/alerts/alerts-list/query-builder.ts#L177-L179

So when the user decides to go back to the query builder, change it and then go back to manual mode, the raw value in the manual mode will be no longer the proper representation of the query builder's value.

Testing
=====

- Go the alerts-list page on the Alerts UI.
- Hit the search button in order to get some result. (At the beginning, there's no result for the current value. The user has to press the search button. It should be in stale mode and show the yellow warning message)
- Build a nice query with the query builder (for example by clicking on cell values in the table)
- Go to manual mode and than go back to the query builder.
- You should be able the see the previous value of the query builder and since nothing has changed, the application shouldn't be in stale mode. 
- If you change the query with the builder by adding of removing new query items, it's gonna turn into stale mode obviously without searching immediately.

Although it solves the performance issue and prevents the users from dosing the servers, there's some limitations in the manual query mode. When you change something in the manual query mode, the application doesn't turn into stale mode however it should. And every time you change something in the manual mode and go back to the query builder, currently, the app is unable to convert the manual query to the value that is accepted by query builder which means we are unable to switch back and forth and always see the proper value in the search field in the appropriate format based on if it's the query builder or the manual mode.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

- [X] Have you ensured that any documentation diagrams have been updated, along with their source files, using [draw.io](https://www.draw.io/)? See [Metron Development Guidelines](https://cwiki.apache.org/confluence/display/METRON/Development+Guidelines) for instructions.

#### Note:

### DIFF
--- a/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.html
+++ b/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.html
@@ -20,7 +20,14 @@
                     <span class="input-group-prepend">
                         <button class="btn btn-secondary btn-saved-searches" type="button" (click)="showSavedSearches()">Searches</button>
                     </span>
-                    <div appAceEditor *ngIf="!isQueryBuilderModeManual()" class="flex-fill" placeholder="Search Alerts" [text]="queryBuilder.displayQuery" (textChanged)="onSearch($event)"> </div>
+                    <div
+                      appAceEditor
+                      *ngIf="!isQueryBuilderModeManual()"
+                      class="flex-fill"
+                      placeholder="Search Alerts"
+                      [text]="queryBuilder.displayQuery"
+                      (textChanged)="onBuilderQueryChanged($event)"
+                    > </div>
                     <div class="flex-fill" *ngIf="isQueryBuilderModeManual()">
                         <input #manualQuery type="text"
                             class="manual-query-input"
--- a/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
+++ b/metron-interface/metron-alerts/src/app/alerts/alerts-list/alerts-list.component.ts
@@ -560,14 +560,15 @@ export class AlertsListComponent implements OnInit, OnDestroy {
       this.queryBuilder.setFilteringMode(FilteringMode.MANUAL);
     } else {
       this.queryBuilder.setFilteringMode(FilteringMode.BUILDER);
-      // FIXME: this could lead to a large blocking load depending on the response time
-      this.queryBuilder.clearSearch();
-      this.search();
     }
   }
 
   queryForTreeView() {
     return this.queryBuilder.query;
+  }
+
+  onBuilderQueryChanged(query: string) {
+    this.staleDataState = true;
   }
 
 }


ruffle1986:
Closing and opening it again to force Travis to re-run the tests.

ruffle1986:
@tiborm @sardell is it really impossible to convert the value when we change between the query builder and manual mode? Now it does it but only for the first time. It seemed okay so we could just simply remove the if condition and let it convert all the time not just when the manual value doesn't have a value.
I gave it a try and it seemed to be working however there was a weird Angular error what we have to get rid of obviously but the conversion worked fine.

sardell:
@ruffle1986 The reason I didn't implement a conversion back to the query builder from manual mode was because I wanted to keep things as simple as possible with the initial implementation. After a quick offline discussion with @merrimanr, I felt that there were a lot of possible edge cases that could arise from user input. In addition, we are doing some query manipulation on the front-end around the meta alerts that complicates things even more. But hey, I'm not against it converting back-and-forth if the effort isn't too great on your end.

ruffle1986:
@sardell The effort is huge already if you want to maintain the things around the query builder. I'm working on saving the manual query into local storage and I'm spending hours to figure what's going on. For the record, it has nothing to do with the conversion stuff. The main reason is having these things poorly implemented at the fundamentals. It's a good example why it hurts if we have multiple components and services tied together and everybody changes everybody's internal state.
I would maintain the manual mode and the query builder separately in order to make it easier to focus on smaller chunks. Yes I understand that we're trying to re-use already existing code but I think it's a common mistake here and harmful to follow the DRY principle. We're adding more code in order to cover different edge cases and it's getting harder to maintain and understand.
I highly recommend to allocate some time to refactor the "search section". I've also heard rumours about getting rid of the query builder and only allow users to type their own lucine queries manually. What about that one?

tiborm:
@ruffle1986 @sardell Thanks for elaborating more on this and sharing more details. I believe this this change brings us closer to a desired state and beside that it covers the actual scope of METRON-2276. I suggest to create ticket(s) out of the followup steps of improvements and let this one in.

sardell:
@ruffle1986 100% agree with everything you say, but I also agree with @tiborm that the proposal of this discussion is out of scope for METRON-2276. 

> I highly recommend to allocate some time to refactor the "search section"

I'm on board with that. It's an area of pain for anyone who touches it. Feel free to start a discussion thread proposing how it should be refactored.

> I've also heard rumours about getting rid of the query builder and only allow users to type their own lucine queries manually.

This has been mentioned offline as an idea between a few of us, but I don't think anything official has been proposed in a discussion thread yet. I guess it comes down to how much value does the query builder bring customers. Maybe we can ask in the user mailing list and the Metron Slack channel to gauge out how important that feature is for users. If users don't find it helpful, I would definitely be on-board with getting rid of it.

mmiklavc:
Good discussion

> Maybe we can ask in the user mailing list and the Metron Slack channel to gauge out how important that feature is for users. If users don't find it helpful, I would definitely be on-board with getting rid of it.

That's a great idea @sardell. I'd broadcast it to user and dev. Give it a few days and go from there. Search the mailing list for deprecation examples in the past.

sardell:
I sent out a brief email. Let's see what others have to say: https://lists.apache.org/thread.html/b795489d0ec35146eb108d2db540bb6314bb2ae8680605fc15bb922f@%3Cdev.metron.apache.org%3E

ruffle1986:
What's the status of this PR?

sardell:
Great question. We were hoping to get some feedback from users regarding whether or not the query builder added any value to them, but, unfortunately, we received no responses.

I do think the focus of this PR discussion got off the rails a little bit. Your work resolves a pretty big potential issue of users' DOSing the servers accidentally. [I've created a Jira](https://issues.apache.org/jira/browse/METRON-2337) to track the stale state inconsistencies that @tiborm mentioned, and I agree that they can be addressed as a separate PR (depending on whether or not we deprecate the query builder).

I'm a +1 on this. Tested with Full Dev. Thanks for the contribution, @ruffle1986!

PR #1539, opened 2019-10-22
METRON-2297 Enrichment Topology Unable to Load Geo IP Data from HDFS

nickwallen:
On the `feature/METRON-2088-support-hdp-3.1` feature branch, the Enrichment topology is sometimes unable to load the GeoIP data from HDFS when using Kerberos authentication.  

### The Problem

The Enrichment topology shows this exception.
```
2019-10-03 18:23:18.545 o.a.h.i.Client Curator-TreeCache-0 [WARN] Exception encountered while connecting to the server : org.apache.hadoop.security.AccessControlException: Client cannot authenticate via:[TOKEN, KERBEROS]
2019-10-03 18:23:18.552 o.a.m.e.a.m.MaxMindDbUtilities Curator-TreeCache-0 [ERROR] Unable to open new database file /apps/metron/geo/default/GeoLite2-City.tar.gz
java.io.IOException: DestHost:destPort metrong-1.openstacklocal:8020 , LocalHost:localPort metrong-7/172.22.74.121:0. Failed on local exception: java.io.IOException: org.apache.hadoop.security.AccessControlException: Client cannot authenticate via:[TOKEN, KERBEROS]
```

* When using Kerberos authentication with no previously cached ticket, a topology is unable to access HDFS, which primarily impacts Enrichment and Batch Indexing.
* If the local credential cache on a Storm worker contains a ticket for the 'metron' user, the topology is able to access HDFS.
* When using Kerberos authentication, all topologies are able to access Kafka. This only impacts HDFS.

### Why?

We use [Storm's AutoTGT](https://github.com/apache/storm/blob/v1.2.1/storm-core/src/jvm/org/apache/storm/security/auth/kerberos/AutoTGT.java) class to unpack the user credentials obtained by the Nimbus and authenticate with HDFS.  

When attempting to authenticate with HDFS, this class uses reflection to load Hadoop's `org.apache.hadoop.security.UserGroupInformation` class at runtime.  I believe Storm uses reflection so `hadoop-common` is not a required dependency for all Storm users. 
 
When [this class is loaded at runtime](https://github.com/apache/storm/blob/d156d25d991311eaa1f5131d3dc34787f87ce684/storm-core/src/jvm/org/apache/storm/security/auth/kerberos/AutoTGT.java#L177), it fails indicating that the class is not found on the classpath.  When this occurs, the error is effectively ignored, an `INFO` message is logged (rather than error or warn), and no details from the exception are logged.  This has made it difficult to understand the cause of the problem.

The JVM is unable to load this class at runtime, despite it being on the classpath in three different places.
1. The class is packaged into into the Enrichment topology Uber jar.
1. The class is packaged into the Stellar common jar which is added to the classpath using `--jars`.
1. The class is added to the classpath at runtime by Storm as part of storm-autocreds.

### The Solution

If we use bundle a custom `AutoTGT` class in the topology uber jars, the custom `AutoTGT` implementation is able to correctly load the `hadoop-common` dependency at runtime.  This avoids the problem.

### Changes

1. Created a custom `AutoTGT` implementation that lives within `metron-common-storm`.  This is a clone of Storm's AutoTGT except that it includes additional commentary and logging should this error occur again.

1. Added `metron-common-storm` as a direct dependency of all project's that create a Storm uber jar.  In all cases except `metron-pcap-backend` this was already a transitive dependency.  We need this to be a direct dependency because without the new AutoTGT implementation, none of the topologies will start when using Kerberos authentication.

1. Created an additional configuration in the Mpack for the topology's `topology.auto-credentials` property.  This can no longer default to the value of Nimbus's `nimbus.credential.renewers.classes`.  This property is only exposed to the user as an "advanced" setting and should rarely need changed by a user.

### Validation

This is difficult (impossible?) to test in the development environment.  This should be tested on a multi-node cluster.  The bug cannot be replicated on a single node because the ticket that is cached when the topology is submitted will be reused by the Storm worker when attempting to access HDFS.  The bug only occurs when there is not already a cached ticket to use.

1. Spin-up Metron on a multi-node cluster.
1. Kerberize the environment.
1. Start the Enrichment topology.
1. Ensure that the Enrichment topology is able to load the GeoIP data from HDFS.
1. Ensure that telemetry is being indexed in HDFS.

## Pull Request Checklist

### DIFF
--- a/metron-analytics/metron-profiler-storm/pom.xml
+++ b/metron-analytics/metron-profiler-storm/pom.xml
@@ -148,6 +148,11 @@
         </dependency>
         <dependency>
             <groupId>org.apache.metron</groupId>
+            <artifactId>metron-common-storm</artifactId>
+            <version>${project.parent.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.metron</groupId>
             <artifactId>metron-writer-storm</artifactId>
             <version>${project.parent.version}</version>
             <exclusions>
--- a/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-security-env.xml
+++ b/metron-deployment/packaging/ambari/metron-mpack/src/main/resources/common-services/METRON/CURRENT/configuration/metron-security-env.xml
@@ -159,7 +159,6 @@
     </value-attributes>
     <on-ambari-upgrade add="true"/>
   </property>
-
   <property>
     <name>metron.ldap.ssl.truststore</name>
     <display-name>LDAP Truststore</display-name>
@@ -239,5 +238,14 @@
     <value>ADMIN</value>
     <description>Name of the role at the authentication provider that provides administrative access to Metron.</description>
     <on-ambari-upgrade add="true"/>
+  </property>
+  <property>
+    <name>topology_auto_credentials</name>
+    <display-name>Topology Auto Credentials</display-name>
+    <description>The value of Storm's `topology.auto-credentials`. A list of plugins used to unpack credentials on the Storm worker. This value is only used when Kerberos has been enabled.</description>
+    <value>['org.apache.metron.storm.security.auth.kerberos.AutoTGT']</value>
+    <value-attributes>
+      <empty-value-valid>true</empty-value-valid>
+    </value-attributes>
   </property>
 </configuration>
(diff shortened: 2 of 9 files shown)


mmiklavc:
It would be great to understand what happened in Storm to change this, however we've used similar workarounds in the past when dealing with classpath related issues in Hadoop, Kafka, Storm, and probably more that I'm forgetting. I looked over everything and it looks clean and well-documented. +1 by inspection from me.

nickwallen:
@mmiklavc I definitely agree with your comments.  Would love to understand true root cause here. Thanks for the review.  This has been merged into the feature branch.

PR #1546, opened 2019-10-29
METRON-2239 Metron Automated backup and restore

mmiklavc:
## Contributor Comments

I created a backup and restore script to assist with Metron upgrades. I'm creating this against master as it should be made available prior to the FB becoming the next release.

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

- n/a Have you ensured that any documentation diagrams have been updated, along with their source files, using [draw.io](https://www.draw.io/)? See [Metron Development Guidelines](https://cwiki.apache.org/confluence/display/METRON/Development+Guidelines) for instructions.

#### Note:

### DIFF
--- a/Upgrade_steps.md
+++ b/Upgrade_steps.md
@@ -1,0 +1,40 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+# Upgrade Steps
+General guidance for upgrading Metron
+
+1. Cut-off all inputs to Metron
+1. Monitor processing until all in-flight data is indexed and stored
+1. Stop Metron service
+1. Run the [Metron Upgrade Helper](./metron-platform/metron-common#metron-upgrade-helper) script in backup mode - does these 2 tasks
+    1. Export zookeeper based metron configs
+    1. Export ambari-based metron configs
+1. Delete the metron service via Ambari
+1. Upgrade OS, if applicable
+1. Major Hadoop platform upgrade steps (example. using HDP 2.6 to 3.x) - https://docs.cloudera.com/HDPDocuments/Ambari-2.7.3.0/bk_ambari-upgrade-major/content/ambari_upgrade_guide.html
+   1. Update Ambari to latest version
+   1. Update to HDP 3.1 using existing Ambari/HDP update documentation (including Solr/ES etc).
+1. Reinstall Metron mpack using "--force" option
+1. Reinstall Metron service
+1. Turn off Metron service
+1. Tweak any configs required to support new Metron version
+   1. Minimally, you should update `metron.home` in `Ambari -> Metron -> Configs -> Advanced metron-env -> Metron home` to point to the new Metron home dir.
+1. Run upgrade helper script in restore mode - does these 2 tasks
+   1. Re-import zookeeper based metron configs
+   1. Re-import ambari-based zookeeper configs
+1. Turn on Metron service
--- a/metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
+++ b/metron-deployment/packaging/docker/rpm-docker/SPECS/metron.spec
@@ -143,6 +143,7 @@ This package installs the Metron common files %{metron_home}
 %{metron_home}/bin/stellar
 %{metron_home}/bin/cluster_info.py
 %{metron_home}/bin/tgt_renew.py
+%{metron_home}/bin/upgrade_helper.sh
 %{metron_home}/config/zookeeper/global.json
 %attr(0644,root,root) %{metron_home}/lib/metron-common-%{full_version}-uber.jar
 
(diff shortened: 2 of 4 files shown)


ottobackwards:
Thanks @mmiklavc!
Can you run `shellcheck upgrade_helper.sh` on this script?
Also, what would be the steps for a current user to use this to prepare for the next big upgrade?

mmiklavc:
@ottobackwards - I ran shellcheck and modified some exit codes. Other than that, it's warnings about shell expansion on variables, which we also have all over in other scripts. I think it should be good to go from that perspective.

WRT added instructions for the upgrade, I've debated where these belong. With the ES upgrade we placed instructions right in the module - https://github.com/apache/metron/blob/b9a130ca96774add38865d3934ed61ca19599b87/metron-platform/metron-elasticsearch/metron-elasticsearch-common/README.md#upgrading-from-233-to-56. We have an existing `Upgrading` document - https://github.com/apache/metron/blob/b9a130ca96774add38865d3934ed61ca19599b87/Upgrading.md, however this is primarily some line items with pointers to other details. What do you think about adding a new README in the project root named something like`Upgrade_steps.md`? In master, we'd start with the existing backup/restore process outline. For the HDP 3.1 feature branch, I'd expand out those details to include things like specifying that we need to upgrade from Centos 6 to 7.

nickwallen:
> @ottobackwards: what would be the steps for a current user to use this to prepare for the next big upgrade?

We will definitely need some notes on upgrading; I agree with @ottobackwards  there.

> @mmiklavc: WRT added instructions for the upgrade, I've debated where these belong....

I think we should handle any upgrading notes in our existing `Upgrading.md` document and do that work as a separate PR.

I think the README changes included in this PR sufficiently covers how to use the script itself.  The expectation being that a user would have to pull down this script from master into their legacy environment to upgrade.

mmiklavc:
Updated this to include a new `Upgrade_steps.md" readme. This should provide more context on how/where to run this script.

ottobackwards:
Have we thought about any other possibilities other than "go to current master and download and run this script"?  That seems a little weird.

mmiklavc:
I could link to it in our wiki. I had considered if we had another release, however most of the work done since our last release has been on this upgrade, so I'm a bit hesitant.

mmiklavc:
@ottobackwards Any other thoughts on this?

ottobackwards:
I have not run spellcheck again, does it pass now?

mmiklavc:
@ottobackwards Any thoughts on this comment also? https://github.com/apache/metron/pull/1546#issuecomment-548510246

WRT shellcheck, I fixed a couple issues, as indicated in my earlier comments. Here's the current output - https://gist.github.com/mmiklavc/bb9cd52994a0ca6c598a960787871d66. This script currently matches our style and patterns on escaping/variable expansion in all our other shell scripts. Anything I'm missing you're specifically requesting? I don't think I agree with the checker that we should go quoting every variable in this case, but maybe there's a good reason for it. Minimally, if we think it's worthwhile to use shellcheck as a passing standard, I think that's worth a discuss thread about the tool, how it compares to others, what we do/don't expect of it, etc.

ottobackwards:
@JonZeolla and I have been using shellcheck for all of our bro testing scripts, and I used it for the RC scripts and the common script as well.  I don't think I would call the existing bash scripts style and practice.

I think it is a good practice, and makes it really clear for everyone contributing to script code what is required.

That being said, it is _not_ the project standard, so I can't make you do it.

ottobackwards:
As far as linking to the wiki.... I don't think people will find Metron or the update from the wiki.  I would like to it from the main site if anything.

This seems like it should be it's own thing.  Having folks download from git doesn't seem right.  But other than releasing another artifact and putting it on a.o.dist i'm not sure what to do.

mmiklavc:
>  I don't think I would call the existing bash scripts style and practice.

Wouldn't call them what?

> I think it is a good practice, and makes it really clear for everyone contributing to script code what is required.

I think it's a good practice to have standards also - we typically discuss them on the dev list. I just haven't heard of or used that tool before, so I'm unclear what the alternatives are.

> That being said, it is _not_ the project standard, so I can't make you do it.

I ran it and shared the remaining results in my comment - all I see at the moment are warnings regarding double quotes. Is there anything in that list (or all of it?) you think we should do? It's not a hard change to make, but I do think it makes the script messier for something that we have never run into afaik. I don't have a strong opinion on it either way.

mmiklavc:
> As far as linking to the wiki.... I don't think people will find Metron or the update from the wiki. I would like to it from the main site if anything.

It's probably worth spinning up or revisiting an existing discuss thread on this. I think we'll want the script in the source code, of course. But having it available in another place while the transition occurs is not unreasonable either.

> This seems like it should be it's own thing. Having folks download from git doesn't seem right. But other than releasing another artifact and putting it on a.o.dist i'm not sure what to do.

I could also create a gist for it or link to the raw source with some basic instructions on the wiki, and linked to on the website. The wiki requires no explicit vote, website update is a simple PR, and then we can wrap this up by referencing this in the final vote on merging the FB. That way we can be sure everything gets covered.

tigerquoll:
+1 (non-binding)
I gave these scripts a test when updating Metron 0.7.1 (HDP 2.6) -> Metron HDP 3.1 support branch.  They worked well, though I only setup Metron 0.71 with a handful of tutorial/walkthroughs beforehand, so the updated fresh Metron install was similar in respects to the configuration I updated it to.  There could be edge cases not covered by a simple upgrade test (such as a larger version upgrade jump).  But it's definitely better having the scripts than not having them.  For any serious production upgrade I would recommend:
1. Use the script to pull the configuration of the existing Metron install
2. Using the script to pull the configuration down from a freshly installed Metron update, 
2. Manually comparing / updating the differences in the pulled configuration files
3. Using the script to push the updated configurations into the upgraded cluster that has the freshly installed Metron/Mpack on it.

ottobackwards:
+1
I don't understand the resistance to shellcheck.  But I'm not going to belabor the point.
I don't think we have a good enough idea of how to make this discoverable and usable overall, although we have some places where we can 'stick this', but that can be a follow on discussion etc.

PR #1542, opened 2019-10-24
METRON-2233 Deprecate CentOS 6 Development Environment

nickwallen:
With the HDP-3.1 upgrade the existing CentOS 7 development environment will become the gold standard for developers.  This will replace the CentOS 6 development environment, which will no longer be needed.

## Changes

* Removed the CentOS 6 development environment.
* Updated all documentation references to point to the CentOS 7 development environment.
* Added a README to the CentOS 7 development environment to match what we had for CentOS 6.
* Updated the Docker container on which the RPMs are built to CentOS 7.

## Acceptance Testing

This PR should be tested using the centos7 development environment.  

1. Start up the centos7 dev environment.
    ```
    cd metron-deployment/development/centos7
    vagrant destroy -f
    vagrant up
    ```

### Basic Indexing and Enrichment

Ensure that we can continue to parse, enrich, and index telemetry.  Verify data is flowing through the system, from parsing to indexing

1. Open Ambari and navigate to the Metron service http://node1:8080/#/main/services/METRON/summary

1. Open the Alerts UI.  Verify alerts show up in the main UI - click the search icon (you may need to wait a moment for them to appear)

1. Go to the Alerts UI and ensure that an ever increasing number of telemetry from Bro, Snort, and YAF are visible by watching the total alert count increase over time.

1. Ensure that geoip enrichment is occurring.  The telemetry should contain fields like `enrichments:geo:ip_src_addr:location_point`.

1. Head back to Ambari and select the Kibana service http://node1:8080/#/main/services/KIBANA/summary

1. Open the Kibana dashboard via the "Metron UI" option in the quick links

1. Verify the dashboard is populating

## Pull Request Checklist

- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel).
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
- [x] Have you included steps or a guide to how the change may be verified and tested manually?
- [x] Have you ensured that the full suite of tests and checks have been executed in the root metron folder via:
- [x] Have you written or updated unit tests and or integration tests to verify your changes?
[...]

PR #1551, opened 2019-10-30
METRON-2223 Reconcile Versions in Root POM with HDP 3.1

nickwallen:
This change includes a few clean-up items that are needed prior to the `feature/METRON-2088-support-hdp-3.1` feature branch being merged into master.  I validated each of the component versions that we are building against to ensure they match against [HDP-3.1](https://docs.cloudera.com/HDPDocuments/HDP3/HDP-3.1.0/release-notes/content/comp_versions.html).

Each of these changes are included in separate commits to make review a bit simpler.  These include.

* Removed the legacy `HDP-2.5.0.0` profile from the root POM.
* Changed the Spark dependency from 2.3.1 to 2.3.2 to match HDP-3.1
* Removed a `storm-kafka` dependency from the root POM that was never needed.
* Moved all properties defining versions of core platform components to the `HDP-3.1` profile.
* Reorganized existing version properties into separate sections for (1) build dependencies (2) test dependencies and (3) build tooling
* Removed unused properties.
* Pulled out some of the build tooling versions into properties to match what we had done in other places.  

I validated this by comparing the version numbers output by `mvn dependency:tree` from before and after the change.  The only version changes introduced by this PR involve the Spark change. This in addition to the basic acceptance testing were used to validate the change.

I'd be happy to break any of these changes out into separate PRs should a reviewer think that is needed.

## Acceptance Testing

This PR should be tested using the centos7 development environment.  

1. Start up the centos7 dev environment.
    ```
    cd metron-deployment/development/centos7
    vagrant destroy -f
    vagrant up
    ```

### Basic Indexing and Enrichment

Ensure that we can continue to parse, enrich, and index telemetry.  Verify data is flowing through the system, from parsing to indexing

1. Open Ambari and navigate to the Metron service http://node1:8080/#/main/services/METRON/summary

1. Open the Alerts UI.  Verify alerts show up in the main UI - click the search icon (you may need to wait a moment for them to appear)

1. Go to the Alerts UI and ensure that an ever increasing number of telemetry from Bro, Snort, and YAF are visible by watching the total alert count increase over time.

1. Ensure that geoip enrichment is occurring.  The telemetry should contain fields like `enrichments:geo:ip_src_addr:location_point`.

1. Head back to Ambari and select the Kibana service http://node1:8080/#/main/services/KIBANA/summary

1. Open the Kibana dashboard via the "Metron UI" option in the quick links

1. Verify the dashboard is populating

#### Batch Profiler

1. Install Spark2 using Ambari.

    1. Stop Metron and Sensor-Stubs to free-up resources.

    1. Install the MySQL Connector and ensure Ambari can find it.
       ``` 
       yum -y install mysql-connector-java
       ln -s /usr/share/java/mysql-connector-java.jar /var/lib/ambari-server/resources/mysql-connector-java.jar
       ```

    1. Install Spark2 using Ambari. YARN must be kept running for the install to complete successfully.

    1. Ensure that Spark can talk with HBase.
        ```
        cp /etc/hbase/conf/hbase-site.xml /etc/spark2/conf/
        ```

1. Use the Batch Profiler to back-fill your profile.  To do this, follow the direction [provided here](https://github.com/apache/metron/tree/master/metron-analytics/metron-profiler-spark#getting-started).

1. Retrieve the entire profile, including the back-filled data.

    ```
    [Stellar]>>> PROFILE_GET("hello-world", "global", PROFILE_FIXED(30, "DAYS"))
    [1203, 2849, 2900, 1944, 1054, 1241, 1721]
    ```

## Pull Request Checklist


mmiklavc:
This is labeled "METRON-2232", but is this actually "METRON-2223"? It's not syncing with Jira currently.

pom.xml
```diff
@@ -138,38 +141,17 @@
                 <activeByDefault>true</activeByDefault>
```
nickwallen: I was torn as to whether we should even define a profile in the root POM (like HDP-3.1) now that we support only one set of versions.  I left the existing HDP-3.1 profile, which is the default, so whether you build with the profile or without you get the same result.

nickwallen: The profile is where all the core "platform" versions are specified.

pom.xml
```diff
@@ -185,11 +167,6 @@
                 <artifactId>mockito-core</artifactId>
                 <version>${global_mockito_version}</version>
             </dependency>
-            <dependency>
```
nickwallen: I don't understand why `storm-kafka` was defined as a dependency here originally.  After removing it, I compared all the dependencies that maven reports via `dependency:tree` and nothing changed.

pom.xml
```diff
@@ -185,11 +167,6 @@
...
-                <groupId>org.apache.storm</groupId>
-                <artifactId>storm-kafka</artifactId>
```
nickwallen: I don't understand why `storm-kafka` was defined as a dependency here originally.  After removing it, I compared all the dependencies that maven reports via `dependency:tree` and nothing changed.

justinleet: I'm not sure why we have it here, but it's not a dependency directly. It's in the dependencyManagement section, so it's theoretically defining a dependency that children can use without having to specify the version (i.e. instead of the various global params we use, you could also define it here, and the children would just say the group/id). I believe it can also be useful for doing things like moving exclusions to be declared in a single place (e.g. always removing logging deps), but my recollection is a bit hazy.

If it were actually used however, the build would have broken. I have no idea why we have the section in general, because we don't actually use it the way it would be intended.

And I'm also scared to look at the git blame, because there's an extremely nonzero chance I did it.

mmiklavc: @justinleet @nickwallen - nope, looks like that was me - https://github.com/mmiklavc/metron/blame/7f1160401897c49738dcb79609661419d39e112d/pom.xml#L116. The way the dependency management section works is to lay down versions and other details about a dependency so you can include it in your child projects without having to explicitly specify the version. Ideally, this would be the way we declare all shared dependencies in Metron. Submodules are free to override the default if/when they choose by explicitly declaring the version in their pom dep.

mmiklavc: If this is not needed, then I think it's fine to remove it. We might consider a separate task at some point of refactoring our poms to use this idiom rather than having to leverage version properties all over. It should clean up our poms considerably.

mmiklavc:
I see a file here named manual install centos 6 - after https://github.com/apache/metron/pull/1542, should this file be renamed? Other than that, this looks good.

nickwallen:
Yeah it looks like there are some centos6 installation files in Ansible that are effectively dead code now.  I can remove those.

nickwallen:
@mmiklavc I believe I cleaned-up all the centos6 Ansible files that you referenced.  Let me know if I missed something.

mmiklavc:
This file is still showing up as "metron-deployment/other-examples/manual-install/Manual_Install_CentOS6.md" - https://github.com/apache/metron/pull/1551/files#diff-c547d25b027d9643c35d4a9d78894112

nickwallen:
Ok, removed those bits.

mmiklavc:
@nickwallen - sorry, I take back my comments on that manual install file. I started looking at the details and it hasn't been touched since 0.4. I think we should leave it alone (name and all) for now and tackle that as a follow-on.

mmiklavc:
+1 on this (pending bringing that manual install file back)

nickwallen:
Yes, agreed.  We can handle that in a follow-on.  I reverted the last commit (aka restored the file.)

nickwallen:
This has been merged into the feature branch.

PR #1553, opened 2019-11-02
METRON-2306: [UI] Grok parsers' have duplicate timestampField

sardell:
## Contributor Comments
WIP!

Link to original ASF Jira ticket: https://issues.apache.org/jira/browse/METRON-2306

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/metron-interface/metron-config/src/app/model/sensor-parser-config.ts
+++ b/metron-interface/metron-config/src/app/model/sensor-parser-config.ts
@@ -23,7 +23,7 @@ export class SensorParserConfig {
   writerClassName: string;
   errorWriterClassName: string;
   invalidWriterClassName: string;
-  parserConfig: {};
+  parserConfig: { timestampField?: string };
   fieldTransformations: FieldTransformer[];
   numWorkers: number;
   numAckers: number;
@@ -35,13 +35,11 @@ export class SensorParserConfig {
   errorWriterNumTasks: number;
   spoutConfig: {};
   stormConfig: {};
-  timestampField: string;
 
   constructor() {
     this.parserConfig = {};
     this.fieldTransformations = [];
     this.spoutConfig = {};
     this.stormConfig = {};
-    this.timestampField = 'timestamp';
   }
 }
--- a/metron-interface/metron-config/src/app/sensors/sensor-parser-config/sensor-parser-config.component.ts
+++ b/metron-interface/metron-config/src/app/sensors/sensor-parser-config/sensor-parser-config.component.ts
@@ -78,6 +78,7 @@ export class SensorParserConfigComponent implements OnInit {
   grokStatement = '';
   patternLabel = '';
   currentSensors = [];
+  paramsID: string;
 
   editMode: boolean = false;
 
@@ -191,8 +192,8 @@ export class SensorParserConfigComponent implements OnInit {
 
   ngOnInit() {
     this.route.params.subscribe(params => {
-      let id = params['id'];
-      this.init(id);
+      this.paramsID = params['id'];
+      this.init(this.paramsID);
     });
     this.createForms();
     this.getAvailableParsers();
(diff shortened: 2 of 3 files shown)


metron-interface/metron-config/src/app/model/sensor-parser-config.ts
```diff
@@ -23,7 +23,7 @@ export class SensorParserConfig {
   writerClassName: string;
   errorWriterClassName: string;
   invalidWriterClassName: string;
-  parserConfig: {};
+  parserConfig: { [key: string]: any; }
```
sardell: Typing using `{}` just extends the Object interface. What we really want here is an object literal that's editable by users.

metron-interface/metron-config/src/app/sensors/sensor-parser-config/sensor-parser-config.component.ts
```diff
@@ -191,8 +192,8 @@ export class SensorParserConfigComponent implements OnInit {
 
   ngOnInit() {
     this.route.params.subscribe(params => {
-      let id = params['id'];
-      this.init(id);
+      this.paramsID = params['id'];
```
sardell: Storing the id in a class property allows me to use it when evaluating whether or not to populate the parserConfig object with the timestampField property and default `timestamp` value.

mmiklavc:
@sardell - I'm a little confused - the only place I think we deal with timestamp fields and/or properties is in the parserConfig itself (the very specifically named object within the broader parser configuration - yeah, calling something a "parserConfig" within a parser config is a bit confusing to try to discuss accurately). Anyhow, where is/was the other "TIMESTAMP" property being set if not in the "parserConfig" object? Specifically for Grok parsers, (e.g. Yaf, Squid) isn't this just a matter of requiring the "timestampField" property in the config section?

It's possible this is a problematic request bc of how that parserConfig object is handled in the UI, but you might also consider making "TIMESTAMP" readonly and simply grab the value from "timestampField", with the error message reflecting the same. I'm not sure I completely understand the problem in the UI, so maybe we can discuss this a bit further.

sardell:
@mmiklavc 
> Anyhow, where is/was the other "TIMESTAMP" property being set if not in the "parserConfig" object?

When you fill in the TIMESTAMP input (ie. the input outside of the parser config object UI area) and submit to create a new parser, the request looks like this:
```
{
    "parserConfig": {
        "grokPath": "/apps/metron/patterns/test",
        "patternLabel": "TEST"
    },
    "fieldTransformations": [],
    "spoutConfig": {},
    "stormConfig": {},
    "timestampField": "timestamp",
    "parserClassName": "org.apache.metron.parsers.GrokParser",
    "sensorTopic": "yaf"
}
```

That's because the model we have implemented on the front-end expects timestampField to be a property of the parser, but instead it's a property of parserConfig. This also explains why we do not see the value of an existing parser in that field. The validator for that input looks like this:

```
group['timestampField'] = new FormControl(
      this.sensorParserConfig.timestampField,
      Validators.required
    );
```
That first parameter in FormControl is what a field's default value should be. It should be looking at `this.sensorParserConfig.parserConfig.timestampField` instead given what is currently returned. I 
 haven't looked into whether this is the result of miscommunication at implementation or a change in what the REST endpoint is returning, but this is why it currently doesn't work.

> Specifically for Grok parsers, (e.g. Yaf, Squid) isn't this just a matter of requiring the "timestampField" property in the config section?

That's exactly what I implemented in this PR.

> you might also consider making "TIMESTAMP" readonly and simply grab the value from "timestampField", with the error message reflecting the same

I could do this. Given the way we currently present the parserConfig object in the UI, I felt it was better to keep the property in the parserConfig. However, I can be swayed to take this approach if others feel it's better to have a separate input field.

When I made this PR, I made a lot of assumptions based on what is implemented in the UI. This is why I originally mentioned that the REST API should be updated to have timestampField as a property on the Grok parser itself. However, if it belongs inside the parserConfig object, I would rather include it as part of the parserConfig section in the UI. Unless anyone objects, I'm going to proceed to add validation onto the work I did and update here when I have that done.

sardell:
@mmiklavc I managed to update the parserConfig to include a validation message for the timestampField. Whenever you get a chance, take a look and let me know if this is an acceptable solution for you and if I cleared up the questions you had with my previous comment.

PR #1554, opened 2019-11-04
METRON-2307: Migrate to JUnit5

justinleet:
## Contributor Comments
Moves to JUnit 5. This came out of my forever ago desire to look into JUnit5 extensions to share test infra (Kafka, ZK, etc.) across our test classes, rather than spin up for each class.

- Migrates everything to JUnit5 annotations.
- Migrates everything to JUnit5 assertions. This includes making everything `assertThrows`
- The only `@Rule`s left are for temporary directories. This feature is experimental in JUnit5 (although I do use it once for a `@ClassRule` one.  We could potentially migrate over, it's not like it'd be a huge deal to upgrade to whatever they choose in the future.
- PowerMock is gone. Doesn't work in JUnit 5. Various methods are added to source to be able to mock out / spy on / whatever to get it to work. Tried to be pretty minimal about this change, since it hits outside of tests.
- I believe a single test got deleted: One for a two line method in `MetronRestApplication`, iirc. Didn't seem super helpful to test that it ran the `SpringApplication`.
 
Build works (potentially barring the last couple cleanup commits which are going through my personal Travis), but I still need to spin it up on dev.

I'm not super thrilled about this being a large PR, but it was an enormous pain that kept finding things backwards in modules I thought I'd already cleared. It could probably be split out, but I'll be honest: If I see that request, I'll probably ignore it for awhile.

## Pull Request Checklist

Thank you for submitting a contribution to Apache Metron.  
Please refer to our [Development Guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61332235) for the complete guide to follow for contributions.  
Please refer also to our [Build Verification Guidelines](https://cwiki.apache.org/confluence/display/METRON/Verifying+Builds?show-miniview) for complete smoke testing guides.  

In order to streamline the review of the contribution we ask you follow these guidelines and ask you to double check the following:

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? If not one needs to be created at [Metron Jira](https://issues.apache.org/jira/browse/METRON/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel).
- [x] Does your PR title start with METRON-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.
- [x] Has your PR been rebased against the latest commit within the target branch (typically master)?

### For code changes:
- [x] Have you included steps to reproduce the behavior or problem that is being changed or addressed?
[...]

PR #1568, opened 2019-11-20
METRON-2324: verify_licenses.sh and list_dependencies.sh don't properly handle maven failures

justinleet:
## Contributor Comments
Discovered in #1554

Travis should work the same as local now. Couple changes happened
  - Readded JUnit dep, since it does occur as part of metron-integration-test
  - Run a build step before verify is called
  - Adjusted the npm scripts to not fail if we provide `-Dskip.npm`. This ensures that all the modules are built for the verify, but we don't bother with anything related to npm.
  - Adjusted the build scripts to properly relay issues if maven fails. We could probably add various quiet options to maven if we want.

## Pull Request Checklist

### For all changes:

### For code changes:
  ```
  ```

  ```
  cd site-book
  mvn site
  ```

#### Note:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,7 @@ matrix:
 
     - name: Verify Licenses
       script:
+        - time mvn install -T 2C -q -DskipTests=true -Dmaven.javadoc.skip=true -B -V -Dskip.npm
         - time ./dev-utilities/build-utils/verify_licenses.sh
 
 before_cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,12 +36,12 @@ matrix:
   include:
     - name: Unit Tests
       script:
-        - time mvn install -T 2C -q -DskipTests=true -Dmaven.javadoc.skip=true -B -V -pl '!:metron-config,!:metron-alerts'
+        - time mvn install -T 2C -q -DskipTests=true -Dmaven.javadoc.skip=true -Dskip.npm -B -V
         - time mvn surefire:test@unit-tests -T 2C
 
     - name: Integration Tests
       script:
-        - time mvn install -T 2C -q -DskipTests=true -Dmaven.javadoc.skip=true -B -V -pl '!:metron-config,!:metron-alerts'
+        - time mvn install -T 2C -q -DskipTests=true -Dmaven.javadoc.skip=true -Dskip.npm -B -V
         - time mvn surefire:test@integration-tests
 
     - name: Alerts UI Tests
@@ -60,7 +60,6 @@ matrix:
 
     - name: Verify Licenses
       script:
-        - time mvn install -T 2C -q -DskipTests=true -Dmaven.javadoc.skip=true -B -V -Dskip.npm
         - time ./dev-utilities/build-utils/verify_licenses.sh
 
 before_cache:
--- a/dev-utilities/build-utils/list_dependencies.sh
+++ b/dev-utilities/build-utils/list_dependencies.sh
@@ -16,4 +16,16 @@
 #  limitations under the License.
 #
 
-{ mvn dependency:list || { echo "ERROR:  Failed to run mvn dependency:list" ; exit 1 ; } ; mvn dependency:list -PHDP-2.5.0.0 || { echo "ERROR:  Failed to run mvn dependency:list -PHDP-2.5.0.0" ; exit 1 ; } ; } | grep "^\[INFO\]   " | awk '{print $2}' | grep -v "org.apache" | grep -v "test" | grep -v "provided" | grep -v "runtime" | grep -v ":system" |  sort | uniq
+DEPS=$(mvn dependency:list)
+rc=$?
+if [[ $rc != 0 ]]; then
+  echo "ERROR:  Failed to run mvn dependency:list"
+  DEPS=$(mvn dependency:list -PHDP-2.5.0.0)
+  rc=$?
+  if [[ $rc != 0 ]]; then
+    echo "ERROR:  Failed to run mvn dependency:list -PHDP-2.5.0.0"
+    exit $rc
+  fi
+fi
+
+echo "$DEPS" | grep "^\[INFO\]   " | awk '{print $2}' | grep -v "org.apache" | grep -v "test" | grep -v "provided" | grep -v "runtime" | grep -v ":system" | sort | uniq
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -60,6 +60,7 @@ matrix:
 
     - name: Verify Licenses
       script:
+        - time mvn install -T 2C -q -DskipTests=true -Dmaven.javadoc.skip=true -B -V -Dskip.npm
```
nickwallen: (1) What does `-Dskip.npm` do?

(2) Why do we need to build before running `verify_licenses.sh`?

(3) Assuming we do need to build, would it make more sense to put the build step in `dev-utilities/build-utils/list_dependencies.sh`?  If I just manually run that script, we should expect it to work also, right?

dev-utilities/build-utils/list_dependencies.sh
```diff
@@ -16,4 +16,16 @@
 #  limitations under the License.
 #
 
-{ mvn dependency:list || { echo "ERROR:  Failed to run mvn dependency:list" ; exit 1 ; } ; mvn dependency:list -PHDP-2.5.0.0 || { echo "ERROR:  Failed to run mvn dependency:list -PHDP-2.5.0.0" ; exit 1 ; } ; } | grep "^\[INFO\]   " | awk '{print $2}' | grep -v "org.apache" | grep -v "test" | grep -v "provided" | grep -v "runtime" | grep -v ":system" |  sort | uniq
+DEPS=$(mvn dependency:list)
+rc=$?
```
nickwallen: Is this just a refactor for readability?  Or was the functionality changed at all here?

justinleet:
@nickwallen Check out the ticket itself, I explain a bit more exactly what's happening there, re: why build.

https://jira.apache.org/jira/browse/METRON-2324

.travis.yml
justinleet: 1. Just skips the UI portion of the build. it's already built-in. The catch is our build currently explodes if we do it, so a couple scripts needed adjustment. It just avoids doing the UI build work when we aren't actually looking to do it.
1. Check out the ticket itself.
1. In practice, I don't know that anyone actually runs the script without already having relevant builds (after all, messing with deps file is pretty much the last thing anybody does). We could split the difference and add a flag to run the build to the script and kick out a warning if it's not done or something?

dev-utilities/build-utils/list_dependencies.sh
justinleet: It's one of the main fixes, because of the incorrect failure handling described in the ticket. It has to do with exiting during pipes not stopping the chained call.

.travis.yml
nickwallen: (3) 

> In practice, I don't know that anyone actually runs the script without already having relevant builds ...

If I get a Travis failure on the license verification, then switch to my local machine to figure out why, I would just run the script.  This could put me in a situation where the outputs are different and it would be confusing.  

It seems to have caused you some headaches in the JUnit 5 upgrade, so I think we should do more to avoid causing someone else the same headaches in the future.

> We could split the difference and add a flag to run the build to the script and kick out a warning if it's not done or something?

If you want to add a flag to the script, that works for me.  Just calling `mvn install ... ` from inside the script seems even simpler to me, but I'd be happy with a flag and check to avoid the problem.

nickwallen: (1) 

> Just skips the UI portion of the build. it's already built-in. 

Is this functionally different than `-pl '!:metron-config,!:metron-alerts'` that is already used elsewhere in our `.travis.yml`?  

If it is functionally the same, then why not just use the existing approach?

ottobackwards:
Can't we just run it as part of the root pom?  Like the last module in the list or something?  And forget about the if's and but's?

.travis.yml
justinleet: That's a good point, I don't think it's functionally different unless we were to restructure things in the future. Having said that, I'd actually make the opposite argument. If the point is to avoid npm, we should be using the built-in skip.npm and building all the modules, rather than making exceptions.

Calling maven inside the script it probably fine. I was thinking that running maven takes a lot longer than it actually does.

nickwallen: > Having said that, I'd actually make the opposite argument. If the point is to avoid npm, we should be using the built-in skip.npm and building all the modules, rather than making exceptions.

I am just interested in consistency.  Using one approach is just the most obvious thing to do here and leaves less open questions for future maintainers.

If you can update the PR to use one approach or the other, I'd be a happy camper.

justinleet:
I moved over to skip.npm in the Travis builds. I haven't run that locally, I'm just letting Travis do it.

I also added the build step @nickwallen suggested to the file directly.  If we have another way to do it, I'm open to options. As noted above, I like the idea of getting it done in the pom, but I'm not sure how to do it cleanly.

ottobackwards:
https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle
`verify` ?

justinleet:
I believe `package` is insufficient, it needs to be `install`, because the deps are pulled from the local repo when another module is depended on for `dependency:list`.

Even then the issue is still that the parent module actually runs first in our case (and this is not a directly controllable or guaranteed thing to the best of my knowledge; it's entirely reactor's call). This means that `verify` would essentially be immediately called and fail because the build hasn't been done.

nickwallen:
All the changes are looking good.  I am just seeing one issue with the validation. I took this branch and removed the `junit` addition to `dependencies_with_url.csv`.  I would expect validation to fail in this case.

Unfortunately, when running in Travis CI the job does not fail. 
*  See https://travis-ci.org/nickwallen/metron/builds/621740129.

When running locally everything builds correctly and I see an error message, but the return code is still 0, when it should probably be non-zero.
```
$ time ./dev-utilities/build-utils/verify_licenses.sh
...
Determining dependencies
Traceback (most recent call last):
  File "dev-utilities/build-utils/verify_license.py", line 44, in <module>
    raise ValueError("Unable to find these components: \n  " + "\n  ".join(components_not_found) + "\nin the acceptable list of components: " + sys.argv[1])
ValueError: Unable to find these components:
  junit:junit:jar:4.12:compile
in the acceptable list of components: ./dependencies_with_url.csv
Finished validating dependencies.

real	2m29.260s
user	9m8.852s
sys	0m32.922s

$ echo $?
0
```

justinleet:
@nickwallen Should be fixed with a proper exit code in the main script and a more useful message.

https://travis-ci.org/justinleet/metron/builds/621750687 succeeds.

https://travis-ci.org/justinleet/metron/jobs/621751002 fails (off a branch that is missing the JUnit dep. It doesn't have the changed message commit, but I tested that locally afterwards).

nickwallen:
+1 Thanks @justinleet

justinleet:
@ottobackwards Are you okay with this approach, or should we keep looking for an alternative?

ottobackwards:
No, I am fine with _any_ improvement.

